PR #7, opened 2026-05-06
feat(cli/register): agent registering logic

guillaume-hexamind:
fortify/cli:
- manifest creation from a specific agent. Supported: fortify agent, openai, google, langchain, pydantic_ai
- agent register cli: creation of a manifest from path of specified agent, POST manifest to platform /v1/agents

platform/api:
- creation of tool and agent version table
- POST /v1/agents endpoint: from agent manifest, check if the agent already exists. If not, create an agent and a first version. Otherwise, create a new version if the manifest has changed

Ruff formatting


guillaume-hexamind:
### Code review

Found 3 issues:

1. `create_openai_manifest` calls `_to_tool_definition` on every item in `agent.tools`, but only `FunctionTool` exposes `params_json_schema`. An OpenAI agent that uses any hosted tool (e.g. `FileSearchTool`, `WebSearchTool`, `ComputerTool`) will crash with `AttributeError` at registration time. There is no `isinstance(t, FunctionTool)` guard.

https://github.com/HexamindOrganisation/coolagents/blob/3dc8d50a0589d0bf5126c3241ec2d42e9c60f773/fortify/cli/register/openai.py#L21-L29

2. `create_pydantic_ai_manifest` passes `agent.name` straight into `AgentManifest(name=...)`, which is typed as a non-optional `str`. Pydantic AI's `Agent.name` is `str | None` and defaults to `None`, so unnamed agents will fail with a `pydantic.ValidationError`. The `fortify` adapter handles this with `agent.name or "fortify_agent"` — the same fallback is missing here.

https://github.com/HexamindOrganisation/coolagents/blob/3dc8d50a0589d0bf5126c3241ec2d42e9c60f773/fortify/cli/register/pydantic_ai.py#L26-L32

3. `POST /v1/agents` does not call `ensure_default_project(session)` before `register_manifest`, while every peer agent endpoint (`api_list_agents`, `api_get_agent`, `mint_token`) does. `require_project` only validates the bearer token; it does not lazily create the project row. If this SDK endpoint is the first one hit on a fresh DB (a likely scenario for the register CLI), `_get_or_create_agent` will fail with an FK `IntegrityError` on `project_id`.

https://github.com/HexamindOrganisation/coolagents/blob/3dc8d50a0589d0bf5126c3241ec2d42e9c60f773/platform/api/main.py#L242-L257

quanghexa94:
Design's right - content-hashed versioning + bearer→project upsert is clean, and the per-framework adapter split mirrors `adapters/` nicely. But two blockers before this can merge:

1. **Tests don't run.** `tests/register/manifest.py` is missing the `test_` prefix, so pytest discovery skips the whole file. Rename to `test_manifest.py` - the 6 framework tests have never actually executed.

2. **`fortify register` requires all 4 framework SDKs installed.** [`register/models.py`](fortify/cli/register/models.py) and [`manifest.py`](fortify/cli/register/manifest.py) do top-level `from agents import …`, `from google.adk … `, `from langgraph …`, `from pydantic_ai …`. A user registering an OpenAI agent gets `ModuleNotFoundError: google.adk`. The existing `adapters/` keeps these per-file for a reason - please match. Same fix for [`register/__init__.py`](fortify/cli/register/__init__.py), which eagerly pulls all five adapters. (just a reminder, no matter what we will revamp how the dependency works so that user no need to install all 5 sdks)

Other things worth addressing:
- New `requests` dep (add to pyproject.toml?) or - rest of fortify uses `httpx` / stdlib `urllib.request`. And no `timeout=` on the POST.
- `load_dotenv()` inside `register_agent()` - surprising side effect when used as a library; move to the CLI entry.
- `models.py` defines `AgentManifest` before `ToolDefinition` - works via Pydantic rebuild but fragile, please reorder.
- `langchain.py` silently falls back to `"langchain_agent"` for unnamed graphs - should raise like `fortify.py` / `pydantic_ai.py` do.
- `tests/.../manifest.py:200` constructs `CoolAgent` with raw kwargs - couples to the constructor; go through `create_agent(...)`.

Nits: `_load_agent` is a redundant alias of `_load_spec`; consider sorting tools by name in `compute_manifest_hash` for strict idempotency; the schema duplication between SDK and platform deserves a CI equality test.

Once #1 and #2 are fixed the rest is mechanical. Nice work overall.

quanghexa94 (approved):

PR #8, opened 2026-05-06
feat: add biscuit token logic + repo restructure

quanghexa94:


guillaume-hexamind:
### Code review

Found 2 issues:

1. `load_fortify_agent` passes `tags=tags or ["fortify", "fortify", config.project_id]` — the string `"fortify"` is duplicated. The original tag list pre-rename was `["coolagents", "fortify", config.project_id]`; the `coolagents → fortify` mass rename collapsed the first two slots into the same value. Whatever was meant to occupy that slot (likely an env/category tag) is now silently dropped from Langfuse traces.

https://github.com/HexamindOrganisation/coolagents/blob/ca06e84679665e012c4d48cea8098c768f329333/fortify/agents/loader.py#L311-L315

2. `FortifyClient` fetches the platform's signing public key from `GET /v1/.well-known/keys` on first use (TOFU) and uses that fetched key to verify Biscuit signatures, with no pinning, no TLS requirement, and a default `base_url=http://localhost:8000`. Any attacker positioned on the network or who can influence `FORTIFY_API_URL` can serve their own JWKS plus a forged Biscuit, turning the verification step into a no-op. The docstring acknowledges this is "TOFU for POC" but the verification path is wired in with no mitigation, so the POC default ships an exploitable signature-verification bypass for users who don't set `FORTIFY_PUBLIC_KEY`.

https://github.com/HexamindOrganisation/coolagents/blob/ca06e84679665e012c4d48cea8098c768f329333/fortify/cloud/client.py#L160-L195

guillaume-hexamind:
Score 75:                                                                                                                                                                    
  - _verified flag persists across config changes (fortify/cloud/client.py) — self._verified = True is set once and never re-checked. If client.config.api_key is mutated after
   first verification, subsequent calls skip verification. Real but latent: requires caller to mutate config, no current call site does this.                                  
  - keystore.public_key_bytes() may run before ensure_keypair() (platform/api/main.py well_known_keys) — ensure_keypair() is only called inside lifespan. If the endpoint is   
  hit without lifespan (e.g. TestClient without with block), it raises RuntimeError("keystore not initialised"). Real edge case but uncommon in production.                    
  - Missing ensure_default_project on api_update_agent (platform/api/main.py) — PUT /v1/projects/{project_id}/agents/{name} doesn't lazily create the project row, while       
  api_list_agents/api_get_agent do. Pre-existing — not introduced by this PR; flagged for consistency with PR #7's review but out of scope here.                        
  - Auth not enforced on mutation endpoints (platform/api/main.py) — optional_dev_token is wired only to GET /agents/{name}. POST /tokens, DELETE /tokens, PUT /agents/{name}  
  accept no auth at all. The optional_dev_token docstring describes this as deliberate POC behavior ("Tighten to required once dashboard auth lands"), so this looks like an
  intentional concession, not an oversight.                                                                                                                                    
                                                                          
  Score 50:                                                                                                                                                                    
  - authorize_token rejects all tokens when policies="" (platform/api/biscuits.py) — Biscuit authorizers default-deny without an allow if rule. The function doesn't validate
  that policies is non-empty, so a caller passing "" would get TokenAuthorizationError for valid tokens. Real but no current caller passes empty policies, and the parameter is
   required (no default).                                                                                                                                                      
  - keystore._private_key_bytes() called from application code (platform/api/main.py) — mint_dev_token(..., signing_key_bytes=keystore._private_key_bytes()) reaches into a    
  leading-underscore method whose own docstring warns: "A future KMSKeyStore would not implement this method; callers that target multiple keystore backends should use sign()
  instead." Real architectural debt, but works today since FileKeyStore is the only backend.

fortify/agents/loader.py
```diff
@@ -266,6 +268,54 @@ def resolve_agent_source(name: str, base_dir: str | Path | None = None) -> Agent
...
+    project_id: str | None = None,
+    base_url: str | None = None,
+    api_key: str | None = None,
+    session_id: str | None = None,
+    user_id: str | None = None,
+    tags: list[str] | None = None,
+    extra_tools: Mapping[str, Any] | None = None,
+    model: str | None = None,
+) -> tuple[AgentGraph, CallbackHandler]:
+    """Fetch an agent from Fortify and return it with policy enforcement applied.
```
guillaume-hexamind, line 283: Good for current demo, but reconstructing an agent from platform yaml config is a logic that will need to be removed later!

fortify/cloud/biscuit.py
```diff
@@ -0,0 +1,78 @@
...
+class TokenError(RuntimeError):
+    """Base class for token-related failures."""
+
+
+class TokenSignatureError(TokenError):
+    """Raised when a Biscuit's signature does not chain to the expected key."""
+
+
+def parse_envelope(envelope: str) -> tuple[str, str, str]:
+    """Parse ``fty_<env>_<project>_<biscuit_b64>`` into ``(env, project, biscuit_b64)``.
```
guillaume-hexamind, line 37: very important to enforce no underscore in our project ids if we wanna keep this pattern

quanghexa94:
1. Fixed — went with ["fortify", "fortify-cloud", project_id] to keep the original 3-slot intent (SDK / source / project). Pre-dates this PR, sed in #4 collapsed it. Happy to drop to ["fortify", project_id] if you'd rather lose the source tag.

2. Acknowledged, called out as POC in the docstring. Surface is narrow today: localhost default, FORTIFY_PUBLIC_KEY env var for self-hosters, build-time pinned pubkey planned for hosted Fortify Cloud. Tracking with the rest of the auth-hardening work — opening a follow-up issue.

3. Thanks for the deep read. Triage:

_verified mutation risk — theoretical (dataclass not mutated). Will add frozen=True.
public_key_bytes() race — not reachable; lifespan calls ensure_keypair() before serving. Can add an assert if you want.
ensure_default_project on PUT — real gap, fixing with mutation auth.
Auth on mutation endpoints — intentional POC, lands with user-token attenuation.
authorize_token empty policies — correct default-deny, but will raise ValueError to fail loud.
_private_key_bytes() exposure — refactoring to keystore.sign() when KMS lands.

platform/api/main.py
```diff
@@ -84,6 +115,31 @@ def v1_health() -> dict[str, str]:
     return {"status": "ok", "service": "fortify-api", "version": "v1"}
 
 
+@v1.get("/.well-known/keys")
```
guillaume-hexamind, line 118: Je n'avais d'avis forts sur le naming, mais voici ce que claude m'a répondu:

Mixed feelings — the path is doing two things at once and neither is quite right.                                                                                            
                                                                                                                                  
  What .well-known actually is. RFC 8615 reserves the /.well-known/ namespace for origin-level discovery URIs — the whole point is that a client can construct the URL from the
   origin alone, no prior knowledge of your API layout. By nesting it under /v1/, you've defeated that property: a client now has to know your version scheme to find it, which
   is exactly what .well-known exists to avoid. Auth0, Okta, Google, AWS Cognito all serve /.well-known/jwks.json and /.well-known/openid-configuration straight off the       
  origin, never under a version prefix.                                                                                                                                        
                                                                                                                                                                               
  The suffix keys isn't a thing. The IANA-registered conventions are:                                                                                                          
  - /.well-known/jwks.json — JSON Web Key Set (the universal JWT public-key endpoint)
  - /.well-known/openid-configuration — OIDC discovery doc that points to a jwks_uri                                                                                           
                                                                                    
  Plain keys isn't registered with IANA and doesn't match anything tooling expects. So you get the cost of looking like a standard endpoint without the benefit that any       
  standard client can talk to it.                                                                                                                                              
                                                                                                                                                                               
  The response isn't actually JWKS-shaped. Looking at the handler, it returns:                                                                                                 
  {"kty": "OKP", "crv": "Ed25519", "x": ...}                              
  That's a single JWK, not a JWK Set. RFC 7517 requires {"keys": [...]} and a kid (key id) per entry — without kid, clients can't do key rotation. The x value should also be  
  unpadded base64url per RFC 7515 §2 (your urlsafe_b64encode includes padding by default). So a real JWKS client would fail to parse it.                                       
                                                                                                                                                                               
  There's also a protocol mismatch. JWKS exists for JWT verification keys. Biscuit isn't JWT — different signature framing, different verification flow. Naming a              
  biscuit-verification endpoint after the JWKS convention invites integrations that try to plug it into JWT libraries and fail in confusing ways.                              
                                                                          
  Two clean options:                                                                                                                                                           
                                                                          
  A — commit to the convention. Move it to /.well-known/jwks.json mounted at the app level (not under the v1 router), and emit a real JWKS:                                    
  {"keys": [{"kty": "OKP", "crv": "Ed25519", "kid": "sha256:abc123…", "x": "<unpadded b64url>"}]}
  The kid is the keystore fingerprint you already compute. This way any JWKS client library can fetch and cache it for the key fetch step, even though biscuit verification    
  itself stays custom. Cleanest path forward.                                                                                                                                  
                                                                                                                                                                               
  B — drop the .well-known masquerade. Call it what it is: /v1/auth/public-keys or /v1/biscuit/keys. Explicit, version-controlled, and you're not creating false expectations. 
  Use this if you don't expect anyone to plug a JWKS library against it.                                                                                                       
                                                                          
  The current name is the worst of both — it claims discovery semantics it doesn't satisfy, claims a format it doesn't match, and lives at a path where discovery clients won't
   look. I'd go with A: it's a small change (move the route, wrap in {"keys": [...]}, add kid, drop padding) and you get real ecosystem compatibility for the key-fetch half of
   the flow.

platform/api/biscuits.py
```diff
@@ -0,0 +1,269 @@
...
+        f'project("{_escape(request.project_id)}")',
+        f'token_id("{_escape(request.token_id)}")',
+        f'name("{_escape(request.name)}")',
+        f'env("{_escape(request.env)}")',
+        f"issued_at({_datalog_datetime(issued_at)})",
+    ]
+    facts.extend(f'scope("{_escape(s)}")' for s in request.scopes)
+
+    builder_source = "; ".join(facts) + ";"
+    if request.ttl_seconds is not None:
```
guillaume-hexamind, line 114: edge case, but maybe add 'and request.ttl_seconds > 0' ? Or enforce only positive integers on the MintRequest

guillaume-hexamind (commented):
A few minor comments but ok with the logic!

My main issue would come from injecting the project_id into the api_key and letting the sdk resolve it to contact the platform /project/:project_id endpoint -> some security and maintanability issue (hard to migrate our db or change the api_key pattern later), and using /project/:project_id/agents when the project_id information is already contained within the bearer token is redundant

PR #11, opened 2026-05-18
feat: fact extraction and revamp ui

quanghexa94:
adds M1 — user attenuation + role-aware policies + dashboard

what's in:

- `User` async context manager — `async with User(role="billing"):` scopes an agent run, runtime lazily mints a per-request biscuit
- per-role policies inline in `policy.yaml` (`inherits:`, `is_mixin:`) — runtime picks the role's policy based on the active User scope
- Rego-shaped string constraints over `args.*` (`==`, `<=`, `in`, etc.) — will compile to Rego when M2 lands
- `/policies` dashboard page: agent dropdown + YAML tab + node-edge graph (role × tool, mode-colored edges, dashed inheritance arrows)
- `/agents` slimmed to a read-only manifest view; policy editing now lives only in `/policies`
- playground "Acting as" dropdown drives per-turn attenuation through `fortify --serve`
- `refund_order` demo tool + seeded `support_bot` agent with `read_only` / `default` / `support` / `billing` roles so the policy story is concrete
- `/v1/projects/{p}/agents/{n}/validate` runs the same YAML + constraint parsers the SDK enforces with
- cleanup: `CoolAgent` → `FortifyAgent`, `COOLAGENTS_DEFAULT_*` → `FORTIFY_DEFAULT_*`

heads-up:

- wire format collapses from 4 fields to 3: `{agent_yaml, policy_yaml, system_md}`. Roles are inline in `policy.yaml`, the `roles_json` column is gone — existing dev DBs need `rm platform/api/fortify.db` to pick up the new shape.
- single-canonical-engine design — today's pydantic constraint evaluator is the no-OPA fallback. M2's OPA WASM adapter drops into the same `GuardedTool` contract without rewriting the runtime.
- same Ed25519 root that signs biscuit tokens will sign policy bundles in M2 — one trust chain.

tests: SDK 470 / platform 46 / dashboard tsc — all green.

next: M2 — Rego/WASM adapter + signed policy bundles. tracking issue to follow.


guillaume-hexamind:
### Code review

Found 1 issue:

1. In `_resolve_inheritance`, a child role that explicitly declares `default_policy: { mode: deny }` (same shape as `BaseToolPolicy()`'s default) is silently overridden by the parent's `default_policy`. The guard `if own.default_policy != BaseToolPolicy()` can't distinguish "child set the explicit default" from "child left it unset", so the child inherits whatever the parent had — e.g. a permissive `allow` — even though the child stated `deny`. Since `BaseToolPolicy()` defaults to `mode: "deny"` (see [models.py L23](https://github.com/HexamindOrganisation/coolagents/blob/bdddd6aa1c052b271250a746106b42d06b3dfbda/fortify/security/models.py#L20-L25)), this directly contradicts the docstring's "self wins" precedence and is fail-open. Consider tracking whether `default_policy` was explicitly provided (e.g. Pydantic's `model_fields_set`) rather than comparing against a freshly constructed default.

https://github.com/HexamindOrganisation/coolagents/blob/bdddd6aa1c052b271250a746106b42d06b3dfbda/fortify/security/policy_set.py#L234-L241

fortify/cloud/biscuit.py
```diff
@@ -76,3 +104,66 @@ def verify_biscuit(token_b64: str, public_key_bytes: bytes) -> None:
...
+            if stripped.startswith("check ") or "<-" in stripped:
+                # checks and rules — biscuit's authorizer enforces these;
+                # the SDK policy engine only wants ground facts.
+                continue
+            match = _FACT_LINE_RE.match(line)
+            if match is None:
+                continue
+            name = match.group("name")
+            if match.group("str") is not None:
+                value: str | int = match.group("str").encode("utf-8").decode(
```
guillaume-hexamind: unicode_escape interprets UTF-8 bytes as Latin-1 escape sequences, so "café" will become
  cafÃ© -> maybe replace \" with " and \\ with \ instead, or use biscuit library own parser?

fortify/runtime/context.py
```diff
@@ -47,3 +74,67 @@ def set_current_tool_use_context(
 def reset_current_tool_use_context(token: Token[ToolUseContext | None]) -> None:
     """Restore the previous tool-use context after a run completes."""
     _CURRENT_TOOL_USE_CONTEXT.reset(token)
+
+
+class User(BaseModel):
```
guillaume-hexamind, line 79: Other possibility: move the contextvar lifecycle off `User` and into the runtime entry points.

  Today `User` doubles as data + async context manager: `__aenter__/__aexit__` set
  `_CURRENT_USER`, and a private `_tokens` stack handles nested re-entry. This
  forces `User` to be async-only, makes the same instance unsafe to share across
  concurrent tasks (the stack would interleave), and ties identity propagation to
  a caller-managed `async with` block.

  Alternative — `User` becomes plain data; the runtime owns the scope:

      async def stream_agent(agent, handler, input, *, user: User | None = None):
          token = _CURRENT_USER.set(user) if user is not None else None
          try:
              ...
          finally:
              if token is not None:
                  _CURRENT_USER.reset(token)

  Wins:
  - Deletes `__aenter__/__aexit__` and the `_tokens` stack.
  - `User` works in sync contexts (no async-only constraint).
  - Per-call scope, so no shared-instance footgun across concurrent tasks.
  - Downstream `get_current_user()` callers (policy engine, tool meta-injection)
    are unchanged — the contextvar still carries identity ambiently.

  Tradeoff:
  - Callers pass `user=` per agent call instead of bracketing many calls in one
    `async with User(...)`. Natural for FastAPI per-request flows; one extra kwarg
    for batch scripts.

fortify/tools/refund.py
```diff
@@ -0,0 +1,49 @@
+"""Demo refund tool — a stub the support-bot playground uses to exercise role policies.
```
guillaume-hexamind, line 1: Useful for current demos, but still think this file, and other demo logic will need to be moved in examples/ folder

platform/api/main.py
```diff
@@ -230,23 +233,24 @@ def revoke_token(
         raise HTTPException(status_code=404, detail="token not found")
 
 
+def _agent_read(agent) -> AgentRead:  # noqa: ANN001 — Agent model not imported at top
```
guillaume-hexamind: Missing import Agent from models here

guillaume-hexamind (commented):
A couple of comments above to check, but the structure and core logic look good for a first version

I still really wonder as explained in the comment which between propagating user context through a context manager, or through an additional argument is the best approach, both for simplicity of use and simplicity of logic

Some next steps I would see:

- Check how to unify the logic so that policy enforcement works for adapter framework agents. Right now the logic is quite coupled with langchain BaseTool. I think a refactor to obtain a tool-shape agnostic PolicyEnforcer gate class is possible if done smartly, returning a typed decision; enabling to have adapters for the wrapping logic based off the decision for each framework + our FortifyOne, I will look into that
- Move multiple cli serve and chat logic into /cli/ (loaders, bootstrap, builtin) will be a good way to divide the actual sdk facing core logic from cli logic
- Move multiple demo and examples logic into examples/ would be a good refactor too

fortify/runtime/context.py
quanghexa94: To think carefully and worth another PR. Two approaches are defensible

fortify/tools/refund.py
quanghexa94: will later revamp the demo structure in another PR

platform/api/main.py
quanghexa94: solved

PR #14, opened 2026-05-21
feat(fortify): new policy enforcer logic

guillaume-hexamind:
New core
  - fortify/security/decision.py — Decision dataclass with ALLOW / DENY / NEEDS_APPROVAL outcomes, plus as_error_payload() (dict) and as_error_message() (string
  with marker).
  - fortify/security/enforcer.py — stateless PolicyEnforcer.decide() reads User.role from contextvar, runs authorize_tool_call, returns a typed Decision. Never
  executes.

  Old module deleted
  - fortify/agents/security.py (−378 lines): the previous GuardedTool, _wrap_tool, wrap_tools_with_*, and the with_before_action / with_approval_handler standalone
   transforms are gone.
   
  Per-adapter wrapping, same Decision flow
  - LangChain: GuardedTool (for SDK-built graphs) + install_enforcer_on_tool (in-place, for BYO CompiledStateGraph). Renders non-allow as a dict.
  - Google ADK: wrap_tool copies + replaces run_async. Renders as string.
  - OpenAI Agents: same pattern, replaces on_invoke_tool. Renders as string.
  - Pydantic AI: copies tool + function_schema, replaces function_schema.call. Raises ModelRetry.

  User scope unified
  - User.sync_scope() added for sync entry points (CLI, run_sync).
  - All four adapter proxies (FortifyLangchainAgent, FortifyRunner ×2, FortifyPydanticAgent) take user= per call and open/close the scope around the framework
  call.
  - Redundant policy.py files in adapters/langchain and adapters/pydantic_ai removed.
  
  API consolidation
  - FortifyAgent.enforce_policy(policy, *, approval_handler=None) is the single entry point. with_approval_handler / with_before_action standalone functions
  removed.
  - All load_*_agent loaders forward approval_handler; load_registered_agent re-stamps existing GuardedTools via _apply_approval_handler.
  - CLI: build_approval_handler returns True / False for auto modes, (action, context) -> bool for ask; prompt_for_approval reads role from the contextvar.

  Tests
  - New tests/security/test_decision.py and tests/adapters/langchain/test_guarded_tool.py. Existing adapter tests rewritten around the Decision contract.

  ---
  Why the approval-handler refactor was needed
  
  The old API had three orthogonal transforms chained on the agent: enforce_policy(p).with_approval_handler(h).with_before_action(b). Each call rebuilt the
  GuardedTools and the previous wrap's fields had to be hand-preserved through an if isinstance(tool, GuardedTool) branch. That created four real problems:

  - Composition order silently mattered, and any miss in the unwrap path silently dropped hooks (exactly the bug we just fixed in _apply_approval_handler).
  - Only LangChain had a GuardedTool — the other three adapters did ad-hoc wrapping that bypassed the security module, so with_approval_handler was effectively
  LangChain-only despite looking generic.
  - Approval was modeled as an exception, not a value, so every adapter had to repeat the same try/except dispatch, and tests had to assert on exception flow
  rather than data.
  - The CLI's --approval-mode couldn't ship one callable that works for every framework without knowing each adapter's wrapping.

  The fix: collapse to one Decision-returning enforcer + one approval_handler field, both living on the same per-adapter wrapper. Adapters only have to render
  three outcomes into their host's tool-result idiom. DENY short-circuits without consulting the handler (semantically correct — "deny" means never, not "maybe
  ask"). NEEDS_APPROVAL is the only path that asks the handler. The handler signature stays (action, context) -> bool | Awaitable[bool] so existing callers keep
  working; the CLI's ask mode produces exactly that shape.

  Net result: one diagram works for all four adapters, idempotent re-wrap is a single branch in GuardedTool.wrap, and the CLI ships one handler that lights up
  everywhere.


guillaume-hexamind:
### Code review

Found 1 issue:

1. The CLI's `ask`-mode approval handler signature does not match the resolver call site, which will raise `TypeError` at runtime on every `NEEDS_APPROVAL` outcome.

`build_approval_handler` returns a closure taking a single positional arg (`decision: Decision`):

https://github.com/HexamindOrganisation/coolagents/blob/a1dc466840ab21c74c9c1dae0629c1fd5575d28a/fortify/cli/_common.py#L149-L154

But `_resolve_approval_sync` / `_resolve_approval_async` invoke the handler with two positional args (`action`, `context`), called from `GuardedTool._arun` / `_run` as `_resolve_approval_*(self.approval_handler, self._build_action(kwargs), self._build_context())`:

https://github.com/HexamindOrganisation/coolagents/blob/a1dc466840ab21c74c9c1dae0629c1fd5575d28a/fortify/adapters/langchain/tools.py#L42-L66

Any `NEEDS_APPROVAL` triggered while running `fortify chat`/`serve` with `--approval-mode=ask` will fail with `TypeError: approval_handler() takes 1 positional argument but 2 were given`. Auto-approve/auto-deny paths short-circuit via the `isinstance(handler, bool)` check and are unaffected.

quanghexa94:
LGTM. The `PolicyEngine` protocol is cleaner than the `isinstance` dispatcher we had. And the WASM / `PolicyBundle` path is preserved end to end: `PolicyBundle.evaluate` plugs into the protocol, `violations` flow through `Verdict` → `Decision` → `as_error_payload`, and the loader bits (`_local_policy_override`, `_platform_bundle`, signature verification) are intact.

one small things to confirm before I tick approve:
- the `as_error_payload` flatten (no more `{ok: false, error: {...}}` outer wrapper) - anything downstream still parsing on those keys? Worth a sanity check on the dashboard playground rendering and the serve-mode JSON relay.

guillaume-hexamind:
> LGTM. The `PolicyEngine` protocol is cleaner than the `isinstance` dispatcher we had. And the WASM / `PolicyBundle` path is preserved end to end: `PolicyBundle.evaluate` plugs into the protocol, `violations` flow through `Verdict` → `Decision` → `as_error_payload`, and the loader bits (`_local_policy_override`, `_platform_bundle`, signature verification) are intact.
> 
> one small things to confirm before I tick approve:
> 
> * the `as_error_payload` flatten (no more `{ok: false, error: {...}}` outer wrapper) - anything downstream still parsing on those keys? Worth a sanity check on the dashboard playground rendering and the serve-mode JSON relay.

as_error_messag is only returned as output for the adapter logics; the consumer of`{ok: false, error: {...}}` outer wrapper will still work with this PR -> {"ok": False, "error": decision.as_error_payload()}

PR #15, opened 2026-05-22
refactor(fortify): remove before action logic, 

guillaume-hexamind:
Removed
  - FortifyAgent.with_before_action(...) and its module-level shim.
  - Type aliases BeforeActionHook, ActionPayload, ActionContext, ContextProvider — their only consumer was Gate 2.
  - The before_action, context_provider, agent_name fields on GuardedTool plus the four helpers feeding them (_build_action, _build_context, _before_action_denied,
   _check_before_action).
  - The before_action requires async RuntimeError branch in GuardedTool._run (dead with the hook gone).
  
  Tightened
  - ApprovalHandler collapses from (ActionPayload, ActionContext) -> bool | Awaitable[bool] to (Decision) -> bool | Awaitable[bool]. _resolve_approval_{sync,async}
   take the Decision directly.
  - GuardedTool.wrap() keeps idempotent re-wrap but only has enforcer and approval_handler left to preserve.
  - prompt_for_approval(console, decision) reads tool_name, arguments, role straight off the Decision — no more get_current_user() lookup or (action, context)
  shape juggling.
  
  Tests — deleted the Gate 2 suites, updated approval-handler tests to assert the Decision-shaped callable.

----------------------------------------------------------------------------------------------------------------

  Why before_action is better off removed

  - Zero in-tree consumers. Not the CLI, not the platform, not the examples — only the hook's own dedicated tests. Shipping a public contract for users who don't
  exist means every future refactor has to keep proving it works.
  - Redundant with Gate 1 and the tool body. Anything the hook could do, a developer can do inside the tool or via a richer Gate 1 constraint. The one "unique" use
   case — a host injecting checks into a developer's tool without consent — is a hosting concern, not an SDK one.
  - Awkward contract. Vetoing meant raising; the awaited return value was discarded silently. Two-way semantics ("raise to veto, but we also await your coroutine
  for no observable reason") tripped up even the code review on the prior branch.
  - Undefined composition with Gate 1. Chaining with_before_action after enforce_policy physically stacked a second GuardedTool over the first with empty policy
  fields — "whichever runs first wins" is not a contract.
  - Cascading simplification. Removing the hook deletes its four supporting type aliases and lets ApprovalHandler drop to a single-argument (Decision) -> bool —
  the same value Gate 1 already produces. Net ~250 LOC of SDK surface gone, one fewer concept to reason about, zero loss of expressiveness: pre-execution checks
  still live inside the tool body or in a host-side wrapper outside the SDK.


quanghexa94:
LGTM. Try to run `make test` and `make platform-api-test` to make sure every test passes

PR #17, opened 2026-05-26
feat: add rego wasm

quanghexa94:
## What

Adds the WASM enforcement pipeline for agent policies (milestone M2). Policies are still authored as `policy.yaml`, but they now compile to a signed WASM bundle that the runtime evaluates — with the existing pydantic engine kept as a fallback.

The full path: author YAML → platform compiles to Rego → WASM via OPA → signs the bundle with the root key → SDK pulls it, verifies the signature against the same key it already trusts for biscuit tokens → enforces via wasmtime. There's a `FORTIFY_LOCAL_POLICY` escape hatch so you can build + test bundles locally without the platform in the loop.

## Highlights

- `fortify policy {build,validate,show-rego,test,keygen}` CLI for local iteration
- Two engines, identical decisions (parity-tested). WASM buys a portable signed artifact + structured deny reasons (the exact constraint strings from your YAML)
- One root key signs both tokens and bundles — no new key distribution
- Ed25519 detached signatures; warn-by-default, refuse on `FORTIFY_BUNDLE_REQUIRE_SIGNATURE`

## Testing

- SDK 639 passing, platform 64 (1 pre-existing unrelated failure)
- Cross-seam parity test: platform-compiled bundles decide identically to the pydantic engine
- Verified the live platform demo end to end (save a policy → it compiles + signs → agent pulls + verifies + enforces)

## Notes for reviewers

- Needs `opa` on PATH wherever bundles compile (platform at save time, or local `fortify policy build`). Missing opa degrades gracefully to pydantic.
- Only the LangChain adapter dispatches through WASM today; OpenAI/Google/Pydantic-AI still use pydantic — follow-up.
- Adds 3 columns to the agent table; wipe `fortify.db` (no prod data yet).
- Bumped the control plane's `requires-python` to `>=3.13` to match the SDK.


guillaume-hexamind:
### Code review

platform/api/models.py
```diff
@@ -43,6 +43,20 @@ class Agent(SQLModel, table=True):
     system_md: str = ""
     updated_at: datetime = Field(default_factory=utcnow)
 
+    # Compiled + signed WASM bundle, produced from policy_yaml at save time
+    # (see services.compile_bundle). Null when opa is unavailable or the
+    # policy fails to compile — the SDK then falls back to the pydantic
+    # engine on policy_yaml. The signature is over bundle_manifest's exact
+    # bytes, signed by the platform's root key (the same key that signs
+    # biscuits), so the SDK verifies it against the published JWKS pubkey.
+    compiled_wasm: Optional[bytes] = Field(
```
guillaume-hexamind, line 52: I don't know if it should be in this PR, but we'll definitely need to create separate tables Policy or Bundle to separate responsabilities, especially since we will want to have multiple yaml per role for a single agent

Mutliple ways to implement this, a single Policy table with increasing int version and one way relation to agent_id could do the trick

fortify/security/rego_wasm.py
```diff
@@ -0,0 +1,213 @@
...
+
+This is the second half of the M2 compile pipeline: Phase 1 emits Rego
+text (``fortify.security.rego``), Phase 3 turns that text into a
+self-contained ``policy.wasm`` blob the runtime can evaluate without
+embedding an interpreter.
+
+The implementation shells out to ``opa build -t wasm`` rather than
+linking against an opa library — opa is the canonical reference
+implementation and pinning to a release line gives us a clean spec to
+test against. Discovery checks PATH first, then ``$FORTIFY_OPA_BIN``;
```
guillaume-hexamind: _discover_opa() does the other way around: FORTIFY_OPA_BIN check, then path -> docstring to change

fortify/cli/policy/main.py
```diff
@@ -0,0 +1,546 @@
+"""`fortify policy` subcommand — author + inspect + dry-run policy documents.
+
+Wraps the M2 phase 1 compiler library and the existing pydantic engine in
+a four-verb CLI: ``build``, ``validate``, ``show-rego``, ``test``. Every
+verb is a thin wrapper — the heavy lifting lives in
+:mod:`fortify.security`. That symmetry lets the platform's save flow use
+the same code without duplication.
+
+Phase 3 will extend ``build`` with the ``opa build -t wasm`` step;
+Phase 4 will extend ``test`` to evaluate against the compiled WASM
```
guillaume-hexamind: Both of these seem to already be implemented, simple docstring change

fortify/cli/policy/main.py
```diff
@@ -0,0 +1,546 @@
...
+
+    parser.set_defaults(func=_main_unknown)
+
+
+def main(args: argparse.Namespace) -> int:
+    """Entry point used by the top-level dispatcher in fortify/cli/__init__.py."""
+    return args.func(args)
+
+
+def _main_unknown(args: argparse.Namespace) -> int:
```
guillaume-hexamind: dead code because of the subparser's required=True; probably better to remove either the code or the required=True

platform/api/services.py
```diff
@@ -102,6 +107,61 @@ def get_latest_agent_versions_map(
...
+    manifest = {
+        "version": 1,
+        "source": "policy.yaml",
+        "source_hash": source_hash,
+        "rego_hash": hashlib.sha256(rego.encode("utf-8")).hexdigest(),
+        "wasm_hash": hashlib.sha256(wasm).hexdigest(),
+    }
+    # Serialize with the SAME convention as `fortify policy build` so the
+    # bytes are reproducible; sign these exact bytes.
+    manifest_text = json.dumps(manifest, indent=2, sort_keys=True) + "\n"
```
guillaume-hexamind: Seems dangerous to me to duplicate same logic in 2 different places, expecting them to stay in sync -> I think a refactor to add a fortify/security/bundle.py including a dataclass

@dataclass(frozen=True)
  class SignedBundle:
      """The artifacts `fortify policy build` and the platform both emit.

      ``manifest_bytes`` are the *exact* bytes the signature covers — callers
      must persist/transmit these verbatim, never a re-serialization of
      ``manifest`` (that's what keeps verification canonicalization-free, the
      same invariant :class:`PolicyBundle` relies on).
      """

      rego_text: str
      wasm_bytes: bytes | None          # None when compile_wasm=False (--no-wasm)
      manifest: dict
      manifest_bytes: bytes
      signature: bytes | None           # None when no `sign` callback was given

      @property
      def source_hash(self) -> str | None:
          return self.manifest.get("source_hash")

      @property
      def wasm_hash(self) -> str | None:
          return self.manifest.get("wasm_hash")

And an exported function:

def build_signed_bundle(
      policy_yaml: str,
      *,
      source_name: str = "policy.yaml",
      sign: Callable[[bytes], bytes] | None = None,
      compile_wasm: bool = True,
      opa_bin: str | None = None,
  ) -> SignedBundle:

Would allow to not repeat the logic, and to import build_signed_bundle in both this method, and in sdk cli _main_build's manifest block

fortify/security/wasm_engine.py
```diff
@@ -0,0 +1,381 @@
...
+        """Lower-level: evaluate with an arbitrary input dict.
+
+        Mostly useful for tests that want to probe odd shapes. Production
+        callers should use :meth:`decide`.
+        """
+        store = self._store
+        mem = self._memory
+        exports = self._exports
+
+        # Reset to the base heap pointer so this call doesn't leak into the next.
```
guillaume-hexamind: Seems safer to force a threading.Lock wrap of the evaluation sequence: wrap_tools_with_policy attaches the same PolicyBundle in every tool -> every guarded tool evaluates through one store, race condition posssible for parallel tool calls

guillaume-hexamind (commented):
Some comments to address, but I approve of the overall rego and wasp enforcement logic; 

let's discuss how we dispatch the next steps so we do not create unwanted conflict!

platform/api/models.py
quanghexa94: Agreed. We will refine the db table models schema in another PR. For now it's simply for the prototype and the db is not even postgres yet, just SQLite

PR #18, opened 2026-05-28
feat(platform): clickhouse schema and deployment

guillaume-hexamind:
Lays the foundation for the policy-decision audit log: a local ClickHouse service plus the `policy_decision` table. Pure infrastructure — no
  Python or SDK changes; the ingest endpoint and SDK sender land in follow-up PRs.

  - `platform/docker-compose.yml` — ClickHouse 24.10 with healthcheck; binds host ports **8124/9001** (offset from defaults to coexist with other
   local ClickHouse instances such as a Langfuse-bundled one).
  - `platform/clickhouse/init/schema.sql` — `policy_decision` table: 8-column shared envelope (event_id, occurred_at, received_at, project_id,
  agent_name, agent_version_id, session_id, user_id) + decision-specific columns, `MergeTree` partitioned monthly, 90-day TTL.
  - `Makefile` — `clickhouse-{up,down,logs,cli,reset}` targets.
  - `README.md` — setup section, with an explicit note that the init dir is **not** a migration system (runs once on an empty volume; future 
  schema changes need a real runner).
  
  The envelope columns are intentionally shaped so future audit event tables (`tool_invocation`, `tool_completion`, ...) can reuse the same 
  prefix and keep cross-event correlation cheap.


guillaume-hexamind:
### Code review

quanghexa94:
Claude review: 
Tight, well-shaped first PR - the commit message explains the "why," the init-script-not-a-migration-runner caveat is called out proactively in both the README and the SQL header, and the envelope-columns design clearly leaves room for `tool_invocation` to slot in cleanly. Findings below, ordered by what I'd hold the merge on vs. what's fine as a follow-up.

---

### 1. Compose binds ClickHouse to `0.0.0.0` with a known dev password

`platform/docker-compose.yml`:
```yaml
ports:
  - "8124:8123"
  - "9001:9000"
```
Docker's default expands those short-form mappings to 0.0.0.0:8124->8123, so the moment a dev runs make clickhouse-up on a coffee-shop wifi the rest of the network can reach the HTTP endpoint and try fortify:fortify-dev-password — which is committed in the same file. With CLICKHOUSE_DEFAULT_ACCESS_MANAGEMENT: 1 enabled, that's full SQL access on success.

One-line fix:


ports:
  - "127.0.0.1:8124:8123"
  - "127.0.0.1:9001:9000"
2. TTL + PARTITION BY use occurred_at (client-supplied)
platform/clickhouse/init/schema.sql:


PARTITION BY toYYYYMM(occurred_at)
ORDER BY (project_id, agent_name, outcome, occurred_at)
TTL toDateTime(occurred_at) + INTERVAL 90 DAY
occurred_at comes from the SDK. A buggy clock (or hostile actor) submitting 1970-01-01 gets the row TTL'd on the next merge; a 2099 timestamp keeps it forever and blows up the future partition fan-out. The whole point of received_at (server-stamped, DEFAULT now64(3)) is to be the trustworthy anchor — switch the TTL and partition key to it:


PARTITION BY toYYYYMM(received_at)
ORDER BY (project_id, agent_name, outcome, occurred_at)
TTL toDateTime(received_at) + INTERVAL 90 DAY
(Keeping occurred_at in the ORDER BY is fine — within-partition time scans still want the event-time ordering.)

## 🟡 Worth a thread, not a blocker
3. Hardcoded credentials in the compose file
platform/docker-compose.yml. The "LOCAL DEV ONLY" comment is the right move, but the conventional dev-compose idiom keeps it overridable so the same file works for prod:


CLICKHOUSE_PASSWORD: ${FORTIFY_CLICKHOUSE_PASSWORD:-fortify-dev-password}
4. Dedup semantics for SDK retries
Once the audit SDK lands, network blips will cause the producer to retry. Plain MergeTree keeps both copies. The two clean options:

ReplacingMergeTree(received_at) with event_id in the ORDER BY — last-write-wins on background merges, but queries need FINAL (or accept eventual consistency) to see dedup'd rows.
Plain MergeTree + idempotent ingest — the ingest endpoint refuses duplicate event_id (needs an in-memory cache or OPTIMIZE ... DEDUPLICATE periodically).
Either is fine but the choice should be made now and documented — it affects how the ingest PR is written.

5. Port collision still possible
8124/9001 are hardcoded — two checkouts on the same dev box or parallel CI runs collide. Either env-override them (${FORTIFY_CLICKHOUSE_HTTP_PORT:-8124}:8123) or call it out as a known limitation. Low priority for a single-dev local setup.

## 🟢 Polish (nice-to-haves)
platform/docker-compose.yml — no restart: policy. unless-stopped would survive Docker Desktop restarts; for a dev box that's a nicer default than the current "container disappears after a crash you didn't notice."
platform/clickhouse/init/schema.sql — reason String has no codec; reasons tend to be longish and repetitive ("role X not in allowed roles Y"). Adding CODEC(ZSTD(3)) matches what hint and arguments already do.
platform/clickhouse/init/schema.sql — worth one line of comment: "no FK to platform Postgres on purpose — audit log preserves history even if the upstream project/agent rows are deleted." Saves the next reader a git blame.
platform/clickhouse/init/schema.sql — Enum8 with three outcomes will need an ALTER TABLE ... MODIFY COLUMN to extend. Worth a comment so the follow-up PR author knows the playbook.

## ✅ Things to keep
Envelope-first column ordering shared across future event tables → cross-table joins/correlations stay cheap.
The README's "Not a migration system" paragraph + the same warning in the SQL header → exactly the kind of land-mine documentation that pays off on month six.
Port-offset choice (8124/9001) so Langfuse-bundled CH doesn't fight this one.
LowCardinality(String) for project_id/agent_name/role/outcome is correct — those are bounded sets.
event_id as required (no DEFAULT) so the producer can't shove rows in without an id to dedup on later.

guillaume-hexamind:
Updated branch and fixed claude's findings 1 to 4, which seem to me like the only blocking ones

PR #19, opened 2026-05-29
feat(platform): add /v1/audit/decisions endpoint with ClickHouse insert

guillaume-hexamind:
Adds the platform-side audit ingest path: a single endpoint that accepts one policy decision per request, stamps server-resolved fields, and
  writes to the ClickHouse `policy_decision` table from PR 1.

  - `POST /v1/audit/decisions` — returns 202 + echoed `event_id`. `project_id` resolved from the bearer, `received_at` stamped by ClickHouse's
  column default, `agent_version_id` looked up from the relational store by `(project_id, agent_name)`. None are trusted from the body.
  - Three-layer split: `clickhouse.py` (infra: client + ping), `audit.py` (domain: row construction, byte caps, custom exceptions), `main.py`
  (HTTP: time-window validation, exception → status mapping).
  - `settings.py` introduced for `FORTIFY_CLICKHOUSE_*` config; defaults match the local docker-compose service.
  - `/health` and `/v1/health` now report `clickhouse: ok|unreachable`; startup logs a warning on unreachable rather than failing.
  - Bundled fixes: `make platform-api-install` corrected to `uv sync --group dev`

  ## Test plan

  - [x] `make platform-api-test` → 69 passed, 12 skipped (default skips integration)
  - [x] `pytest -m integration` with `make clickhouse-up` running → 1 passed (real ClickHouse round-trip)
  - [x] `curl GET /health` shows `clickhouse: ok`
  - [x] `curl POST /v1/audit/decisions` with a real bearer returns 202; `SELECT` confirms `project_id`, `received_at`, and `agent_version_id` are
   server-stamped


guillaume-hexamind:
### Code review

Found 3 issues:

1. The `/ready` and `/v1/ready` readiness endpoints return HTTP 200 even when ClickHouse is unreachable — `_readiness()` only sets `"clickhouse": "unreachable"` in the body while the status code stays 200. Kubernetes readiness probes key off the status code, not the body, so a pod whose ClickHouse is down stays in the load-balancer rotation and every `/v1/audit/decisions` write returns 503. This undercuts the liveness/readiness split introduced in `1c1f70f`. Consider returning 503 when `clickhouse_ping()` is false.

https://github.com/HexamindOrganisation/coolagents/blob/1c1f70f720495be24511dadcb8e7476c28ac51fd/platform/api/main.py#L164-L169

2. The integration-test skip guard uses exact string equality `config.getoption("-m") == "integration"`, which fails for any compound marker expression (e.g. `pytest -m "integration and not slow"`). In that case the guard falls through and integration tests are silently skipped even though the user explicitly selected them. A substring check (`"integration" in config.getoption("-m", default="")`) or `config.option.markexpr` is more robust.

https://github.com/HexamindOrganisation/coolagents/blob/1c1f70f720495be24511dadcb8e7476c28ac51fd/platform/api/tests/conftest.py#L7-L9

3. Stale comment: the lifespan startup comment says `/health surfaces it`, but after this PR's refactor `/health` is a dependency-free liveness probe and ClickHouse reachability is surfaced by `/ready` / `/v1/ready`. The comment should reference `/ready`.

https://github.com/HexamindOrganisation/coolagents/blob/1c1f70f720495be24511dadcb8e7476c28ac51fd/platform/api/main.py#L77-L79

quanghexa94:

### 1. `async_insert_deduplicate=1` is a no-op on this schema

`platform/api/audit.py`:
```python
_DECISION_INSERT_SETTINGS = {
    "async_insert":             1,
    "wait_for_async_insert":    1,
    "async_insert_deduplicate": 1,
}
```
Two reasons this setting provides no protection here:

It only applies to ReplicatedMergeTree — PR #1's schema is plain MergeTree. ClickHouse silently no-ops the flag on non-replicated tables.
Even on replicated tables it needs an insert_deduplication_token. Without one, dedup keys on the block contents; with single-row inserts each carrying a unique event_id, every block is unique and nothing ever dedups.
The setting reads like protection against SDK retry duplicates, but in practice if the SDK retries on a 503 → 202 race, you get two rows with the same event_id and the audit log silently double-counts.

Two clean fixes (pick one and document):

insert_deduplication_token=str(event.event_id) per insert — works even on plain MergeTree. Cheapest change.
Switch the table to ReplacingMergeTree(received_at) with event_id in the ORDER BY in PR #1 — last-write-wins on background merges, but queries need FINAL or accept eventual consistency.
Either way, the current async_insert_deduplicate=1 should either be removed (it's misleading) or paired with the per-row token.

### 🟡 Worth a thread, not a blocker
2. violations: list[str] has no byte cap
platform/api/schemas.py:


violations: list[str] = Field(default_factory=list, max_length=64)
64 items is the only limit. Each string is unbounded, so a hostile client can send 64 × 1MB strings (64 MB body) and get past Pydantic validation. hint and arguments are byte-capped in audit.insert_decision (MAX_HINT_BYTES, MAX_ARGS_BYTES) — violations should get the same treatment, either as a per-item Field(..., max_length=N) or a post-serialization byte cap.

3. Default password lives in the codebase as a prod footgun
platform/api/settings.py:


clickhouse_password: str = "fortify-dev-password"
A misconfigured prod that forgets FORTIFY_CLICKHOUSE_PASSWORD will silently start up against the (presumably absent) fortify-dev-password user — or, worse, against a ClickHouse instance that does accept it. Two safer defaults:

clickhouse_password: str | None = None and refuse to start with a clear error message when None in prod.
Or split the field: clickhouse_password: str = "" and have get_clickhouse() log a loud warning if it's empty when clickhouse_secure=True.
4. No idempotency story for event_id (companion to #1 above)
Same root cause, separate concern: even after #1 is resolved at the storage layer, the application layer should be explicit about what the SDK should do on retry. Worth one paragraph in the route docstring: "the SDK SHOULD/MAY retry with the same event_id; the ingest path is idempotent because X." Without it, a future SDK author will reinvent the wheel.

5. Time-window validation leaks into the HTTP handler
platform/api/main.py defines CLOCK_SKEW_FUTURE, RETENTION_WINDOW, and the occurred_at check inline in ingest_decision. Given the explicit three-layer split (clickhouse.py / audit.py / main.py), domain rules like "events older than retention" feel like they belong in audit.py as a validation function that raises a new exception class — same pattern as AuditPayloadTooLarge. main.py would map the exception to 400. Not a blocker, just an inconsistency given the cleanliness of the rest of the layering.

### 🟢 Polish
get_latest_agent_version_id returns "" for unknown agents — silent fallback rather than a 4xx. Probably the right call for an audit log (a logger that 404s on the agent it's logging about is broken) but worth a sentence in the route docstring so the next reader doesn't think it's a bug. No test currently covers this path either.
/ready uses response.status_code = 503 via injected Response — works at runtime but OpenAPI advertises 200. return JSONResponse(content=body, status_code=503) makes the doc honest.
Logger naming: clickhouse.py uses logging.getLogger(__name__) → "clickhouse". The rest of the codebase namespaces under fortify.platform.* (e.g. auth.py uses "fortify.platform.auth"). Tiny consistency thing.
No model-level invariant linking outcome="deny" to a non-empty error_type / reason — accepts {"outcome":"deny", "error_type":"", "reason":""} today. For a security audit log that's a strange shape to allow. A @model_validator(mode="after") would catch it.
agent_version_id lookup is one DB query per audit event — will need a (project_id, agent_name) → version_id cache once throughput matters. Fine for v1.

### ✅ Things to keep
Three-layer split (clickhouse.py infra / audit.py domain / main.py HTTP) with custom exceptions bubbling cleanly — this is the right shape.
Trust boundary made explicit in both code (# bearer-resolved, # platform-resolved comments on the row tuple) and tests (test_agent_version_id_comes_from_platform_lookup proves a body-injected agent_version_id is ignored).
Liveness vs readiness split with the dependency-cascade rationale spelled out in the docstring.
wait_for_async_insert=1 to surface failures synchronously — exactly what an audit log needs.
Error-code mapping is correct: transient (OperationalError) → 503 + Retry-After, permanent (DataError/ClickHouseError) → 422 with no retry header.
field_validator normalizing naive occurred_at to UTC + a test (test_naive_occurred_at_accepted_as_utc) that proves it.
@pytest.mark.integration opt-in via conftest hook, with a working real-CH round-trip test that cleans up after itself via ALTER TABLE … DELETE.
Bundled --extra dev → --group dev fix in the Makefile — right kind of opportunistic cleanup.
_DECISION_COLUMNS comment ("Order matches schema.sql; received_at absent — server-stamped via column default") — exactly the kind of inline land-mine doc that pays off six months later when someone adds a column.

guillaume-hexamind:
PR now points to main; claude review comments 1 to 5 fixed

PR #20, opened 2026-05-29
feat(fortify): emit audit events from PolicyEnforcer to /v1/audit/decisions

guillaume-hexamind:
SDK-facing side of the audit pipeline: every `PolicyEnforcer.decide()` now emits an audit event to the platform's `/v1/audit/decisions` endpoint when
  `FORTIFY_KEY` is set. Fire-and-forget HTTP POST per decision, bounded concurrency, silently inert when no key is configured.

  - `fortify/security/decision.py` — `Decision` gains auto-stamped `event_id: UUID` and `occurred_at: datetime` (UTC) at construction. Names mirror the
  platform's `AuditEnvelope`.
  - `fortify/audit.py` (new) — `AuditEvent` dataclass wrapping a `Decision` plus envelope (`user_id`, `session_id`) with `as_payload()`. `AuditSender` posts
  each event via `httpx.AsyncClient` + `asyncio.create_task`, bounded by `asyncio.Semaphore(32)`, with one 503 retry, periodic drop-warning, and graceful
  `close()` drain. Module-level `configure(api_key=None, base_url=None)` / `get_sender()` / `shutdown()` singleton — env-aware (`FORTIFY_KEY` /
  `FORTIFY_API_URL`), idempotent, returns `None` when no key resolvable.
  - `fortify/security/enforcer.py` — after building each `Decision`, emits an `AuditEvent` to the configured sender (no-op when none). `user_id` /
  `session_id` sourced from the existing `User` contextvar; `None` normalized to `""` at the wire boundary to match the platform's defaults.
  - `fortify/bootstrap.py` — calls `audit.configure()` after `load_dotenv()`. One line.
  - All four adapter wrappers (`langchain`, `openai`, `google`, `pydantic_ai`) + `FortifyAgent.enforce_policy()` — auto-configure audit at wrap-time so
  end-users get emission without ceremony.
  - `examples/*.py` (5 files) — dropped placeholder `api_key="sk-..."` strings so the adapter env fallback kicks in.

  ## Test plan

  - [x] `make test` — 25 audit + enforcer-audit tests pass (`tests/audit/`, `tests/security/test_enforcer_audit.py`)
  - [x] `pytest -m integration` with `make clickhouse-up` + `make platform-api` running — round-trip tests pass against live platform
  - [x] `make serve --agent support_bot` + Playground chat — decisions appear in ClickHouse within ~1s; visible at `localhost:5173/audit`
  - [x] `project_id` server-resolved from bearer; `agent_version_id` server-resolved from registration; `user_id` / `session_id` / `role` populated from
  `User` scope


quanghexa94:
The lifecycle model (`configure()` per api_key, loop-rotation rebuild, drop-on-saturation) is mostly thoughtful, and the "Enforcement is authoritative; audit is observational" principle in `docs/audit-pipeline.md` is exactly the right invariant to anchor the rest on. But there are a few real design holes — one of them a correctness bug that's invisible until production traffic hits the right tool — plus some scope/data-classification questions that should at least be acknowledged before merge.

---

### 1. TOCTOU on `arguments` — the audit payload races with the caller

`fortify/security/enforcer.py`:

```python
args_snapshot = dict(arguments)             # shallow copy
decision = Decision.from_verdict(..., arguments=args_snapshot)
...
self._audit_sender.emit(AuditEvent(decision=decision, ...))
return decision
```

`fortify/audit.py`:

```python
def emit(self, event: AuditEvent) -> None:
    ...
    task = asyncio.create_task(self._send(event), ...)   # scheduled, not awaited

async def _send(self, event: AuditEvent) -> None:
    async with self._semaphore:
        payload = event.as_payload()                     # builds dict, still by reference
        response = await self._client.post(..., json=payload)  # httpx serializes HERE
```

The `dict(arguments)` is **shallow**. Any nested dict / list / object inside the arguments stays a live reference all the way through `Decision` → `AuditEvent` → `payload` → httpx's `json=` serializer. Between the `create_task` call and httpx actually serializing the body (which can be deferred by the semaphore wait, the loop scheduler, the network), the agent's tool function has had time to mutate the same dict the caller passed in. Common offenders:

- Tools that pop/insert keys into their args dict before calling out.
- Tools that mutate a nested config dict in place.
- Anywhere the policy engine itself enriches `args` (less likely here, but the pattern is general).

When this fires, the audit log shows the **post-mutation** state instead of what the policy actually saw. The decision was correct; the audit record lies about what it was made on — which is the single thing audit must not do.

Two fixes, both fine:

- **`copy.deepcopy(args_snapshot)` in the enforcer** — bulletproof but pays the deep-copy cost on every `decide()` even when no audit sender is wired. Pay it only when `audit_sender is not None`.
- **Eager JSON serialization in `emit()`** — turn the payload into `bytes` *before* `create_task`, store bytes on the queued work, and pass `content=` to httpx instead of `json=`. Costs a JSON encode on the hot path but pins the payload at the moment of decision. This is the closer match to "audit is a snapshot."

Adding a test that proves it: mutate an arg dict after `decide()` returns and assert the captured payload still reflects the pre-mutation value.

### 2. Tool arguments flow to ClickHouse verbatim — no SDK-side redaction hook

`AuditEvent.as_payload()` ships `d.arguments` straight onto the wire. The platform byte-caps it at 8KB (`MAX_ARGS_BYTES`), but it never filters content. Concrete worry list:

- A `send_email(to=..., body=...)` tool puts customer email bodies into ClickHouse.
- A `query_db(sql=...)` tool puts the SQL — and any inline literals — into ClickHouse.
- An `authenticate(username, password)` tool (yes, this happens in real codebases) puts a plaintext password into ClickHouse.

ClickHouse holds these for 90 days under whatever access controls the audit DB has. For a security audit log this is the wrong default — security teams expect audit logs to be readable by people who shouldn't see PII / secrets.

Minimum bar before merge:

- A **redaction protocol on the SDK side**: per-tool-name allow/deny key lists, or a callable `redact: dict -> dict` passed into `configure()`, defaulting to "strip values for keys matching `password|secret|token|api_key`".
- A loud doc note in `docs/audit-pipeline.md` listing what *will* be captured if no redactor is configured, so operators make an informed choice rather than discover it three months later.

Either of those would let me sleep on this. The current state — "we capture everything tool callers send, indefinitely modulo TTL, with no opt-out" — is a data-protection footgun.

---

## 🟡 Worth a thread, not a blocker

### 3. `Decision` is becoming a god-object

`fortify/security/decision.py`:

```python
@dataclass(frozen=True, slots=True)
class Decision:
    outcome: DecisionOutcome
    agent_name: str
    tool_name: str
    role: str | None = None
    reason: str = ""
    error_type: str | None = None
    hint: dict[str, Any] | None = None
    violations: tuple[str, ...] = ()
    arguments: dict[str, Any] | None = None
    event_id: UUID = field(default_factory=uuid4)
    occurred_at: datetime = field(default_factory=lambda: datetime.now(timezone.utc))
```

`Decision` is the return type of `enforcer.decide()` — it's the public API every adapter, every approval handler, and every test consumes. Folding audit-emission identifiers (`event_id`, `occurred_at`) onto it does two things:

- Every `decide()` call mints a UUID + a `datetime.now()` whether audit is configured or not. Cheap individually but it's "free work" on the hot path that the no-audit consumer can never opt out of.
- It couples the audit subsystem into the domain dataclass. A future consumer reading `Decision` sees fields they have no use for and have to wonder about.

Cleaner shape: keep `Decision` clean of audit fields, and let `AuditEvent.__post_init__` stamp `event_id` + `occurred_at` on itself (it's already a dataclass). The platform-side `AuditEnvelope` lives in `schemas.py` precisely as a separate wrapper — the SDK should mirror that.

### 4. PR description ≠ diff: dashboard + debug GET aren't actually in this PR

The body claims:

> `platform/api/main.py` — new `GET /v1/audit/decisions?limit=N` debug endpoint returning raw rows as JSON.
> `platform/dashboard/src/routes/Audit.tsx` — minimal page fetching the debug endpoint and dumping raw rows for inspection.

Actual state: `Audit.tsx` is 10 lines of static markup ("Every policy decision, streamed live.") that doesn't fetch anything, and `platform/api/main.py` has no new GET endpoint. The "unauth POC read" diagram in `docs/audit-pipeline.md` references infrastructure that doesn't exist in the diff. Either drop those claims from the description (and the diagram), or land the actual code.

Tangent worth raising while it's still cheap to change: an **unauth** debug read endpoint that returns rows by `project_id` is a hard no for a security audit log. When that endpoint does get added (next PR? this one?), it must require auth and scope rows to the bearer's project — same trust envelope as `POST /v1/audit/decisions`.

### 5. Bearer token shared between policy fetch and audit write

The same `FORTIFY_KEY` resolved by `FortifyClient` (used to fetch policy bundles, register agents, etc.) is also the bearer the SDK uses to POST audit events. Two scoping concerns flow from that:

- **Audit-forgery via key exfiltration.** An attacker who lifts the key from a deployed agent can spam `POST /v1/audit/decisions` to flood the log with garbage and bury real malicious activity in noise — the platform side has no rate limit on this endpoint that I can see in PR #19.
- **Future biscuit scoping.** The platform already uses biscuits for SDK tokens. The natural fix is to attenuate audit-emit-only sub-tokens, so a key restricted to "fetch policy" can't write audit, and vice versa. Out of scope for this PR but worth filing as a follow-up.

### 6. No SDK-side size limit — big-arg calls silently lose their audit event

The platform rejects payloads over 8KB args / 4KB hint with a 413. The SDK's `_send()` just logs the 4xx and moves on:

```python
if response.status_code >= 400:
    _log.error("audit ingest failed: %s %s", response.status_code, response.text[:200])
```

So a single big-arg tool call (e.g., a tool that takes a large document blob) loses its decision audit completely, with only a log line as a trace. Two options:

- **Pre-truncate on the SDK** (the schema comment already calls it "SDK-truncated JSON snapshot") so events always land, even if lossy.
- **Distinguish 413 from generic 4xx**: keep a minimal "decision happened but args dropped" record by retrying with `arguments={}` + a marker like `"_args_dropped": true`.

Either way the principle is "audit is observational; observational data missing entirely is worse than observational data being lossy."

---

## 🟢 Polish / design observations

- **`_send()`'s 503 retry has no jitter**: `await asyncio.sleep(min(self._http_timeout, 2.0))` — a fleet of 200 SDKs all hitting a platform 503 sleep exactly 2 seconds and retry simultaneously. Add `random.uniform(0, 1)` to the sleep.
- **`_senders: dict[str, AuditSender]` is unbounded.** A long-lived process that rotates keys (multi-tenant gateway, key-per-request pattern) leaks one sender + one httpx pool per unique key. Either bound the registry with an LRU + evict via `await close()`, or document the assumption "small fixed key set per process."
- **Sync agents silently get no audit.** `asyncio.get_running_loop()` raises and `emit()` no-ops with a one-time warning. Adapters that drive agents from purely sync code (LangChain `.invoke()` instead of `.ainvoke()`) produce zero audit events. Document this in `docs/audit-pipeline.md` so it's not a surprise.
- **Drain semantics on process exit**: `bootstrap.py` docstring says "the httpx client is otherwise closed by Python's garbage collector at exit." At process exit the running event loop is already tearing down, so pending `asyncio.create_task` background sends are cancelled, not finished. The honest framing: "events emitted within the last ~drain_timeout window before exit may be lost unless `await audit.shutdown()` is called." Worth one accurate sentence in place of the GC claim.
- **`@dataclass(frozen=True, slots=True)` on `AuditEvent` + `decision=Decision` field** — the AuditEvent itself is immutable, but the `decision` field references a `Decision` whose `arguments` is a `dict`, which is mutable. Same root cause as item #1; would be obviated by item #1's fix.
- **Tests don't cover the deep-copy / mutation scenario** (see #1) or the case where the agent's tool function mutates an arg dict between `decide()` and the background task running. Worth adding once the fix lands.

---

## ✅ Things to keep

- **Per-key registry, not first-wins singleton.** The "one sender per resolved api_key" choice with the comment explaining why (one process can wrap several tenants' agents) is the right shape — and the commit history shows this was iterated to, not arrived at by accident.
- **Loop-rotation rebuild (`_ensure_loop_state`)** is a thoughtful response to a real failure mode: tests / scripts that `asyncio.run()` multiple times in one process previously crashed on "bound to a different event loop" — and there's a test (`test_rebuilds_loop_bound_state_across_event_loops`) that pins the fix.
- **Drop-on-saturation with periodic-warning logging** is the correct trade-off for an *observational* system. The 100-drop warning cadence is sane (loud enough to notice, quiet enough to not drown other logs).
- **`docs/audit-pipeline.md`'s three design principles** (enforcement authoritative, server owns identity, one envelope many event types) are exactly the right invariants to write down up front — these are the kind of decisions that get accidentally inverted six months later if they're only in someone's head.
- **Trust boundary preserved end-to-end.** The SDK sends some fields as empty strings (`agent_version_id`, `received_at`); the platform server-stamps them; the test `test_agent_version_id_comes_from_platform_lookup` (in PR #19) and the doc both say so explicitly.
- **Adapter wiring is consistent across all four** (`langchain`, `openai`, `google`, `pydantic_ai`) — each calls `configure()` with its resolved key and injects the sender into the enforcer. Easy to extend to a fifth adapter without forgetting a step.
- **Integration tests gated behind `@pytest.mark.integration`** and skip cleanly when `FORTIFY_KEY` isn't set — opt-in is the right default for "needs a live platform" tests.

guillaume-hexamind:
Fixes 1 to 6 + Polish / design observations all addressed
Only issue 5 may be worth exploring in a future PR

PR #21, opened 2026-06-01
feat: policy hot-reload via PolicySource — platform If-None-Match + local yaml/bundle-dir refresh

quanghexa94:
# Policy hot-reload: refresh-on-every-run for platform + local sources

Policy used to load once at agent boot. After this PR, the runtime consults a `PolicySource` at the top of **every** agent run — so a YAML edit in the dashboard, or a `policy.yaml` saved on disk, takes effect on the next chat message. No restart, no tool re-wrap, no rebuild step in between.

## What changed

**`PolicySource` protocol** — one tiny interface, three implementations:

| Source | Wakes up on | Cheap path |
|---|---|---|
| `PlatformPolicySource` | `FORTIFY_KEY` set | `If-None-Match` → `304 Not Modified`, cached `PolicyBundle` reused |
| `BundleDirPolicySource` | `FORTIFY_LOCAL_POLICY=<dir>` | `stat()` the manifest; identity-reuse when mtime unchanged |
| `YamlPolicySource` | `FORTIFY_LOCAL_POLICY=*.yaml` | `stat()` the yaml; recompile only when mtime changed |

**Runtime seam** — `FortifyAgent.refresh_policy()` swaps `PolicyEnforcer.policy` in place when the source returns a new bundle instance. Tools hold a reference to the enforcer, not the bundle, so one update covers every guarded tool with zero re-wrapping. `invoke_agent` / `stream_agent_raw` call it via `asyncio.to_thread` and swallow refresh failures (a flaky source never kills a turn).

**Content-addressed WASM cache** — `WasmPolicy.from_bytes_cached(wasm, wasm_hash)` shares wasmtime stores across loads of the same hash. Refresh that didn't actually change → no wasmtime re-instantiation. LRU(16) with per-hash locks against stampede.

**Platform endpoint** — `GET /v1/projects/{p}/agents/{a}` now sets `ETag: "<wasm_hash>"` and returns `304` with empty body when `If-None-Match` matches.

**Dev signing** — `FORTIFY_BUNDLE_SIGN_KEY_PATH=<*.private>` opts a YAML source into signing each recompile, so downstream gates that check `bundle.is_signed` see what they expect.

## Dev-facing surface

Zero new code in user-land. The whole story:

```bash
# Production: edit YAML in the dashboard, next chat picks it up
export FORTIFY_KEY=fty_test_...

# Tight dev loop: edit policy.yaml in $EDITOR, save, next chat sees the new policy
export FORTIFY_LOCAL_POLICY=./policy.yaml

# Production-shaped local testing: rebuild between turns
fortify policy build policy.yaml --out ./bundle
export FORTIFY_LOCAL_POLICY=./bundle/
```

Same `User(role=...)` scope around the call, same `enforce_policy` wrapping, same everything else.

## Tests

- **SDK 694/694 green**, **platform 70/70 green**, lint clean.
- New: `tests/security/test_local_sources.py` (18), `tests/security/test_source.py` (7), 3 wasm-cache tests, 3 platform ETag/304 tests, 3 `refresh_policy` seam tests, 4 parity-with-old-API tests in `test_local_policy_override.py`.

## Out of scope

- Policy hot-reload during a single agent turn — refresh is at turn boundaries, not mid-tool-call.
- Approval handler hot-reload — only the policy swaps; handlers wired into `enforce_policy(...)` stay where they were.

## Commits

- `377f214` — phase 8a: platform ETag/304 + WASM cache + `refresh_policy` seam
- `d43e9d0` — phase 8b: local yaml + bundle-dir sources
- `95faa9a` — README: build-an-agent walkthrough, drop stale `before_action` section

### DIFF
--- a/fortify/agents/factory.py
+++ b/fortify/agents/factory.py
@@ -2,6 +2,8 @@
 
 from __future__ import annotations
 
+import asyncio
+import logging
 from collections.abc import AsyncIterator, Awaitable, Callable, Mapping, Sequence
 from pathlib import Path
 from typing import Any, Self, TypeAlias
@@ -367,6 +369,14 @@ class FortifyAgent:
         client = getattr(self, "fortify_client", None)
         if client is not None:
             rebuilt.fortify_client = client
+        # Phase 8: keep the enforcement seam alive across with_tools rebuilds
+        # so refresh_policy() can still swap engines after the rebuild.
+        enforcer = getattr(self, "_enforcer", None)
+        if enforcer is not None:
+            rebuilt._enforcer = enforcer
+        source = getattr(self, "_policy_source", None)
+        if source is not None:
+            rebuilt._policy_source = source
         return rebuilt
 
     def enforce_policy(
@@ -414,7 +424,40 @@ class FortifyAgent:
                 )
             else:
                 wrapped.append(tool_spec)
-        return self.with_tools(wrapped)
+        rebuilt = self.with_tools(wrapped)
+        # Stash the enforcer so refresh_policy() can swap its policy in
+        # place when the platform serves a new bundle, without rebuilding
+        # the tool wrappers.
+        rebuilt._enforcer = enforcer
+        return rebuilt
+
+    def refresh_policy(self) -> None:
+        """Pull the current policy from the attached source and swap it in.
+
+        Called at the top of every agent run (see :func:`stream_agent` /
+        :func:`invoke_agent` / serve mode) so policy edits land at the
+        next run no matter how the user invokes the agent. The
+        :class:`~fortify.security.source.PlatformPolicySource` does the
+        ETag/304 dance, and the content-addressed
+        :class:`~fortify.security.wasm_engine._WasmPolicyCache` makes the
+        unchanged case a single small HTTP round-trip — no wasmtime
+        re-instantiation, no signature re-verify.
+
+        No-op when no source is attached (programmatic callers that
+        constructed the agent without one) or no enforcer exists (no
+        policy was applied). Failures from the source propagate so
+        callers see them at run boundary, not as silent staleness.
+        """
+        source = getattr(self, "_policy_source", None)
+        enforcer = getattr(self, "_enforcer", None)
+        if source is None or enforcer is None:
+            return
+        new_policy = source.fetch()
+        if new_policy is None or new_policy is enforcer.policy:
+            # Source has nothing to offer, or the same object came back
+            # (e.g. PlatformPolicySource returns the cached bundle on 304).
+            return
+        enforcer.policy = new_policy
 
 
 AgentGraph: TypeAlias = FortifyAgent
@@ -501,6 +544,29 @@ def create_agent(
     return agent, handler
 
 
+_logger = logging.getLogger("fortify.agents.factory")
+
+
+async def _refresh_policy_safely(agent: "AgentGraph") -> None:
+    """Pull the latest policy from the agent's attached source, off the loop.
+
+    Called at the top of every async invocation so policy changes land at
+    the next run without anyone having to know which entry point they're
+    going through. No-op when no source is attached (programmatic
+    construction). Failures log a warning and keep the previous policy —
+    a transient network blip should never crash a chat turn.
+    """
+    refresh = getattr(agent, "refresh_policy", None)
+    if refresh is None:
+        return
+    try:
+        await asyncio.to_thread(refresh)
+    except Exception as exc:  # noqa: BLE001 — refresh failures must not crash the run
+        _logger.warning(
+            "policy refresh failed: %s — keeping previously loaded policy", exc
+        )
+
+
 @observe(name="invoke_fortify_agent")
 async def invoke_agent(
     agent: AgentGraph,
@@ -510,6 +576,7 @@ async def invoke_agent(
     tool_use_context: ToolUseContext | None = None,
 ) -> dict[str, Any]:
     """Invoke the agent for one normalized input payload."""
+    await _refresh_policy_safely(agent)
     token = set_current_tool_use_context(
         _resolve_tool_use_context(agent, tool_use_context)
     )
@@ -530,6 +597,7 @@ async def stream_agent_raw(
     tool_use_context: ToolUseContext | None = None,
 ) -> AsyncIterator[LangChainStreamEvent]:
     """Stream raw LangChain events from the agent runtime."""
+    await _refresh_policy_safely(agent)
     config = get_langfuse_runnable_config(handler)
     config["run_id"] = new_root_run_id()
     token = set_current_tool_use_context(
--- a/fortify/agents/factory.py
+++ b/fortify/agents/factory.py
@@ -6,7 +6,15 @@ import asyncio
 import logging
 from collections.abc import AsyncIterator, Awaitable, Callable, Mapping, Sequence
 from pathlib import Path
-from typing import Any, Self, TypeAlias
+from typing import TYPE_CHECKING, Any, Self, TypeAlias
+
+if TYPE_CHECKING:
+    # Optional seam-attribute types — referenced only in __init__ signatures
+    # / annotations. Imported under TYPE_CHECKING to avoid the runtime cycle
+    # (security.* and cloud.* both eventually import from this module).
+    from fortify.cloud.client import FortifyClient
+    from fortify.security.enforcer import PolicyEnforcer
+    from fortify.security.source import PolicySource
 
 from langchain.agents import create_agent as create_langchain_agent
 from langchain.agents.middleware.types import AgentMiddleware
@@ -193,7 +201,7 @@ def _resolve_user_facts(agent: "FortifyAgent") -> dict[str, list[str | int]] | N
     user = get_current_user()
     if user is None:
         return None
-    client = getattr(agent, "fortify_client", None)
+    client = agent.fortify_client
     if client is None:
         # Local agent or test stub — User scope is set but there's nothing to
         # attenuate against. Surface a single warning so devs see why their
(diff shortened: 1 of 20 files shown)


guillaume-hexamind:
## Code review — policy hot-reload via `PolicySource`

Reviewed across correctness, removed-behavior, cross-file, and cleanup/altitude angles. Findings below with a severity score **/100** (weighted: correctness/security impact highest, then behavior regressions, then cleanup). Each was verified against the code.

### 🔴 Correctness & behavior

**[58/100] The content-addressed WASM cache is never reached in production.**
`from_bytes_cached` / `_WasmPolicyCache` are invoked only from tests — the sole eval path is `policy.py:134` `bundle.policy().decide(...)`, and `PolicyBundle.policy()` (`bundle.py:275`) unconditionally calls `WasmPolicy.from_bytes()`. So when `refresh_policy()` swaps in a new bundle instance with an unchanged `wasm_hash` (whitespace-only yaml recompile, bundle-dir touch), the first `decide()` still pays the full ~50–100ms wasmtime instantiation the cache was built to avoid. The PR's headline benefit ("refresh that didn't change → no re-instantiation") does not actually fire. → wire `policy()` to `from_bytes_cached(self.wasm_bytes, self.wasm_hash)`.

**[55/100] Committed caller broken by the `get_agent()` signature change.**
`examples/fortify_agent_demo.py:70` still does `payload = client.get_agent(agent_name)` then `payload["policy_yaml"]` / `payload['updated_at']`. `get_agent()` now returns `tuple[dict|None, str|None]`, so the demo raises `TypeError: tuple indices must be integers`. → unpack `payload, _ = client.get_agent(...)`.

**[45/100] A signed local bundle with no pubkey now loads silently announced as "signed", with the old "signature NOT verified" warning gone.**
Old `_verify_bundle_signature` printed `[fortify] warning: ... signature NOT verified` for (signed bundle, `FORTIFY_BUNDLE_PUBKEY_PATH` unset, `REQUIRE_SIGNATURE=false`). New path: `_resolve_pubkey_for_verification` returns `None`, `BundleDirPolicySource(verify_with=None)` skips verification, and `_announce_local_override` prints "signed". A dev sees "signed" and reasonably assumes authenticity was checked when it wasn't. (Verification behavior is unchanged from before — only the warning was lost — but it's security-relevant and now misleading.)

**[35/100] `serve` no longer hot-reloads system prompt / agent spec / tools — only the policy.**
The removed `_refresh_runtime`/`_rebuild` called `build_runtime → load_agent` each turn, re-fetching `system_md`, `agent_yaml`, and tools. The replacement only swaps `enforcer.policy`. A dashboard edit to an agent's system prompt or spec mid-session, which previously took effect at the next turn boundary, is now silently ignored until restart. May be intended (PR scope is policy reload), but worth calling out as a narrowing.

### 🟡 Cleanup & altitude

**[22/100] Override-load block copy-pasted across three loaders.**
The `bundle, source = override; enforced = enforce_policy(...); enforced._policy_source = source` sequence appears in `load_builtin_agent`, `load_local_agent`, and `load_fortify_agent`. Future changes to how the source attaches must touch all three or they drift.

**[20/100] `BundleDirPolicySource.fetch` and `YamlPolicySource.fetch` duplicate the same stat-mtime → compare → reload → cache state machine** (~40 lines incl. OSError handling and identity-reuse). No shared base/helper; refresh discipline must be maintained twice.

**[15/100] Signature-policy enforcement is split between construction (`_resolve_pubkey_for_verification`) and fetch (`_verify_local_source_signature_policy`), and the latter's BundleDir branch is a no-op.** The full `REQUIRE_SIGNATURE` matrix is spread across three locations, making it hard to confirm every cell is covered.

### Verified clean (dropped)
- `_WasmPolicyCache` `_inflight` "deadlock" on loader exception — **not a bug**: `with slot:` releases on exception and the entry self-heals on next load.
- `assert payload is not None` under `python -O` — **not reachable**: the first `get_agent` never sends `If-None-Match`, so 304 is impossible.
- ETag quoting — **consistent** end-to-end (platform emits `"<sha256>"`, client echoes verbatim, `PlatformPolicySource` fallback matches, platform compares with `.strip()`).
- `enforcer.policy` swap seam — **correct**: plain mutable attribute read fresh in `authorize`, so GuardedTools pick up the new bundle without re-wrapping.

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code)</sub>

fortify/agents/factory.py
```diff
@@ -367,6 +369,14 @@ def with_tools(self, tools: Sequence[ToolSpec]) -> Self:
         client = getattr(self, "fortify_client", None)
         if client is not None:
             rebuilt.fortify_client = client
+        # Phase 8: keep the enforcement seam alive across with_tools rebuilds
+        # so refresh_policy() can still swap engines after the rebuild.
+        enforcer = getattr(self, "_enforcer", None)
```
guillaume-hexamind: Seam attributes should be first-class. _enforcer / _policy_source / fortify_client are attached via setattr and read with getattr(self, ..., None) 
  ▎ everywhere. A typo or a missed propagation path silently disables refresh, and since the path is best-effort nobody notices. Make them real __init__ 
  ▎ fields (default None) so the type checker covers them. Same for getattr(agent, "refresh_policy", None) — agent is always a FortifyAgent, so call it 
  ▎ directly.

fortify/agents/factory.py
```diff
@@ -414,7 +424,40 @@ def enforce_policy(
...
-        return self.with_tools(wrapped)
+        rebuilt = self.with_tools(wrapped)
+        # Stash the enforcer so refresh_policy() can swap its policy in
+        # place when the platform serves a new bundle, without rebuilding
+        # the tool wrappers.
+        rebuilt._enforcer = enforcer
+        return rebuilt
+
+    def refresh_policy(self) -> None:
+        """Pull the current policy from the attached source and swap it in.
```
guillaume-hexamind, line 461: Docstring vs. behavior. refresh_policy says failures "propagate ... not silent staleness," but _refresh_policy_safely catches Exception → 
  ▎ logger.warning, which is silent staleness. For a security control that's a defensible choice, but let's say so in the docstring and make it observable (a
  ▎ counter / on_refresh_error hook / last_refreshed_at) rather than a log line.

fortify/agents/factory.py
```diff
@@ -510,6 +576,7 @@ async def invoke_agent(
     tool_use_context: ToolUseContext | None = None,
 ) -> dict[str, Any]:
     """Invoke the agent for one normalized input payload."""
+    await _refresh_policy_safely(agent)
```
guillaume-hexamind: Single refresh seam. Refresh lives on invoke_agent/stream_agent_raw but not on FortifyAgent.ainvoke/astream_events. A direct method caller silently 
  ▎ gets no hot-reload. Worth funneling both through one place.

fortify/cloud/client.py
```diff
@@ -238,20 +251,45 @@ def _fetch_public_key(self) -> bytes:
...
             "User-Agent": "fortify-sdk/0.1",
         }
         if authorize:
             headers["Authorization"] = f"Bearer {self.config.api_key}"
+        if if_none_match is not None:
+            headers["If-None-Match"] = if_none_match
         request = urllib.request.Request(url, headers=headers)
         try:
             with urllib.request.urlopen(request, timeout=self.timeout) as response:
+                etag = response.headers.get("ETag")
```
guillaume-hexamind, line 307: self.timeout defaults to 10s, and this GET is now on the per-turn hot path: refresh_policy awaits PlatformPolicySource.fetch → get_agent at the top of every turn. So a slow/hung platform blocks every chat turn up to 10s before responding — for a refresh that's best-effort and falls back to the cached bundle on failure anyway. Suggest a separate tight timeout (~1–2s) for the conditional if_none_match path so platform degradation can't tax every turn.

fortify/security/source.py
```diff
@@ -0,0 +1,347 @@
...
+        self._client = client
+        self._agent_name = agent_name
+        # Pre-seed when the caller already fetched + verified the bundle
+        # (typical at agent load time). Avoids a redundant 200 round-trip
+        # on the first refresh — that call will send If-None-Match and
+        # get a cheap 304.
+        self._cached_bundle: PolicyBundle | None = initial_bundle
+        self._cached_etag: str | None = initial_etag
+
+    def fetch(self) -> PolicyBundle | None:
```
guillaume-hexamind, line 105: fetch() mutates self._cached_bundle and self._cached_etag with no lock, and it runs on a to_thread worker via refresh_policy. If one agent is shared 
  across concurrent turns, two fetch() calls can interleave and pair the bundle from one response with the etag from another → a later spurious 200/304. 
  The wasm engine serializes its store access carefully; this stateful source should match it. Either guard the cached-field updates with a lock, or 
  document explicitly that a source instance is single-threaded / one-per-agent and the refresh seam must never call it concurrently.

fortify/security/wasm_engine.py
```diff
@@ -163,6 +165,28 @@ def from_bytes(
             base_heap_ptr=base_heap_ptr,
         )
 
+    @classmethod
+    def from_bytes_cached(
```
guillaume-hexamind, line 169: Internal note for the future: this cache is not yet wired into the eval path — PolicyBundle.policy() (bundle.py:275) still calls from_bytes() directly, so 
from_bytes_cached is currently exercised only by its own unit tests. We should not assume a policy refresh hits this cache until policy() is switched over.

guillaume-hexamind (commented):
A couple of comments to check (mine and the top ones of claude code), but ok with the implementation overall

We'll need to add a similar logic to the adpaters in a following PR

quanghexa94:
for the claude comment at the beginning, fixed first 3 comments, the four one is intentional: serve only does hot-reload on policies, not on agent's definition (prompt, tools etc..)

PR #10, opened 2026-05-13
refactor(fortify/cli): split fortify [...] command into fortify chat [...] and fortify serve [...]

guillaume-hexamind:
CLI restructure: fortify becomes a multi-subcommand entry point
  - Split the single-command flag-routed CLI into three explicit subcommands: fortify chat, fortify serve, and fortify register.
  - Replace the manual argv dispatch in fortify/cli/__init__.py with a proper argparse parser; each subcommand registers itself via
  add_parser(subparsers) and wires its handler with set_defaults(func=...).
  - Rename fortify/cli/app.py to fortify/cli/chat.py so the module name matches the new subcommand.

  Extract shared CLI plumbing into fortify/cli/_common.py
  - New module hosts AgentRuntime, build_runtime(), load_agent_script(), prompt_for_approval(), build_approval_handler(), and
  add_shared_agent_flags() — previously duplicated or living inside app.py.
  - chat.py and serve.py now import these helpers instead of reimplementing them.

  Promote register to a first-class subcommand
  - fortify/cli/register/main.py switches from main(argv: list[str]) to the subcommand convention main(args: argparse.Namespace) -> 
  int and exposes its own add_parser().
  - fortify/cli/register/__init__.py re-exports add_parser.

PR #22, opened 2026-06-03
feat(platform/dashboard): audit dashboard page

guillaume-hexamind:
Audit dashboard + read API

  Adds the read/aggregation half of the audit pipeline: project-scoped query endpoints over policy_decision and the Audit
  dashboard that consumes them. Builds on the SDK emit + ingest work in the base branch.
  
  Platform API (read path)

  - New project-scoped aggregation endpoints (query-time GROUP BY, no rollups):
    - GET /v1/projects/{id}/audit/summary — totals + breakdowns by agent/role/tool (single GROUPING SETS scan).
    - GET /v1/projects/{id}/audit/timeseries — per-bucket outcome counts.
    - GET /v1/projects/{id}/audit/decisions — filterable, paginated detail rows.
  - Shared scope filters (agent/role/tool) narrow every read consistently; table adds outcome/session_id. Windows
  24h/7d/30d/90d, bounded by the 90-day TTL.
  - Removed the old unscoped GET /v1/audit/decisions debug dump.
  - ClickHouse client created with autogenerate_session_id=False so the dashboard's concurrent reads don't hit "concurrent
  queries within the same session".

  Dashboard (Audit page)

  - New command-center Audit view: KPI cards w/ sparklines, stacked area chart + donut, switchable breakdown
  (tools/agents/roles, by volume or denials), filterable + paginated decisions table, slide-over detail drawer
  (reason/violations/arguments/envelope + same-session events), JSONL export.
  - Charts are hand-built SVG (no charting dependency); styling reuses existing design tokens.
  - Filters drive react-query; KPIs/charts/table all reflect the same scoped slice.
  - Removed the placeholder Dashboard page (/ now redirects to Agents) and the stale 24h badge on the Audit nav item.

  Docs

  - Updated docs/audit-pipeline.md read-path section to describe the new endpoints.

  Note: read endpoints remain unauthenticated POC (project-scoped), matching the other dashboard reads — read_audit scope gate
  is a follow-up (TODO(auth) in place).


quanghexa94:
Now that M3 (#23), the audit ingest (#19) and the SDK emitter (#20) have all landed on `main`, this PR can be reviewed against the conventions those PRs set up — not against an empty room. The backend read-path design is sound (single GROUPING SETS scan, parameterized SQL, `count() OVER ()` for paged totals, sparse timeseries). The frontend page works, looks polished in isolation, and the design intent (KPIs + chart + breakdown + table + drawer) is the right shape for an audit view.

But: the page is built as if the dashboard didn't already have conventions for everything it does. The handlers drift from the surrounding code on auth, async, types, tags, and module layering; the React side rebuilds UI primitives, icons, and CSS that already exist; and zero tests ship on either side despite the scaffolding being in place for both. Most of the work to fix this is mechanical — but it has to happen before merge or it becomes the precedent for the next page.

---

### 1. The three new endpoints must be gated by `require_org_member` — the helper already exists

Every other project-scoped dashboard read on `main` uses one auth dep:

```python
@v1.get("/projects/{project_id}", tags=["projects"])
async def api_get_project(
    project_id: str,
    _user: User = Depends(require_org_member),    # ← canonical pattern
    session: AsyncSession = Depends(get_session),
) -> ProjectRead:
    ...
```

`require_org_member` ([platform/api/main.py:450](platform/api/main.py#L450)) resolves the project's `org_id`, confirms the active user belongs to that org, returns `404` for unknown projects (no ID-enumeration leak) and `403` for non-members. It's used by `/agents`, `/agents/manifest`, `/agents/{name}`, `/tokens`, `GET /projects/{id}`. Cookie session OR bearer token both flow through it cleanly.

This PR's three handlers take `project_id` from the path and trust it. The `TODO(auth)` comment was honest when written, but the helper landed in #23. Drop-in change per handler:

```python
@v1.get("/projects/{project_id}/audit/summary",
        response_model=AuditSummary,
        tags=["audit"])
async def api_audit_summary(
    project_id: str,
    _user: User = Depends(require_org_member),   # ← add
    window: AuditWindow = "24h",
    agent: str | None = None,
    role: str | None = None,
    tool: str | None = None,
    clickhouse_client=Depends(require_clickhouse),
) -> AuditSummary:
    ...
```

Same change on `/audit/timeseries` and `/audit/decisions`. After that, the read endpoints carry the **same trust envelope as the write endpoint** in PR #19 (`require_project` is the SDK-bearer variant; `require_org_member` is the cookie-or-bearer human-side variant) — exactly the symmetry `docs/audit-pipeline.md` documents as the design intent ("server owns identity").

Without this change the dashboard ships a serve-all-projects-to-anyone read API for the one endpoint where the data is most sensitive. With it, the diff per handler is one line.

### 2. Zero tests on either side — and both scaffolds are already in place

**Backend:** Every endpoint on `main` has tests in `platform/api/tests/test_<feature>.py`. `tests/test_audit.py` already exists (PR #19); the conftest fixture, `dependency_overrides` pattern, and `@pytest.mark.integration` skip-by-default machinery are ready. Three pure-Python tests against a `MagicMock` client (no live ClickHouse needed) cover what's easy to break:

- **`_GROUPING_SETS` result classification.** Hand a fixture of `result_rows` tuples into `summarize`'s post-processing loop and assert the `totals` / `by_agent` / `by_role` / `by_tool` dict shape. One wrong `not` and the grand total leaks into a breakdown — exactly the kind of bug you'd notice in dashboards a week later, if at all.
- **`count() OVER ()` window with pagination.** Pin the contract: page 1 of 3 reports `total=N`; page 4 (past end) falls through to the separate `count()` branch.
- **`_scope()` WHERE-clause composition.** Enumerate (agent only), (role only), (all three), and verify the params dict + WHERE list. The dynamic SQL string-concat is exactly the surface where injection-shape bugs hide.

A fourth, easy one: hit each handler through the FastAPI `TestClient` with `require_org_member` overridden, and assert 401 → 403 → 200 in the auth matrix.

**Frontend:** The dashboard ships seven `*.test.tsx` files on `main`:

```
src/components/InviteMemberDialog.test.tsx
src/components/NoProjectEmptyState.test.tsx
src/components/OrgProjectSwitcher.test.tsx
src/lib/active.test.tsx
src/routes/AcceptInvitation.test.tsx
src/routes/OrgMembers.test.tsx
src/routes/OrgSettings.test.tsx
src/routes/Orgs.test.tsx
```

All of them follow a single pattern (`renderWithProviders` from `@/test/render`, `vi.spyOn(window, 'fetch')` for stubbing, vitest + `@testing-library/react`). The Audit page is the largest new page in months — over 300 LOC of routing logic alone — and adds zero tests. Easy first ones (mirroring `Orgs.test.tsx`):

- Filter selection updates the query: change agent dropdown, assert the next `fetch` URL contains `agent=foo`.
- Outcome KPI cards toggle the `outcome` filter when clicked.
- Detail drawer opens on row click and closes on Esc.
- "Same session" cross-link drills into a filtered list.

None of these need a live backend; the existing stub helper handles them.

---

## 🟡 Backend consistency drift — worth a thread

### 3. Sync `def` handlers in an `async def` codebase

Every `api_*` handler in `main.py` is `async def`:

```
$ grep -nE "^(async )?def api_(list|get)_" main.py | head -6
async def api_list_agents(...)
async def api_get_project(...)
async def api_list_invitations(...)
async def api_list_projects(...)
async def api_list_members(...)
async def api_list_orgs(...)
```

The audit read handlers in this PR are plain `def`. FastAPI runs sync handlers in a threadpool, which means each audit call hops off the event loop — fine while traffic is low, lossy as soon as it isn't. The `clickhouse_connect.Client` is a sync client, so the handler doesn't gain anything from `async def` *internally* without `asyncio.to_thread` around `client.query(...)`. Two consistent fixes:

- **Cheapest:** make the handlers `async def` and wrap the blocking ClickHouse call in `await asyncio.to_thread(summarize, client, ...)`. Matches the surrounding code, doesn't block the loop.
- **Or:** swap to `clickhouse-connect.async_client.AsyncClient`. Bigger blast radius (touches `clickhouse.py` and PR #19's insert path) but ends up with a fully async stack.

### 4. `Optional[str]` instead of `str | None`

```
$ grep -c "Optional\[" platform/api/main.py
0
$ grep -c "str | None" platform/api/main.py
~30
```

This PR's three handlers reach for `Optional[str]` and import `Optional` from `typing`. Trivial mechanical change; the value is just consistency for the next reader running `grep`.

### 5. No `tags=` on the three new routes

Every route in `main.py` has a tag (`tags=["projects"]`, `tags=["orgs"]`, `tags=["agents"]`). The three audit routes have none, so they land under the default group in OpenAPI. Add `tags=["audit"]`.

### 6. Read-path query helpers belong in `services.py`, not `audit.py`

Layering convention on `main`:

- `services.py` — domain functions for relational reads/writes (`list_agents`, `list_orgs_for_user`, `accept_invitation`, …).
- `audit.py` — persistence layer for the ClickHouse write path (the docstring even says so: *"Persistence layer for audit events"*).

This PR adds `summarize` / `timeseries` / `list_decisions` to `audit.py`, which makes the module half persistence-write and half query-read. Two options:

- Keep one `audit.py` but rewrite the module docstring to describe both halves.
- Split into `audit_writes.py` and `audit_reads.py` — mirrors how `services.py` would split if it ever got large.

Either is fine; today's state is the docstring lying about what's in the file.

### 7. `autogenerate_session_id=False` is process-wide — verify the write path

The same client is used by `insert_decision` from PR #19, where `_DECISION_INSERT_SETTINGS` enables `async_insert_deduplicate=1`. The flag is a no-op without a session anyway (as I noted on PR #18), so functionally nothing changes — but worth one assertion in an integration test that an INSERT still acks with this flag on.

### 8. `"(none)"` as a wire-format sentinel for empty role

Clever — drives the no-role drill-down cleanly. But the literal string `"(none)"` now leaks across the wire as a reserved value. A policy author defining a role literally named `"(none)"` would have their data routed to the empty bucket and become unfilterable. Either reserve formally (validate at policy creation), or send `role=""` over the wire and let the dropdown locally display `"(none)"`.

---

## 🟡 Frontend consistency drift — also worth a thread

### 9. The page rebuilds primitives that already exist in shadcn/ui

Every existing dashboard page on `main` imports shadcn/ui primitives:

```typescript
// Tokens.tsx, Agents.tsx, Orgs.tsx, OrgMembers.tsx, Policies.tsx, …
import { Button } from '@/components/ui/button'
import { Badge } from '@/components/ui/badge'
import { Input } from '@/components/ui/input'
import { Dialog, DialogContent, DialogHeader, … } from '@/components/ui/dialog'
import { Select, … } from '@/components/ui/select'
import { Table, TableRow, … } from '@/components/ui/table'
```

PR #22 imports none of them. Instead, `platform/dashboard/src/styles/audit.css` (+333 LOC) hand-rolls a parallel set:

- `fty-btn`, `fty-iconbtn`, `fty-btn.secondary`, `fty-btn.ghost.sm` ← shadcn `<Button variant=… size=…>`
- `fty-badge`, `fty-badge.muted` ← `<Badge>`
- `fty-select` ← `<Select>`
- `fty-seg`, `fty-seg button.active` ← could be `<ToggleGroup>` or a `<Tabs>` variant
- `fty-card`, `fty-card.pad-lg`, `fty-card-hd`, `fty-card-title` ← `<Card>` / `<CardHeader>` / `<CardTitle>`
- `fty-page`, `fty-page-hd`, `fty-page-title`, `fty-page-sub` ← the page-frame wrapper Agents/Tokens use directly

`components.json` is already configured for shadcn (`components.json` in the dashboard root), and `npx shadcn@latest add toggle-group tabs` would cover what's missing. The audit-CSS file deletes itself.

### 10. Inline `style={{...}}` is endemic in this PR — and exists nowhere else in the dashboard

```
$ for f in Agents Tokens Orgs OrgMembers Policies Playground; do
    count=$(git show origin/main:platform/dashboard/src/routes/$f.tsx | grep -c "style={{")
    echo "  $f.tsx → $count"
  done
  Agents.tsx → 0
  Tokens.tsx → 0
  Orgs.tsx → 0
  OrgMembers.tsx → 0
  Policies.tsx → 0
  Playground.tsx → 0

$ git show <pr22>:platform/dashboard/src/routes/Audit.tsx | grep -c "style={{"
39
$ git show <pr22>:platform/dashboard/src/components/audit/pieces.tsx | grep -c "style={{"
39
$ git show <pr22>:platform/dashboard/src/components/audit/charts.tsx | grep -c "style={{"
29
```

**107 inline-style props in the new page; zero in the rest of the dashboard.** The reason matters less than the number — the existing code uses Tailwind utility classes (`text-sm font-medium text-muted-foreground`) and CSS variables (`hsl(var(--card))`) almost exclusively. The Audit page mixes inline `style={{ color: 'hsl(var(--muted-foreground))', fontSize: 12.5, padding: '5px 0' }}` with the new `fty-*` classes, both of which sit alongside but don't compose with the existing system.

The fix is mechanical: replace inline-style maps with Tailwind classes and lift recurring patterns into the shadcn primitives. The audit-specific colour tokens (`--semantic-deny-soft`, `--semantic-approval`) belong in `index.css` next to the existing tokens, not duplicated across `audit.css`.

### 11. Custom `Icon` component instead of `lucide-react`

Every existing page imports icons from `lucide-react`:

```typescript
import { Bot, FileText, ShieldCheck, Check, X, Filter, KeyRound, … } from 'lucide-react'
```

`lucide-react` is already in `dashboard/package.json`. PR #22 builds `components/audit/icon.tsx` (51 LOC) as a name-keyed wrapper rendering inline SVG paths. The wrapper API (`<Icon name="check" size={12} />`) is fine, but every icon in the file is already in lucide. Delete the wrapper, import `{ Check, X, Filter, Activity, CircleDashed, … }` from lucide like everything else.

### 12. `DEFAULT_PROJECT_ID = 'support-bot'` instead of the M3 active-project hook

`platform/dashboard/src/lib/api.ts`:

```typescript
export const DEFAULT_PROJECT_ID = 'support-bot'
```

M3 (#23) shipped the canonical project-scope hook at [platform/dashboard/src/lib/active.ts](platform/dashboard/src/lib/active.ts):

```typescript
export function useProjectScoped(): ProjectScope { ... }
// returns { projectId: string | null, status: 'loading' | 'no-project' | 'ready' }
```

Every project-scoped page on `main` already uses it:

```typescript
// Agents.tsx
const scope = useProjectScoped()
const manifests = useQuery({
  queryKey: ['agent-manifests', scope.projectId],
  queryFn: () => api.listAgentManifests(scope.projectId as string),
  enabled: !!scope.projectId,
})
if (scope.status === 'no-project') return <NoProjectEmptyState resource="agents" />
```

The Audit page should follow exactly this pattern: derive `projectId` from `useProjectScoped()`, gate all four queries on `enabled: !!scope.projectId`, render `<NoProjectEmptyState resource="audit" />` when there's no project. Otherwise the page silently shows `support-bot` data even after the user switches orgs, which becomes a data-leak bug on top of the auth gap in #1.

### 13. State patterns: `useState` vs the established `zustand` pattern

The page's filter state lives in a local `useState<Filters>` — fine. But every other multi-screen state on `main` lives in zustand (`useActive`, `useOrgs`, `useProjects`). For pagination + filter state that the user expects to persist across drawer-open/close (and arguably across tab switches), a `useAuditFilters` zustand store would match the rest of the dashboard and free Audit.tsx of the "reset table-limit on filter change" workaround. Borderline — flagging because the existing pattern is consistent enough that the deviation is worth a sentence in code.

### 14. Hand-rolled SVG charts

This is the one I'd actually defend: `recharts` / `visx` would have been the obvious "add a dep" answer, but SVG-by-hand for an audit page that needs only an area chart + a donut + sparklines is fine — and avoids ~80KB of charting library. The inline styles inside the chart components are the real problem (see #10), not the choice to draw SVG by hand.

That said: extracting the chart components into the same `components/ui/` directory as the rest of the primitives, named neutrally (`AreaChart`, `Donut`, `Sparkline`), would let them be reused on a future page without anyone having to grep into `components/audit/`.

---

## 🟢 Polish

- **`count() OVER ()` cost on long windows.** Scans the full match set per page. Worth a "measured X ms at Y rows" line in the PR description.
- **Three identical fetches when no filter is set.** `optionsQ` and `summaryQ` issue the same request when scope reduces to range-only. Compute dropdown options off `summaryQ.data` when the filter set is empty.
- **`request<T>()` swallows the status code.** Define `class ApiError extends Error { status: number }` so react-query can distinguish "audit-temporarily-unavailable" from "permission-denied" without parsing the message. Grep main for an existing pattern — the M3 PR likely has one.
- **`refetchInterval: 30_000` × 3 queries × N tabs.** `refetchIntervalInBackground: false` for background tabs.
- **JSONL export is page-bounded.** Worth a one-line note in code so the next reader doesn't expect "Export" to mean "everything matching."
- **Branch typo `platoform_audit_dashboard`.** Lives in `git log --merges` forever once merged. `git branch -m` + force-push if you care.

---

## ✅ Things to keep

### Backend
- **Single GROUPING SETS scan for the summary** instead of four separate queries.
- **`count() OVER ()` for paged total** with inline rationale.
- **`_scope()` shared helper** keeps filter semantics identical across the three endpoints.
- **`autogenerate_session_id=False`** with the "concurrent queries within the same session" comment.
- **Sparse timeseries** server-side.
- **Removed the old unscoped `GET /v1/audit/decisions` debug dump** — closes the loop on the PR #20 review note.

### Frontend
- **One shared `scope` object drives all three queries** — UI changes refire all three in sync, `placeholderData: keepPreviousData` keeps the page from flickering during refetch. This part *is* exactly the pattern the rest of the dashboard's react-query usage settles on.
- **Detail drawer's "Same session · N"** — correlated events without a route hop. Right UX for incident response.
- **The page's information architecture** (range selector → KPIs → chart pair → breakdown → table → drawer) is sound. The work to bring it into convention is mechanical, not design-level.
- **No new charting dep** — hand-SVG was the right call for the limited surface here.

guillaume-hexamind:
All 15 claude caude comments solved - branch up to date and pointing towards main

PR #23, opened 2026-06-03
feat: multi tenant platform

quanghexa94:
Multi-tenant control plane
Goes from "single-tenant demo with hardcoded project + agent" to a real multi-tenant SaaS shape: real auth, orgs and projects, the dashboard's whole UI rebuilt around active-tenant state, token-implicit project resolution for the SDK, and a uvicorn-style fortify serve module:attr that closes the dev loop.

What's actually new
Real auth. FastAPI Users provides cookie sessions, registration, email verification, password reset, and Google OAuth. The dashboard talks to the platform via a fortify_session httpOnly cookie; the SDK keeps using bearer biscuits. Two surfaces, no hybrid dependency.

Orgs, projects, invitations. Every signup auto-creates a personal default org. Owners can invite teammates by email (magic-link accept, last-owner guard, role escalation prevention). Each org owns N projects; the OrgProjectSwitcher in the dashboard header is the single source of truth for "what tenant am I in." Projects gate every project-scoped route on the backend; the dashboard's Tokens / Agents / Policies / Playground pages all read activeProjectId from a zustand store with persisted localStorage.

Token-implicit project for the SDK. The biscuit envelope (fty_<env>_<project>_<biscuit>) already carried the project id; we stopped requiring the CLI to repeat it in URLs. POST /v1/agents, GET /v1/agents/{name}, WS /v1/serve, and GET /v1/me/key are now bearer-only routes that derive project from the token. The WS handshake uses subprotocol auth (browsers can't set custom headers on WS upgrades) with percent-encoded bearers — works around the RFC 7230 token grammar refusing = padding.

Register-time default policy. A brand-new agent registered via fortify register lands with a generated four-role starter policy (read_only mixin → default / member / admin), classified from the manifest's tool list via a name-shape heuristic. Shells gate even for admin; unknown tools fail-closed for member. The dashboard's Playground "Acting as" dropdown lights up immediately. Subsequent re-registers preserve the operator's edits — policy belongs to the operator after first create.

Uvicorn-style fortify serve. One command, no flags, no env vars:

uv run fortify serve examples.customer_bot:agent
The CLI imports the module, derives the manifest in one call (name, tools, model, system prompt come off the object), auto-registers on the platform (idempotent — content-hash short-circuits no-ops), fetches the operator's policy, applies enforcement to the local agent, and opens the relay. The Python file is authoritative for code; the platform is authoritative for policy. Edit the policy in /policies, save, the next chat turn picks it up via the existing ETag refresh contract.

Notable removals
DEFAULT_PROJECT_ID constant from the dashboard's lib/api.ts — every page now reads activeProjectId from the store.
FORTIFY_AGENT_NAME env var, the --agent / --use / --tools / --model / --system-prompt flags on fortify serve. The module:attr spec replaces all of them.
require_user_or_sdk_token dual-auth helper. The two surfaces stay separate by design.
Legacy WS /v1/projects/{id}/serve route, the bearer half of the old dual-auth GET /v1/projects/{id}/agents/{name} route, the legacy /agents page's wrap_langchain_agent demo, and two other examples that no longer earned their slot.
What's not in this scope
Step 2's positional-spec migration is LangChain / FortifyAgent first. OpenAI Agents / Pydantic-AI / Google ADK still go through the legacy --agent flow on the register side and aren't yet wired into fortify serve <spec>. The dispatch sites exist in fortify/cli/register/{openai,pydantic_ai,google}.py; a future step generalizes serve.
fortify chat keeps its --agent + --use shape for now. Same uvicorn-style migration sits in a follow-up.
No "list project's agents" endpoint on the dashboard side; the AgentPicker on /agents and /policies still queries per page. Cheap to add when the dropdown wants a global view.
Two notable bugs caught and fixed mid-flight
bootstrap.py had load_dotenv(override=True) — which clobbered shell-set env vars with whatever was in .env. Flipped to override=False to match the convention every other tool (uvicorn, vite, cargo, npm) follows. The user was hitting this because their .env had a stale FORTIFY_AGENT_NAME overriding their fresh shell prefix.
register/main.py called load_dotenv() inside add_parser — which fired on parser construction, polluting the process env whenever the top-level CLI was built (including for unrelated subcommand tests). Moved to main() via bootstrap() so it only fires at actual invocation.
Dashboard verify-email page hung at "Verifying…" forever under React 18+ StrictMode. useMutation + fire-on-mount + StrictMode's simulated unmount/remount tore down the mutation observer mid-flight, and the remount's fresh observer never reconnected to the in-flight request. Converted to useQuery keyed by token; query state lives on the global QueryClient and survives the cycle.
Verification

Platform API     234 passed
Dashboard       28 passed (Vitest + RTL)
Top-level       702 passed across the repo
No DEFAULT_PROJECT_ID / FORTIFY_AGENT_NAME / resolve_agent_name / require_user_or_sdk_token references left in code (only intentional "this was removed" mentions in docs).

Commit ledger
17 commits across 5 phases, ordered as shipped:

Phase 3b–c — email verification, password reset, Google OAuth.
Phase 4 (5 steps) — orgs + projects + members + invitations on the platform.
Phase 5 (5 steps) — every dashboard page rebuilt around active org/project state.
Phase 6 (3 steps) — token-implicit project for CLI routes; legacy {project_id} URL + dual-auth helper retired.
Phase 7 (2 steps) — register-time default policy + signed bundle; uvicorn-style serve.
Plus two surgical fixes (fix(dashboard): verify-email page stuck and fix(cli, platform): percent-encode bearer for WS subprotocol) caught during smoke testing.


guillaume-hexamind:
### Code review

Scope: `/platform` only. Found 7 issues, ordered by severity.

**1. `ws_chat` WebSocket has no authentication** (bug)

`ws_serve` was upgraded to require a bearer token via `ws_require_project`, but the sibling consumer socket `ws_chat` still calls `websocket.accept()` with no auth check and relays received JSON into the project's serve process via `relay_to_serve`. Any unauthenticated client who knows/guesses a `project_id` can connect, eavesdrop on agent traffic, and inject payloads — bypassing every org-membership check this PR adds.

https://github.com/HexamindOrganisation/coolagents/blob/5c8b2a7f7ec9de18a67f3d929d38f0ca88f93ac9/platform/api/main.py#L961-L975

**2. Privilege escalation: any admin can promote to `owner`** (bug)

`PATCH /v1/orgs/{org_id}/members/{user_id}` is gated by `require_org_admin` (admin or owner) and calls `change_member_role`, which only guards against demoting the last owner — there is no caller-rank >= target-role check. An admin can PATCH their own membership to `{"role":"owner"}` and seize the org. The invitation path enforces this via `_can_invite_role`, so the two paths are inconsistent.

https://github.com/HexamindOrganisation/coolagents/blob/5c8b2a7f7ec9de18a67f3d929d38f0ca88f93ac9/platform/api/services.py#L259-L285

**3. Invitation accept leaks the invitee's email address** (bug)

`InvitationEmailMismatch` is raised with a message embedding the invited email, and the handler forwards `str(exc)` verbatim as the HTTP 403 detail. Any authenticated user holding an invitation id can probe `/invites/{id}/accept` and have the server confirm the exact email of the intended invitee (PII disclosure).

https://github.com/HexamindOrganisation/coolagents/blob/5c8b2a7f7ec9de18a67f3d929d38f0ca88f93ac9/platform/api/services.py#L485-L489

**4. `cookie_secure=False` hardcoded with no env toggle** (bug)

The session cookie (and the OAuth CSRF cookie at L350) hardcode `cookie_secure=False`. The adjacent comment says "Toggle via env when we deploy," but no such override exists. Deployed over HTTPS as-is, the session JWT cookie ships without the `Secure` flag. Note `FORTIFY_ALLOW_DEV_USER_HEADER` already has an env gate — this should too.

https://github.com/HexamindOrganisation/coolagents/blob/5c8b2a7f7ec9de18a67f3d929d38f0ca88f93ac9/platform/api/auth.py#L238-L243

**5. `require_project` skips biscuit signature verification** (bug)

`_validate_sdk_token` runs both a signature gate (`parse_envelope` + `verify_token`) and a revocation lookup, and its docstring claims it is used "indirectly by `require_project`." But `require_project` only calls `find_token_by_secret` (a plain secret-string lookup) — the signature gate is absent, while `ws_require_project` and `optional_dev_token` both verify. SDK write routes gated by `require_project` (e.g. `POST /v1/agents`) accept any string matching a stored secret without verifying the biscuit chains to the platform key. Defense-in-depth/consistency gap plus an inaccurate docstring.

https://github.com/HexamindOrganisation/coolagents/blob/5c8b2a7f7ec9de18a67f3d929d38f0ca88f93ac9/platform/api/main.py#L171-L187

**6. `accept_invitation` silently drops role upgrades for existing members** (bug)

When the accepting user is already a member, the existing-member branch reuses the row but never sets `member.role = invitation.role`. Re-inviting an existing `member` as `admin` and accepting marks the invite consumed but leaves the role unchanged — a silent no-op the members UI implies should work.

https://github.com/HexamindOrganisation/coolagents/blob/5c8b2a7f7ec9de18a67f3d929d38f0ca88f93ac9/platform/api/services.py#L491-L507

**7. New user can be left with no org if `create_org` fails** (bug)

`on_after_register` comments that the personal-org creation shares "the same transaction lineage as the User insert so any downstream failure rolls back together." But FastAPI-Users commits the user row before invoking the hook, and `create_org` issues its own `commit()`. A failure inside `create_org` leaves a committed user with no personal org — the exact empty state the hook exists to prevent; the "rolls back together" invariant does not hold.

https://github.com/HexamindOrganisation/coolagents/blob/5c8b2a7f7ec9de18a67f3d929d38f0ca88f93ac9/platform/api/auth.py#L144-L172

platform/api/main.py
```diff
@@ -1,5 +1,7 @@
 import base64
+import hashlib
```
guillaume-hexamind, line 2: Non blocking but main is starting to get huge and holding multiple unrelated domains in one module. A refactor will be needed at some point to keep it maintainable, for example something like:

main.py            # app, lifespan, CORS, router mounting only (~120 lines)
dependencies.py    # require_user / require_org_* / require_project_* / ws_require_project
routers/
  agents.py        # /projects/{id}/agents/* + SDK /agents/*
  tokens.py        # /projects/{id}/tokens/*
  orgs.py          # /orgs CRUD + members
  invites.py       # /orgs/{id}/invites + /invites/{id}/*
  projects.py      # /orgs/{id}/projects + /projects/{id}
  ws.py            # /serve + /chat
  keys.py          # /.well-known/keys + /me/key

platform/api/main.py
```diff
@@ -505,4 +974,645 @@ async def ws_chat(websocket: WebSocket, project_id: str) -> None:
...
+
+
+# ---------------------------------------------------------------------------
+# M3 Phase 4 step 5 — Project CRUD
+#
+# Create + list live under /orgs/{org_id}/projects (need to know the
+# org); read + rename live under /projects/{project_id} (the project
+# row knows its own org). DELETE intentionally not shipped here —
+# cascade across Agent / DevToken / AgentVersion / Tool needs its
+# own focused commit.
```
guillaume-hexamind, line 1642: Indeed, good to keep in mind

guillaume-hexamind (commented):
A couple of the claude code comments to address;
But the logic looks good, it seems good to merge once the comments are addressed!

platform/api/main.py
quanghexa94: Yes will refactor in a follow-up PR

quanghexa94:
All comments are resolved or commented

PR #24, opened 2026-06-05
feat(fortify): Adapter policy bundle pull — real platform enforcement

guillaume-hexamind:
feat(fortify): Adapter policy bundle pull — real platform enforcement

  Summary

  Replaces the allow-all placeholder policy in all four adapters (LangChain, OpenAI, Google ADK, pydantic-ai) and create_agent
  with real platform policy pull, enforcement, and refresh. Adds a PolicyBinding abstraction that holds an enforcer and the
  source that keeps it current.

  Key changes

  - security/binding.py (new): PolicyBinding pairs a PolicyEnforcer with an optional PolicySource. refresh() / refresh_async()
  hot-swap enforcer.policy in place, so tools that closed over the enforcer pick up the new policy at the next run boundary.
  Adds ResolvedPolicy and resolve_policy().
  - resolve_policy(): precedence is FORTIFY_LOCAL_POLICY override → platform (client / api_key / FORTIFY_KEY) → raise. Eager
  and fail-loud — a platform 404 propagates.
  - Refresh is fail-soft: fetch/verification failures log a warning and keep the last verified policy; it never downgrades to
  an unverified one.
  - build_enforcer() (new in enforcer.py): the single place that pairs an engine with its audit sink, so the six surfaces (four
  adapters, the OpenAI runner, FortifyAgent.enforce_policy) don't each repeat the audit.configure wiring.
  - PolicySource protocol + implementations in security/source.py: PlatformPolicySource (ETag/304-aware),
  BundleDirPolicySource, YamlPolicySource.
  - Single source of truth for FORTIFY_LOCAL_POLICY: the resolution logic and the SignaturePolicy / REQUIRE_SIGNATURE matrix
  now live only in security/source.py (loader re-imports), removing a divergent duplicate — loader.py drops ~286 lines.

  Breaking changes

  1. Adapters are deny-by-default. Code relying on the old allow-all behavior now needs platform connectivity; unlisted tools
  are denied.
  2. Agents must be registered before first use. A missing agent surfaces the platform 404 loudly — register with fortify 
  register. Auto-register on construction was removed because it was a silent remote write that masked typos and wrong-scope
  keys.

 Deferred

  Opt-in auto-register (FORTIFY_AUTO_REGISTER) with a uniform param across all surfaces and WARNING-level logging with scope.


quanghexa94:
## 🔴 Block on these

### `_should_bind_policy` heuristic misfires on shared `FORTIFY_KEY`

`fortify/agents/factory.py` — `_should_bind_policy`.

A dev with `FORTIFY_KEY` set (for *some other* agent on the platform) calls `create_agent(name="my_local_prototype")` for an unregistered agent → silent auto-bind → 404 → **raise at construction**. Previous behaviour was "construct, enforce_policy later." The error appears far from the cause (after the LangChain graph is built).

**Fix**: default `bind_policy=False`, force opt-in. Or require both `FORTIFY_KEY` AND a `FORTIFY_BIND_AGENTS=1` toggle. Either works.

### `_should_bind_policy` ignores `FORTIFY_LOCAL_MODE` from PR #25 (I will update PR #25 to cover this new change no worry)

Same function. Anyone with `FORTIFY_KEY` + `FORTIFY_LOCAL_MODE=1` set and calling `create_agent(name="X")` directly hits the auto-bind path → platform fetch fires → local-mode intent silently violated. This re-opens the mode-B half-state PR #25 was meant to close. (I will adapt PR #25 to cover this point but worth informing) 

**Fix**: one-line guard before the env-detect branch:

```python
if os.environ.get("FORTIFY_LOCAL_MODE", "").strip().lower() in ("1", "true", "yes", "on"):
    return False
```

Plus `test_create_agent_skips_auto_bind_under_FORTIFY_LOCAL_MODE`. Cleaner: lift `audit._local_mode_active()` into a public helper and call from both sites.

---

## 🟡 Worth a thread

### New import cycle papered over with TYPE_CHECKING

`fortify/audit.py` now defers `from fortify.security.decision import Decision` to `TYPE_CHECKING`. Reason: the new `binding.py` creates `audit → security/__init__ → binding → enforcer → audit`.

**Fix**: either (a) add `# why this TYPE_CHECKING is here` comment so the next maintainer doesn't "fix" it back, OR (b) restructure `binding.py` to import `enforcer` lazily inside `resolve_policy`'s function body so the cycle never forms.

### `enforce_policy(source=X)` asymmetric defaults are misuse-prone

`fortify/agents/factory.py::FortifyAgent.enforce_policy` now accepts `source=None`. The (policy, source) matrix has four cells but the docstring only covers three.

**Fix**: add a test asserting `enforce_policy(None, source=X)` is rejected (or explicitly documented as no-op). Currently a caller passing `source=X` and forgetting `policy` will silently get no enforcement.

### `bind_policy=True` validation lives in private helper

`_should_bind_policy` raises `ValueError("create_agent(bind_policy=True) requires name=...")` from inside a dispatch helper. The message is good but conventionally argument validation goes at the public API boundary.

**Fix**: hoist the check into `create_agent` itself, before the dispatch.

---

## End-state checklist

Before merge:
- [ ] `_should_bind_policy` heuristic flipped to opt-in
- [ ] coordinate `bootstrap.py` conflict with #25 author
- [ ] TYPE_CHECKING smell addressed (comment or restructure)
- [ ] `enforce_policy(source=, policy=None)` test or doc
- [ ] `bind_policy=True` validation moved to public surface

Otherwise the logic is good

guillaume-hexamind:
All comments (besides FORTIFY_LOCAL_MODE) fixes, checks passed

PR #25, opened 2026-06-08
feat: local mode and chat polish

quanghexa94:
# Local mode + chat polish

Closes two rough edges that have been bugging the inner-loop dev experience.

## Why

Before this PR:

1. **`fortify chat` could quietly post audit events to the platform** if you had `FORTIFY_KEY` in your `.env` from an earlier session. The command was nominally "local" (it loaded the policy from disk), but every adapter wrapper called `audit.configure()` independently, picked the key up from env, and shipped events to a platform that didn't know about the agent. Mode B from the audit-pipeline doc — silent, confusing.

2. **The chat REPL threw away the structured `Decision`** that the policy enforcer produced. When a tool got denied you saw a generic tool-error string in the agent's response, but the `reason` / `violations` / `hint` / `error_type` — all the data the dashboard renders — were captured by the enforcer and immediately discarded. So the platform got nice diagnostics; the dev at the terminal got nothing.

Two smaller polish items also got folded in: `fortify chat`'s `--agent` flag now accepts the same `module:attr` spec as `fortify serve` (was a surprising asymmetry), and a small README block clarifies when to use which workflow.

### 🔒 `fortify chat` is now full-local end-to-end

- New `FORTIFY_LOCAL_MODE` env flag. When set, `audit.configure()` returns `None` even with a key in env — no sender created, no requests made.
- `bootstrap()` gains a kwarg-only `local_only=False`. When `True`, it sets `FORTIFY_LOCAL_MODE=1` **before** `audit.configure()` runs. Any later adapter wrapper that re-configures audit checks the same env var and stays inert.
- `fortify chat` passes `local_only=True` automatically.
- One INFO log line — `audit suppressed: FORTIFY_LOCAL_MODE=1 (...)` — fires the first time a key was actually present, so you know why audit went quiet. No log spam for OSS users who never set a key.
- New WARNING when `FORTIFY_KEY` *and* `FORTIFY_LOCAL_POLICY` are both set — almost always a forgotten env entry.

### ⛔ Inline policy-decision panel in the chat REPL

After every turn, denies and needs-approval calls render as a rich panel showing reason / role / violations / hint / error_type. Allows are muted by design — a panel per tool call would be noise; the point of the feed is "what just got blocked." A `--show-allow` flag can land later if anyone asks.

```
> read /etc/passwd

[agent's response here]

╭─ ⛔ deny · read_file ─────────────────────────────────────╮
│ path outside allowed glob                                 │
│ error_type: policy_denied                                 │
│ role: analyst                                             │
│ violations:                                               │
│   • path_outside_glob                                     │
│ hint: {"glob": "/workspace/**"}                           │
╰───────────────────────────────────────────────────────────╯
```

Mechanically: `PolicyEnforcer` gains a second injection slot (`decision_observer`) alongside `audit_sender`. Chat injects `deque.append` as the observer; the chat loop drains the deque between turns. Broken observers can't break enforcement (try/except + log).

### 🔗 `module:attr` spec parity with `fortify serve`

Both commands now accept the same two forms:

```bash
fortify chat --agent researcher                           # by id
fortify chat --agent examples.customer_bot:agent          # by module:attr
fortify serve --agent examples.customer_bot:agent         # (already worked)
```

The spec path is local-only on chat — no manifest re-registration, no cloud round-trip, no policy fetch. The user's module is responsible for having called `.enforce_policy(...)` before exporting; chat just wires its observer + approval handler into the agent's existing enforcer.

### 🧭 README "Which path do I pick?" block

New section between the two Quickstarts answering the question every new dev asks in week one — chat (inner loop, no platform) vs. serve + Playground (team loop, full platform). Small table covering the four cells (needs platform / audit destination / policy edits visible at / best for).

### 🩹 Drive-by

Flipped `load_dotenv(override=True)` → `override=False` in `bootstrap.py` to match a pre-existing test (`test_bootstrap_loads_requested_env_file`) that was failing on main. The test's docstring already pinned the "shell wins over .env, matching uvicorn/vite/cargo/npm" contract; the code had drifted.

## Commits

| | |
|---|---|
| `2f779e2` | `feat(fortify): explicit local mode; audit honors it` |
| `ece0e6b` | `feat(cli/chat): inline policy-decision panel` |
| `dd77207` | `feat(cli/chat): accept module:attr spec like serve` |
| `a09ad1c` | `docs: chat vs Playground, local vs platform` |
| `354efd1` | `fix(local-mode-pr): review-pass tidying` |

Each commit is independently buildable + ruff-clean + pytest-green, so `git bisect` and PR-by-commit review both work.

## Test plan

- [x] `ruff check fortify tests` → clean
- [x] `pytest tests/` → **773 passed, 2 skipped** (+13 new tests vs. main; the 2 skipped are integration tests gated behind the `integration` marker)
- [x] Manual smoke checks:
  - [x] `unset FORTIFY_KEY && fortify chat --agent example_agent` → runs, no audit, quiet startup
  - [x] `FORTIFY_KEY=test fortify chat --agent example_agent` → runs, one INFO line "audit suppressed", no requests to platform
  - [x] `FORTIFY_KEY=test FORTIFY_LOCAL_POLICY=/tmp/... fortify chat ...` → adds the WARNING line, still local
  - [x] `fortify chat --agent examples.customer_bot:agent` → loads via spec, welcome banner shows `(spec)` source
  - [x] Ask the agent to call a denied tool → red ⛔ panel renders with reason + violations + JSON hint
  - [x] `fortify serve --agent ... &&` (platform running) → audit fires (regression check)

## Files changed

```
17 files changed, +1243 net LOC
```

| Area | Files |
|---|---|
| Audit gate | `fortify/audit.py`, `fortify/bootstrap.py`, `docs/audit-pipeline.md` |
| Enforcer + loader plumbing | `fortify/security/enforcer.py`, `fortify/agents/factory.py`, `fortify/agents/loader.py`, `fortify/cli/_common.py` |
| Chat REPL | `fortify/cli/chat.py` |
| Docs | `README.md` (Quickstart block), `docs/audit-pipeline.md` (local-mode subsection) |
| Tests | new `tests/security/test_enforcer_observer.py`, expanded `tests/cli/test_chat.py` + `tests/audit/test_configure.py` + `tests/test_bootstrap.py`, +4 stub updates in `tests/agents/test_*.py` |

## Known follow-ups (deliberately out of scope)

- `bootstrap(local_only=True)` mutates `os.environ` permanently — once set, no toggle-off within the same process. Realistic workflow doesn't hit this; one line in the docstring would be belt-and-braces.
- `--use` + `--agent module:attr` is undefined behaviour today (the registered agents from `--use` are silently ignored on the spec path). Needs a small UX decision — warn / error / docs.
- `_apply_approval_handler` has no test for its no-GuardedTool warning path. Pre-existing gap I mirrored for symmetry with the new `_apply_decision_observer`; worth a small follow-up.

Happy to address any of these in the review thread.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,8 @@ Useful next commands:
 
 ```bash
 fortify chat --list-agents
-fortify chat --agent researcher
+fortify chat --agent researcher                                      # by id
+fortify chat --agent examples.customer_bot:agent                     # by module:attr (same shape as `fortify serve`)
 fortify chat --use examples/file_agents.py --agent workspace_explorer
 fortify chat --use examples/research_agents.py --agent update_researcher
 ```
--- a/README.md
+++ b/README.md
@@ -76,7 +76,7 @@ The two Quick Starts above aren't competing — they answer different questions.
 
 | Path | Needs platform? | Audit destination | Policy edits visible at | Best for |
 |------|-----------------|-------------------|--------------------------|----------|
-| `fortify chat --agent ...` | No | Local terminal panel | YAML/bundle reload from disk | Inner loop, policy authoring |
+| `fortify chat --agent ...` | No | Local terminal panel | Edit + restart (hot-reload only when `FORTIFY_LOCAL_POLICY` is set) | Inner loop, policy authoring |
 | `fortify serve --agent ...` + Playground | Yes | ClickHouse via `/v1/audit/decisions` | Per-turn fetch from dashboard | Team review, demos, integration testing |
 
 Both commands accept either a plain agent id (`--agent researcher`) or a uvicorn-style `module.path:attr` spec (`--agent examples.customer_bot:agent`), so the same entry-point string works in both workflows.
--- a/README.md
+++ b/README.md
@@ -66,6 +66,21 @@ The included local agent lives in `examples/example_agent/`, and the CLI can als
 - code-defined agents registered from `examples/file_agents.py`
 - code-defined research agents registered from `examples/research_agents.py`
 
+## 🧭 Which path do I pick?
+
+The two Quick Starts above aren't competing — they answer different questions.
+
+**Inner loop — `fortify chat`.** A single-process REPL against a local or builtin agent. No platform, no Docker, no browser. The chat command sets `FORTIFY_LOCAL_MODE=1` automatically so audit stays on your machine even if `FORTIFY_KEY` lives in your `.env` from an earlier session. Denies and approval-required calls render as inline panels in the terminal — same `Decision` data the platform would log, surfaced where you're iterating. Reach for `chat` when you're authoring a policy YAML, tweaking a tool, or shaping a system prompt.
+
+**Team loop — `fortify serve` + dashboard Playground.** Same agent code, but the policy + decisions round-trip through the platform. You get auditable decisions in ClickHouse, the shared Playground UI, and live policy edits via the dashboard. Reach for `serve` when you're collaborating on an agent's behaviour, debugging a production-like trace, or demoing.
+
+| Path | Needs platform? | Audit destination | Policy edits visible at | Best for |
+|------|-----------------|-------------------|--------------------------|----------|
+| `fortify chat --agent ...` | No | Local terminal panel | YAML/bundle reload from disk | Inner loop, policy authoring |
+| `fortify serve --agent ...` + Playground | Yes | ClickHouse via `/v1/audit/decisions` | Per-turn fetch from dashboard | Team review, demos, integration testing |
+
+Both commands accept either a plain agent id (`--agent researcher`) or a uvicorn-style `module.path:attr` spec (`--agent examples.customer_bot:agent`), so the same entry-point string works in both workflows.
+
 ## 🚀 Quick Start — Platform
 
 To run the full Fortify control plane locally (FastAPI backend + dashboard + your local agent serving over WebSocket), you need **three terminals**. The Makefile has a target that prints the recipe:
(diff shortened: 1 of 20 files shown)


guillaume-hexamind:
### Code review

Found 1 issue:

1. `_apply_local_override` silently drops the new `decision_observer`, so `fortify chat` renders no decision panels when `FORTIFY_LOCAL_POLICY` is set.

`load_builtin_agent` / `load_local_agent` accept `decision_observer` and forward it to `enforce_policy(...)` on the normal path, but when `FORTIFY_LOCAL_POLICY` is set they return early through `_apply_local_override(agent, approval_handler)`, which neither accepts nor forwards `decision_observer`. The observer is dropped before reaching `enforce_policy`, so the headline decision-panel feature is disabled — deterministically — for exactly the local-policy-override dev workflow this PR targets.

https://github.com/HexamindOrganisation/coolagents/blob/14842e7fe573684567fb8a384b7a30563ab1d195/fortify/agents/loader.py#L56-L78

guillaume-hexamind:
### Code review (additional findings)

Two further issues, lower confidence than the one above (each was deliberately deferred by the author or is a milder concern), included on request:

1. `--use` + `--agent module:attr` is undefined: when the agent name contains a colon, `build_runtime` branches to `_build_runtime_from_spec` and silently ignores any agents registered by `--use`, with no warning or error. The PR description acknowledges this as an out-of-scope follow-up.

https://github.com/HexamindOrganisation/coolagents/blob/14842e7fe573684567fb8a384b7a30563ab1d195/fortify/cli/_common.py#L81-L88

2. Bare `assert payload is not None` in production code. Under `python -O`/`-OO` asserts are stripped, so if the invariant were ever violated the next line would raise a confusing `TypeError: 'NoneType' object is not subscriptable` instead of the clear message. A `raise` would be more robust (note: this duplicates a pre-existing pattern in `_common.py`).

https://github.com/HexamindOrganisation/coolagents/blob/14842e7fe573684567fb8a384b7a30563ab1d195/fortify/agents/loader.py#L477-L479

guillaume-hexamind (commented):
Logic looks good to me! A couple of comments from claude

quanghexa94:
Yes they're corrected!

PR #26, opened 2026-06-08
feat: add GitHub actions and fix linting

quanghexa94:
Adds two GitHub Actions workflows:

- **`tests.yml`** — runs on every push to `main` and every PR
  targeting `main`. Three parallel jobs: SDK (uv + ruff + pytest +
  opa), Platform API (uv + pytest), Dashboard (pnpm: lint +
  typecheck + tests).

- **`release.yml`** — manual trigger only. Bumps the SDK version via
  `uv version --bump`, re-runs the full test matrix as a gate, builds
  sdist + wheel via `uv build`, publishes to PyPI via trusted
  publishing (OIDC, no token), tags + commits, opens a GitHub release.
  Supports TestPyPI dry-runs via the `target` input.

Includes the supporting changes needed for day-one green:

- Renames the SDK's PyPI distribution name from `fortify` (taken on
  PyPI since 2014) to `hexgate`. Import path stays `from fortify
  import ...` (Pillow → PIL pattern).
- Catches the codebase up on `ruff format` (49 files, pure whitespace).
- Fixes a pre-existing failing test (`bootstrap` `override=True → False`).
- Resolves the dashboard's lint blockers (1 unused var, 2 explicit
  `any`, 1 unused eslint-disable) and demotes two opinionated React 19
  / fast-refresh rules to warnings.

## One-time setup before the first release run

PyPI trusted publisher registration:
- https://pypi.org/manage/account/publishing/ → "Add a new pending publisher"
  - Project Name: `hexgate`
  - Owner: `HexamindOrganisation`
  - Repository: `coolagents`
  - Workflow: `release.yml`
  - Environment: `pypi-release`
- Repeat on https://test.pypi.org/manage/account/publishing/

GitHub Environment:
- Settings → Environments → new env `pypi-release`
- Required reviewers + `main`-only deployment branch

## Test plan
- [x] `make check` → 735 pass / 2 skip
- [x] `make platform-api-test` → 300 pass / 1 skip
- [x] `pnpm {lint,typecheck,test}` (dashboard) → 33 pass, 0 errors

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -158,7 +158,19 @@ jobs:
       - name: Bump version
         id: bump
         run: |
-          if [ -n "${{ inputs.version }}" ]; then
+          if [ "${{ inputs.target }}" = "testpypi" ]; then
+            # TestPyPI runs use an ephemeral dev version tied to the
+            # workflow run number. Each dry-run gets its own slot
+            # (`0.0.0.dev<run>`) so iterating against TestPyPI never
+            # collides on re-upload, and the version is obviously
+            # non-canonical so it can't be mistaken for a real release.
+            # We deliberately ignore inputs.version / inputs.version-type
+            # here — the user's choice is what they intend to ship to
+            # *real* PyPI, and that decision shouldn't be consumed by a
+            # dry run. The pyproject.toml change is never committed back
+            # to main on this path (see step 6).
+            uv version "0.0.0.dev${{ github.run_number }}"
+          elif [ -n "${{ inputs.version }}" ]; then
             # Explicit version (e.g. "1.2.3") overrides version-type.
             uv version "${{ inputs.version }}"
           else
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,234 @@
+name: release
+
+# Manual trigger only — click "Run workflow" in the Actions tab.
+# Pick a version-type (or explicit version), the workflow runs the full
+# test matrix as a gate, bumps the SDK version, builds sdist + wheel,
+# publishes to PyPI via trusted publishing (OIDC — no token), commits
+# the bump, tags it, and opens a GitHub release with auto-generated notes.
+#
+# ----------------------------------------------------------------------
+# PyPI trusted-publisher registration — REQUIRED before the first run:
+#
+#   https://pypi.org/manage/account/publishing/  →  "Add a new pending publisher"
+#     PyPI Project Name:  hexgate
+#     Owner:              HexamindOrganisation
+#     Repository name:    coolagents
+#     Workflow name:      release.yml
+#     Environment name:   pypi-release
+#
+# Repeat on https://test.pypi.org/manage/account/publishing/ for the
+# TestPyPI dry-run path.
+#
+# Also create a GitHub Environment named `pypi-release`:
+#   Settings → Environments → New environment → "pypi-release"
+#     - Required reviewers: at least one maintainer
+#     - Deployment branches: main only
+# ----------------------------------------------------------------------
+
+on:
+  workflow_dispatch:
+    inputs:
+      version-type:
+        description: "Bump kind (ignored when 'version' is set)."
+        type: choice
+        required: true
+        default: patch
+        options:
+          - patch
+          - minor
+          - major
+      version:
+        description: "Explicit version like 1.2.3. Overrides version-type."
+        type: string
+        required: false
+        default: ""
+      dry-run:
+        description: "Run everything except publish, commit, and tag."
+        type: boolean
+        required: false
+        default: false
+      target:
+        description: "Where to publish. Use 'testpypi' for a dry run on the test index."
+        type: choice
+        required: true
+        default: pypi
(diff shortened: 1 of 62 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,234 @@
...
+# Pick a version-type (or explicit version), the workflow runs the full
+# test matrix as a gate, bumps the SDK version, builds sdist + wheel,
+# publishes to PyPI via trusted publishing (OIDC — no token), commits
+# the bump, tags it, and opens a GitHub release with auto-generated notes.
+#
+# ----------------------------------------------------------------------
+# PyPI trusted-publisher registration — REQUIRED before the first run:
+#
+#   https://pypi.org/manage/account/publishing/  →  "Add a new pending publisher"
+#     PyPI Project Name:  hexgate
```
guillaume-hexamind, line 13: I think the team agreed on HX Gate (or hxgate), not hexgate here.

.github/workflows/release.yml
```diff
@@ -0,0 +1,234 @@
...
+#     Owner:              HexamindOrganisation
+#     Repository name:    coolagents
```
guillaume-hexamind, line 15: Will need to be renamed too later!

pyproject.toml
```diff
@@ -3,9 +3,14 @@ requires = ["setuptools>=68", "wheel"]
 build-backend = "setuptools.build_meta"
 
 [project]
-name = "fortify"
+# PyPI distribution name (`pip install hexgate`). The import name stays
+# `fortify` (`from fortify import ...`) — same shape as Pillow → PIL or
+# beautifulsoup4 → bs4. The bare `fortify` slot on PyPI was claimed in
+# 2014 by an unrelated abandoned project, so we publish under the brand
+# name and let import-time keep the existing identifier.
+name = "hexgate"
```
guillaume-hexamind, line 11: Same comment!

guillaume-hexamind:
### Code review

Found 3 issues:

1. `release.yml`: the release job pushes to `main` with no in-workflow branch guard. `workflow_dispatch` has no branch restriction, so the job can be dispatched from any branch; tests then run against that branch and `git push origin HEAD:main` overwrites `main` with it. The only protection is the external `pypi-release` GitHub Environment branch rule (repo settings, not the YAML) — a one-line `if: github.ref == 'refs/heads/main'` on the job would add defense-in-depth.

https://github.com/HexamindOrganisation/coolagents/blob/5c3471e08d1c97a383938fb023cdd14a1673162f/.github/workflows/release.yml#L216-L225

2. `eslint.config.js`: `react-hooks/set-state-in-effect` is not a rule exported by the pinned `eslint-plugin-react-hooks@^7.1.1` (that version ships only `rules-of-hooks` and `exhaustive-deps`). Flat config silently ignores the unknown rule, so it never fires — the intended demotion-to-`warn` has no effect. Either bump the plugin to a version that ships the rule, or drop the entry.

https://github.com/HexamindOrganisation/coolagents/blob/5c3471e08d1c97a383938fb023cdd14a1673162f/platform/dashboard/eslint.config.js#L34-L36

3. `release.yml`: a non-dry-run `testpypi` release bumps and publishes a version but never commits it. The "Bump version" step runs unconditionally and "Publish to TestPyPI" runs for `target == 'testpypi'`, but "Commit + tag" is gated to `target == 'pypi'`. The bump is lost, so the next real PyPI release re-derives the same version that was already consumed on TestPyPI.

https://github.com/HexamindOrganisation/coolagents/blob/5c3471e08d1c97a383938fb023cdd14a1673162f/.github/workflows/release.yml#L206-L217

guillaume-hexamind (commented):
Outside of claude minor comments;  I approve of the logic!
The only point to discuss is the name of the project name -> hxgate rather than hexgate was what we aggreed on, but still up to discussion

.github/workflows/release.yml
winlp4ever: doesn't matter for now. later we can change the name if needed. but claude insists hx-gate is not a good name at all.

.github/workflows/release.yml
winlp4ever: same

pyproject.toml
winlp4ever: same, this PR is only to have an end-to-end testing - release - publishing on pypi. The naming is not yet finalized so it's not really that important

winlp4ever:
The naming is only for having something on pypi to have full pipeline merged - test -release - publish. Creating a lib on pypi is easy so later we can change to whatever name we want

.github/workflows/release.yml
guillaume-hexamind: makes sense, I am good with that

winlp4ever:
For claude comments:

1. The pypi-release GitHub Environment is configured with main-only deployment branches (Settings → Environments → pypi-release → "Selected branches and tags" rule). The job declares environment: pypi-release, so a workflow_dispatch from any non-main branch gets rejected at environment-access time. So a protection is already in place 

2. It's false positive from Claude
3. Yes i will correct this

PR #30, opened 2026-06-11
fix(adapters): propagate user scope to tool execution in OpenAI streaming and Google ADK sync runs

guillaume-hexamind:
Fix: propagate User scope to tool execution in OpenAI streaming and Google ADK sync runs

  **Problem**

  Both adapters lost the User context var (_CURRENT_USER) at the point tools actually execute, so enforcers fell back to no active role:

  - OpenAI run_streamed — Runner.run_streamed synchronously spawns the agent loop as a background asyncio.Task (which snapshots context at creation via
  copy_context()). The scope was only opened later inside the stream_events() wrapper, so the task — where tools fire — captured _CURRENT_USER = None.
  - Google ADK sync run — ADK's Runner.run drives the agent loop in a separate OS thread (threading.Thread + asyncio.run). Context vars don't cross thread
  boundaries, so user.sync_scope() in the calling thread was invisible to the tools running in the worker thread.

  **Fix**

  Align the execution context with the scope in both adapters:

  - OpenAI: wrap the Runner.run_streamed(...) call itself in user.sync_scope() so the spawned background task inherits _CURRENT_USER in its context
  snapshot. The existing stream_events wrapper is kept for consumer-side scope / exit semantics.
  - Google: drive Runner.run_async inline on a per-call event loop (instead of ADK's threaded Runner.run), keeping tool execution in the scoped calling
  thread. Matches ADK's per-call-loop sync semantics and cleans up via aclose().

  run/run_sync/run_async paths that already held the scope across execution were unaffected.

  **Tests**

  - OpenAI: assert the User is active at the moment Runner.run_streamed is invoked (the task's context-snapshot point).
  - Google: assert sync run delegates to run_async under a live scope, and that the User stays visible across await points (the actual tool-call sites).
  Both new Google tests were verified to fail on the old implementation and pass on the fix.

  Full adapter suite green (16 OpenAI + 14 Google).

  **Note**

  Calling Google's sync run from within an already-running event loop now raises (can't nest loops on one thread) — an anti-pattern that ADK documents
  sync run as not intended for ("local testing and convenience only").

### DIFF
--- a/hexgate/adapters/google/runner.py
+++ b/hexgate/adapters/google/runner.py
@@ -79,16 +79,32 @@ class HexgateRunner:
         user: User,
         **kwargs: Any,
     ) -> Generator[Any, None, None]:
-        """Run the Google ADK agent synchronously, yielding events."""
+        """Run the Google ADK agent synchronously, yielding events.
+
+        ADK's ``Runner.run`` drives the agent loop in a worker thread whose
+        context cannot see our :class:`User` scope, so the tools' enforcers
+        lose the active role. We drive ``run_async`` inline on a per-call loop
+        instead, keeping execution in this scoped thread.
+        """
         self._setup_observability()
         self._binding.refresh()  # per-run policy pull; 304 when unchanged
         with user.sync_scope(), self._propagate(user):
-            yield from self._runner.run(
+            agen = self._runner.run_async(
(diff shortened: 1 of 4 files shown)


quanghexa94:
LGTM

PR #35, opened 2026-06-17
fix(hexgate/audit): deliver events emitted from executor threads

guillaume-hexamind:
## What
  `AuditSender.emit()` dropped events when called from a thread with no running
  loop. Sync LangChain tools run their policy decision (and the audit emit) on a
  `run_in_executor` worker thread under `ainvoke`, so every audit event from a
  sync-tool agent was silently lost.
 
  ## Fix
  - Capture the build-time loop in `__init__` (adapters build the sender inside
    the caller's async context).
  - When `emit()` is called off-loop, route the send to that loop via
    `loop.call_soon_threadsafe` instead of dropping it.

  On-loop behavior (async tools, OpenAI/Google runners) is unchanged.
  
  ## Test
  Added a regression test that emits via `run_in_executor` (as `BaseTool.ainvoke`
  runs a sync tool) and asserts the event reaches the wire. Verified end-to-end
  through deepagents: a sync tool under `ainvoke` now audits correctly.

### DIFF
--- a/hexgate/audit.py
+++ b/hexgate/audit.py
@@ -149,7 +149,13 @@ class AuditSender:
         # latch onto the first loop that drives them and reject any other
         # (e.g. a second asyncio.run()). Build them eagerly so configure()
         # stays sync, but track the loop and rebuild if it rotates.
-        self._loop: asyncio.AbstractEventLoop | None = None
+        #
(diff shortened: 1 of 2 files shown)


guillaume-hexamind:
### Code review

quanghexa94 (approved):

PR #36, opened 2026-06-17
fix(langchain): correct astream_events streaming + User scope teardown

guillaume-hexamind:
Two related fixes for streaming through the LangChain proxy (`HexgateLangchainAgent`). Both surfaced via the HexUI agent-server `/stream` route.

## 1. `astream_events` signature + forwarding

`astream_events` raised `TypeError: got multiple values for argument 'config'` on every call. The proxy exposed `version` as a positional 2nd arg and forwarded it positionally, but the wrapped graph's signature is `astream_events(input, config=None, *, version='v2', ...)` — `version` is keyword-only and the 2nd positional is `config`, so `version` bound to `config` and collided with the explicit `config=`.

Fix: make the proxy's `version` keyword-only with a `"v2"` default (matching base langchain and the sibling `ainvoke`/`astream` methods), and forward `config`/`version`/`**kwargs` by keyword. The recording-graph test stub now mirrors langgraph's real signature so a positional-`version` regression fails the test.

## 2. `User` scope teardown across Contexts

After (1), streaming surfaced `ValueError: Token was created in a different Context` as an unretrieved task exception. `User` set `_CURRENT_USER` via a `ContextVar` token and reset it on exit — but a token can only be reset in the Context it was created in, and async-generator finalizers (a consumer breaking early, then `aclose()` running in the event loop's finalizer task) run `__aexit__` in a different Context.

Fix: save/restore the previous value instead of using a token. `set()` carries no Context binding, so the foreign-context exit is harmless; for this var (default `None`) it's equivalent to a token reset, and nesting/normal completion are unchanged. Added regression tests that enter and exit in distinct copied Contexts (both fail on the token-based code).

## Verification
- 644 tests pass across runtime/adapters/agents/security/audit/streaming; ruff clean.
- Standalone repro streams cleanly; the "Token was created in a different Context" / unretrieved-task-exception noise is gone.

> Based on `gp/fix/audit_emition_loop`; merge after that lands (or retarget to `main`).

### DIFF
--- a/hexgate/adapters/langchain/agent.py
+++ b/hexgate/adapters/langchain/agent.py
@@ -2,7 +2,7 @@
 
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Any, AsyncIterator, Iterator
+from typing import TYPE_CHECKING, Any, AsyncIterator, Iterator, Literal
 
 from langchain_core.runnables import RunnableConfig
 from langfuse import get_client, propagate_attributes
@@ -134,10 +134,10 @@ class HexgateLangchainAgent:
     async def astream_events(
         self,
         input: dict[str, Any],
-        version: str,
         *,
         user: User,
         config: RunnableConfig | None = None,
+        version: Literal["v1", "v2"] = "v2",
         **kwargs: Any,
     ) -> AsyncIterator[dict[str, Any]]:
         """Stream the agent events asynchronously inside a User scope."""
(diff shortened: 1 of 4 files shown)


guillaume-hexamind:
### Code review

quanghexa94 (approved):

quanghexa94:
LGTM!

PR #38, opened 2026-06-18
feat(audit): add script to seed clickhouse table

victorludvig:
This script adds 820 rows of mock data to the audit table.
This mock data will be used to design the anomaly detection system.

800 rows of normal data are added, with tools randomly selected with 80% allow / 10% deny / 10% needs_approval.
The anomaly consists in 20 rows of denied access tools in a span of 5 minutes by a single user.

[Notion task](https://app.notion.com/p/Add-test-data-with-anomalies-in-ClickHouse-audit-db-382fb45dbae280738b65c141c7f51ecf?v=337fb45dbae2800ca10e000c2073872d&source=copy_link)

Screenshot 2026-06-17 at 17 36 25

### DIFF
--- a/platform/scripts/seed_audit.py
+++ b/platform/scripts/seed_audit.py
@@ -1,0 +1,179 @@
+from __future__ import annotations
+
+import random
+from datetime import datetime, timedelta, timezone
+from uuid import uuid4
+
+from clickhouse_connect.driver.client import Client
+
+import sys
+from pathlib import Path
+
+sys.path.insert(0, str(Path(__file__).parent.parent / 'api'))
+
+from services import DEFAULT_PROJECT_ID, DEFAULT_PROJECT_NAME
+from audit import _DECISION_COLUMNS, _DECISION_INSERT_SETTINGS
+from clickhouse import get_clickhouse
+
+# ── Agent & users ─────────────────────────────────────────────────────────────
+# Uses the dev default project id (imported above) so data is visible in the dashboard.
+# clear() scopes deletes to USER_IDS so real audit rows are not affected.
+
+AGENT_NAME = DEFAULT_PROJECT_NAME
+USER_IDS = ['Alice', 'Bob', 'Charlie']
+ANOMALY_USER = 'Bob'
+
+# ── Traffic shape ─────────────────────────────────────────────────────────────
+# Normal: 800 rows over 30 days, outcomes weighted 80% allow / 10% deny / 10% needs_approval.
+# Anomaly: Bob spikes 20 denies in a 5-minute window 10 days ago,
+# probing restricted tools (refund_customer, create_ticket).
+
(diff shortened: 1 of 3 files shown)


platform/scripts/seed_audit.py
```diff
@@ -0,0 +1,178 @@
...
+
+
+def build_rows() -> list[list]:
+    now = datetime.now(timezone.utc)
+    rng = random.Random(0)
+    return generate_normal_data(rng, now) + generate_anomalies(rng, now)
+
+
+def seed(client: Client) -> int:
+    clear(client)
```
guillaume-hexamind, line 140: Minor issue, but with clickhouse mutations being async, running twice a make seed-audit could create a race condition and 1600 entries instead of 800

Doing somethings like passing settings={'mutations_sync': 2} to the client.command in clear could solve the issue

platform/scripts/seed_audit.py
```diff
@@ -0,0 +1,178 @@
...
+        '',
+        user_id,
+        rng.choice(TOOL_NAMES),
+        'default',
+        outcome,
+        '' if outcome != 'deny' else 'permission_denied',
+        '' if outcome != 'deny' else 'User does not have permission',
+        [] if outcome != 'deny' else ['unauthorized_action'],
+        '',
+        '{}',
```
guillaume-hexamind: Minor, but current audit passes empty strings when no arguments are provided, that is then decoded into None. Does not change the anomaly and aggregation logics, but passing "" instead would align with real empty arguments tool calls.

platform/scripts/seed_audit.py
```diff
@@ -0,0 +1,178 @@
...
+        '',
+        ANOMALY_USER,
+        rng.choice(RESTRICTED_TOOLS),
+        'default',
+        'deny',
+        'permission_denied',
+        'User does not have permission',
+        ['unauthorized_action', 'policy_violation'],
+        '',
+        '{}',
```
guillaume-hexamind: same

guillaume-hexamind (commented):
A couple of minor comments easily fixable, but I approve the logic! We can merge once addressed

platform/scripts/seed_audit.py
victorludvig: For the next revision, I added the `mutations_sync` setting:
```
    client.command(f"ALTER TABLE policy_decision DELETE WHERE user_id IN ({users}) AND project_id = '{DEFAULT_PROJECT_ID}'",
                   settings={'mutations_sync': '2'})
```

platform/scripts/seed_audit.py
victorludvig: For the next revision, I changed the arguments from `'{}'` to `''`.

guillaume-hexamind (approved):

PR #39, opened 2026-06-19
Add start_date and end_date to audit API

victorludvig:
## Rev 2
- add `_ensure_utc` and `_prepare_date_range` functions to pre-process datetimes objects
- add corresponding unit tests

## What is changing
To filter decisions by time range in the audit window, we need to query the backend with a time range. Hence `start_date` and `end_date` are added to the audit API GET endpoints that allow filtering by date.
If these parameters are not passed, the date is filtered using the `since_hours` parameter, like it did before.

[Notion task](https://app.notion.com/p/Modify-backend-to-enable-start-date-end-date-filters-383fb45dbae28074b01bdc394054d47d?v=337fb45dbae2800ca10e000c2073872d&source=copy_link)

## Tests
- unit tests: `make platform-api-test`
- integration test: `uv run pytest -m integration tests/test_audit.py::test_real_clickhouse_round_trip -v`
- manual tests: tested that the UI correctly shows the graph for all fixed existing time ranges (24h, 7d, 30d, 90d):

Screenshot 2026-06-19 at 12 14 19

### DIFF
--- a/platform/api/audit.py
+++ b/platform/api/audit.py
@@ -12,11 +12,14 @@ HTTP-agnostic — exceptions map to status codes in main.py.
 from __future__ import annotations
 
 import json
+import logging
 from datetime import datetime, timedelta, timezone
 
 from clickhouse_connect.driver.client import Client
 
 from schemas import DecisionEvent
+
+_log = logging.getLogger(__name__)
 
 
 class AuditPayloadTooLarge(Exception):
@@ -132,22 +135,40 @@ def insert_decision(
 # Dashboard windows → hours; 90d is the 90-day TTL ceiling.
 WINDOW_HOURS: dict[str, int] = {"24h": 24, "7d": 24 * 7, "30d": 24 * 30, "90d": 24 * 90}
 
-# Bucket size per window (~24-90 points): 24h→hourly, 7d→6h, 30d/90d→daily.
-_WINDOW_BUCKET_MINUTES: dict[str, int] = {
-    "24h": 60,
-    "7d": 360,
-    "30d": 1440,
-    "90d": 1440,
-}
 
+def bucket_minutes_for_timedelta(delta: timedelta) -> int:
+    """Bucket size (minutes) for a free-form date range.
 
-def bucket_minutes_for(window: str) -> int:
-    """Bucket size (minutes) for a window key (KeyError on unknown window)."""
-    return _WINDOW_BUCKET_MINUTES[window]
+    ≤30min→1min, ≤1h→5min, ≤6h→15min, ≤12h→30min, ≤24h→60min, ≤7d→360min, else→1440min.
+    """
+    if delta <= timedelta(minutes=30):
+        return 1
+    elif delta <= timedelta(hours=1):
+        return 5
+    elif delta <= timedelta(hours=6):
+        return 15
+    elif delta <= timedelta(hours=12):
+        return 30
+    elif delta <= timedelta(hours=24):
+        return 60
+    elif delta <= timedelta(days=7):
+        return 360
+    else:
+        return 1440
 
 
 def _zero_counts() -> dict[str, int]:
     return {"all": 0, "allow": 0, "deny": 0, "needs_approval": 0}
+
+
+def _date_range_valid(start: datetime | None, end: datetime | None) -> bool:
+    """True if both dates are present and start <= end."""
+    if not (start and end):
+        return False
+    if start > end:
+        _log.warning(f"Date range invalid, start > end: {start} > {end}")
+        return False
+    return True
 
 
 def _scope(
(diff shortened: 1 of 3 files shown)


platform/api/audit.py
```diff
@@ -132,39 +135,67 @@ def insert_decision(
...
 
 def _zero_counts() -> dict[str, int]:
     return {"all": 0, "allow": 0, "deny": 0, "needs_approval": 0}
 
 
+def _date_range_valid(start: datetime | None, end: datetime | None) -> bool:
+    """True if both dates are present and start <= end."""
+    if not (start and end):
+        return False
+    if start > end:
```
guillaume-hexamind, line 168: **Severity ~64/100 — correctness (HTTP 500).** `start > end` isn't tz-normalized, so a request mixing a naive and a tz-aware bound raises `TypeError: can't compare offset-naive and offset-aware datetimes` (reproduced locally).

FastAPI parses `?start_date=2025-01-07T00:00:00&end_date=2025-01-01T00:00:00Z` as one naive + one aware datetime. The `TypeError` is **not** a `ClickHouseError`, so it escapes the `try/except` in `api_audit_summary`/`timeseries`/`decisions` → **HTTP 500** instead of the intended `since_hours` fallback.

Suggest normalizing both bounds to UTC at the API boundary before they reach this comparison.

platform/api/audit.py
```diff
@@ -132,39 +135,67 @@ def insert_decision(
...
         "project_id = {pid:String}",
-        "occurred_at >= now() - INTERVAL {hrs:UInt32} HOUR",
     ]
-    params: dict[str, object] = {"pid": project_id, "hrs": since_hours}
+    params: dict[str, object] = {"pid": project_id}
+    if _date_range_valid(start_date, end_date):
+        where.append(
+            "occurred_at >= {start_date:DateTime} AND occurred_at <= {end_date:DateTime}"
+        )
+        params["start_date"] = start_date
```
guillaume-hexamind, line 194: **Severity ~58/100 — correctness (silent wrong data).** A *naive* `start_date`/`end_date` bound to `{start_date:DateTime}` is reinterpreted in the **API host's local timezone**, not UTC.

clickhouse-connect 1.1.1 `format_bind_value` (binding.py:244) calls `value.astimezone(server_tz)` unconditionally, and `astimezone()` treats a naive datetime as system-local. With `?start_date=2025-01-01T00:00:00` (no offset) on a host where `TZ != UTC`, the window is silently shifted by the host offset. Since `occurred_at` is `DateTime64(3,'UTC')`, the query returns the wrong slice with no error.

Coerce naive inputs to `tzinfo=UTC` at the endpoint. (Latent today: the dashboard doesn't send these params yet, and UTC containers mask it.)

platform/api/audit.py
```diff
@@ -132,39 +135,67 @@ def insert_decision(
...
     """Shared WHERE + params for the scope filters (project/window/agent/role/
     tool) that all reads narrow by. Pass role="" for the no-role bucket."""
     where = [
         "project_id = {pid:String}",
-        "occurred_at >= now() - INTERVAL {hrs:UInt32} HOUR",
     ]
-    params: dict[str, object] = {"pid": project_id, "hrs": since_hours}
+    params: dict[str, object] = {"pid": project_id}
+    if _date_range_valid(start_date, end_date):
+        where.append(
```
guillaume-hexamind, line 191: **Severity ~50/100 — efficiency / DoS surface.** The date-range path injects `start_date`/`end_date` straight into the `WHERE` with no retention/future/max-span cap, dropping the implicit ≤90d ceiling that the old `now() - INTERVAL {hrs} HOUR` clause always enforced.

`?start_date=2000-01-01Z&end_date=2100-01-01Z` triggers a full-partition scan/aggregation on `policy_decision` — a single audit read can scan the whole table. The write path already guards exactly this via `validate_event_window` (`RETENTION_WINDOW = 90d`); the read path should too. Rows past the 90d TTL also don't exist, so an over-wide range looks valid but returns partial data.

Suggest clamping the range to `[now-90d, now]` and rejecting (400) spans beyond it.

---
_The highest-leverage fix for this and the two other comments is to normalize + validate the range once at the FastAPI boundary: coerce to UTC, require `start ≤ end`, clamp to the 90d retention window, else 400._

guillaume-hexamind (commented):
I approve of the logic

Claude auto-review found a couple of minor datetime issues, we can merge once addressed!

platform/api/audit.py
victorludvig: With `_ensure_utc` (from previous comment), the datetimes are forced to be utc datetimes, so they will not be interpreted in the API host's local timezone.

platform/api/audit.py
victorludvig: For the next revision, I added this function to check and convert the datetime to timezone-aware datetimes if necessary:
```
def _ensure_utc(dt: datetime | None) -> datetime | None:
    if dt is None:
        return None
    if dt.tzinfo is None:
        return dt.replace(tzinfo=timezone.utc)
    return dt.astimezone(timezone.utc)
```

platform/api/audit.py
victorludvig: For the next revision I added this function to clamp start_date to end_date-90 days:
```python
def _prepare_date_range(
    start_date: datetime | None, end_date: datetime | None
) -> tuple[datetime | None, datetime | None]:
    start_date = _ensure_utc(start_date)
    end_date = _ensure_utc(end_date)

    if start_date and end_date:
        start_date = max(start_date, end_date - RETENTION_WINDOW)

    return start_date, end_date```

victorludvig:
### Code review

guillaume-hexamind (approved):

PR #41, opened 2026-06-22
feat(dashboard): syntax-highlighted YAML editor with inline diagnostics

quanghexa94:
Replaces the plain <textarea> on /policies with a CodeMirror 6 editor:

  * YAML syntax highlighting via @codemirror/lang-yaml
  * Line numbers + lint gutter for inline error markers
  * Server validation results pushed into the gutter via setDiagnostics (clears on edit; the existing role-level error list above the editor still surfaces no-line errors)
  * Theme built from the dashboard's CSS vars so colors are tied to the palette and follow any future tweak — keys = primary blue, strings = allow green, numbers/bools = approval amber
  * basicSetup tuned for ~50-100 line policies: line numbers on, fold gutter off, bracket matching off (YAML), autocomplete off (deferred until JSON-schema-driven completions are wired)
  * Dark-only for now — dashboard hard-codes dark mode at main.tsx:9; light variant is a ~10-line createTheme mirror when we add a toggle

Lint diagnostics come from the existing /policies/validate endpoint; CodeMirror's lint extension is just the rendering surface, validation stays server-side (single source of truth).

Bundle: ~180 KB gzipped of CodeMirror, behind route-based code split, so only /policies pays the cost.

Tests:
  * diagnostics.test.ts — pure adapter: line-anchoring, role prefix, clamping out-of-range lines, dropping role-only errors
  * PolicyEditor.test.tsx — smoke: initial value renders, lint gutter mounted, readOnly honored. Edit-lifecycle coverage delegated to @uiw/react-codemirror's own suite.

41/41 dashboard tests pass; 0 type errors; 10 pre-existing lint warnings unchanged.


platform/dashboard/src/lib/auth.ts
```diff
@@ -124,7 +125,15 @@ export function useLogout() {
...
   return useMutation({
     mutationFn: logoutRequest,
-    onSuccess: () => qc.invalidateQueries({ queryKey: USER_QUERY_KEY }),
+    onSuccess: () => {
+      // Tear down the playground module-level store + socket so the
+      // next signed-in user on this browser tab doesn't see the prior
+      // user's chat history. There's no full page reload on logout
+      // (just a cookie invalidation), so without this the module
+      // globals survive.
+      resetPlayground()
```
victorludvig, line 134: The resetPlayground() call in useLogout's onSuccess is the critical line preventing user data leaking across sign-ins on the same tab. Might be worth a test that verifies it gets called on logout success.

victorludvig (approved):
I suggested adding a test. It can be added in a follow-up PR. Approving.

platform/dashboard/src/lib/auth.ts
quanghexa94: I've added a test to cover this

victorludvig (approved):

PR #43, opened 2026-06-22
build(dashboard): add prettier formatter and format entire codebase

victorludvig:
## What is changing

Adds prettier as a dev dependency with `format` and `format:check` pnpm scripts. Extends make fmt and make fmt-check to cover platform/api (ruff) in addition to the SDK. Adds standalone dashboard-fmt, dashboard-fmt-check, dashboard-lint, and dashboard-typecheck make targets.

Runs prettier across all dashboard src files for the first time.

Also add make commands to test / format the entire codebase all at once.

## Tests
- `make check-all`

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -147,6 +147,7 @@ jobs:
           pnpm install --frozen-lockfile
           pnpm lint
           pnpm typecheck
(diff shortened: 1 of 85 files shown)


quanghexa94 (approved):
LGTM

PR #44, opened 2026-06-22
feat(platform/api): optional Postgres backend (DATABASE_URL, SQLite fallback)

guillaume-hexamind:
## Summary

Make the control-plane API run on **Postgres**, driven by `DATABASE_URL`, while keeping **SQLite as the zero-setup fallback** so dev and the test suite are unchanged. Schema is still created with `create_all` — **no Alembic / no migration system** (beta decision: data is disposable across schema changes, wipe rather than migrate).

"Use Postgres" and "build migrations" are deliberately separated — this PR does the first and skips the second. Doing the engine swap now, while data is disposable, is the cheapest time to do it.

## Changes

- **`db.py`** — engine URL comes from `DATABASE_URL`; unset falls back to the local SQLite file. Bare `postgres://` / `postgresql://` URLs are normalized to the `asyncpg` driver the async engine needs. `pool_pre_ping=True` for managed-PG dropped connections. `load_dotenv()` runs here too, since the engine is built at import (before `main`'s own `load_dotenv`), so `DATABASE_URL` works from `platform/api/.env`. Public interface (`init_db`/`get_session`/`async_session_factory`) unchanged.
- **`pyproject.toml` + `uv.lock`** — add `asyncpg`.
- **`models.py`** — every `datetime` column now uses `sa_type=DateTime(timezone=True)`. **This fixes a real PG blocker:** `utcnow()` is tz-aware but the columns mapped to `TIMESTAMP WITHOUT TIME ZONE`; SQLite tolerated it, asyncpg rejected the insert (so the lifespan seed would have failed at startup on PG). `timestamptz` on Postgres. On SQLite, read-backs are now tz-aware to match writes (previously naive). Project code uses `utcnow()` (tz-aware) everywhere and already normalizes DB-loaded datetimes before comparison (`_ensure_utc_aware` in `services.py`, `_ensure_utc` in `audit.py`), so there's no observable change in behavior.
- **`docker-compose.yml`** — local `postgres:16` service (loopback `5433:5432`, dev creds, healthcheck) + `postgres-data` volume.
- **`Makefile`** — `postgres-up` (waits healthy) / `postgres-stop` / `postgres-psql` / `postgres-reset`, and `platform-api-pg` (starts PG, runs the API against it). `make platform-api` stays SQLite.
- **`.env.sample`** — `DATABASE_URL` section with the local-PG DSN.
- **`tests/test_postgres_smoke.py`** — opt-in (skipif on `DATABASE_URL`) smoke test: `init_db` + round-trip against a live PG. The rest of the suite builds its own SQLite engines, so this is the only coverage of the asyncpg path.
- **`.github/workflows/tests.yml`** — isolated `platform-api-postgres` job (PG service container, `DATABASE_URL` set) running only the smoke test, kept separate so it can't destabilize the SQLite suite.
- **`README.md`** — "Control-plane database (SQLite / Postgres)" section.

## Validation

- SQLite suite: **318 passed, 2 skipped** (unchanged; the PG smoke test skips without `DATABASE_URL`).
- Against a live local Postgres: smoke test passes, and the full lifespan seed lands cleanly (orgs=1, users=1, projects=1, agents=3 — exercises JSON + LargeBinary columns), tz-aware timestamps preserved.
- `ruff` clean; workflow YAML validated.

## Reviewer note

The **`models.py` timestamptz change** is the substantive part — it touches every table's datetime columns and was surfaced by the smoke test (not in the original plan). Worth a closer look.

## Out of scope

Alembic, the production compose (Caddy/ClickHouse/TLS), and parametrizing the existing test fixtures to run on PG (they stay SQLite — the smoke test is the PG coverage).

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -88,6 +88,47 @@ jobs:
       - name: Tests
         run: make platform-api-test
 
+  platform-api-postgres:
+    # Isolated from the platform-api job on purpose: that job runs the whole
+    # suite on SQLite (its fixtures build their own engines), so it never
+    # exercises asyncpg. This job sets DATABASE_URL and runs ONLY the opt-in
+    # smoke test, which points the real engine at Postgres and round-trips a
+    # row — the only coverage that the schema actually builds on PG. No OPA
+    # needed (the smoke test doesn't compile bundles).
+    name: Platform API (Postgres smoke)
+    runs-on: ubuntu-latest
+    services:
+      postgres:
+        image: postgres:16
+        env:
+          POSTGRES_DB: hexgate
+          POSTGRES_USER: hexgate
+          POSTGRES_PASSWORD: ci
+        ports:
+          - "5432:5432"
+        options: >-
+          --health-cmd "pg_isready -U hexgate -d hexgate"
+          --health-interval 5s --health-timeout 3s --health-retries 10
+    env:
+      DATABASE_URL: postgresql+asyncpg://hexgate:ci@localhost:5432/hexgate
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@v6
+        with:
+          enable-cache: true
+          cache-dependency-glob: "platform/api/uv.lock"
+
+      - name: Set up Python 3.13
+        run: uv python install 3.13
+
+      - name: Sync platform-api deps
+        run: cd platform/api && uv sync --group dev
+
+      - name: Postgres smoke test
+        run: cd platform/api && uv run pytest tests/test_postgres_smoke.py -v
+
   dashboard:
     name: Dashboard
     runs-on: ubuntu-latest
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -89,12 +89,9 @@ jobs:
         run: make platform-api-test
 
   platform-api-postgres:
-    # Isolated from the platform-api job on purpose: that job runs the whole
-    # suite on SQLite (its fixtures build their own engines), so it never
-    # exercises asyncpg. This job sets DATABASE_URL and runs ONLY the opt-in
-    # smoke test, which points the real engine at Postgres and round-trips a
-    # row — the only coverage that the schema actually builds on PG. No OPA
-    # needed (the smoke test doesn't compile bundles).
+    # The platform-api job runs the whole suite on SQLite, never exercising
+    # asyncpg. This job sets DATABASE_URL and runs only the opt-in smoke test
+    # — the sole coverage that the schema builds on Postgres.
     name: Platform API (Postgres smoke)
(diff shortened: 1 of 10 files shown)


quanghexa94:
The PR is "optional Postgres backend" but the diff also includes:

Makefile: fmt / fmt-check dropped platform/api from their target list (platform-api Python no longer gets auto-formatted by make fmt)
Makefile: check-all target deleted entirely (was running full-stack lint + typecheck + tests)
Makefile: dashboard-fmt, dashboard-fmt-check, dashboard-lint, dashboard-typecheck all deleted
.github/workflows/tests.yml: Lint step removed from platform-api job
.github/workflows/tests.yml: Format check step removed from dashboard job

Why these changes?

Secondly, about postgres

1. "No change to SQLite behavior or Python comparison semantics" - not 100% true

The PR description says the SQLite read-back is unchanged. That's not quite right: DateTime(timezone=True) on SQLite makes read-backs tz-aware instead of tz-naive. So any code that compared a DB-fetched datetime against a naive one (e.g. an older datetime.utcnow() without .replace(tzinfo=utc)) will start raising TypeError: can't compare offset-naive and offset-aware datetimes.

Mitigating: your utcnow() helper already returns tz-aware, so internal code is consistent. Risk is in third-party deps (fastapi-users in particular) that may have introspected the old shape. The full test suite still passing on SQLite (318) is the best evidence this is fine in practice — but the framing in the PR body undersells what changed. Worth a one-line clarification: "SQLite read-backs are now tz-aware to match writes; project code already uses utcnow() everywhere, so no observable change."

2. pool_pre_ping=True is wasted on SQLite.

Beneficial for managed Postgres (catches dropped connections). On SQLite-via-aiosqlite, it issues a pointless SELECT 1 per checkout. Comment says "harmless on SQLite", agreed it's not broken, but technically maybe we can:


engine_kwargs: dict = {"echo": False}
if "postgresql" in url:
    engine_kwargs["pool_pre_ping"] = True
engine = create_async_engine(url, **engine_kwargs)

guillaume-hexamind:
> 3\. pool_pre_ping=True is wasted on SQLite.
> 
> Beneficial for managed Postgres (catches dropped connections). On SQLite-via-aiosqlite, it issues a pointless SELECT 1 per checkout. Comment says "harmless on SQLite", agreed it's not broken, but technically maybe we can:
> 
> engine_kwargs: dict = {"echo": False} if "postgresql" in url: engine_kwargs["pool_pre_ping"] = True engine = create_async_engine(url, **engine_kwargs)

For the changes mentionned in the first paragraph, these are introduced by #43 and will auto-resolve after merge; I still set my branch up to date so they do not appear anymore in the diff.

1. -> you are right, the description of this PR has been changed accordingly

2. -> seems minor but does make sense, the suggested change has been applied

quanghexa94 (approved):
LGTM

PR #45, opened 2026-06-23
feat(platform/dashboard): Add custom date filter in Audit Dashboard

victorludvig:
## What is changing

- The range toggle (24h / 7d / 30d / 90d) in the page header gains a fifth option: Custom.
- Selecting Custom expands a date range picker row directly below the header, above the filter bar and KPIs, making it unambiguous that the date filter scopes the entire page (KPIs, charts, breakdown, and decisions table), not just the Decisions table.
- Selecting a preset clears any active custom dates and dismisses the picker row. 
- The default label shown in the picker before any dates are selected is derived from the active range preset (30 days by default).

https://github.com/user-attachments/assets/9e9b6710-9062-43e2-a8ab-d086ba94c110

[Notion Task](https://app.notion.com/p/Add-audit-filters-383fb45dbae28025bc32f0a4fed3a96b?source=copy_link)

## Tests

- make check-all


guillaume-hexamind (commented):
I approve of the logic and final UI/UX with custom button

Here are a couple of potential issues found by claude, if you can check them:

---
  Comment 1 — platform/dashboard/src/routes/Audit.tsx (~line 1488 / onSelect)

  **🔴 Single-day selection is silently ignored, but the UI claims the filter is applied.**

  When the user clicks a single day in the calendar, `onSelect` fires with `range.from` set
  and `range.to` undefined, so we store `start_date` with `end_date = null`. The backend's
  `_date_range_valid` (`platform/api/audit.py`) requires **both** dates and silently falls
  back to the `f.range` window. The table and KPIs then show full-range data (e.g. 30d)
  while the trigger button reads `"Jun 1, 2026 → …"` — the displayed data doesn't match the
  filter the user set.

  Suggested fix: don't commit a partial range to the store. Either
  - only set `start_date/end_date` once both ends are picked (treat a single click as
    in-progress), or
  - when only `from` is present, treat it as a single-day range (`end_date = endOfDay(from)`)
    so the backend actually filters.

  ---
  Comment 2 — platform/dashboard/src/routes/Audit.tsx (~line 1390 / optionsQ, also ~348)

  **🔴 The filter-options query is now scoped to the custom date range, which can make
  filters unreachable.**

  `optionsQ`'s queryKey/queryFn now include `start_date/end_date` (via `optionsScope`).
  This turned the previously date-unscoped source of agent/role/tool options into a
  date-scoped one. If the user narrows the range to a window where a given agent had no
  activity, that agent **disappears from the filter dropdown** — so they can't filter by it
  — and the unscoped "X of Y total" denominator shrinks too.

  Suggested fix: keep `optionsQ` on the window-only scope (`{ window: f.range }`) so the
  dropdown always lists the full set of selectable entities, independent of the custom date
  range. Only the data/summary queries should carry `start_date/end_date`.

  ---
  Comment 3 — platform/dashboard/src/components/audit/pieces.tsx (~line 200, ActiveChips clear)

  **🔴 Clearing filters leaves the Custom date row stuck open and the button implying an
  active filter.**

  `ActiveChips` "clear" resets `start_date/end_date` to null in the store, but it can't reset
  the parent's local `showDateRow` state (`Audit.tsx:327`). After clearing while Custom is
  active, the date row stays visible and the trigger falls back to
  `impliedRangeLabel(f.range)` — telling the user a custom range is active when the query has
  actually reverted to the plain window.

  Root cause: "is custom mode on" is duplicated between the local `showDateRow` flag and the
  store's `start_date/end_date`, and the two desync.

  Suggested fix: derive the row's visibility from the store
  (`const showDateRow = !!(f.start_date || f.end_date)` or a dedicated `mode` field) instead
  of separate local state, so clearing the dates automatically collapses the row and resets
  the label. If a local flag is needed for the "Custom toggled but no date yet" case, sync it
  in an effect on `start_date/end_date`.

  ---
  Comment 5 — platform/dashboard/src/routes/Audit.tsx (~line 1488, date boundary serialization)

  **🟠 Date boundaries are offset by the browser timezone.**

  Boundaries are taken at the browser's local timezone (local midnight for start, local
  `endOfDay` for end) and then serialized with `toISOString()` to UTC. For a user in UTC-5,
  picking June 1 sends `start_date = 2026-06-01T05:00:00Z`, so decisions between 00:00–05:00
  UTC on June 1 are shifted/excluded. The filtered result is off by up to a day at the edges
  relative to the day the user picked.

  Suggested fix: decide on one timezone contract with the backend and apply it consistently —
  e.g. send date-only strings (`yyyy-MM-dd`) and resolve the day boundaries server-side, or
  explicitly construct UTC midnight/end-of-day from the picked calendar date instead of
  relying on the local-time `Date` → `toISOString()` conversion.

victorludvig:
- Comment 1: This can never happen. When clicking on a single date start date, the end date defaults to the start date:
```
// node_modules/react-day-picker/dist/esm/utils/addToRange.js
if (!from && !to) {
    // the range is empty, add the date
    range = { from: date, to: min > 0 ? undefined : date };
}
```
Screenshot 2026-06-24 at 14 19 52

- Comment 2: This is intended behavior. When selecting a custom time range, we only show in the agent selection dropdown list agents that had some actions during this period. We also count the number of actions in this period, so the denominator in "X/Y" is shrunk. This is what is already happening when filtering with the preset time ranges: 24h|7d|30d|90d

- Comment 3: This is a rare and not really harmful edge case but I fixed it. Added customMode boolean to the Zustand store. showDateRow is now derived as f.customMode, no local state. range stays as the last active preset throughout, so the implied range label always shows the correct window. "Clear all" in ActiveChips resets customMode: false alongside the other filters, collapsing the date row atomically. Added corresponding unit test.

- Comment 5: This is intended. When a user selects a date, he selects it on his timezone. The frontend uses [toISOString](https://github.com/HexamindOrganisation/hexgate/blob/1ab37bb50372f12dbcf1f517c4dafb3b5781988a/platform/dashboard/src/routes/Audit.tsx#L341-L342) to convert the date to UTC, then sends it to backend, which stores all dates in UTC format.

guillaume-hexamind (approved):

PR #48, opened 2026-06-24
build(platform): production deployment stack (api, compose, docs)

guillaume-hexamind:
## What

Makes the platform deployable on a single machine for **prod and staging**, with no container registry and no separate edge/TLS container. The dashboard SPA is baked into the API image and served **same-origin** alongside `/v1/*`; TLS is terminated by the box's reverse proxy upstream.

## Changes

**Deployment stack**
- **`platform/api/Dockerfile`** — control-plane API image (amd64, OPA pinned by SHA, single worker, `/health` HEALTHCHECK). A first build stage runs `pnpm build` and bakes `platform/dashboard/dist` into the image at `/app/static`. `Dockerfile.dockerignore` overrides the repo-root `.dockerignore` to keep the dashboard source in the build context.
- **`platform/docker-compose.deploy.yml`** — one stack (postgres, clickhouse, api) parameterized per environment; the api publishes only on loopback for the upstream proxy. Replaces the old prod-only compose + Caddy edge.
- **`Makefile`** — `platform-{prod,staging}-{up,down,logs}` targets (build locally, `up -d --build`), each on its own compose project + env file.
- **`platform/.env.prod.sample` / `.env.staging.sample`**, **`.gitignore`**, **`platform/DEPLOY.md`** — per-env templates (real `.env.*` gitignored) and a deploy runbook.

**Same-origin SPA serving**
- **`platform/api/spa.py`** (new) — shared module that mounts the hashed-asset static dir + the `/{full_path:path}` catch-all → `index.html`.
- **`platform/api/main.py`** — SPA + Google-OAuth router mounting moved into the FastAPI `lifespan` (after the keystore is ready), with the SPA catch-all registered **last** so it can't shadow `/v1/auth/google/*`.
- **`platform/api/demo.py`** — demo mode now delegates SPA serving to `spa.mount_spa` instead of duplicating it.

**Hardening / correctness**
- **DB URL** (`db.py`) — `DATABASE_URL` is now assembled and **percent-encoded** by the app from `HEXGATE_POSTGRES_*` components, so the password may contain any characters (no URL-safe restriction).
- **SPA catch-all** — unknown API paths (`/v1/*`, `/health`, …) return FastAPI's default `404 {"detail":"Not Found"}` JSON instead of `index.html` with `200`, so SDK clients aren't fed HTML.
- **Missing/misconfigured build** — `mount_spa` logs a loud `WARNING` when the dashboard build is absent (instead of failing silently), and `dashboard_dist()` warns on a set-but-empty `HEXGATE_DASHBOARD_DIST` rather than quietly using the dev path.
- **Preview banner** — `VITE_PREVIEW_BANNER` wired as a Docker **build arg** (Vite inlines it at `pnpm build`), sourced from `.env.<env>`, so staging/prod can show the beta strip.
- **Email config** — `RESEND_API_KEY` + `HEXGATE_EMAIL_FROM` documented in the env samples.

**Tests**
- `tests/test_spa.py`, `tests/test_db.py` (new) and an OAuth route-shadowing guard in `tests/test_auth_oauth.py`.

## Notes

- **Same-origin by design** — no CORS, and `Secure` cookies work behind the proxy.
- **ClickHouse port is 8123** (HTTP, `clickhouse_connect`), not native 9000.
- **Single API instance only** — the in-process WS relay can't scale horizontally.

## Testing

Full non-integration suite passes (332 passed, 2 integration skipped). End-to-end local deploy brings up postgres/clickhouse/api healthy with the SPA served same-origin; `/v1/health` and `/v1/ready` reachable, deep-link routes (`/playground`) resolve. TLS issuance is handled by the upstream proxy and covered by the runbook.


quanghexa94:
It's good but do we need to have env.staging.sample and env.prod.sample ?? they are almost duplicate no? would be nice to just have an argument that tell what stage you launch in. same for many commands platform-{stage}-up can really be platform-up and stage is a param. much less commands and easily to scale up otherwise every time we need to add a new command we may need to do it 6 7 times

guillaume-hexamind:
> It's good but do we need to have env.staging.sample and env.prod.sample ?? they are almost duplicate no? would be nice to just have an argument that tell what stage you launch in. same for many commands platform-{stage}-up can really be platform-up and stage is a param. much less commands and easily to scale up otherwise every time we need to add a new command we may need to do it 6 7 times

Fully agree with the make commands, I merged them into single ones where we pass STAGE={stage} as arg. For .env, I'd rather keep them as is in this PR to not confuse the staging and prod 7000 and 7200 ports + URLs that are coupled with actual nginx config of prod01 machine. I will change that in my next PR that adds a secret pull from scaleway secret manager!

quanghexa94 (approved):
LGTM!

PR #50, opened 2026-06-24
fix(security): per-turn refresh works on the pydantic-fallback path

quanghexa94:
## The bug

In the demo notebook flow (Modal container, no \`opa\` installed), editing a tool's mode in \`/policies\` and saving had **no effect on the running agent**. The next chat turn used the same enforcement as before the edit. Restart was required.

## Root cause

\`PlatformPolicySource.fetch()\` returned \`PolicyBundle | None\`. When the platform can't compile (no opa on the host), every \`GET /v1/agents/:name\` returns 200 with the new \`policy_yaml\` but null bundle fields. \`decode_and_verify_platform_bundle\` returns \`None\`, \`fetch()\` returns \`None\`, and \`PolicyBinding.refresh()\` short-circuits at:

\`\`\`python
if new_policy is None or new_policy is self.enforcer.policy:
    return  # nothing served, or same cached object (304)
\`\`\`

The pydantic engine built at agent-load time by \`platform_policy_from_payload\` was the only engine the agent ever ran with. The load-time path knew the no-bundle fallback shape; the refresh-time path didn't.

## The fix

Broadened \`PolicySource.fetch()\` to return \`PolicyEngine | None\` (was \`PolicyBundle | None\`). \`PolicyBundle\` still satisfies the protocol, so the WASM path is unchanged.

\`PlatformPolicySource.fetch()\` now branches:

- **Bundle present** → WASM path, ETag-based 304 caching (unchanged).
- **Bundle absent** → hash the \`policy_yaml\`, construct a fresh \`PolicySet\` only when the hash changes. Identity preservation keeps the binding's \`is\` check from triggering a needless rebind when nothing changed.

Transition cases covered:
- yaml-hash sentinel cleared when we land on the bundle path, so a future opa-came-back-online response doesn't reuse a stale hash.
- \`platform_policy_from_payload\` computes the load-time yaml hash from the same payload it parsed into the initial PolicySet, seeding the source so the first refresh hits the cache instead of constructing a content-equivalent fresh PolicySet.
- \`initial_bundle=\` kept as a back-compat alias for the new \`initial_engine=\` kwarg.

## Tests

- \`test_no_bundle_in_payload_falls_back_to_pydantic_engine\` (renamed from \`test_no_bundle_in_payload_returns_none\`, which pinned the buggy behavior)
- \`test_no_bundle_unchanged_yaml_returns_same_engine_instance\` — identity preservation
- \`test_no_bundle_changed_yaml_returns_new_engine\` — **the regression test for the demo bug**
- \`test_no_bundle_then_bundle_clears_yaml_hash_state\` — opa-came-back-online transition
- \`test_refresh_swaps_policy_on_pydantic_fallback_path\` — end-to-end binding flow: load with deny, refresh with allow, verify the new decision applies

## Test plan

- [x] \`make check\` — 837 passed, 2 skipped
- [ ] **Manual** in the demo container: start playground, edit a tool to allow, send a new message, verify it routes through

## Why now

The demo deployment is the trigger that makes this matter — production deployments will run with opa installed, so the bundle path catches everything. This fix unblocks the Modal demo without requiring opa in the container, and also makes the pydantic-fallback path symmetric with the load-time fallback behaviour.

## Scope

Single change: \`PlatformPolicySource.fetch\` + \`platform_policy_from_payload\` seeding. The WASM path, both other source implementations (\`BundleDirPolicySource\`, \`YamlPolicySource\`), and the binding's refresh logic are unchanged.

### DIFF
--- a/deploy/boot.py
+++ b/deploy/boot.py
@@ -44,7 +44,11 @@ _procs: list[subprocess.Popen] = []
 
 def _spawn(cmd: list[str], *, cwd: Path, env: dict[str, str]) -> subprocess.Popen:
     print(f"[boot] starting: {' '.join(cmd)}  (cwd={cwd})", flush=True)
-    p = subprocess.Popen(cmd, cwd=str(cwd), env=env)
+    # start_new_session=True isolates the child in its own process group, so the
+    # terminal's Ctrl-C goes only to boot.py — boot owns shutdown and signals
+    # each child once, avoiding the SIGINT+SIGTERM race that left uvicorn
+    # half-shutdown (port 8000 still bound) when boot exited too fast.
+    p = subprocess.Popen(cmd, cwd=str(cwd), env=env, start_new_session=True)
     _procs.append(p)
     return p
 
@@ -125,10 +129,37 @@ def run(dash_url: str | None = None) -> None:
     _block_until_exit()
 
 
+_TERM_GRACE = 5.0  # seconds; long enough for uvicorn graceful shutdown
+
+
+def _signal_group(p: subprocess.Popen, sig: int) -> None:
+    """Signal the child's whole process group — covers uvicorn workers, etc."""
+    try:
+        os.killpg(os.getpgid(p.pid), sig)
+    except ProcessLookupError:
+        pass  # already gone
+
+
 def _block_until_exit() -> None:
-    def _shutdown(*_a):
+    shutting_down = False
+
+    def _shutdown(*_):
+        nonlocal shutting_down
+        if shutting_down:
+            return  # already in progress — ignore re-entrant signal
+        shutting_down = True
         for p in _procs:
-            p.terminate()
+            if p.poll() is None:
+                _signal_group(p, signal.SIGTERM)
+        # Wait for graceful exit, then SIGKILL stragglers. Without this
+        # the parent died first and uvicorn was orphaned with port still bound.
+        for p in _procs:
+            try:
+                p.wait(timeout=_TERM_GRACE)
+            except subprocess.TimeoutExpired:
+                print(f"[boot] pid={p.pid} did not exit in {_TERM_GRACE}s; killing", flush=True)
+                _signal_group(p, signal.SIGKILL)
+                p.wait()
         sys.exit(0)
 
     signal.signal(signal.SIGTERM, _shutdown)
--- a/hexgate/security/__init__.py
+++ b/hexgate/security/__init__.py
@@ -74,6 +74,7 @@ from hexgate.security.rego_wasm import (
 from hexgate.security.source import (
     BundleDirPolicySource,
     PlatformPolicySource,
+    PolicyContentError,
     PolicySource,
     YamlPolicySource,
 )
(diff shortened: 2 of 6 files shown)


guillaume-hexamind:
Claude review:

  🔴 Critical
  
  1. Silent strict-signature downgrade on refresh — source.py:179 (CONFIRMED) · 88/100
  HEXGATE_BUNDLE_REQUIRE_SIGNATURE is enforced only at load time, not in fetch(). With REQUIRE_SIGNATURE=1, an agent boots fine on a signed WASM bundle;
  if opa later goes down mid-session and get_agent starts returning policy_yaml with null bundle fields, the next per-turn refresh takes the no-bundle
  branch, builds a PolicySet from raw yaml with no signature check, and PolicyBinding.refresh swaps it in. The agent then enforces an unsigned, unverified
  (possibly tampered or stale) control-plane policy in a deployment that explicitly demanded signatures — exactly the downgrade
  decode_and_verify_platform_bundle was written to refuse. Re-establish the REQUIRE_SIGNATURE refusal on the fetch/refresh path, not just at load.

  ---
  🟠 Medium

  2. Malformed/invalid edited policy → silent stale enforcement — source.py:179 (CONFIRMED) · 58/100
  fetch() now runs yaml.safe_load + load_policy_set_from_dict on the per-turn path. An edit that parses at the platform but fails the SDK's pydantic
  validation (PolicySetError, missing default role, etc.) makes fetch() raise; PolicyBinding.refresh catches it (BLE001), logs a warning, and keeps the
  old engine. The dashboard shows the policy as saved while the agent silently keeps enforcing the old one — the same "edit silently ignored at runtime"
  class this PR set out to fix, now reachable via an invalid edit, with no hard error surfaced.

  3. Non-null ETag on a no-bundle response swallows policy edits — source.py:184 (PLAUSIBLE) · 50/100
  If any platform build computes an ETag from the agent record rather than from compiled_wasm, the no-bundle 200 carries an ETag; fetch() caches it, the
  next request sends If-None-Match, the server returns 304, and fetch() returns the cached engine without ever reaching the yaml-hash comparison — a
  dashboard edit never takes effect. Currently masked only because platform/api/main.py sets etag=None when compiled_wasm is None; the SDK has no guard of
  its own and breaks if that platform invariant ever changes. Worth a defensive guard in the SDK rather than relying on the server contract.

  4. Full GET + parse on every turn on the fallback path — source.py:183 (CONFIRMED) · 45/100
  On the no-opa / Modal deployment shape this PR targets, the server never sends an ETag, so _cached_etag stays None and every agent turn issues a
  full-payload GET plus sha256 + yaml.safe_load on the hot path (binding.refresh runs at the top of every run). The old code returned None cheaply on the
  no-bundle path. Adds per-turn latency and increases exposure to transient platform slowness/outages on exactly the target deployment.

guillaume-hexamind (commented):
Logic Looks good overall! 
You can check claude's potential issues, the critical one seems like the one important fix to apply

quanghexa94:
all four problems addressed!

guillaume-hexamind (approved):

PR #51, opened 2026-06-24
chore: wire test coverage + Codecov badge across SDK / API / dashboard

quanghexa94:
## Summary

Add test-coverage measurement to all three surfaces (Python SDK, Python platform API, TypeScript dashboard), upload to Codecov from CI, and surface the SDK number as a badge in the README. Coverage is treated as a regression guard, not a target: each flag has a floor with a small tolerance threshold, ratcheted by hand as the codebase matures.

## Current numbers (full local runs)

| Surface | Branch coverage | Tests |
|---|---|---|
| **SDK** | **84.1%** | 833 |
| **Platform API** | **79.7%** | 318 |
| **Dashboard** | **26.9%** | 51 |

The dashboard is honest-low because most components don't have tests today; lib/ is well-covered, components are mostly not. The codecov.yml policy reflects this with separate per-flag targets — sdk 75% / platform-api 70% / dashboard 25%, with the dashboard flag set to \`informational\` (notifies but doesn't block PRs) so we can ratchet it up over time as we add component tests instead of blocking every PR on coverage that we haven't yet built.

## What's wired

**Per-surface coverage tooling:**
- SDK + Platform API: \`pytest-cov\` with branch coverage. Config lives in each project's \`pyproject.toml\` under \`[tool.coverage.run]\` / \`[tool.coverage.report]\`.
- Dashboard: \`@vitest/coverage-v8\` with lcov + text reporters, scoped to \`src/\` excluding tests, shadcn primitives, the entry point, type stubs.

**CI uploads** (\`.github/workflows/tests.yml\`):
- Each job now produces a coverage report and uploads via \`codecov-action@v5\` with a flag matching the surface (\`sdk\`, \`platform-api\`, \`dashboard\`).
- \`fail_ci_if_error: false\` so a Codecov outage can't block merges — the test suite is the gate, the upload is reporting.

**codecov.yml at repo root:**
- Per-flag project targets with 1-2% thresholds so refactor noise doesn't trigger PR failures.
- Patch target of 70% on changed lines — more meaningful than the whole-repo number for a single PR.
- \`ignore:\` list excludes \`docs/\`, \`examples/\`, \`notebooks/\`, the stale \`fortify/\` tree, dashboard \`dist/\`.

**Local UX:**
- \`make coverage\` — pytest-cov with terminal + XML output.
- \`make coverage-html\` — same plus a browsable \`htmlcov/\` report.
- \`pnpm test:coverage\` (dashboard) — vitest with v8 coverage.
- \`.gitignore\` already aware of all coverage artifacts.

## Action required before merge

**Add \`CODECOV_TOKEN\` to repo secrets** (Settings → Secrets and variables → Actions → New repository secret).

For public OSS repos, Codecov v5 supports tokenless OIDC uploads, but a token is more reliable and matches what they document. Without the token, the upload steps will skip gracefully (no failure thanks to \`fail_ci_if_error: false\`) but the badge won't update.

## Why the dashboard floor is intentionally low

The dashboard's actual coverage today is ~27%. If we set the floor at 50%, every PR would fail on coverage until we backfill component tests — which is the kind of activation energy that delays merging real features. Start with the bar where the codebase actually is, ratchet by hand as we add tests. The \`informational: true\` flag means Codecov posts the number on PRs (so we see drops) but doesn't block.

## Test plan

- [x] \`make coverage\` — 833 passed, 84.1% branch coverage
- [x] \`cd platform/api && uv run pytest --cov\` — 318 passed, 79.7%
- [x] \`pnpm test:coverage\` (dashboard) — 51 passed, 26.9%
- [x] \`make check\` (no coverage involvement) — still green
- [ ] After merge: confirm CI Codecov uploads land in the repo's Codecov dashboard
- [ ] After merge: confirm the badge in the README renders the SDK percentage

## Scope notes

- README badge points at the SDK flag specifically (the most marketable number). Per-flag breakdowns visible in the Codecov dashboard for anyone who wants them.
- Coverage thresholds in codecov.yml only, not in vitest.config.ts or pyproject.toml — keeps the "fail PR on coverage drop" lever in one place across all three surfaces.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,6 +51,19 @@ test-one: ## Run one test path: make test-one T=tests/security/test_bundle.py
 	@test -n "$(T)" || (echo "Set T=<path>, e.g. make test-one T=tests/security/test_bundle.py" && exit 1)
 	$(UV) pytest $(T) -v
 
+.PHONY: coverage
+coverage: ## Run the SDK suite with branch coverage (terminal + xml for CI)
+	# `uv run` without --active so pytest-cov resolves from the project's
+	# .venv. If you drive uv with UV_PROJECT_ENVIRONMENT (micromamba etc.)
+	# you'll need pytest-cov installed there too — `uv sync --extra dev`
+	# against the same env.
+	uv run pytest --cov --cov-report=xml --cov-report=term $(TESTS)
+
+.PHONY: coverage-html
+coverage-html: ## Coverage with a browsable HTML report under htmlcov/
+	uv run pytest --cov --cov-report=html --cov-report=term $(TESTS)
+	@echo "Open htmlcov/index.html in a browser."
+
 .PHONY: lint
 lint: ## Static check via ruff
 	$(UV) ruff check hexgate tests
--- a/Makefile
+++ b/Makefile
@@ -81,11 +81,19 @@ fmt-check: ## Check formatting without writing changes
 	$(UV) ruff format --check hexgate tests platform/api
 
 .PHONY: check
-check: lint fmt-check test ## Python CI parity: lint + fmt-check + test
+check: lint fmt-check test ## Python CI parity: lint + fmt-check + test (no coverage overhead)
 
 .PHONY: check-all
-check-all: check dashboard-lint dashboard-typecheck dashboard-fmt-check ## Full stack check: Python + dashboard lint, typecheck, fmt, tests
-	cd platform/dashboard && pnpm test --run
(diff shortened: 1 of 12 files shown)


guillaume-hexamind (approved):

Makefile
victorludvig: Can we add coverage here?

victorludvig (approved):
Approved, we can add coverage to check-all but it can be done in a follow-up PR.

Makefile
quanghexa94: done! it's added

PR #52, opened 2026-06-25
feat(platform/api): wire Resend for real verification + reset emails

quanghexa94:
## Summary

Wires [Resend](https://resend.com) behind the existing `EmailSender` Protocol seam — verification, password-reset, and invitation emails flow through real delivery in production. Dev mode is unchanged (still prints to stderr). All four high-priority findings from the inline `/code-review` pass are addressed.

## What's behind the seam

```
platform/api/mailer.py
  ├─ EmailSender (Protocol)           — unchanged
  ├─ StderrEmailSender                — dev default, prints to stderr
  └─ ResendEmailSender                — NEW. wraps resend.Emails.send
```

`main.py`'s lifespan picks the sender at boot:

| `RESEND_API_KEY` | `HEXGATE_EMAIL_FROM` | Result | Log level |
|---|---|---|---|
| set | set | `ResendEmailSender` | INFO |
| unset | unset | `StderrEmailSender` (dev) | INFO |
| set XOR set | — | `StderrEmailSender` + warning | **WARNING** naming the missing var |

The three-way differentiation is deliberate: a half-rotated deployment with only `RESEND_API_KEY` set should NOT read as a benign dev-mode notice. The else-branch also explicitly resets to `StderrEmailSender` so a re-config that clears env vars can't leave a stale Resend sender silently shipping mail.

## Failure contract

- **`ResendEmailSender.send` re-raises** after logging ERROR. The `fastapi-users` `on_after_request_verify` + `on_after_forgot_password` hooks wrap the call in `try/except` so the `/v1/auth/...` endpoints still return 200. The invitation flow in `main.py:1796` already had a `try/except` — re-raise restores the org-scoped log signal it was waiting for.
- **Per-instance api_key** on the wrapper, not on the `resend` module global. `_send_sync` sets `resend.api_key` under a module-level lock immediately before each call, so two senders with different keys can't race. Throughput cost is irrelevant at our rate (single-digit emails/sec at peak).
- **PII-redacted error logs**. `_redact_email` keeps the first 2 chars of the local-part + the domain (\`al***@example.com\`); fully masks local-parts shorter than 5 chars; never leaks more than ~40% of an address. Domain is preserved so \"all our gmail users are failing\" stays grep-able. Body is never logged — that would expose the magic-link token. Same helper is reused for the invitation-flow log in \`main.py:1796\`.

## Env config

\`\`\`
# platform/api/.env
RESEND_API_KEY=re_…
HEXGATE_EMAIL_FROM=\"Hexgate <noreply@yourdomain.com>\"
HEXGATE_DASHBOARD_URL=https://app.yourdomain.com   # link host inside the email
\`\`\`

From-address must be on a verified Resend domain (see [Resend → Domains](https://resend.com/domains)).

## Tests

\`platform/api/tests/test_mailer.py\` — 17 cases:

- \`_redact_email\` parametrized over 8 cases (incl. short local-part + malformed input)
- Payload shape (from / to-list / subject / text)
- Per-instance api_key: A's send uses A's key, B's send uses B's
- \`to_thread\` offload verified by observed thread id
- Re-raise contract: \`RuntimeError\` propagates AND ERROR log emitted
- PII redaction: local-part and tag NOT in log; partial form IS; token NOT
- Wiring: assert \`_from\` + \`_api_key\` values (not just \`isinstance\`)
- Three-way log: WARN on partial config, INFO on clean dev mode
- Re-config reset: clearing env vars after wiring Resend reverts to stderr

## Code-review fixes folded in

This PR includes the four findings from \`/code-review xhigh\` on the initial draft:

1. **#1 PII redaction** — recipient redacted in mailer.py and main.py:1796 (invitation flow). Body never logged.
2. **#2 Re-raise** — provider failures now reach the caller; auth hooks wrap them.
3. **#3 Per-instance api_key** — module-global mutation serialized under a lock.
4. **#5 Three-way log discrimination** — partial config emits WARNING.

## Action required before this is useful

Add \`RESEND_API_KEY\` and \`HEXGATE_EMAIL_FROM\` to the production environment. Until then the platform stays on \`StderrEmailSender\` (no behavior change).

## Test plan

- [x] \`platform/api\`: 335 passed, 2 skipped
- [x] SDK: 844 passed, 2 skipped
- [x] \`ruff check\` clean on touched files
- [ ] After merge: configure Resend domain, set the two env vars, register a test account, confirm the link in the inbox routes to the dashboard.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -40,7 +40,12 @@ jobs:
       # Pin a known-good version so an upstream OPA change can't break CI overnight.
       - name: Install OPA
         run: |
-          curl -fsSL -o opa https://openpolicyagent.org/downloads/v0.70.0/opa_linux_amd64_static
+          # --retry-all-errors covers curl 56 (recv reset mid-download) and
+          # other transient TCP hiccups against openpolicyagent.org's CDN
+          # that --retry alone won't catch. ~3min worst case before the job
+          # gives up, much better than a one-shot fail forcing a re-run.
+          curl -fsSL --retry 5 --retry-all-errors --retry-delay 2 --retry-max-time 60 \
+            -o opa https://openpolicyagent.org/downloads/v0.70.0/opa_linux_amd64_static
           chmod +x opa
           sudo mv opa /usr/local/bin/opa
           opa version
@@ -94,7 +99,12 @@ jobs:
       # ETag header. Same pin as the SDK job.
       - name: Install OPA
         run: |
-          curl -fsSL -o opa https://openpolicyagent.org/downloads/v0.70.0/opa_linux_amd64_static
+          # --retry-all-errors covers curl 56 (recv reset mid-download) and
+          # other transient TCP hiccups against openpolicyagent.org's CDN
+          # that --retry alone won't catch. ~3min worst case before the job
+          # gives up, much better than a one-shot fail forcing a re-run.
+          curl -fsSL --retry 5 --retry-all-errors --retry-delay 2 --retry-max-time 60 \
+            -o opa https://openpolicyagent.org/downloads/v0.70.0/opa_linux_amd64_static
           chmod +x opa
           sudo mv opa /usr/local/bin/opa
           opa version
--- a/README.md
+++ b/README.md
@@ -243,6 +243,22 @@ Integration tests (`pytest -m integration`) round-trip rows through the live Cli
 
 The dashboard's `/policies` page lets you edit each agent's policy. `hexgate serve` re-fetches at every turn boundary, so your edits take effect on the next chat message without a restart.
 
+### Outbound email (Resend)
+
+Verification emails and password-reset links go through [Resend](https://resend.com). The platform API has two senders:
+
+- **Dev (default)** — `StderrEmailSender` prints the mail body (including the magic link) to stderr. No keys needed; click the link out of the terminal to exercise the flow end-to-end.
+- **Production** — `ResendEmailSender`, picked up automatically when both `RESEND_API_KEY` and `HEXGATE_EMAIL_FROM` are set. The from-address must be on a verified Resend domain.
+
+```bash
+# platform/api/.env
+RESEND_API_KEY=re_…
+HEXGATE_EMAIL_FROM="Hexgate <noreply@yourdomain.com>"
+HEXGATE_DASHBOARD_URL=https://app.yourdomain.com   # ← link host inside the email
+```
+
+Partial config (only one var set) keeps the dev stderr sender so an operator notices the misconfig in the startup log line instead of silently failing every send at delivery. Resend API failures are logged at error level but never 5xx the calling endpoint — the user can request a new link.
+
 ## ✨ Core Primitives
 
 The two main primitives are:
--- a/docs/docs.json
+++ b/docs/docs.json
@@ -66,7 +66,8 @@
               "platform/control-plane",
               "platform/dashboard",
               "platform/live-demo",
-              "platform/audit-clickhouse"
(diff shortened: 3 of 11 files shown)


guillaume-hexamind:
Claude review:

🔴 1 — Silent send failures: endpoints return 200 even when no email was sent — 72/100

platform/api/auth.py:225–231, 254–255 · CONFIRMED

on_after_forgot_password and on_after_request_verify wrap the send in try/except Exception that only logs. During a Resend outage the user gets a normal "check your inbox" response, but no email arrives — no retry, no alert, no user-facing signal. For password reset this can lock a user out of account recovery for the whole outage, with only an operator tailing logs aware.

▎ Note: returning 200 regardless of whether the account exists is intentional anti-enumeration. The defect is the absence of any failure path (retry queue / metric / alert), not the 200 itself. Worth a follow-up rather than a blocker.

---
🟠 2 — Registration still logs full email in cleartext — 65/100

platform/api/auth.py:196 (also main.py:1817 context) · CONFIRMED

on_after_register writes user registered: <full-email> (id=...) to the platform.auth logger. The PR added _redact_email everywhere else specifically to keep addresses out of log stores — but the happy path still ships full PII into log aggregation, defeating the stated boundary. (This line sits in a function the PR didn't touch, so it's adjacent scope — but it directly undercuts the PR's own goal.)

---
🟠 3 — Global lock held across the blocking HTTP send serializes all outbound mail — 55/100

platform/api/mailer.py:143–147 · CONFIRMED

_resend_send_lock is held across both the cheap resend.api_key = ... assignment and the blocking resend.Emails.send() network round-trip. When Resend is slow, one in-flight send stalls every verification/reset/invitation email process-wide until it times out, even though the sends are independent. Fix: hold the lock only around the api_key mutation (snapshot/set/restore) and release before the network call. (A harsher variant claiming "no timeout → blocks forever" was refuted — the SDK does pass a timeout.)

guillaume-hexamind (commented):
Logic looks good! You can check the claude warnings to see if they are relevant

winlp4ever:
Thanks for the review! Triaged the three Claude findings:

**#2 — full email in `on_after_register` log → fixed in `91ab459`.**
Reused `_redact_email`. One-line change, aligned with the PII boundary the rest of the PR established.

**#1 — silent send failures, no failure path → deferring to a follow-up.**
Agreed the gap is real — during a Resend outage we log ERROR and the user sees a normal 200. But as the review note itself flags, the 200 is load-bearing (anti-enumeration on `/auth/forgot-password`), so the fix isn't user-facing: it's wiring the mailer's ERROR log to an alert source (Sentry / metric counter that pages on-call). That's its own change with no overlap to this PR. Will open a follow-up issue.

**#3 — global lock across the blocking send → kept, with a comment.**
The simpler "release before the HTTP call" fix doesn't work as suggested — the resend SDK reads `resend.api_key` inside `Request.__init__` when building the Authorization header. Releasing the lock before `.perform_with_content()` would let a concurrent sender overwrite the global between auth-header construction and the actual HTTP send. The clean fix that *does* work reaches into `resend.request.Request` private API to construct under the lock + send outside, which is fragile against any SDK refactor.

Today there's only one `ResendEmailSender` instance per process (created once in `main.py`'s lifespan), so the lock never actually sees contention — it's belt-and-suspenders for the multi-tenant case where we'd need to swap keys. I added a comment in `mailer.py` explaining the constraint + the future volume-driven refactor path, so the next maintainer doesn't have to re-derive it. At our volume (single-digit emails/sec) the serialization cost is invisible.

Re-running CI now — the OPA-curl-retry commit in this PR should also help any future flakes on the install step.

guillaume-hexamind (approved):

PR #54, opened 2026-06-26
feat(platform/api) Add audit/anomalies API endpoint

victorludvig:
## What is changing
A new API endpoint is added for anomaly detection.
A sliding-window function is added to detect per-user anomalies.
An anomaly is defined by 50% of denies in a 1-hour time window containing more than 5 decisions.
The approach is straightforward and not fully optimised for simplicity, refer to code comments for more details.

## Test
make check-all

### DIFF
--- a/platform/api/audit.py
+++ b/platform/api/audit.py
@@ -14,10 +14,11 @@ from __future__ import annotations
 import json
 import logging
 from datetime import datetime, timedelta, timezone
+from collections import deque
 
 from clickhouse_connect.driver.client import Client
 
-from schemas import DecisionEvent
+from schemas import AnomalySeverity, AuditAnomaly, AuditOutcome, DecisionEvent
 
 _log = logging.getLogger(__name__)
 
@@ -42,6 +43,12 @@ MAX_HINT_BYTES = 4 * 1024
 # older than retention — rows past TTL would be merged away on arrival.
 CLOCK_SKEW_FUTURE = timedelta(minutes=5)
 RETENTION_WINDOW = timedelta(days=90)
+
+_ANOMALY_MIN_REQUESTS = 5
+_TIMEDELTA_ANOMALY_HOURS = 1
+_WINDOW_TD = timedelta(hours=_TIMEDELTA_ANOMALY_HOURS)
+_DENY_RATE_MEDIUM = 0.3
+_DENY_RATE_HIGH = 0.5
 
 
 def validate_event_window(occurred_at: datetime) -> None:
@@ -158,7 +165,7 @@ def bucket_minutes_for_timedelta(delta: timedelta) -> int:
 
 
 def _zero_counts() -> dict[str, int]:
-    return {"all": 0, "allow": 0, "deny": 0, "needs_approval": 0}
+    return {"all": 0, **{e.value: 0 for e in AuditOutcome}}
 
 
 def _date_range_valid(start: datetime | None, end: datetime | None) -> bool:
@@ -338,7 +345,7 @@ def timeseries(
     points: dict[object, dict] = {}
     for t, outcome, n in result.result_rows:
         point = points.setdefault(
-            t, {"bucket": t, "allow": 0, "deny": 0, "needs_approval": 0}
+            t, {"bucket": t, **{e.value: 0 for e in AuditOutcome}}
         )
         if outcome in point:
             point[outcome] = int(n)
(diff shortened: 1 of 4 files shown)


platform/api/audit.py
```diff
@@ -431,6 +437,129 @@ def list_decisions(
...
+            _flush(best_burst)
+            curr_user = user
+            curr_deny = 0
+            window.clear()
+            best_burst = None
+
+        while window and timestamp - window[0][0] > timedelta(
+            hours=_TIMEDELTA_ANOMALY_HOURS
+        ):
+            _, old_decision = window.popleft()
```
victorludvig, line 462: **Bug: strict `>` comparison freezes `last_seen` when the window slides at constant rate and size**

When an old DENY is evicted from the left and a new DENY enters on the right, `deny_rate` and `window_length` stay identical. `(deny_rate, window_length) > (best_burst["deny_rate"], best_burst["all"])` is `False`, so `best_burst` is never updated — `last_seen` stays frozen at the moment the window first hit threshold.

Reproduced: 5 DENYs at T+0..T+20 min, 4 more at T+65..T+80 min (each evicts one old DENY). Reported `last_seen = T+20`, actual burst end `T+80` — off by 60 minutes.

**Fix:** change `>` to `>=`:
```python
if best_burst is None or (deny_rate, window_length) >= (
    best_burst["deny_rate"],
    best_burst["all"],
):
```
The existing `test_when_deny_rate_decreases_within_burst_then_peak_is_not_overwritten` still passes because the dip events produce `deny_rate < best_burst["deny_rate"]`, so the tuple comparison stays `False` regardless.

victorludvig (commented):
## Review

This PR adds a  endpoint backed by two ClickHouse queries and a pure-Python O(n) sliding-window burst detector. The algorithm is well-designed and the test suite is thorough. Six findings below — one correctness bug worth fixing before merge, the rest are cleanup and future-proofing.

Findings are inline where possible, general ones are in top-level comments below each inline thread.

platform/api/audit.py
```diff
@@ -431,6 +437,129 @@ def list_decisions(
...
+            candidate = dict(
+                user_id=curr_user,
+                severity=severity,
+                deny=curr_deny,
+                all=window_length,
+                deny_rate=deny_rate,
+                first_seen=window[0][0],
+                last_seen=window[-1][0],
+            )
+            if best_burst is None or (deny_rate, window_length) > (
```
victorludvig: **Bug: strict `>` comparison freezes `last_seen` when the window slides at constant rate and size**

When an old DENY is evicted from the left and a new DENY enters on the right, `deny_rate` and `window_length` stay identical. `(deny_rate, window_length) > (best_burst["deny_rate"], best_burst["all"])` evaluates to `False`, so `best_burst` is never updated — `last_seen` stays frozen at the moment the window first hit threshold.

Reproduced: 5 DENYs at T+0..T+20 min, 4 more at T+65..T+80 min (each evicts one old DENY). Reported `last_seen = T+20`, actual burst end `T+80` — off by 60 minutes.

**Fix:** change `>` to `>=`:
```python
if best_burst is None or (deny_rate, window_length) >= (
    best_burst["deny_rate"],
    best_burst["all"],
):
```
The existing `test_when_deny_rate_decreases_within_burst_then_peak_is_not_overwritten` still passes because dip events produce `deny_rate < best_burst["deny_rate"]`, so the tuple comparison stays `False` regardless.

platform/api/audit.py
```diff
@@ -431,6 +437,129 @@ def list_decisions(
...
+
+    for user, timestamp, decision in rows:
+        if user != curr_user:
+            _flush(best_burst)
+            curr_user = user
+            curr_deny = 0
+            window.clear()
+            best_burst = None
+
+        while window and timestamp - window[0][0] > timedelta(
```
victorludvig: **Efficiency: `timedelta` reconstructed on every iteration**

`timedelta(hours=_TIMEDELTA_ANOMALY_HOURS)` is called on every row processed, even rows where the eviction loop body never runs. At 1 M rows that's ~1 M avoidable allocations.

**Fix:** hoist to a module-level constant alongside the other thresholds:
```python
_WINDOW_TD = timedelta(hours=_TIMEDELTA_ANOMALY_HOURS)
```
Then replace `timedelta(hours=_TIMEDELTA_ANOMALY_HOURS)` in the loop with `_WINDOW_TD`.

platform/api/main.py
```diff
@@ -1264,6 +1266,33 @@ async def api_audit_decisions(
...
+    response_model=list[AuditAnomaly],
+    dependencies=[Depends(require_org_member)],
+    tags=["audit"],
+)
+async def api_audit_anomalies(
+    project_id: str,
+    window: AuditWindow = "24h",
+    start_date: datetime | None = None,
+    end_date: datetime | None = None,
+    clickhouse_client=Depends(require_clickhouse),
```
victorludvig, line 1280: **Missing agent / role / tool scope filters**

The other three audit endpoints (`summary`, `timeseries`, `decisions`) all accept `agent`, `role`, and `tool` query params via `_scope()`. This endpoint doesn't, so callers that add `?agent=my-agent` expecting scoped anomalies will silently get org-wide results.

When the dashboard adds agent filtering to this panel (consistent with the filter bar already built for the others), a new API version will be needed to add the params. Adding them now as optional `None` defaults costs two lines and would keep the surface consistent.

victorludvig:
**`main.py:1239` — Stale `Literal` missed during `AuditOutcome` consolidation**

This PR adds `AuditOutcome` to centralise the `"allow" | "deny" | "needs_approval"` triple, but the `outcome` query param on `api_audit_decisions` (line 1239) was not updated and is now the only remaining raw `Literal["allow", "deny", "needs_approval"]` in the codebase.

```python
# platform/api/main.py:1239
outcome: AuditOutcome | None = None,
```
FastAPI will coerce the query string value through the StrEnum, so the behaviour is identical.

victorludvig:
**`audit.py:167,347` — `_zero_counts()` and the `timeseries()` inline dict hardcode the outcome key set**

Both sites spell out `"allow"`, `"deny"`, `"needs_approval"` as string literals. Now that `AuditOutcome` owns this set, adding a fourth value to the enum won't update either of these automatically — silent omission in summary and timeseries responses.

```python
# platform/api/audit.py:166
def _zero_counts() -> dict[str, int]:
    return {"all": 0, **{e.value: 0 for e in AuditOutcome}}

# platform/api/audit.py:347
point = points.setdefault(
    t, {"bucket": t, **{e.value: 0 for e in AuditOutcome}}
)
```

platform/api/audit.py
victorludvig: I fixed it and added a test with the proposed data.

platform/api/audit.py
victorludvig: Same comment as above. It's the same fix.

platform/api/audit.py
victorludvig: Fixed.

platform/api/main.py
victorludvig: Not necessary for this PR, discarded.

victorludvig:
> **`main.py:1239` — Stale `Literal` missed during `AuditOutcome` consolidation**
> 
> This PR adds `AuditOutcome` to centralise the `"allow" | "deny" | "needs_approval"` triple, but the `outcome` query param on `api_audit_decisions` (line 1239) was not updated and is now the only remaining raw `Literal["allow", "deny", "needs_approval"]` in the codebase.
> 
> ```python
> # platform/api/main.py:1239
> outcome: AuditOutcome | None = None,
> ```
> 
> FastAPI will coerce the query string value through the StrEnum, so the behaviour is identical.

Fixed.

guillaume-hexamind (approved):
LGTM. As pointed out in the comment, we'll have to keep in mind the scalability issue once the problem becomes relevant

PR #55, opened 2026-06-29
feat(platform/api): Add by_user to audit API endpoint

victorludvig:
## What is changing
The user field is added to summary, timeseries and list_decisions audit API endpoint and scope function.
It will be used to filter by user when clicking on the anomaly card.
The summary sql construction is also consolidated with a list, to prevent spacing issues.

## Test
make check-all


victorludvig (commented):
**Code review generated by Claude**

This PR adds `user` as a filter dimension across the audit stack — `_scope()`, `summarize()`, `timeseries()`, and `list_decisions()` — and exposes a `by_user` breakdown in the summary response. The SQL construction refactor from implicit string concatenation to a list-join is a clear improvement.

Two actionable findings below, plus one cleanup note.

platform/api/audit.py
```diff
@@ -205,6 +206,9 @@ def _scope(
     if tool:
         where.append("tool_name = {tool:String}")
         params["tool"] = tool
+    if user:
```
victorludvig: **[Generated by Claude]** `if user:` should be `if user is not None:` for consistency with `role` (line 203 uses `if role is not None:`). The `_scope()` docstring explicitly documents `role=""` as the "no-role bucket" drill-down, and `DecisionEvent.user_id` defaults to `""` — so an empty-string user_id bucket is possible in the data. An `if user:` guard silently skips that case, widening the filter to all users instead of the no-user-id subset.

platform/api/audit.py
```diff
@@ -266,17 +284,19 @@ def _add(store: dict[str, dict[str, int]], key: str, outcome: str, n: int) -> No
         agent,
         role,
         tool,
+        user,
```
victorludvig, line 294: **[Generated by Claude]** The loop unpacks this row column into a local variable named `user`, which shadows the `user: str | None` filter parameter from the function signature. Consider renaming it to `user_id` (matching the column name) so the two concepts don't collide. As written, any code inside the loop that references `user` gets the row's value, not the filter — a latent bug on the next edit.

platform/api/main.py
```diff
@@ -1165,6 +1165,7 @@ async def api_audit_summary(
     agent: str | None = None,
     role: str | None = None,
     tool: str | None = None,
+    user: str | None = None,
```
victorludvig, line 1170: **[Generated by Claude]** `user` is correctly wired into `api_audit_summary` here, but `api_audit_timeseries` and `api_audit_decisions` don't expose it as a query param and don't forward it to their underlying `timeseries()` / `list_decisions()` calls — even though both those functions now accept it. A future user drill-down passing `?user=Alice` on all three endpoints would get filtered summary data but unfiltered chart and table data (silent split-brain).

platform/api/audit.py
victorludvig: Fixed.

platform/api/audit.py
victorludvig: This is not really an issue because we only use user to build the sql query, and it is consistent with what is done with agent, role and tool. Ignoring.

platform/api/main.py
victorludvig: Fixed.

guillaume-hexamind (approved):
LGTM!

PR #57, opened 2026-06-29
feat(mcp): auto-wrap MCP server tools as policy-enforced BaseTools

quanghexa94:
## Summary

PR #1 of a planned MCP series. Introduces `hexgate.mcp` — attach any [Model Context Protocol](https://modelcontextprotocol.io) server, every tool the server exposes auto-registers under the namespace `mcp-<server>-<tool>` as a LangChain `BaseTool`. The existing `enforce_policy(agent, policy)` pass wraps each one in `GuardedTool`, so MCP tools become indistinguishable from native `@agent_tool` functions to the rest of the runtime: same policy enforcement, same audit, same approval flow, same `{"ok": ..., "content"/"error"}` return envelope.

## Usage

```python
from hexgate import create_agent, enforce_policy
from hexgate.mcp import MCPServerConfig, MCPToolset

slack = MCPServerConfig(
    name="slack", transport="stdio",
    command="slack-mcp-server",
    env={"SLACK_TOKEN": "..."},
)

async with MCPToolset(slack) as mcp:
    agent, handler = create_agent(model="gpt-5.4", tools=mcp.tools)
    agent = enforce_policy(agent, "policy.yaml")
    await agent.ainvoke({"messages": [...]}, config={"configurable": {}})
```

Policy.yaml references MCP tools by their qualified names:

```yaml
roles:
  default:
    tools:
      "mcp-slack-list_channels": { mode: allow }
      "mcp-slack-send_message":
        mode: approval_required
      "mcp-github-create_issue":
        mode: allow
        constraints: ["args.repo == 'hexgate'"]
```

## Try it

```bash
make demo-mcp
```

Runs `examples/mcp_demo.py` — one self-contained file that spawns a tiny FastMCP server, attaches it via `MCPToolset`, and walks through one call per policy outcome (allow / deny / approval-required). No external services, no LLM key required.

## Design decisions

- **Hyphen separator, not colon** — OpenAI Function Calling rejects `:` in tool names. Format `mcp-<server>-<tool>` works across OpenAI + Anthropic.
- **Server name validated to `^[a-z0-9-]{1,32}$`** — keeps qualified names within OpenAI's 64-char tool-name limit.
- **Zero coupling to `PolicyEnforcer`** in `hexgate.mcp` — the proxy just produces `BaseTool` instances; policy gets grafted on by the existing `enforce_policy(...)` pipeline.
- **`MCPToolset` is an async context manager** — explicit lifecycle, atomic registration via `AsyncExitStack` (no orphaned subprocesses under `CancelledError`), symmetric teardown.
- **Output envelope matches `@agent_tool`** — `{"ok": True, "content": ...}` / `{"ok": False, "error": {...}}`, so callers can discriminate uniformly across origins.

## Hardening — addressed all 15 findings from an xhigh code-review pass

| # | Finding | Fix |
|---|---|---|
| 1 | README example doesn't unpack `create_agent`'s tuple | Fixed in README + Mintlify page |
| 2 | Docstring example calls `mcp_tools(name=..., ...)` — no such signature | Deleted helper, fixed docstring to show `MCPToolset` |
| 3 | `env=dict(env) or None` silently collapses `env={}` (sandbox) to `None` (inherit) | `env: dict \| None` — `None`/inherit and `{}`/empty now distinct |
| 4 | `structuredContent` (MCP's typed return path) discarded | Rendered as JSON text in the envelope |
| 5 | No `read_timeout_seconds` on `call_tool` — wedged server stalls forever | `call_timeout_seconds` field on config, default 30s |
| 6 | `list_tools` ignores `nextCursor` — tools past page 1 silently lost | Pagination loop |
| 7 | Proxy doesn't catch `RuntimeError` from SDK — aborts agent run | Catches everything, returns error envelope |
| 8 | Cleanup masks original handshake exception | `contextlib.suppress` on teardown, chain with `raise from` |
| 9 | `exc_info` dropped in `__aexit__` | Forwarded to inner stack |
| 10 | Post-close call returns cryptic `MCPClient` error | Sentinel `_ToolsetState.open`; clear envelope referring to `MCPToolset` |
| 11 | `CancelledError` between client `__aenter__` and `append` leaks subprocess | `AsyncExitStack.enter_async_context` is atomic |
| 12 | Duplicate `name` configs produce duplicate tool names (OpenAI BadRequestError) | Reject at construction with clear error |
| 13 | `args_schema=dict` skips LangChain validation — partial args reach policy | jsonschema validation in the proxy returns error envelope; closes the round-trip waste, partial-policy-args caveat documented as follow-up |
| 14 | Proxy returns bare `str`; native tools return envelope dict | Unified envelope shape |
| 15 | `mcp_tools()` helper returned entered toolset with no `aclose` — leak by design | Deleted; `async with MCPToolset(...)` is the only entry point |

## Test plan

- [x] 42 cases in `tests/mcp/test_{config,proxy}.py` (was 30 before review; +12 covering the fixes)
- [x] Full SDK suite: 886 passed, 2 skipped (was 844 pre-PR, +42 new)
- [x] \`ruff check\` clean
- [x] \`make demo-mcp\` runs end-to-end (allow / deny / approval-required all render correctly)
- [ ] After merge: try with the real \`@modelcontextprotocol/server-filesystem\` to confirm npx-style stdio servers work

## Scope + roadmap

This PR is the runtime proxy. Explicit follow-ups (separate PRs):

- **2** — MCP tool registration in the platform manifest (so \`/agents\` view + auto-policy generation see MCP tools)
- **3** — Catalog threat scanner (typosquatting + prompt-injection pattern detection at registration time — mirrors Microsoft AGT's \`McpSecurityScanner\`)
- **4** — Declarative \`mcp_servers:\` block in \`agent.yaml\`
- **5** — Response-side gating (post-call evaluator + redact primitive)
- **6** — Pydantic-model arg validation upstream of policy (closes the "policy sees partial args" caveat — today validation only fires after policy)


victorludvig:
Can you check these findings from Claude please:

## 1) hexgate/mcp/proxy.py:843 — falsy-zero check silently replaces {} inputSchema with a more-constrained fallback

schema = mcp_tool.inputSchema or {"type": "object", "properties": {}} — an empty dict {} is falsy in Python, so a server that returns {} for inputSchema (valid JSON Schema meaning "accept anything") triggers the fallback. The replacement {"type": "object", "properties": {}} is semantically different: it adds a type: object constraint and is also passed to StructuredTool.from_function(args_schema=schema), changing the LLM-visible argument spec. The fix is a one-character change: if mcp_tool.inputSchema is None.


## 2) hexgate/mcp/client.py:502 — MCPClient.__aexit__ drops the return value of stack.__aexit__(), breaking exception suppression

if stack is not None:
    await stack.__aexit__(exc_type, exc, tb)   # return value discarded

The async CM protocol requires __aexit__ to return a truthy value to suppress an exception; returning None unconditionally means any inner CM that wants to suppress (e.g., a custom transport that handles a graceful disconnect) is ignored and the exception propagates anyway. Should be return await stack.__aexit__(exc_type, exc, tb). The same pattern appears in MCPToolset.__aexit__ at line 814.

## 3) hexgate/mcp/proxy.py:815 — _states not cleared in __aexit__, accumulates stale entries on re-entry

__aexit__ does self._tools.clear() but never clears self._states. There is no guard preventing __aenter__ from being called twice on the same MCPToolset instance. On each re-entry, new _ToolsetState objects are appended to the existing list; old closed states with open=False persist indefinitely. Any code that iterates mcp._states (including the test at line 2166 that already does so) would see phantom entries. Fix: add self._states.clear() alongside self._tools.clear().

## 4) hexgate/mcp/proxy.py:994 — tool_error (server-side isError=True) marked retryable=True, but no existing code reads the field

"retryable": kind in {"transport_error", "tool_error"},

MCP's isError=True covers deterministic application failures (invalid input accepted past schema validation, resource not found, permissions denied) that will reproduce identically on retry — marking them retryable is misleading. More concretely: no caller in the current codebase reads retryable (confirmed by search — normalize.py reads only ok and error.message), so the field is inert today but diverges from the native Decision.as_error_payload() shape (which always sets retryable: False). If a future consumer is added expecting the native shape, tool_error cases will get the wrong value. Consider dropping retryable from the MCP envelope or constraining it to transport_error only.

quanghexa94:
Thanks @victorludvig — all four findings addressed in `eac6737`.

**#1 — empty `{}` inputSchema falsy-replaced** ✅
One-character fix: `is None` instead of `or`. Regression test: `test_proxy_preserves_empty_input_schema` (a server that returns `inputSchema={}` now reaches LangChain unchanged).

**#2 — `__aexit__` discards stack suppression** ✅
Fixed in BOTH `MCPClient.__aexit__` and `MCPToolset.__aexit__` — `return await stack.__aexit__(...)`. Pinned by two new tests that monkeypatch a `_SuppressingSession` / `_SuppressingClient` returning `True` from `__aexit__` and assert the with-block does NOT re-raise.

**#3 — `_states` not cleared** ✅
Now `_tools.clear()` + `_states.clear()` symmetrically. Regression test asserts both internal lists are empty after exit.

**#4 — `tool_error` retryable=True** ✅
You're right that it diverges from native `Decision.as_error_payload`. Constrained to `transport_error` only. Renamed + flipped the existing test that had pinned the wrong behavior (`test_error_envelope_only_transport_error_is_retryable`).

Full SDK suite: 911 passed (was 907; +4 net new tests covering the four fixes). CI re-running on the push.

victorludvig (approved):

PR #58, opened 2026-06-30
feat(dashboard): Add anomaly detection card

victorludvig:
## What is changing
The anomaly detection card is added to the audit dashboard.
When clicking on an anomaly, a user filter and a date filter are applied to the whole page:
- date filter: day of the anomaly
- user filter: user responsible for the detected anomaly

The anomaly endpoint only accepts time range filter, so a narrower AnomalyScope is created.

https://github.com/user-attachments/assets/2b5403e1-56a3-4a9b-8e1d-91bcb463f59e

## Tests
make check-all

### DIFF
--- a/platform/dashboard/src/components/audit/pieces.tsx
+++ b/platform/dashboard/src/components/audit/pieces.tsx
@@ -5,9 +5,12 @@ import {
   CircleDashed,
   Download,
   List,
+  TriangleAlert,
   X,
 } from "lucide-react";
 import type {
+  AnomalySeverity,
+  AuditAnomaly,
   AuditBreakdownRow,
   AuditDecisionRow,
   AuditOutcome,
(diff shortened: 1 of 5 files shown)


victorludvig (commented):
**EDIT**: All findings fixed for revision 2.

**Reviewed by Claude** (claude-sonnet-4-6)

Three findings from this review:

---

**1.  accepts  but backend only uses 3 fields**
`api.ts:350` — `getAuditAnomalies` takes a full `AuditScope` (which now includes `user`, `agent`, `role`, `tool`) and spreads it into the query string, but `GET /audit/anomalies` only accepts `window`, `start_date`, `end_date`. The current call site is correct (`anomalyScope` is intentionally narrowed), but the function signature over-promises — a future caller passing the full `scope` would silently send params the backend ignores with no compile-time warning. Consider a dedicated `AnomalyScope` type or at least a comment on the function.

---

**2. `anomalyScope` intentional narrowing is undocumented**
`Audit.tsx:392` — The object deliberately excludes `user`/`agent`/`role`/`tool` because the anomaly endpoint is cross-user by design. Without a comment, a future contributor adding a new filter dimension to `scope` is likely to copy the pattern from `summaryQ`/`decisionsQ` and accidentally start passing it here. A one-line comment would prevent this.

---

**3. PR title missing colon**
Title reads `feat(dashboard) Add anomaly detection card` — should be `feat(dashboard): Add anomaly detection card` per Conventional Commits (`type(scope): description`).

guillaume-hexamind (approved):
Looks good to me!

PR #60, opened 2026-07-01
refactor(hexgate): HEXGATE_KEY -> HEXGATE_API_KEY rename, HEXGATE_API_URL default to saas.

guillaume-hexamind:
Consolidates the SDK's environment-variable handling behind a new `hexgate/config/env.py` resolver.

## Changes

- **Rename `HEXGATE_KEY` → `HEXGATE_API_KEY`** across code, tests, docs, deploy scripts, and examples — for symmetry with `HEXGATE_API_URL` and the `api_key` param, and to disambiguate from the signing-key family (`HEXGATE_KEYSTORE_PATH`, `HEXGATE_PUBLIC_KEY`).
- **`HEXGATE_API_URL` now defaults to Hexgate Cloud** (`https://app.hexgate.ai`) instead of `localhost:8000`, centralized in a single constant. Self-hosters set it to `http://localhost:8000`.
- **Bootstrap no longer hard-requires `OPENAI_API_KEY` / `LINKUP_API_KEY` / `TAVILY_API_KEY`** — each is enforced lazily by the tool or model provider that reads it. Removed the now-empty `Settings.validate_required_keys()`.

## Migration

- Rename `HEXGATE_KEY` → `HEXGATE_API_KEY` in your `.env` / secret store.
- SaaS users can drop `HEXGATE_API_URL`; self-hosters point it at their local platform.

Tests: 842 passing, lint + format clean.

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -8,3 +8,9 @@ LANGFUSE_HOST=https://cloud.langfuse.com
 
 HEXGATE_DEFAULT_MODEL=openai:gpt-5.4
 HEXGATE_DEFAULT_SEARCH_ENGINE=linkup
+
+# Platform credential + URL. Unset URL defaults to Hexgate Cloud
+# (https://app.hexgate.ai). For local platform runs, point it at localhost —
+# a key only verifies against the platform instance that minted it.
+HEXGATE_API_KEY=
+# HEXGATE_API_URL=http://localhost:8000
--- a/Makefile
+++ b/Makefile
@@ -283,7 +283,7 @@ AGENT_SPEC ?= examples.customer_bot:agent
 
 .PHONY: serve
 serve: ## Run `hexgate serve` on the customer_bot demo (override with AGENT_SPEC=)
-# Reads HEXGATE_KEY from asianf/.env at startup. Uvicorn-style spec —
+# Reads HEXGATE_API_KEY from asianf/.env at startup. Uvicorn-style spec —
 # the agent name + tools come from the loaded object, no env vars to
 # keep in sync.
 	$(UV) hexgate serve $(AGENT_SPEC)
@@ -303,7 +303,7 @@ demo-platform: ## Print 3-terminal instructions for the full platform demo
 	@echo ""
 	@echo "  Terminal 3 — your local agent bridged to the platform:"
 	@echo "      1. Open  http://localhost:5173/tokens  and mint a dev token"
-	@echo "      2. Add to asianf/.env:  HEXGATE_KEY=fty_live_..."
+	@echo "      2. Add to asianf/.env:  HEXGATE_API_KEY=fty_live_..."
 	@echo "      3. make serve  (or: hexgate serve <your.module:agent>)"
 	@echo ""
 	@echo "Then chat with the live agent at  http://localhost:5173/playground"
--- a/README.md
+++ b/README.md
@@ -331,7 +331,7 @@ Same enforcement seam, same `User` scope. The difference is whose system of reco
 
 | What dev sets | What changes |
 |---|---|
-| `HEXGATE_API_KEY=fty_live_<project>_…` | Wakes up the platform path. Without it, adapters / `load_agent` fall back to local / builtin. (Formerly `HEXGATE_KEY`, still honored with a deprecation warning.) |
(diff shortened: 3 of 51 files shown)


victorludvig:
Here are some findings from Claude, can you check them please:

1. hexgate/bootstrap.py:55 / hexgate/config/settings.py — OPENAI_API_KEY loses its error path entirely, contradicting the PR's own claim. validate_required_keys() is deleted with no replacement. For LINKUP_API_KEY/TAVILY_API_KEY, hexgate/tools/websearch.py/fetch.py still raise a clear RuntimeError at use-time — but no equivalent exists for OPENAI_API_KEY anywhere in the repo. The model string flows straight into LangChain, which raises whatever generic third-party error it produces. Failure: any deployment missing OPENAI_API_KEY now boots fine and fails deep inside a model call with an opaque error, instead of the clean upfront RuntimeError it used to get — the opposite of the README's claim that "each raises a clear error at use-time."

2. hexgate/cli/chat.py:349 — missing-key errors for web_search/fetch now surface mid-conversation instead of at startup, and this hits the default agent path. build_runtime unconditionally wires web_search/fetch into any plain-id agent, including the default. Failure: a user runs hexgate chat with LINKUP_API_KEY unset, haon, and only discovers the missing key when the LLMhappens to invoke web_search — worse discoverability than the old upfront check.

3. hexgate/config/env.py:24 — the PR-promised legacyt exist. The PR description says "the old name is nolonger read; a one-time warning fires if it's set alone" — but resolve_api_key(), bootstrap.py, and audit.py contain zero rethe old HEXGATE_KEY name (confirmed: git history shoed in an early commit, then explicitly deleted in alater one, without updating the PR description). Failure: a user renames HEXGATE_KEY→HEXGATE_API_KEY in most places but miss.env/secret store/CI config — the app silently fallsudit with zero signal the old var name is the cause.

4. hexgate/config/env.py:33 — an empty-string HEXGATE_API_URL silently resolves to the production cloud URL. (explicit or   os.environ.get(API_URL_ENV) or DEFAULT_API_URL) treale template leaves HEXGATE_API_KEY= empty right nextto a commented-out HEXGATE_API_URL line, making an accidentally-empty URL plausible. Failure: a self-hoster with HEXGATE_API(empty) gets silently routed to https://app.hexgate.tform; their key (minted by their own instance) thenfails Biscuit signature verification with a confusing, unrelated-looking error.                                             5. Commit history — 3 of 5 commits violate this repot. refactor(hexgate): remove HEXGATE_KEY fallback andrefactor(hexgate): don't raise an error when LINKUP, TAVILY or OPENAI... use scope hexgate, which isn't in the allowed list (platform-api, dashboard, sdk, cli, clickhouse); refgate saas platform omits a scope entirely. Cost: these titles land as the permanent commit/squash-merge message on main, and future contributors copying the pattern will keep inventing ad hoc scopes.

6. hexgate/agents/factory.py:679 and hexgate/cli/_common.py — resolve_api_key() used as a bare boolean gate, then redundantly re-resolved seconds later via HexgateConfig.from_env_bind_policy checks resolve_api_key() then immediately builds HexgateConfig.from_env() (re-reading the same env var), and build_runtime_from_local_agent does the same around its
auto-register branch. Cost: minor duplicated work tohird key-resolution source is added later (e.g.keyring), 6 independent call sites need auditing instead of one HexgateConfig abstraction.

guillaume-hexamind:
@victorludvig thanks, Here are the feedback

1. fixed the doc claim, which was the real inaccuracy. One correction on the failure mode though: I tested it, and a missing OPENAI_API_KEY doesn't fail "deep inside a model call with an opaque error." It fails at create_agent(...) construction — before any model call — with a key-named message:
openai.OpenAIError: The api_key client option must be set either by passing
api_key to the client or by setting the OPENAI_API_KEY environment variable

2. Fixed but scoped to the CLI rather than reviving the bootstrap check. build_runtime (shared by hexgate chat/serve) is the one spot that unconditionally wires web_search/fetch, so it now prints a clear stderr warning at startup for any of LINKUP_API_KEY/TAVILY_API_KEY that's unset, naming the exact key and tool. That restores upfront discoverability without the mid-conversation surprise.

3. description mismatch, updated now

4 and 5: Seem like quite minor issues, I am fine with current version of the branch

victorludvig (approved):

PR #61, opened 2026-07-01
fix(dashboard): render inline-roles policies on the overview Graph

quanghexa94:
## What broke

The overview Graph page rendered every edge under `default_policy.mode` (usually deny → all red) OR missed tools entirely, on every seeded multi-role agent. Root cause: `parsePolicy()` only walked the flat `tools:` shape and never traversed `roles.<name>.tools`. `buildAgentView` compounded it by reading tools off `agent.yaml`, missing anything only declared in a role. `AgentView.missingInPolicy` was also poisoned — falsely flagging every tool as unpoliced.

The **wasm bundle at runtime is correct** — it resolves inheritance + per-role tool maps. This is a **dashboard-visualization bug only**; no runtime policy enforcement is affected.

## Fix

`parsePolicy` now handles both shapes:
- **Flat** (`tools:` at top level) — unchanged behaviour
- **Inline-roles** (`roles.<name>.tools`) — merges every concrete role's tool map, keeping the **worst-case** mode when a tool differs across roles (`deny > approval_required > allow`)

Mixins (`is_mixin: true`) are excluded from the merge — they compose INTO concrete roles via `inherits:`, and treating them first-class would double-count decisions. `ParsedPolicy.roles: Record<role, tools>` is also exposed for a future "filter by role" affordance.

`buildAgentView` tool list = union of `agent.yaml` tools + every tool in any role's policy. Without this, a role-only rule (`billing.refund_order`) was invisible on the graph.

Also **removed dead code**: `buildAgentGraph` (~40 LOC, zero callers, leftover from an earlier per-agent detail view).

## Known limitation, documented in-test

Mixin-only tools (`web_search` inherited via `read_only`) still fall through to `default_policy=deny` because we don't resolve `inherits:` client-side. The wasm bundle handles it correctly; only the dashboard's cosmetic edge color is wrong for this case. Pinned by `test_falls_back_to_default_policy_for_mixin_only_tools`; a future inheritance-aware resolver would flip that test's expectation. Not blocking for this fix — it's a strict improvement over "everything renders as deny."

## Tests

- **11 new regression cases** in `platform/dashboard/src/lib/policy.test.ts`
- Flat + inline-roles parsing, worst-case merge, mixin filter (parseRolesFromPolicy behavior pinned unchanged), tool-list union, effectiveMode routing, missingInPolicy fallback
- **67/67 dashboard tests pass** (+11 new)
- `pnpm typecheck` clean, `pnpm lint` clean on touched files (13 pre-existing warnings on other files, unrelated)

## Test plan

- [x] Unit tests cover both YAML shapes
- [x] `pnpm typecheck` clean
- [ ] After merge: eyeball `/graph` on a fresh env with the seeded `support_bot` — refund_order edge should render deny (was: missing, or same edge painted default green)

### DIFF
--- a/platform/dashboard/src/lib/graph.ts
+++ b/platform/dashboard/src/lib/graph.ts
@@ -4,6 +4,7 @@ import {
   buildAgentView,
   effectiveMode,
   MODE_COLOR,
+  worstMode,
   type AgentView,
   type Mode,
 } from "./policy";
@@ -90,11 +91,14 @@ export function buildOverviewGraph(agents: AgentRead[]): OverviewGraph {
   const toolCenterY = ((toolList.length - 1) * ROW_H) / 2;
   const shift = agentCenterY - toolCenterY;
   toolList.forEach((toolName, i) => {
-    // Determine the "worst" mode for the left strip on the tool node
+    // Determine the "worst" mode for the left strip on the tool node.
+    // Shares MODE_STRENGTH ordering with parsePolicy's cross-role merge
+    // via the exported worstMode helper — one source of truth for
+    // "which mode wins when they disagree."
     const modesForTool: Mode[] = agentViews
       .filter((v) => v.tools.includes(toolName))
       .map((v) => effectiveMode(v, toolName));
-    const mode = pickStripMode(modesForTool);
+    const mode = worstMode(modesForTool) ?? "default";
     nodes.push({
       id: `tool:${toolName}`,
       type: "tool",
@@ -116,11 +120,4 @@ function edgeStyle(mode: Mode): React.CSSProperties {
     return { ...base, strokeWidth: 1.5, opacity: 0.85 };
   }
   return base;
-}
-
-function pickStripMode(modes: Mode[]): Mode | "default" {
-  if (modes.includes("deny")) return "deny";
-  if (modes.includes("approval_required")) return "approval_required";
-  if (modes.includes("allow")) return "allow";
-  return "default";
 }
--- a/platform/dashboard/src/lib/graph.ts
+++ b/platform/dashboard/src/lib/graph.ts
@@ -107,49 +107,6 @@ export function buildOverviewGraph(agents: AgentRead[]): OverviewGraph {
   return { nodes, edges, agentViews };
 }
 
-/** Build a per-agent local graph: the agent + its tools only. */
-export function buildAgentGraph(agent: AgentRead): {
-  nodes: Node[];
-  edges: Edge[];
-  view: AgentView | null;
-} {
-  const view = buildAgentView(agent);
-  if (!view) return { nodes: [], edges: [], view: null };
-
-  const nodes: Node[] = [];
-  const edges: Edge[] = [];
-
-  const centerY = ((view.tools.length - 1) * ROW_H) / 2;
-
-  nodes.push({
-    id: "agent:self",
-    type: "agent",
-    position: { x: 40, y: centerY },
-    data: { name: view.name, model: view.model, toolCount: view.tools.length },
-    draggable: false,
-  });
-
-  view.tools.forEach((toolName, i) => {
-    const mode = effectiveMode(view, toolName);
-    nodes.push({
-      id: `tool:${toolName}`,
-      type: "tool",
-      position: { x: 380, y: i * ROW_H },
-      data: { name: toolName, mode },
-      draggable: false,
-    });
-    edges.push({
-      id: `e:agent->${toolName}`,
-      source: "agent:self",
-      target: `tool:${toolName}`,
-      style: edgeStyle(mode),
-      animated: mode === "approval_required",
-    });
-  });
-
-  return { nodes, edges, view };
-}
-
 function edgeStyle(mode: Mode): React.CSSProperties {
   const base = { stroke: MODE_COLOR[mode], strokeWidth: 1.75 };
   if (mode === "approval_required") {
--- a/platform/dashboard/src/lib/policy.test.ts
+++ b/platform/dashboard/src/lib/policy.test.ts
@@ -1,0 +1,209 @@
+import { describe, expect, it } from "vitest";
+import {
+  buildAgentView,
+  effectiveMode,
+  parsePolicy,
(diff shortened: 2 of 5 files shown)


quanghexa94:
All 10 code-review findings addressed — pushed as `c14aec7`.

**Status:** all correctness issues resolved. 90/90 dashboard tests pass; typecheck + fmt-check + lint clean on touched files. Ready for another look.

### What changed

**Correctness fixes:**

- **#1 fixed** — `ParsedPolicy.roles` now agrees with `parseRolesFromPolicy` on which roles exist (empty-tools roles kept as `{}` entries).
- **#2 fixed** — flat top-level `tools:` is authoritative; roles no longer silently upgrade `allow → deny` on a flat entry. New `mergedTools(policy)` helper for callers that want the cross-role union.
- **#3 fixed** — `AgentView.tools` = `agent.yaml` verbatim, declared order. Confirmed runtime tool-gating comes from agent.yaml (not policy); role-only entries no longer create phantom capability edges.
- **#7 fixed** — new `isMixinSpec()` predicate coerces truthy variants (`is_mixin: "true" / 1 / "yes"` all counted); reused in all three call sites.
- **#10 fixed** — no more alphabetical sort of `AgentView.tools`; declaration order preserved.

**One-source-of-truth extractions (#4, #6, #8, #9):**

- `MODE_STRENGTH` + `worstMode()` exported → `graph.ts`'s `pickStripMode` deleted; one ordering.
- `readToolMap` exported → `policy_graph.ts` routes through it; capital `Allow` gets dropped consistently in both graph views.
- `isMixinSpec` reused across `parsePolicy` + `parseRolesFromPolicy` + `policy_graph.ts`.

**Documented + punted (#5):** mixin-only tools still fall to `default_policy` because we don't resolve `inherits:` client-side (wasm bundle does at runtime). Test explicitly pins the current behavior — a future ~50 LOC topological-walk PR flips the assertion.

### Test coverage

29 policy cases (was 23). Regressions pinned for #1, #2, #3, #7, #10.

Codecov patch coverage was also boosted above the threshold in an earlier commit (`f2d135e`).

guillaume-hexamind (commented):
Logic seems OK; 

below 4 claude issues found:

① inherits: is never resolved — inherited-mixin tools render as denied — policy.ts:248 (also :164, :176)  → 90/100
parsePolicy drops mixins (if (isMixinSpec(spec)) continue;) and nothing resolves inherits:. For the exact policies this PR targets (e.g. seeded support_bot whose web_search is allow in the read_only mixin that concrete roles inherit), the overview Graph draws web_search red/deny even though the agent is actually allowed to call it. This is the opposite of the fix's intent. policy.test.ts:213 even pins this as deny, acknowledging it as a known limitation — but it's still user-visible wrong output. Verified CONFIRMED by 3 independent verifiers.

② Per-role default_policy is declared but never applied — policy.ts:163 (also :246)  → 78/100
RoleSpec declares default_policy, but parsePolicy only reads each role's tools map; the per-role default is never parsed. A role that raises its own baseline (default_policy: {mode: allow}) but doesn't list a tool still shows it as the global default (deny). Graph contradicts runtime for any role that overrides its own default.

③ Flat top-level tools: overrides a stronger role mode — policy.ts:106  → 75/100
mergedTools rule #1 lets a flat top-level tools: entry win over a role's stronger mode. A policy with a top-level allow and a role deny on the same tool renders allow-green on the overview edge/strip — the user believes the tool is universally allowed and misses that a role blocks it.

④ effectiveMode worst-cases across ALL roles — mis-colors per-agent edge — policy.ts:222  → 72/100
effectiveMode returns the worst-case mode over every role (deny > approval > allow). In INLINE_ROLES_POLICY, refund_order is deny in one role and allow in billing; the overview edge is drawn red for every role including billing. A user concludes billing cannot refund orders when it can.

⑤ Tools with missing/invalid mode are dropped instead of shown as deny — policy_graph.ts:196 (also :191, :101)  → 68/100
Mode edges now iterate the readToolMap-filtered map, which skips entries whose mode isn't a valid Mode. On main a mistyped/omitted mode (e.g. 'Deny' capitalized) rendered a deny-colored edge; now the edge disappears. The user sees a tool node with no incoming edge and may assume it's unconstrained rather than mistyped.

quanghexa94:
All 5 addressed. `719817a` on the branch.

- **[90/100] `inherits:` never resolved** — fixed. `parsePolicy` now runs a two-pass resolver: collect all role specs (mixins included as ancestors), walk each concrete role's inheritance chain with cycle guard, own tools override inherited. web_search declared only in the read_only mixin now renders allow on the overview for every role that inherits it, matching the wasm engine. The `known limitation` test at policy.test.ts:213 is flipped and pins the new correct behavior.
- **[78/100] per-role `default_policy` never applied** — fixed. `ParsedPolicy.roles[<name>]` now carries the resolved `default_policy` (inherited from ancestors when absent), and `effectiveMode` counts each role's default as a voice for unlisted tools. A role that raises its baseline to allow no longer collapses to the global deny.
- **[75/100] flat top-level tools overrides a stronger role mode** — fixed. `mergedTools` no longer treats the flat block as authoritative. Every voice (flat + each role) contributes to a worst-case merge, so a role deny stays visible even when the flat block says allow. Matches runtime: at call time the role IS what gates the call.
- **[72/100] effectiveMode worst-cases per-agent edge** — worst-case semantics stayed (the docstring calls it "worst case any caller could hit," and the runtime matches: if any role denies, that role denies), but the wider-fix landing above means the merge is now honest about which voice is contributing. Callers wanting per-role detail already have the Graph tab (`policy_graph.ts`) where each edge encodes one role.
- **[68/100] tools with invalid mode disappear from the graph** — fixed. `policy_graph.ts` now iterates raw tool names (not the readToolMap-filtered view) and renders invalid-mode entries with an `invalid mode · deny` label plus fail-closed deny styling. Users see the typo instead of a missing edge. New `policy_graph.test.ts` pins both the missing-edge regression and the mixed valid/invalid case.

Total: **+8 new tests, all 98 dashboard tests pass**, typecheck + lint + fmt clean.

guillaume-hexamind (approved):

PR #62, opened 2026-07-01
feat(dashboard): persist agent selection via ?agent= + Settings shell

quanghexa94:
## What

Two small UX paper-cuts from the design brief on PR #61's review pass.

### 1. Picker selection now survives navigation between Agents ↔ Policies

Before: click **edit policy →** on `/agents` from `support_bot`, land on `/policies` with the first agent auto-picked instead of `support_bot`. Same trip in reverse. Each route held its own local `useState` + first-agent auto-pick effect.

After: `?agent=` in the URL is the source of truth. New shared `useAgentParam(knownNames)` hook wraps `useSearchParams` with three properties:

- **Stale-URL guard:** exposes `null` when `?agent=` doesn't match a known name, so a URL to a deleted agent doesn't leave the picker in a broken "selected but not in the dropdown" state.
- **No back-button pollution:** uses `{ replace: true }` — picker clicks don't create a history entry per selection.
- **Preserves other query params:** future `?tab=graph` or `?role=admin` won't be stomped.

Companion `useAutoSelectFirstAgent(agent, knownNames)` preserves the current "land on the first agent when nothing is selected" behavior — extracted so every consumer isn't re-writing the same effect. Policies + Agents wired through this pair.

The `edit policy →` link on Agents.tsx now forwards `?agent=<name>` — the actual UX win this PR is built around.

### 2. Settings shell — no more blank page

`/settings` was a title + one sentence. Now six section cards covering the eventual scope, each with a "Coming soon" badge:

| Section | Eventual scope |
|---|---|
| General | Project name, delete project (confirm-typing) |
| Bundle & runtime | Default model, framework, bundle pinning, decision timeout, approval TTL |
| Webhooks | `decision.denied` / `decision.needs_approval` / `agent.registered` URLs + signing secret + delivery log |
| Retention | Audit retention window, PII scrubbing on `arguments` |
| Environment variables | Key/value pairs injected into the policy eval context |
| Danger zone | Rotate signing key · Revoke all tokens · Purge audit history |

Danger-zone card gets a red border + destructive-tinted icon so the visual language is already right when real controls land. When each section ships, replace its `Coming soon` badge with the form — the card + description doubles as inline documentation next to the controls.

## Deferred to future PRs (from the same design brief)

- **Graph health strip** — top chips: *N agents · M unregistered · K deny-only · P approval-required*, chip-click filters nodes
- **Live decision stream on Audit** — WebSocket toggle, sliding rows with color pulse
- **Real Tokens page** — remove useless copy-masked-value button, ship Rotate + stale-token warning
- **Policy Diff Preview signature moment** — on Save, replay the last N=500 audit events against the draft, show what decisions would change

## Tests

- **10 new cases in `agent_param.test.tsx`** — URL param round-tripping, stale-guard, no-history-pollution, other-param preservation, auto-select-first once + never overrides
- **71/71 dashboard tests pass**
- `pnpm typecheck` + `pnpm format:check` clean
- `pnpm lint` clean on touched files (9 pre-existing warnings on unchanged code)

## Test plan

- [x] Automated: agent selection round-trips through the URL
- [x] Automated: `edit policy →` link carries the ?agent= param
- [ ] After merge: eyeball — pick `support_bot` on /agents, click edit policy →, land on /policies with support_bot preselected
- [ ] After merge: eyeball /settings on a fresh env — six section cards, danger zone in red

### DIFF
--- a/platform/dashboard/src/lib/agent_param.test.tsx
+++ b/platform/dashboard/src/lib/agent_param.test.tsx
@@ -1,16 +1,20 @@
 /**
- * Tests for the ``useAgentParam`` + ``useAutoSelectFirstAgent`` pair
- * that back ``?agent=`` selection sync across /agents and /policies.
+ * Tests for ``useAgentSelection`` — the single hook that backs
+ * ?agent= sync across /agents and /policies.
  *
- * The invariants:
- *   1. On mount, ``selected`` reflects the ?agent= param IF it matches
- *      a known agent (otherwise null — stale URL stays inert).
- *   2. ``set(name)`` writes ?agent= via ``replace`` (no back-button
- *      trail for picker changes).
- *   3. ``clear()`` removes ?agent= entirely (project-switch reset).
- *   4. When ``selected`` is null and knownNames is non-empty,
- *      ``useAutoSelectFirstAgent`` runs ``set(names[0])`` exactly once
- *      per (names identity) change — no loops.
+ * The invariants this file pins (many learned in review after the
+ * first draft broke every one of them):
+ *
+ *   1. ?agent= from the URL is preserved on first render, even while
+ *      the agents query is still loading (knownNames=undefined).
+ *   2. When the URL points at a stale/renamed agent, ``selected``
+ *      falls back to the first known name — but the URL is NEVER
+ *      auto-rewritten (bookmarks stay shareable, ``fromUrl`` flags
+ *      the divergence so callers can surface a banner).
+ *   3. ``set()`` writes ?agent= with replace, no back-button pollution.
+ *   4. ``resetOn`` only clears on real transitions — a stable value
+ *      across renders does NOT wipe an incoming URL param at mount.
+ *   5. ``resetOn`` = undefined opts out of reset entirely.
  */
 
 import { act, renderHook } from "@testing-library/react";
@@ -18,7 +22,7 @@ import type { JSX, ReactNode } from "react";
 import { MemoryRouter, useLocation } from "react-router-dom";
 import { describe, expect, it } from "vitest";
 
-import { useAgentParam, useAutoSelectFirstAgent } from "./agent_param";
+import { useAgentSelection } from "./agent_param";
 
 function wrapperAt(initialUrl: string) {
   return ({ children }: { children: ReactNode }): JSX.Element => (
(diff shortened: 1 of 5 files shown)


quanghexa94:
All 10 code-review findings addressed — pushed as `57e829d`.

**Status:** the three critical URL-sync bugs that broke the PR's primary feature are fixed and pinned by regression tests. 73/73 dashboard tests pass; typecheck + fmt-check + lint clean on touched files. Ready for another look.

### What changed

**Rewrote the two-hook API into one `useAgentSelection`** — one call replaces `useAgentParam` + `useAutoSelectFirstAgent` + the duplicated `useMemo(names)` + duplicated `clear()` effect + two `eslint-disable`s. Semantics are stricter and safer:

```ts
const { selected, set, fromUrl } = useAgentSelection(agents.data, {
  resetOn: scope.projectId,
});
```

This fixes:

- **#1 (mount clear wiping ?agent=)** — `resetOn` is ref-tracked: it snapshots the initial value and only fires on real transitions, so mount doesn't wipe the incoming URL param.
- **#2 (loading-race dropped URL value)** — during the loading window (`knownNames === undefined`) the hook trusts the URL rather than treating a missing list as "no match."
- **#3 (auto-select rewrites URL)** — stale URL now stays visible. `selected` falls back to the first agent for rendering, but the URL is never mutated by the fallback. `fromUrl` flags the divergence so a future banner can surface "this agent doesn't exist in this project."

- **#4 (Settings gated behind no-project)** — dropped the gate. Org-owners in fresh orgs (zero projects) now see the roadmap cards + a subtle "select a project to configure" banner. Also covers **#7** (loading-state flash).

- **#6, #8, #9, #10** — subsumed by the single-hook refactor: the over-firing dep list is gone, callers no longer paste boilerplate, no `eslint-disable`s.

### Test coverage

12 cases (was 10). Regressions pinned:
- URL preserved on first render while loading (#2)
- Stale-fallback does NOT rewrite the URL (#3)
- Mount does NOT clear `resetOn` (#1)
- Real `resetOn` transition DOES clear
- `resetOn=undefined` opts out entirely
- `fromUrl` flag correctness

**Not fixed here (#5):** manifest-name vs listAgents-name drift is a real risk but requires backend coordination — filed as follow-up.

guillaume-hexamind (commented):
Logic looks good. Below are three potential issues found by claude:

[88/100] agent_param.ts:124 reset-on-load wipes the incoming ?agent= on every cold load CONFIRMED
resetOn = scope.projectId transitions null â realId during the normal orgs-hydration window (useProjectScoped returns null while useOrgs() loads, even with a persisted project). The ref-guarded reset effect sees null !== realId, treats page load as a "project switch," and strips ?agent=. This defeats the PR's headline feature on every fresh load / refresh / shared link â not an edge case. Tests pass a stable resetOn from mount and never simulate the hydration transition, so they miss it. This is the one to fix first.

[82/100] agent_param.ts:111 loading-window branch fires a cross-project fetch on project switch CONFIRMED
On a project switch, react-query (no keepPreviousData, staleTime 5000) drops agents.data to undefined â namesLoaded=false, so the still-present old ?agent=X passes straight through as selected. Policies renders <YamlEditor agentName={X} projectId={newProject}> and Agents does a cross-project find, firing an API request for an agent that doesn't exist in the new project â 404/wrong-agent flash before the reset clears the param next tick.

[78/100] Policies.tsx:52 fromUrl is computed but never consumed â silent misroute to wrong agent CONFIRMED
The old auto-select + clear-on-mismatch invariant was replaced with a fromUrl=false fallback, but no caller reads fromUrl. A shared /policies?agent=support_bot link after a rename lands the user on the first agent's policy editor with no banner â while the URL still says support_bot. The user can Save changes to the wrong agent. The whole design justifies itself with a banner that was never wired up.

quanghexa94:
All 3 addressed. `c641c60` on the branch.

- **[88/100] hydration transition strips `?agent=`** — fixed. The reset effect now only fires when both endpoints are known project ids; null → realId hydration and realId → null teardown leave the URL alone. New test `does NOT clear the URL when resetOn hydrates from null to a real id` simulates the useProjectScoped hydration transition the original tests missed.
- **[82/100] loading-window race leaks stale `?agent=` on project switch** — fixed. Render-time detection of a switch-in-flight (`isSwitchingProjects`) makes the hook return the new project's first agent (or null) instead of trusting the stale URL during the react-query loading window. New test `does NOT leak the old project's ?agent= into the new project on a switch`.
- **[78/100] fromUrl computed but never consumed** — fixed. Added `requested` and `notFound` to the AgentSelection API and wired an inline banner in `Policies.tsx`: "No agent named X in this project. Showing Y instead." 5 tests pin the new API.

Total: **20/20 dashboard tests pass** on this file (was 13), full suite 81/81, typecheck + lint + fmt clean.

guillaume-hexamind (approved):

PR #63, opened 2026-07-02
feat(playground): route approval_required tool calls to the connected tab

quanghexa94:
> **Walkthrough deck:** [claude.ai/code/artifact/ba44bb68](https://claude.ai/code/artifact/ba44bb68-92d4-4b9e-8fe0-a10753df2d87). Seven slides covering the primitive, the WS wire flow, adapter coverage, framework comparison, fail-closed guarantees, and quickstart. Skimmable if you don't want to read the whole PR.

## What this fixes

Before this PR, `hexgate serve` silently auto-approved every tool call the policy flagged `approval_required`. The playground never saw those prompts — the user's mode setting made no difference and the whole approval feature was a no-op through the local dev loop.

Now: the SDK asks, the playground pops a prompt inline above the composer, and the user's answer comes back down the same WebSocket the chat is already flowing over. If nobody's connected, if the tab closes mid-prompt, or if the user takes too long — the tool call is denied.

## How it works

```
Playground UI          Platform relay          hexgate serve             SDK enforcer
     │                       │                       │                        │
     │─── chat ─────────────▶│──── relay ──────────▶│─▶ stream_agent ─▶ GuardedTool
     │                       │                       │                    NEEDS_APPROVAL
     │                       │                       │                        │
     │◀── approval.request ──│◀─── forward ─────────│◀─ RelayApprovalHandler(decision)
     │                       │                       │      mint decision_id
     │                       │                       │      await asyncio.Event
User clicks Approve          │                       │
     │                       │                       │
     │─── approval.reply ──▶│──── forward ────────▶│─▶ _handle_message
     │  {decision_id,        │                       │      handler.resolve()
     │   allowed:true}       │                       │      Event.set()
     │                       │                       │                    ▼ tool runs
     │◀── tool.end / delta ─│◀─── stream ──────────│◀───────────────────────┘
```

The relay is untouched — it's still a pure JSON forwarder. All the new behavior lives on the SDK and dashboard sides.

## Design choices worth calling out

- **Simple callback, not serialize-and-resume.** The primitive is `async fn(decision) -> bool`. That's the same shape a CLI dev would use with `input()`, so implementing your own approval sink is a five-line job. No `RunState`, no checkpointer, no cross-process durability at v1 — same tradeoff Claude Code makes.
- **Fail-closed everywhere.** TTL expiry, WebSocket disconnect, no bound socket, unknown decision_id, non-bool `allowed`, send failure — all deny.
- **Session-scoped.** No Postgres, no REST endpoints. If `hexgate serve` restarts mid-approval, the prompt is lost. Durability lands later, when webhook / Slack / mobile approvers become a real need.
- **Inline UI, not a modal.** The prompt renders above the composer so it stays in the flow of the conversation, and multiple parallel tool calls stack instead of fighting over the modal slot.
- **Default flipped from auto-approve to ask.** Headless callers (CI, cron) need to opt back in with `--approval-mode auto-approve`. A loud startup line names the flag.

## Scope

All four framework adapters now carry the `approval_handler` callback: LangChain (via `HexgateAgent` / `hexgate serve`), OpenAI Agents (`wrap_openai_agent` + `HexgateRunner`), Pydantic AI (`wrap_pydantic_agent`), and Google ADK (`wrap_google_agent` + `HexgateRunner`). Same behavior on all four: truthy return runs the tool, falsy return or missing handler renders the existing `[approval_required]` marker to the model (string on OpenAI / Google, `ModelRetry` on Pydantic AI).

The playground path itself (which motivated the PR) stays LangChain-only because `hexgate serve` builds a `HexgateAgent`. What the other adapters unlock is: a developer using them directly can now pass their own callback and have it actually fire.

Explicit non-goals for this PR: durable approvals (Postgres), webhook / Slack integration, per-tool RBAC on approvers, public `TerminalApprovalHandler` helper for CLI devs. All tracked as separate work; the callback slot on each adapter's guarded tool is the extension point.

## Documentation

Two new mintlify pages under `docs/concepts/`, wired into the Concepts nav:

- **`approval-required.mdx`** — quickstart, `Decision` field reference, four common handler shapes (CLI, auto-approve for tests, playground, webhook / Slack), points of vigilance.
- **`approval-frameworks-compared.mdx`** — honest comparison against OpenAI Agents SDK, LangGraph, Pydantic AI, and Claude Code. Names where each framework wins and where our callback trades durability for zero infrastructure.

## Tests

- `tests/cli/test_serve_approval.py` — 17 cases covering the playground WS relay: payload shape, resolve unblocks, TTL denies, disconnect fails all pending, concurrent distinct decision_ids, `asyncio.Lock` serializes concurrent sends, unknown decision_id no-op, `_handle_message` routing (happy + string-allowed + missing-id + non-relay-handler), send-failure denies, `_dispatch_message` error-echo, end-to-end `_apply_approval_handler` plumbing, deadlock regression on `_serve_loop`.
- `tests/adapters/openai/test_tools.py`, `tests/adapters/pydantic_ai/test_tools.py`, `tests/adapters/google/test_tools.py` — 5 cases each on the new callback: async handler true, async handler false, bool shortcut, handler ignored on plain deny, sync handler acceptance.
- `platform/dashboard/src/lib/playground.test.tsx` — 10 cases: receive renders, dedupe, respond sends + optimistic remove on success, resolve one preserves others, unknown decision_id no-op, TTL sweep, WS close clears list, reset sends deny-replies, respond fails when closed.
- **SDK: 954 pass. Dashboard: 71 pass.** Lint, fmt-check, typecheck clean.

## Test plan

- [x] Automated regression suites above
- [ ] After merge: `make demo-notebook` → send a prompt that triggers a role with `approval_required` → confirm the prompt appears in the browser and clicking approve/deny lands the right outcome
- [ ] After merge: fire 5+ parallel tool calls in one turn → confirm each prompt has its own decision_id and resolves independently

### DIFF
--- a/deploy/serve_manager.py
+++ b/deploy/serve_manager.py
@@ -70,26 +70,31 @@ def _demo_settings():
 
 def _run(agent_obj) -> None:
     global _loop, _task, _status
-    from hexgate.cli._common import (
-        build_approval_handler,
-        build_runtime_from_local_agent,
-    )
-    from hexgate.cli.serve import run_serve
+    from hexgate.cli._common import build_runtime_from_local_agent
+    from hexgate.cli.serve import RelayApprovalHandler, run_serve
     from rich.console import Console
 
     _loop = asyncio.new_event_loop()
     asyncio.set_event_loop(_loop)
     try:
         settings = _demo_settings()
+        # Same default as `hexgate serve` (see hexgate/cli/serve.py): route
+        # NEEDS_APPROVAL to the connected playground tab via the WS relay.
+        # Without this the notebook BYOK container silently auto-approved
+        # every approval_required tool — the demo would ship the wrong
+        # security contract.
+        approval_handler = RelayApprovalHandler()
         runtime = build_runtime_from_local_agent(
             settings,
             agent_obj=agent_obj,
             description=None,
-            approval_handler=build_approval_handler(Console(), "auto-approve"),
+            approval_handler=approval_handler,
             auto_register=True,  # idempotent register of the agent manifest
             console=Console(),
         )
-        _task = _loop.create_task(run_serve(runtime))
+        _task = _loop.create_task(
+            run_serve(runtime, approval_handler=approval_handler)
+        )
         _status = "running"
         _loop.run_until_complete(_task)
     except asyncio.CancelledError:
--- a/docs/concepts/approval-frameworks-compared.mdx
+++ b/docs/concepts/approval-frameworks-compared.mdx
@@ -69,4 +69,3 @@ The other thing worth calling out is that approval is one of three policy outcom
 - `SlackApprovalHandler`: post an approval request to a channel and treat a reaction as the answer.
 - Per-tool RBAC on approvers, so a policy can require a specific role rather than "any human".
 - Auto-approve rules for repeat identical calls, with an audit note. Opt-in, since some regulated tenants explicitly do not want this.
-- Callback plumbing on the OpenAI Agents, Pydantic AI, and Google ADK adapters.
--- a/docs/concepts/approval-frameworks-compared.mdx
+++ b/docs/concepts/approval-frameworks-compared.mdx
@@ -1,74 +1,72 @@
 ---
-title: "Approval — how other frameworks do it"
-description: "An honest comparison of the approval primitive across OpenAI Agents SDK, LangGraph, Pydantic AI, and Claude Code."
+title: "Approval in other frameworks"
+description: "How OpenAI Agents SDK, LangGraph, Pydantic AI, and Claude Code approach human-in-the-loop, and where Hexgate fits."
 ---
 
-Every serious agent framework has an approval story. They differ on what they optimize for — durability, framework-native ergonomics, protocol coupling — and each design has a cost. This page names the tradeoffs we accepted and the ones we didn't.
+Most serious agent frameworks now ship some form of human-in-the-loop approval. They approach it differently depending on what they own (the whole runtime, a graph, just the tool interface). This page walks through each one so you can pick the right primitive for your setup, or understand what Hexgate is trading off if you're already using one of them.
 
-## TL;DR
+## At a glance
 
-| Framework | Primitive | Durable across process restart | State a dev must configure | Concepts to learn |
+| Framework | Primitive | Durable across a restart | State you configure | Concepts you learn |
 |---|---|---|---|---|
-| **Hexgate** | `async fn(decision) -> bool` callback | No (v1) | None — memory-scoped | 1 (the callback) |
-| **OpenAI Agents SDK** | `needs_approval=True` + `RunState.approve()` after `InterruptionEvent` | Yes (serialize `RunState`) | Pick a serialization sink | 4 (needs_approval flag, InterruptionEvent, RunState, approve/reject) |
-| **LangGraph** | `interrupt()` + `Command(resume=...)` | Yes (via checkpointer) | Configure checkpointer (in-memory / SQLite / Postgres) | 4 (interrupt, thread_id, checkpointer, resume) |
-| **Pydantic AI** | Raise `ModelRetry` from the tool | No | None | 2 (ModelRetry exception, retry handling) |
-| **Claude Code** | Terminal prompt in the loop | No | None | 0 (implicit) |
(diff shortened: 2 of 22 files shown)


victorludvig:
Hello @quanghexa94, I approve the logic. 

Can you check these potential issues found by Claude:

1. hexgate/cli/serve.py:272 — Only the approval-request send is protected by RelayApprovalHandler._send_lock; the chat-stream send (line 272), the reset ack (line 277), and _dispatch_message's error echo (line 408) are not. Since every inbound frame now runs as its own concurrently-dispatched task (line 369), two of these can call ws.send() at the same time. The PR's own docstring states the websockets library "does NOT serialize concurrent send() calls on one socket — interleaved frames would be a silent bug otherwise" — that's exactly the risk left open on every send path except the one it was written for.

2. hexgate/cli/serve.py:369 (interacting with hexgate/cli/state.py's ChatState) — The new per-frame task dispatch means two "chat" frames (or a "chat" + "reset") can now run concurrently against the same unlocked ChatState. start_turn() unconditionally overwrites self.current_run, so if a second chat frame arrives while the first is still streaming, the first task's subsequent apply_event() calls mutate the wrong turn. Nothing on the client prevents this: Playground.tsx's composer submit() only checks !composer.trim() — there's no busy/streaming guard, so a user sending a message before the previous reply finishes triggers this directly.

3. hexgate/cli/serve.py:109-117 (unbind_socket) racing line 196 (resolve) — Both mutate the same box dict for a pending decision without any lock, and the entry is only removed from _pending by the awaiting __call__ coroutine's own finally (line 176), not by resolve(). If a WS disconnect fires between resolve() setting box["allowed"] = True and the parked __call__ coroutine waking up to read it, unbind_socket() finds the (still-present) entry and overwrites it back to False — a legitimately approved tool call gets silently flipped to denied.

4. platform/dashboard/src/lib/playground.ts:429 (respondToApproval) — Treats a successful ws.send() as the end of the story and immediately clears the local pending-approval card; there's no server ack. If the server-side TTL (asyncio.wait_for in serve.py) has already fired — e.g. due to clock drift or network delay — the reply lands on an already-popped decision_id and silently no-ops server-side (resolve() just logs at debug), while the UI still shows the approval as successfully sent with no error.

5. hexgate/adapters/openai/tools.py:38 (also google/tools.py:31, pydantic_ai/tools.py:28) — The bool-shortcut/isawaitable-coercion helper _resolve_approval is copy-pasted verbatim into all three adapters, while hexgate/adapters/langchain/tools.py already implements the same contract more completely (it guards against a sync caller receiving a coroutine, which the three copies don't). This should collapse into one shared helper — the drift between the langchain version and the three copies is itself evidence duplication already caused inconsistency.

quanghexa94:
Thanks @victorludvig — 4 of 5 fixed in `ddb4791`, 1 deferred as follow-up. Full SDK suite green (958 passed).

**Fixed:**

1. **Unlocked ws.send calls (serve.py:272, :277, :408)** — every send on the serve socket now routes through a new `_safe_send(context, ws, msg)` helper that acquires a per-connection `context.send_lock`. `RelayApprovalHandler.bind_socket` takes the same lock so approval-request sends serialize with every other send. One lock across all sends is the invariant.

2. **ChatState race (serve.py:369)** — added per-connection `chat_lock` on ServeContext. The chat branch of `_handle_message` acquires it, so two back-to-back chat frames serialize. Approval replies and resets still run concurrently with streaming (different branches), so the deadlock fix isn't undone.

3. **unbind_socket vs resolve race (serve.py:109/196)** — `resolve()` now pops the entry from `_pending` atomically before mutating `box`. `unbind_socket` can't find a resolved entry to overwrite. The `__call__` coroutine's own `finally` `pop(..., None)` no-ops on the second remove.

5. **Duplicated `_resolve_approval` (openai/pydantic_ai/google adapter)** — extracted to `hexgate/agents/approvals.py` (`resolve_approval_sync`, `resolve_approval_async`). All four adapters (including LangChain) now import from there. The sync-caller-receiving-coroutine guard the three copies were missing is now available to all of them via the shared helper.

4 new regression tests pin each fix.

**Deferred: #4 (respondToApproval optimistic remove-on-send-success).** Fixing right needs a wire-protocol change — either a full `approval.ack` frame or a targeted `approval.stale` frame for the TTL-fired edge case. Tracked as a follow-up PR to keep this one scoped.

victorludvig (approved):

PR #65, opened 2026-07-03
feat(mcp): fan out MCP tool wrapping to every adapter

quanghexa94:
## What this ships

Follow-up to the merged MCP proxy PR (#57). Before this PR, `MCPToolset` produced `langchain_core.tools.BaseTool` and only the LangChain adapter path could reach MCP servers. Now every framework adapter (LangChain, OpenAI Agents, Pydantic AI, Google ADK) has a matching `wrap_mcp_toolset(toolset)` that turns MCP tools into framework-native tool objects, and the existing per-adapter `enforce_policy` gate covers them for free.

Design captured in `_design/MCP_ADAPTERS.md` (kept local, not in this PR).

## Shape

```
hexgate/mcp/
├── client.py                 # unchanged (MCP protocol client)
├── config.py                 # unchanged (MCPServerConfig)
└── proxy.py                  # split: MCPToolProxy dataclass now framework-agnostic

hexgate/adapters/
├── langchain/mcp.py          # wrap_mcp_toolset(toolset) -> list[BaseTool]
├── openai/mcp.py             # wrap_mcp_toolset(toolset) -> list[FunctionTool]
├── pydantic_ai/mcp.py        # wrap_mcp_toolset(toolset) -> list[Tool]
└── google/mcp.py             # wrap_mcp_toolset(toolset) -> list[BaseTool]
```

`MCPToolProxy(qualified_name, description, input_schema, call)` is the canonical descriptor. Each adapter reads only those four fields; nothing about the underlying `MCPClient` or `_ToolsetState` leaks out. The `{"ok": True|False, ...}` envelope is preserved across every path so the LLM sees one shape regardless of framework.

## Usage (identical across adapters)

```python
from hexgate.mcp import MCPServerConfig, MCPToolset

# Pick the adapter that matches your agent:
from hexgate.adapters.openai.mcp import wrap_mcp_toolset      # OpenAI Agents
# from hexgate.adapters.pydantic_ai.mcp import wrap_mcp_toolset  # Pydantic AI
# from hexgate.adapters.google.mcp import wrap_mcp_toolset        # Google ADK
# from hexgate.adapters.langchain.mcp import wrap_mcp_toolset     # LangChain

async with MCPToolset(slack_cfg) as mcp:
    tools = wrap_mcp_toolset(mcp)
    # feed tools into whichever agent you're building, then enforce_policy
```

`mcp.tools` still returns LangChain tools directly as a back-compat shortcut for pre-adapter-split code.

## Schema question — resolved without workarounds

Design doc flagged that Pydantic AI and Google ADK might need dynamic Pydantic-model generation from JSON Schema. Two inline spikes proved neither is needed:

- **Pydantic AI**: `Tool.from_schema(function, name, description, json_schema)` accepts raw JSON Schema. Trivial wrap.
- **Google ADK**: Subclass `BaseTool`, override `_get_declaration()` to return a `FunctionDeclaration` with `parametersJsonSchema=<dict>`. The camelCase alias accepts JSON Schema dicts and Gemini receives them intact.

Both adapters ship the minimal ~30-line wrap.

## Code review — 8 findings addressed

Ran `/code-review` at xhigh before pushing. 15 findings survived verification; 8 fixed in the final commit, 5 deferred with reason:

**Fixed:**
- `MCPToolset.tools` identity instability (cached on first access)
- Post-close `.proxies` / `.tools` silently returned `[]` (now raise RuntimeError)
- Unstable `exc.__class__.__name__` error types normalized to `"unknown"`
- `TimeoutError` bucketed as retryable `transport_error` (was landing non-retryable)
- Google `{}` schema causing Gemini rejection (coerced to minimal object schema)
- Shared mutable `input_schema` dict (defensive shallow copy in `_build_proxy`)
- Stale `hexgate/mcp/__init__.py` docstring (still claimed LangChain-only)
- Passthrough coroutine wrappers dropped across all four adapters
- Test fixture duplication extracted to `tests/mcp/conftest.py`

**Deferred (noted in the commit):** sync `func=` on LangChain (pre-existing), lazy langchain import (langchain is a hard dep today), OpenAI `_tool_origin` observability, Google `tool_context` forwarding, non-object JSON drops (matches existing wrap behavior).

## Tests

- **`tests/mcp/test_proxy_core.py`** — framework-agnostic behavior against `MCPToolProxy` directly (envelope, validation, lifecycle, use-after-close, plus 5 new regression tests pinning the review fixes)
- **`tests/adapters/{langchain,openai,pydantic_ai,google}/test_mcp.py`** — per-adapter tests, thin, focused on the `MCPToolProxy → <framework tool>` conversion
- **`tests/mcp/conftest.py`** — shared helpers (`FakeMCPClient`, `slack_config`, `mcp_tool`, `toolset_stub`, `build_proxy`, `text_result`) imported into all five test files
- **964 SDK tests pass** (was 921 before the fanout, +43 net)
- Ruff, fmt-check clean on every touched file

## Docs

`docs/concepts/mcp.mdx` gains an "Adapter coverage" section with the one-line `wrap_mcp_toolset` call for each of the four adapters.

## Scope + roadmap

Follow-ups noted but not shipping here (all listed in `docs/concepts/mcp.mdx`'s existing "what hexgate does NOT do yet" section):
- Streaming tool responses
- Non-text content blocks (images / binary resources)
- Response-side policy gating
- Reconnection
- Declarative `mcp_servers:` block in `agent.yaml`

## Test plan

- [x] Full SDK suite passes (964 tests)
- [x] Code review addressed
- [ ] Post-merge: `make demo-mcp` runs unchanged on the LangChain path (back-compat)
- [ ] Post-merge: adapt `examples/mcp_demo.py` or add sibling demos for each adapter (small follow-up)

### DIFF
--- a/docs/concepts/mcp.mdx
+++ b/docs/concepts/mcp.mdx
@@ -8,7 +8,7 @@ description: "Auto-wrap third-party Model Context Protocol servers as policy-enf
 [README — MCP servers](https://github.com/HexamindOrganisation/hexgate#-mcp-servers-proxy).
 </Note>
 
-Hexgate's `hexgate.mcp` module wraps any [MCP](https://modelcontextprotocol.io) server as a set of LangChain `BaseTool` objects so they go through the same policy enforcement, audit, and approval flow as native `@agent_tool` functions. Zero glue code: connect the server, every tool it exposes auto-registers under a `mcp-<server>-<tool>` namespace.
+Hexgate's `hexgate.mcp` module wraps any [MCP](https://modelcontextprotocol.io) server as framework-native tool objects so they go through the same policy enforcement, audit, and approval flow as native tool functions. Zero glue code: connect the server, every tool it exposes auto-registers under a `mcp-<server>-<tool>` namespace. Wire the tools through whichever adapter your agent is built on (LangChain, OpenAI Agents, Pydantic AI, Google ADK).
 
 ## Usage
 
@@ -49,6 +49,30 @@ roles:
         mode: allow
         constraints: ["args.repo == 'hexgate'"]
 ```
+
+## Adapter coverage
+
+`MCPToolset` produces framework-agnostic `MCPToolProxy` descriptors. Each adapter has a `wrap_mcp_toolset` that turns those into its own tool type. Same shape everywhere; pick the one that matches your agent.
+
+```python
+# LangChain (also the back-compat default via mcp.tools)
+from hexgate.adapters.langchain.mcp import wrap_mcp_toolset
+tools = wrap_mcp_toolset(mcp)  # list[langchain_core.tools.BaseTool]
+
+# OpenAI Agents
+from hexgate.adapters.openai.mcp import wrap_mcp_toolset
+tools = wrap_mcp_toolset(mcp)  # list[agents.FunctionTool]
+
+# Pydantic AI
+from hexgate.adapters.pydantic_ai.mcp import wrap_mcp_toolset
+tools = wrap_mcp_toolset(mcp)  # list[pydantic_ai.tools.Tool]
+
+# Google ADK
+from hexgate.adapters.google.mcp import wrap_mcp_toolset
+tools = wrap_mcp_toolset(mcp)  # list[google.adk.tools.BaseTool]
+```
+
+Every wrapped tool carries the qualified name, the MCP-advertised description and JSON Schema, and the same `{"ok": True | False, ...}` envelope on invocation. The tools share the toolset's connection lifecycle: once the `async with MCPToolset(...)` block exits, calls return a `use_after_close` envelope rather than crashing.
 
 ## Try it
 
--- a/hexgate/adapters/google/mcp.py
+++ b/hexgate/adapters/google/mcp.py
@@ -58,13 +58,19 @@ class _MCPProxyTool(BaseTool):
         self._call = proxy.call
 
     def _get_declaration(self) -> genai_types.FunctionDeclaration:
-        # Coerce a literal `{}` schema (MCP's "accept anything" shorthand)
-        # to a minimal object schema — Gemini's FunctionDeclaration
-        # validator rejects declarations whose parametersJsonSchema is
-        # missing a top-level `type`, causing the whole tool list to fail
-        # at declaration time. LangChain and OpenAI Agents tolerate `{}`
-        # unchanged, so the coercion is Google-specific.
-        schema = self._input_schema or {"type": "object", "properties": {}}
+        # Gemini's FunctionDeclaration validator rejects any
+        # parametersJsonSchema missing a top-level `type` — literal `{}`,
+        # or a partial like `{"properties": {...}}`, or a bare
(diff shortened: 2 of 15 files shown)


guillaume-hexamind (commented):
Ok with the logic;

If you can check these issues found by claude:

1. Register drops MCP input schemas silent policy mis-gating 88/100

hexgate/cli/register/google.py:78  CONFIRMED
_to_tool_definition reads declaration.parameters, but the new Google MCP adapter (adapters/google/mcp.py:71) populates only parametersJsonSchema, leaving parameters (the Schema field) None. Every MCP tool registers with an empty input schema (properties={}, required=[]). Any generated/default policy that keys on tool arguments (write-vs-read classification, arg-level rules) is computed against an empty arg set â tools are silently mis-gated. This is the security-relevant one.

2. _closed never reset in __aenter__  reused toolset crashes 82/100

hexgate/mcp/proxy.py:138  CONFIRMED
__aexit__ sets _closed=True; __aenter__ rebuilds the stack/proxies/states but never clears _closed. Re-entering one MCPToolset instance in a second async with block opens connections fine, then the first .tools/.proxies access raises RuntimeError('...already exited') despite a live connection â agent can't be built, run aborts. __aexit__'s own clear-on-exit comment implies reuse is meant to be supported.

3. Google empty-schema guard too narrow 74/100

hexgate/adapters/google/mcp.py:67 CONFIRMED
self._input_schema or {"type":"object",...} only fires on a falsy/empty dict. A non-empty schema missing a top-level "type" (e.g. {"properties": {...}} or a bare {"anyOf": [...]}) passes through unchanged and hits the exact Gemini declaration-time rejection the comment claims to prevent failing the whole tool list, agent won't start.

quanghexa94:
All 3 fixed in \`5fddda8\`. Full SDK suite green (1175 passed).

**Fixed:**

1. **[88/100] register/google drops MCP schemas** — \`_to_tool_definition\` now reads \`parametersJsonSchema\` as a fallback when \`declaration.parameters\` is None. Every MCP tool now registers with its actual argument surface, so downstream policy generation sees the real fields instead of an empty schema. Regression test builds a real MCP-wrapped ADK tool with a typed schema and asserts the manifest's InputSchema carries the correct properties + required list.

2. **[82/100] _closed never reset on re-enter** — \`__aenter__\` now sets \`self._closed = False\` at the top before rebuilding the stack. Re-entering the same MCPToolset instance in a second \`async with\` block works on the fresh connection instead of raising "already exited" from the stale flag. Regression test enters → exits → confirms \`.proxies\` raises → re-enters → confirms \`.proxies\` works.

3. **[74/100] Google empty-schema guard too narrow** — replaced \`self._input_schema or {...}\` with an explicit \`"type" not in schema\` check that injects \`type: "object"\` whenever it's absent while preserving every other key. Fixes both the \`{"properties": {...}}\` case and the \`{"anyOf": [...]}\` case the old guard silently passed through. Two regression tests cover both shapes plus one that pins the "well-formed schema is left alone" invariant.

guillaume-hexamind (approved):

PR #66, opened 2026-07-03
refactor(sdk): slim core - remove demo content, fix layering & naming

guillaume-hexamind:
## Summary

Refactor of the `hexgate` SDK to make the core **leaner, vendor-neutral, and correctly layered**: demo/example content no longer ships inside the wheel, and two structural issues (a CLI→SDK dependency inversion and a `register_agent` name collision) are fixed. Net: **−747 / +296 lines**, and the SDK now ships **zero packaged data files** (pure Python).

## Changes

**Remove the builtin agent-source strategy and the `researcher` demo agent**
Deleted `hexgate/agents/builtin/` and the entire "builtin" branch of the loader (`load_builtin_agent`, `list_builtin_agents`, the `builtin` `AgentSource`); renamed `resolve_builtin_tools` → `resolve_tools`. The CLI default now falls back to `example_agent`.
_Why: the only builtin agent was a demo already superseded by `examples/example_agent/`; shipping it in the wheel and maintaining a whole loader strategy for one showcase was dead weight._

**Move the demo `refund_order` tool to `examples/`**
Relocated the stub refund tool out of `hexgate/tools/` and `BUILTIN_TOOLS` (the platform's seeded `support_bot` still references it by name, enforced server-side).
_Why: it was demo-only (its own TODO said "relocate to examples") and leaked domain-specific logic into the public SDK surface._

**Move vendor-bound `web_search`/`fetch` tools to `examples/`**
Relocated the Tavily/Linkup-backed tools to `examples/tools/`, dropped them from `BUILTIN_TOOLS` and the public API, removed the CLI's auto-injection of them into every agent, and retargeted `example_agent` + the default prompt to vendor-neutral filesystem tools.
_Why: they hardcoded specific SaaS vendors + API keys; core should ship vendor-neutral primitives. Side benefit: `hexgate chat` now works zero-config, no Tavily/Linkup keys required._

**Inline the default system prompt, drop the packaged prompts dir**
Turned `DEFAULT_SYSTEM_PROMPT` into an inline string; deleted `hexgate/agents/prompts/` and the `package-data` entry.
_Why: it was one generic paragraph shipped as a data file; inlining removes the file/dir/packaging machinery and leaves the SDK pure-Python._

**Extract a `hexgate/manifest/` package out of `cli/register/` (fix layering leak)**
Moved the manifest models, builder, and per-framework builders to `hexgate/manifest/`; `cli/register/` now consumes from it. `hexgate/__init__.py` no longer imports from `hexgate.cli`.
_Why: the SDK public API was importing from the CLI package — a dependency inversion (the library must not depend on its own CLI)._

**Rename the loader's `register_agent` → `register_agent_factory`**
Renamed the in-process factory-registry function (and `unregister_agent` → `unregister_agent_factory`). The platform `register_agent` backing `hexgate register` is unchanged.
_Why: two different public functions were both named `register_agent` (in-process registry vs publish-to-platform) — a real footgun that had already caused confusion._

**Docs & test-fixture cleanup**
Repointed README/`docs/` off `researcher`/builtin/`load_builtin_agent`, pointed examples at `example_agent`, and updated audit fixtures (SDK + platform + dashboard) to a non-demo agent name.
_Why: keep documentation and fixtures consistent with the removed/renamed surface._

**Untrack demo example agents committed by mistake**
`git rm --cached` for 12 example files (devops/hr/itsm/healthcare/…) that were accidentally committed; kept locally on disk.
_Why: they were added by mistake and don't belong in this PR._

## Verification

- SDK suite: **901 passed**; lint + fmt-check clean; wheel builds.
- Platform agents suite: **81 passed** (server-side policy for `support_bot`/`refund_order` unaffected).
- No SDK→CLI imports remain; `register_agent` collision resolved.

**Accepted coverage loss:** relocating `fetch`/`websearch` to `examples/tools/` drops ~210 lines of tool-wrapper tests (Tavily/Linkup HTTP + env-var validation). Left uncovered on purpose rather than standing up an `examples/tests/` harness for demo-only vendor wrappers that don't ship in the wheel or count toward the SDK coverage gate. (The `tests/cli/test_common.py` deletions are unrelated — they tested the CLI auto-injection this PR removes, so they're correctly gone.)

## ⚠️ Breaking changes

These are intentional consequences of slimming the core; both are called out here as explicit decisions.

- **`web_search`, `fetch`, and `refund_order` are no longer built in.** They moved to `examples/tools/` and were removed from `BUILTIN_TOOLS`, so an `agent.yaml` (or code path) that still lists one now fails at load time. Pass them explicitly via `extra_tools`, e.g. `load_agent("my_agent", extra_tools={"web_search": web_search})`. `resolve_tools` now raises a targeted error naming the new location and the `extra_tools` fix rather than a bare `KeyError`. _(In-repo nothing breaks — `example_agent` uses filesystem tools; the platform-seeded demo agents are Playground-only and enforced server-side, so no executed path is affected.)_
- **No default packaged agent.** With the `researcher` builtin removed, `hexgate chat` in a directory with no local/registered agent now errors ("No agents found …") instead of silently running a bundled agent. Use `example_agent`, add a local `agent.yaml`, or pass `--use` / `--agent`. Migration convenience wins: the default `hexgate chat` experience now works zero-config against the workspace (no Tavily/Linkup keys required).
- **`register_agent` / `unregister_agent` removed from the top-level `hexgate` namespace.** The in-process factory-registry functions are now `register_agent_factory` / `unregister_agent_factory`, so `from hexgate import register_agent` fails with `ImportError`. No deprecation shim ships: the SDK is pre-1.0 with no external consumers yet, and keeping the surface clean is worth more than a compat path for zero known callers. Migration: `from hexgate import register_agent_factory`. (Unrelated: `hexgate.cli.register.register_agent`, which backs `hexgate register`, is unchanged.)


quanghexa94:
Reviewed. One ship-blocker + three real regressions worth flagging, ranked most-severe first.

## 🔴 Blocker — platform-seeded `support_bot` becomes unloadable via the SDK

`platform/api/hexgate_api/features/agents/seed_data.py` seeds `support_bot` with `agent_yaml`:
```yaml
tools:
  - web_search
  - read_file
  - refund_order
```

This PR removes `web_search`, `fetch`, and `refund_order` from `BUILTIN_TOOLS` (`hexgate/agents/loader.py:41-49`) and drops the CLI's auto-injection.

Concrete failure chain:
1. `hexgate serve --agent support_bot` (Playground demo path) → `build_runtime(local_only=False)` → `load_agent(...)` → `load_hexgate_agent(...)` → `resolve_tools(["web_search", "read_file", "refund_order"], extra_tools=None)` at `hexgate/agents/loader.py:452`.
2. `resolve_tools` looks up `web_search` in the shrunk `BUILTIN_TOOLS` → **KeyError** (the friendly "moved to examples/tools/" hint doesn't help the operator here — they can't wire an SDK CLI to `extra_tools`).
3. `hexgate serve` fails at startup.

The PR body says "the platform-seeded demo agents are Playground-only and enforced server-side, so no executed path is affected." That's incorrect: the Playground IS an SDK-driven load. **The OOTB `make demo-notebook` / Playground demo breaks.**

**Fix options**: (a) drop `web_search`/`refund_order` from the seeded `support_bot` agent_yaml too (matches seed to new SDK surface); (b) auto-inject examples/tools in the CLI when they're referenced; (c) keep `refund_order`/`web_search` as stub tools in `BUILTIN_TOOLS` that raise a friendly error only when actually invoked. Option (a) is smallest and most honest.

## 🟠 Hard break — `register_agent` / `unregister_agent` removed with no compat shim

`hexgate/__init__.py` drops both from `__all__` and imports. External `from hexgate import register_agent` for the in-process factory now dies with a bare `ImportError`. The `register_agent_factory` docstring explains the rationale, but the user hitting the import error doesn't see it.

**Fix**: keep `register_agent = register_agent_factory` (with a `DeprecationWarning`) for one release. Same for `unregister_agent`.

## 🟠 Hard break — `web_search`, `fetch`, `refund_order` no longer importable from `hexgate` or `hexgate.tools`

`resolve_tools` gives a targeted error for agent.yaml references, but `from hexgate import web_search` (or `from hexgate.tools import fetch`) crashes at module import with a plain `ImportError` — no hint that the tool moved to `examples/tools/`.

**Fix**: leave a shim or `__getattr__` on `hexgate/tools/__init__.py` that raises with the same "moved to examples/tools/" hint the `resolve_tools` path already emits.

## 🟡 Test coverage loss

`tests/tools/test_tools.py` deletes ~200 lines of fetch + websearch tests. The tools still exist under `examples/tools/` but are now untested — nothing catches drift in the Tavily/Linkup wrappers or the env-var validation. Either move the tests to an `examples/tests/` directory or accept the coverage loss explicitly in the commit message.

## 🟢 Minor — `DEFAULT_SYSTEM_PROMPT` type changed from `Path` to `str`

`hexgate/agents/factory.py:54` inlines the prompt as a string. `load_system_prompt` handles both types (verified by the new test), but any external code holding the constant as a `Path` object and calling `.read_text()` / `.parent` breaks silently. Worth a line in the release changelog.

## What I liked

- The `hexgate.manifest/` extraction cleanly removes the SDK→CLI dependency inversion. `hexgate/__init__.py` no longer imports from `hexgate.cli` — verified.
- The `register_agent_factory` rename with cross-reference docstring is the right disambiguation.
- `package-data` entry in `pyproject.toml` cleanly removed — wheel is pure Python now.
- `resolve_tools` error message specifically names the relocation target and the `extra_tools=` fix path — that's the shape helpful errors should take.

## Recommendation

**Block on the seeded-`support_bot` issue** — that's a real regression against the demo the platform ships and the Playground exercises. Everything else is severity-of-choice: nudging toward compat shims for `register_agent` / `web_search` / `fetch` / `refund_order` (public SDK surface) but defensible as a "clean break in a pre-1.0 rename PR" if the release notes call them out loudly.

guillaume-hexamind:
> Member

Blocker issue -> I dug into the paths and the blocker doesn't hold — the failure chain described doesn't exist:

1. hexgate serve takes a module:attr spec, not a seeded name. Its only positional arg is agent_spec (serve.py:617, e.g. examples.customer_bot:agent) → load_spec() → build_runtime_from_local_agent(). It never calls the name-resolving build_runtime and never hits resolve_tools.
2. The Playground demo builds a local agent, not the seed. deploy/demo_notebook.py build_agent() calls create_agent(tools=TOOLS, name="demo_agent") with tools defined inline in the notebook. The seeded support_bot YAML is never loaded.
3. hexgate chat is local_only=True and ignores HEXGATE_API_KEY (chat.py:352), so it never enters the cloud/load_hexgate_agent branch.
4. The platform never runs the SDK loader — resolve_tools/load_agent/create_agent don't appear anywhere in platform/api/. Seeds are compiled to policy WASM and stored for the /graph canvas; they're not executed server-side.

So make demo-notebook / Playground is unaffected. The only way to reach the KeyError is bespoke code calling load_agent(..., local_only=False) with an API key — not a shipped path.

There's a real-but-minor kernel: the seeded YAML references relocated tool ids (and it's default/read_only too, not just support_bot). Since seeds are only ever compiled/visualized, this is cosmetic, and I'm keeping refund_order because the whole role demo (deny → $50 → $500) is built on it. I'll fix the one genuine bug this surfaced — the stale from hexgate.tools import web_search, fetch hint in the notebook comment.


Hard break 1 -> Good catch — you're right this was a public export. I'm going to keep it as a clean break rather than shim it: we're pre-1.0 (0.2.7), this branch is a deliberate surface refactor, and there are no external SDK consumers yet (the examples/ usage is the unrelated hexgate.cli.register.register_agent). Carrying a deprecation path for zero known callers isn't worth it while we can still keep the surface clean. I'll add a BREAKING: note to the PR body so it's on record.

Hard break 2 -> Fixed — added a __getattr__ on both hexgate and hexgate.tools that raises an ImportError naming the examples/tools/ location (reusing the same relocation map as resolve_tools). Raises rather than shims: the tools stay out of core, but the error is now legible.

Test coverage loss -> Agreed the wrapper logic is now uncovered — accepting it explicitly rather than standing up an examples/tests/ harness for demo-only vendor wrappers that don't ship in the wheel. Noted in the PR body. (The test_common.py deletions are separate: they tested the CLI auto-injection this PR removes, so they're correctly gone.)

quanghexa94:
Verified — blocker withdrawn. I traced the wrong call chain: `hexgate serve` takes `module:attr` (serve.py:617), not a seeded name; the notebook uses inline `create_agent(tools=TOOLS)`, not `load_hexgate_agent`; `hexgate chat` is `local_only=True`; and grep confirms the platform never imports the SDK loader. My mistake — should have walked the CLI's argparse before assuming the failure path.

The other calls all land fine:

- **`register_agent` clean break**: agreed. Pre-1.0, no external consumers, a shim for zero known callers is more clutter than kindness. BREAKING note in the PR body is the right shape.
- **`web_search` / `fetch` / `refund_order` `__getattr__` shim**: perfect — reusing the `_RELOCATED_TOOLS` map means one source of truth for the relocation hint across `resolve_tools` and direct imports.
- **Test coverage loss accepted with a PR-body note**: fine given they're demo-only vendor wrappers, not shipped in the wheel.
- **Stale notebook comment fix**: catches the one real bit my review surfaced.
- **Keeping `refund_order` in the `support_bot` seed** for the deny→$50→$500 role demo: makes sense — the seeds are compiled to policy WASM for the /graph canvas, they're never executed, so the tool name is a policy identifier there, not an SDK dependency.

Good to merge from my side.

quanghexa94 (approved):

PR #67, opened 2026-07-03
fix(platform-api): register_manifest populates agent_yaml so /graph works

quanghexa94:
## The bug

Fresh `hexgate serve`, agents show up on **/agents**, policies fill in on **/policies**, but **/graph** is empty — "No agents yet."

## Root cause

`platform/api/hexgate_api/features/agents/service.py:274-275` — when `hexgate register` creates a new Agent row via `_get_or_create_agent`, it sets `agent_yaml=""`. Then `register_manifest` writes `policy_yaml` (from `_default_policy_for_manifest`) but never touches `agent_yaml`.

The three dashboard pages split on this:

| Page | Reads | Behavior |
|---|---|---|
| /agents | `listAgentManifests` → AgentVersion.manifest | ✓ populated by register |
| /policies | `listAgents.policy_yaml` | ✓ populated by register |
| /graph | `listAgents.agent_yaml` | ✗ empty string → `yaml.load("")` returns undefined → `parseAgent` returns null → `buildAgentView` returns null → `agentViews.length === 0` → "No agents yet." |

The comment at `service.py:263-265` explicitly acknowledged the design ("legacy NOT-NULL columns; code-defined agents leave them empty since the actual content lives on each AgentVersion"). That was fine when the Graph tab didn't parse `agent_yaml`. The current Graph parses it directly, so the code-defined-agent path silently broke the graph.

## Fix

- New `_agent_yaml_from_manifest(manifest)` helper in `compiler.py`, sibling to `_default_policy_for_manifest`. Emits a seed-shaped `agent.yaml` body: name, model (skipped when None), `system_prompt: system.md` (sidecar path), tools list, `policy: policy.yaml` (sidecar path).
- `register_manifest` populates `agent_yaml` on first create AND backfills any legacy row whose column is still `""` (via `or not agent.agent_yaml`).
- Operator-edited `agent_yaml` is preserved — the backfill only fires when the column is empty.

## Migration for existing broken agents

Anyone with a previously-registered agent whose graph is broken just runs `hexgate register` again. The `or not agent.agent_yaml` branch backfills it on the next register. No SQL migration needed.

## Tests

7 new tests:
- `_agent_yaml_from_manifest` — shape, optional-field omission, zero-tools case (unit tests in `test_default_policy.py`)
- `register_populates_agent_yaml_for_graph_page` — end-to-end via TestClient
- `register_backfills_empty_agent_yaml_on_re_register` — legacy auto-heal
- `register_preserves_operator_edited_agent_yaml` — non-empty column left alone

**Full API suite: 382 passed. Ruff + fmt clean.**

## Test plan

- [x] Automated regressions above
- [ ] Post-merge: run `hexgate register` on a previously-broken agent, refresh /graph, confirm it renders
- [ ] Post-merge: fresh `make demo-notebook` → register new agent → /graph shows it immediately

### DIFF
--- a/platform/api/hexgate_api/features/agents/compiler.py
+++ b/platform/api/hexgate_api/features/agents/compiler.py
@@ -233,3 +233,32 @@ def _default_policy_for_manifest(manifest: AgentManifest) -> str:
   admin:
     inherits: [read_only]
 {admin_tools}"""
+
+
+def _agent_yaml_from_manifest(manifest: AgentManifest) -> str:
+    """Build the canonical ``agent.yaml`` body from a registered manifest.
+
+    Populates the ``Agent.agent_yaml`` column that the code-defined
+    registration path used to leave empty. Without a populated
+    agent_yaml the dashboard's Graph tab silently drops the agent —
+    ``yaml.load("")`` returns undefined, which ``parseAgent`` rejects,
+    which ``buildAgentView`` treats as an unparseable agent, which
+    ``buildOverviewGraph`` filters out, which produces the "No agents
+    yet" empty state even when the /agents and /policies pages show
+    the agent fine. Shape mirrors the seeded agents' YAML so the
+    dashboard's YAML editor and downstream tooling see a consistent
+    format across seed-vs-register paths.
+    """
+    lines = [f"name: {manifest.name}"]
+    if manifest.model:
+        lines.append(f"model: {manifest.model}")
+    if manifest.system_prompt:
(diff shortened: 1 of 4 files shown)


quanghexa94:
Closing in favor of a client-side fix that reads the manifest directly instead of synthesizing an `agent_yaml` — see [next PR]. That path avoids duplicating manifest data into a legacy column and the drift bug that would introduce (my `or not agent.agent_yaml` guard only backfills when empty, so a re-register with changed tools would leave `agent_yaml` stale).

PR #68, opened 2026-07-03
fix(dashboard): Graph reads registered manifest as source of tools

quanghexa94:
## The bug

Fresh `hexgate serve`, agents show up on **/agents**, policies fill in on **/policies**, but **/graph** is empty - "No agents yet."

## Root cause

`hexgate register` creates Agent rows with `agent_yaml=""` and populates the manifest instead. The three dashboard pages diverge on which source they read:

| Page | Reads | Behavior |
|---|---|---|
| /agents | `listAgentManifests` → AgentVersion.manifest | ✓ shows agent |
| /policies | `listAgents.policy_yaml` | ✓ shows editor |
| /graph | `listAgents.agent_yaml` | ✗ `yaml.load("")` → undefined → `parseAgent` null → `buildAgentView` null → agent silently dropped |

## Fix (client-side)

Graph now reads the manifest as the source of truth for tools + model on code-registered agents. `agent_yaml` parsing is the fallback path for legacy YAML-edited agents that predate the manifest flow.

- `buildAgentView(agent, manifest?)` — prefers manifest's `.tools[].name` + `.model` when provided; falls back to `agent_yaml` when the caller didn't have a manifest for this agent.
- `buildOverviewGraph(agents, manifests?)` — takes both lists, indexes manifests by name, threads the match into each `buildAgentView`.
- `Graph.tsx` — fetches both `listAgents` and `listAgentManifests` (same project, both required); waits for both before rendering.

## Migration

None needed. Existing broken agents just work after this ships — they had a manifest all along, the client just wasn't looking at it.

## Tests

11 new tests across `policy.test.ts` (extended `buildAgentView` semantics) and a new `graph.test.ts` (overview builder). Key regressions:

- `buildAgentView` prefers manifest tools + model over `agent_yaml`
- `buildAgentView` coerces `manifest.model=null` to `""`
- Falls back to `agent_yaml` when manifest is undefined
- Returns null for empty `agent_yaml` + no manifest (the exact fingerprint of the pre-fix bug — graceful, doesn't crash)
- `buildOverviewGraph` renders manifest-fed agents whose `agent_yaml` is `""`
- Mixes manifest-fed and yaml-fed agents in one project
- Skips a code-registered agent whose manifest isn't loaded yet (query timing race, resolves on re-render)
- Empty manifest tool list → agent node with no outgoing edges

**Full dashboard suite: 139 passed. Lint + typecheck + prettier clean.**

## Test plan

- [x] Automated regression tests above
- [ ] Post-merge: `hexgate serve` → register an agent → /graph shows it immediately (no re-register needed)
- [ ] Post-merge: legacy agent with hand-edited `agent_yaml` still renders correctly

### DIFF
--- a/platform/dashboard/src/lib/graph.test.ts
+++ b/platform/dashboard/src/lib/graph.test.ts
@@ -1,4 +1,4 @@
-import { describe, expect, it } from "vitest";
+import { describe, expect, it, vi } from "vitest";
 import { buildOverviewGraph } from "./graph";
 import type { AgentManifestView, AgentRead } from "./api";
 
@@ -123,5 +123,49 @@ describe("buildOverviewGraph — manifest-fed tools unblock code-registered agen
     expect(graph.edges.filter((e) => e.source === "agent:skeleton")).toEqual(
       [],
     );
+  });
+
+  it("renders an agent whose envelope has manifest=null (registered, no version yet)", () => {
+    // Post-review regression: pre-fix `?.manifest ?? null` collapsed
+    // "no envelope" and "envelope with null manifest" into the same
+    // fallback. The latter is a real state per AgentManifestView docs
+    // (agent row exists, no persisted AgentVersion.manifest yet). We
+    // now render a bare agent node in that case so the operator sees
+    // the agent exists — dropping it silently would reproduce the
+    // exact "No agents yet" bug this PR set out to fix.
+    const envelope = {
+      name: "half_registered",
+      manifest: null,
+      version: null,
+      content_hash: null,
+      updated_at: "2026-07-03T00:00:00Z",
+    } as unknown as AgentManifestView;
+
+    const graph = buildOverviewGraph(
+      [agent("half_registered", "")],
+      [envelope],
+    );
+    expect(graph.agentViews).toHaveLength(1);
+    expect(graph.agentViews[0].name).toBe("half_registered");
+    expect(graph.agentViews[0].tools).toEqual([]);
+  });
+
+  it("warns and keeps the first envelope on duplicate manifest names", () => {
+    // Two envelopes for the same agent name would silently overwrite
+    // each other under a naive Map.set loop, hiding a real backend
+    // bug (join fan-out, stale envelope during rename). Log once and
+    // keep the first so devtools surfaces the problem.
+    const warnSpy = vi.spyOn(console, "warn").mockImplementation(() => {});
+    try {
+      const first = manifestFor("dupe", ["first_tool"]);
+      const second = manifestFor("dupe", ["second_tool"]);
+      const graph = buildOverviewGraph([agent("dupe", "")], [first, second]);
+      expect(graph.agentViews).toHaveLength(1);
+      expect(graph.agentViews[0].tools).toEqual(["first_tool"]);
+      expect(warnSpy).toHaveBeenCalledOnce();
+      expect(warnSpy.mock.calls[0][0]).toMatch(/duplicate manifest envelope/);
+    } finally {
+      warnSpy.mockRestore();
+    }
   });
 });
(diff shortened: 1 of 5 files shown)


guillaume-hexamind (commented):
Logic looks good;

If you can check these potential issues flagged by claude:

1. graphBuildBlocked guard is a no-op for the exact case it was written to fix — 80/100

Graph.tsx:45 — manifests.isFetching && !manifests.isError && manifests.data === undefined. The data === undefined term is only true on the first-ever fetch. On a background refetch (post-hexgate register invalidation, window-focus) the previous manifest array is still cached, so data is not undefined → guard is false → the memo rebuilds buildOverviewGraph(newAgents, staleManifests). A freshly code-registered agent (empty agent_yaml, no envelope in stale map yet) hits the legacy path and is dropped — it flickers out of the graph until the refetch lands. That flicker is precisely what the guard's comment claims to eliminate. The PR's headline goal isn't actually achieved for the register flow.

2. Manifest tools branch has no Array.isArray guard → whole Graph route crashes — 78/100

policy.ts:367 — tools = manifestInfo.tools.map((t) => String(t?.name ?? "")), called unconditionally. The legacy path right above (policy.ts:136) guards with Array.isArray(agent.tools) ? … : []. If listAgentManifests ever returns a manifest whose tools is null / non-array (backend serialization bug, partial record, schema drift) .map throws a TypeError inside the useMemo — no try/catch — so React tears down the tree and the entire Graph page goes blank / hits the error boundary, instead of degrading gracefully like the old code. Strictly worse than the pre-fix "No agents yet."

3. Unparseable policy_yaml + manifest → agent shown as fully denied — 70/100

policy.ts:388,394 — when a manifest envelope exists, an agent with a broken policy_yaml no longer drops out; it renders against the fail-closed {mode:"deny"}, tools:{} policy. Because effectivePolicy.tools is empty, every manifest tool lands in missingInPolicy and every edge draws in deny styling. An operator with a transient YAML parse error sees a fully-configured agent misrepresented as having zero policy coverage — likely to push them to re-author a policy that was fine.

4. Manifests error with all-code-registered agents → contradictory "No agents yet." + error banner — 66/100

On a listAgentManifests 500 in a project where every agent is code-registered (empty agent_yaml), the memo builds with manifests.data === undefined, parseAgent drops them all, agentViews.length === 0, and the body shows "No agents yet." while the top banner simultaneously says manifests failed. Two contradictory messages; the real agents are hidden rather than shown.

5. Error banner shows even when cached manifest data renders a complete graph — 54/100

Graph.tsx:81 — banner is gated on manifests.isError alone. After a good load, a background refetch that 500s sets isError=true while React Query keeps the last-good data; the graph is still built correctly and all agents remain visible — but the red "Code-registered agents may be missing" banner appears anyway. False alarm that erodes trust in the banner.

quanghexa94:
All 5 fixed in \`ca7701f\`. Full dashboard suite: 149 passed.

1. **graphBuildBlocked no-op on refetch** — right, the \`data === undefined\` term only catches the first fetch. Now also compares \`manifests.dataUpdatedAt < agents.dataUpdatedAt\` so a stale manifest array during an invalidate-driven refetch blocks the memo. The freshly-registered agent no longer flickers.
2. **manifest.tools not array → whole route crashes** — added \`Array.isArray\` guard matching the legacy path; degrades to empty tool list, agent still renders.
3. **Broken policy_yaml rendered as fully denied** — AgentView now carries \`policyParseFailed\`; graph skips the tool-edge loop for parse-failed agents, suppresses \`missingInPolicy\`, and excludes them from the tool-node worst-case aggregate so a shared tool isn't poisoned. Agent node receives \`data.policyParseFailed\` so the future AgentNode component can render a warning marker.
4. **Contradictory "No agents yet" + banner** — empty-state text now branches on \`agentsExistButBlocked\` (manifests errored AND agents.data has rows) and explains what happened instead of falsely claiming no agents exist.
5. **False-alarm banner on refetch failure with cached data** — banner gated on \`manifests.isError && manifests.data === undefined\`, so a successful load followed by a background 500 no longer triggers it.

5 new regression tests, one per fix. Lint + typecheck + fmt clean.

guillaume-hexamind (approved):

PR #69, opened 2026-07-06
feat(sdk): expand constraint DSL + policy-in-code, with pydantic↔WASM parity

quanghexa94:
## What

Expands the policy **constraint DSL** and adds **policy-in-code**, while keeping
strict pydantic ↔ WASM/Rego parity — one grammar, two engines, identical
decisions.

### DSL grammar (each tier is a new AST node + Rego case, not a rewrite)
- **Cross-field refs** — `args.amount <= args.limit`
- **`count()`** — `count(args.recipients) <= 10`
- **String functions** — `startswith` / `endswith` / `contains` / `matches` (RE2)
- **Quantifiers** — `every(args.to, endswith(., "@acme.com"))`, `any(...)`, `.field`
- **Boolean composition** — `and` / `or` / `not` + parentheses (Rego has no `or`;
  compiled via helper predicates / rule-ref negation)
- **Named constants** — `consts.<name>` (numbers, strings, lists), merged through
  `inherits:`
- **Call-scope facts** — `role` and `tool` (mirror `input.role` / `input.tool`)

### Policy in code (Option 1)
`PolicyBuilder` + typed `C(...)` constructors that emit the same grammar strings,
validated at the call site, plus `assert_allows` / `assert_denies` /
`assert_needs_approval` for pytest.

### Testing
- **Generative differential fuzzer** (`test_dsl_fuzz.py`) — emits arbitrary valid
  constraints and asserts pydantic == WASM and always-compiles. Deterministic,
  seeded, no `hypothesis` dependency.
- **Enforcement-seam parity** — `PolicyEnforcer.decide()` over both engines with a
  real `User` role scope and complex multi-clause rules.
- Per-tier parity fixtures + named regressions for every divergence found.

### Docs + demo
- New **Writing conditions** guide (example-first); `constraints.mdx` reframed as
  the terse reference.
- `deploy/policy_dsl_demo.py` — a marimo notebook that runs every decision through
  both engines side by side.

## Parity bugs fixed along the way
Both engines must decide identically; these were caught (mostly by the fuzzer) and
fixed + pinned:
- `not`-of-missing (Rego inline negation was 3-valued) → rule-ref negation
- type error inside a quantifier body silently fail-open → parameterized element
  predicate
- cross-type ordered comparison fail-open (`"x" > 10` true in Rego) + bool-as-int
  → type guards on both engines
- `default_policy` dropped by the compiler; unknown/None role fallback ignored by
  WASM; `file_scope` fail-open in WASM

## Test status
1459 passed, 2 skipped (opt-in integration). Lint + format clean. `opa` present,
so all cross-engine parity tests exercised the real `opa` → `wasmtime` path.


guillaume-hexamind (commented):
Logic and new syntax + policy constraints look good, we are now getting more and more complex policies.

If you can check these different claude-raised potential issues:


1. hexgate/cli/policy/main.py:441 — 86/100 · correctness

policy check --engine pydantic calls evaluate_tool_call(policy, tool, tool_args) without role=role, while the WASM path (line 460) and production (enforcer.py:83 → PolicySet.evaluate) both forward it. This is the one pydantic decision path in the PR not updated for the new role fact.
→ hexgate policy check --role admin ... --engine pydantic prints DENY on a role == "admin" constraint that --engine wasm and production ALLOW. The CLI gives a wrong verdict for exactly the role-scoped constraints this PR introduces. Parity tests miss it because they route through PolicySet.evaluate, not the CLI helper.

2. hexgate/security/constraints.py:243 (also :546) — 83/100 · parity

_RE2_INCOMPATIBLE blocklist is incomplete: catches numeric backrefs and lookaround but misses \Z/\A anchors and named backreferences (?P<n>…)(?P=n) — accepted by Python re, rejected by RE2/OPA.
→ matches(args.id, "inv_[0-9]+\Z") passes _validate_re2 and model load, evaluates fine on pydantic (can ALLOW), but the WASM/Rego bundle's regex.match yields undefined under RE2 → DENY. opa build succeeds (regex validated only at runtime), so the divergent bundle ships silently. Same call authorized on one engine, denied on the other.

3. hexgate/security/constraints.py:441 — 80/100 · correctness

Load-time rejection of a bare/unquoted RHS was removed. Old _parse_rhs ran json.loads and raised ConstraintParseError for a bare identifier; new _parse_operand reads any bare identifier as a cross-field Ref, so a forgotten-quotes typo parses silently.
→ args.region != EU (missing quotes) parses as a ref to absent field EU, resolves to _MISSING, and _eval_cmp returns False for both == and !=. An allow-rule args.region != EU therefore never allows any call — surfacing only as mysterious runtime denials with no config-time error.

4. hexgate/security/constraints.py:717 — 70/100 · security footgun

matches() compiles to an unanchored re.search/regex.match, so a pattern meant as a whole-value gate matches any substring.
→ matches(args.file_path, "src/") accepts ../../etc/src/passwd; matches(args.id, "inv_[0-9]+") accepts attacker_inv_9_evil. Documented and consistent across engines (footgun, not a cross-engine bug), but a real enforcement gap on an auth path.

5. hexgate/security/rego.py:164 — 66/100 · parity

consts.<name> reference resolution is validated only in the Rego compile path (_validate_const_refs); the pydantic runtime and the model-level grammar validator never check it.
→ args.amount <= consts.max_refund with max_refund undefined loads cleanly, evaluates as MISSING → DENY on pydantic, but compile_to_rego raises PolicySetError and refuses to build. The two deployment paths disagree on whether the same YAML is even valid — surfaces only at bundle-build time.

6. hexgate/cli/policy/main.py:319 — 60/100 · validation gap

policy validate is not a subset of build: it runs only parse_constraint + schema load, skipping the const-ref / file_scope / RE2 checks that build performs.
→ validate prints "✓ Policy parses cleanly" for policies build then rejects. A dev using validate as their CI gate ships a policy that fails at the platform build step — false confidence rather than mirroring the real pipeline.

quanghexa94:
Thanks — all six looked at. **5 fixed with code + tests, 1 is intentional (explained below). 1470/1470 tests pass, lint + fmt clean.** All in `4c9e0fc`.

**1. `policy test --engine pydantic` dropped `role`** — fixed. `_test_via_pydantic` now forwards `role=role` like the wasm path. Tests: `--role admin` on `role == "admin"` now ALLOWs on both engines; `--role default` denies.

**2. `_RE2_INCOMPATIBLE` incomplete (`\Z`, named backrefs, …)** — fixed. Verified against `opa eval` that `\Z`, `(?P=name)`, `(?#…)`, and `(?(…)…)` are all *undefined* under RE2 while Python matches, then added them to the blocklist. Named-group *definitions* `(?P<n>…)` stay allowed (both engines support them). Rejection tests added for each.

**3. Bare/unquoted RHS parsed silently as a ref** — fixed. A lone identifier is now only valid as a fact (`role`/`tool`); anything else (`== USD`) raises at load with a "did you forget quotes?" message. Dotted cross-field refs (`args.max`) are unaffected.

**5. `consts.<name>` validated only in the Rego path** — fixed. Moved the undefined-const check to `PolicySet.__init__` (shared `iter_const_refs` walker), so the pydantic load path and the compiler now agree on validity — the same YAML is valid or invalid on both, at load time. Removed the duplicated check from `rego.py`.

**6. `validate` not a subset of `build`** — fixed. `policy validate` now runs the full `compile_to_rego` (minus the opa/wasm step), so it catches const conflicts and `file_scope`-in-wasm, not just grammar. Tests assert validate now rejects policies build would reject.

**4. `matches()` is unanchored (substring)** — leaving as-is, intentionally. This mirrors Rego's `regex.match`, which is also unanchored, so it's consistent across engines (not a parity bug — you flagged this too). Auto-anchoring would both diverge from Rego and break legitimate substring gates. It's documented in the Writing-conditions guide and the reference with the `^…$` tip. Happy to add a lint/warning for anchor-less patterns on auth-sensitive tools as a follow-up if you'd like.

guillaume-hexamind (approved):

PR #70, opened 2026-07-06
docs: README and documentation restructuration

guillaume-hexamind:
Restructures the project docs from a single monolithic `README.md` into a structured Mintlify docs site.

## What changed

- **README** slimmed from ~1500 lines to a short landing page that points into the docs site.
- **Docs site** organized by reader intent:
  - **Getting started** — `introduction`, `quickstart`, `two-paths` (chat vs serve)
  - **Guides** — `build-an-agent`
  - **Concepts** — policy decision, user scope, audit trail, sandbox, signed bundles, MCP
  - **CLI** — `chat`, `serve`, `register`, `policy`
  - **Platform** — overview, workflow, hosted (Cloud), control-plane (self-host), dashboard, audit/ClickHouse, email
  - **Policy** — YAML shape, constraints, WASM bundles, signing, local override
  - **Adapters** — OpenAI Agents, LangChain, Google ADK, Pydantic AI
  - **Reference** — environment variables
  - **SDK** — public API index
- **Navigation** (`docs.json`) updated for the new tree.

Documentation only — no code changes.


README.md
```diff
@@ -4,10 +4,10 @@
 
 # Hexgate
 
-**Authorization infrastructure for AI agents.**
-Policy enforcement, signed policy bundles, per-request user scope, audit trail — for OpenAI Agents, LangChain, Google ADK, Pydantic AI, or a native runtime.
+**Runtime authorization for AI agents.**
+On every tool call, Hexgate decides whether *this user*, in *this role*, may run *this tool* with *these arguments* — allow, deny, or require approval. For OpenAI Agents, LangChain, Google ADK, Pydantic AI, or a native runtime.
 
-[**Website**](https://hexgate.ai) · [Docs](https://docs.hexgate.ai) · [PyPI](https://pypi.org/project/hexgate/) · [Discussions](https://github.com/HexamindOrganisation/hexgate/discussions)
+[**Website**](https://hexgate.ai) · [**Docs**](https://docs.hexgate.ai) · [PyPI](https://pypi.org/project/hexgate/) · [Discussions](https://github.com/HexamindOrganisation/hexgate/discussions)
```
victorludvig: The discussion link is broken.
We could create a discord server instead.

README.md
```diff
@@ -31,1501 +27,95 @@ Policy enforcement, signed policy bundles, per-request user scope, audit trail 
...
-
-async with MCPToolset(slack) as mcp:
-    agent, handler = create_agent(model="gpt-5.4", tools=mcp.tools)
-    agent = enforce_policy(agent, "policy.yaml")
-    await agent.ainvoke({"messages": [...]}, config={"configurable": {}})
+pip install hexgate
 ```
 
-Then in `policy.yaml` reference the MCP tools by qualified name:
+**See it enforce — no API keys.** Save a policy that gives two roles different
```
victorludvig, line 52: I understand that this section is different from the notebook because it does not require an API key.
However, should we include this in the quick start notebook and link it here instead?

README.md
```diff
@@ -31,1501 +27,95 @@ Policy enforcement, signed policy bundles, per-request user scope, audit trail 
...
-### Audit log (ClickHouse)
-
-Policy decisions are written to a local ClickHouse instance for the audit dashboard. Requires Docker.
-
-```bash
-make clickhouse-up        # start the server (first run also creates the schema)
-make clickhouse-cli       # interactive SQL shell
-make clickhouse-down      # stop (keeps data)
-make clickhouse-reset     # wipe and recreate (also re-applies the schema)
+   end user (id + role)          tool call (name + args)
```
victorludvig, line 35: Not necessary for this PR but I would add an image diagram here.

README.md
```diff
@@ -31,1501 +27,95 @@ Policy enforcement, signed policy bundles, per-request user scope, audit trail 
 
 Hexgate is two things that move together:
 
-- **`hexgate` — the SDK.** A Python runtime that gates every tool call through a typed `Decision` (allow / deny / approval-required), wraps your existing OpenAI / LangChain / Google ADK / Pydantic AI agent without rewriting it, and threads per-request user identity through tracing + audit.
-- **The Hexgate platform** *(optional)* — a FastAPI control plane + React dashboard for editing policy in a browser, minting per-project tokens, watching live decisions stream from a serving agent, and shipping signed WASM policy bundles to production.
+- **`hexgate` — the SDK.** A Python runtime that gates every tool call through a typed `Decision` (allow / deny / approval-required), resolving the caller's role at call time to pick which rules apply. Wraps your existing OpenAI / LangChain / Google ADK / Pydantic AI agent without rewriting it. Every decision is traced and audited with the caller's identity.
```
victorludvig: The supported Agentic frameworks are already listed in the header, so I think they are redundant here.

Proposition:
- **`hexgate` — the SDK.** A Python runtime that gates every tool call through a typed `Decision` (allow / deny / approval-required), resolving the caller's role to enforce pre-defined rules. Wraps your existing agents without rewriting them. See supported frameworks [INSERT HYPERLINK].

docs/concepts/user-scope.mdx
```diff
@@ -3,26 +3,135 @@ title: "User scope"
...
+@app.middleware("http")
+async def attach_user(request, call_next):
+    auth = await authenticate(request)                       # your auth
+    async with User(
+        user_id=auth.id,
+        role=auth.role,                                       # e.g. "billing"
+        session_id=request.state.session_id,
+        ttl_seconds=300,
+    ):
+        return await call_next(request)
```
quanghexa94: **Bug in the example — this pattern silently strips the User scope from streaming responses.**

`async with User(...): return await call_next(request)` returns the endpoint's `StreamingResponse` and then `__aexit__` immediately resets `_CURRENT_USER`. Starlette iterates `body_iterator` *after* the middleware returns, so every event the endpoint yields runs with `get_current_user() == None` — `PolicyEnforcer.decide` sees role=None, the billing policy is never selected, and the default policy applies.

A reader deploying this pattern gets a silent auth downgrade — no error, wrong policy. Options: bind the scope inside the endpoint itself, or use an ASGI middleware that wraps the send channel.

docs/platform/control-plane.mdx
```diff
@@ -1,22 +1,78 @@
...
 
-- `POST /v1/projects/:id/tokens` — mint a dev token
-- `GET /v1/projects/:id/agents` — list agents
+- `POST /v1/projects/:id/tokens` — mint a dev token (returned in full once)
+- `GET /v1/projects/:id/tokens` — list dev tokens (masked)
+- `DELETE /v1/projects/:id/tokens/:tid` — revoke
+- `GET /v1/projects/:id/agents` — list agents with their YAMLs
+- `GET /v1/projects/:id/agents/:name` — read one agent
 - `PUT /v1/projects/:id/agents/:name` — save agent / policy / system YAMLs
 - `WS /v1/projects/:id/serve` — producer socket (the `hexgate serve` CLI dials here)
```
quanghexa94: Wrong route. The producer socket is `WS /v1/serve`, not `WS /v1/projects/:id/serve`.

- `platform/api/hexgate_api/features/chat/router.py:13` — `@router.websocket("/serve")` under the `/v1` prefix
- `hexgate/cli/serve.py:550` — CLI dials `f"{ws_base}/v1/serve"`

Project is resolved from the bearer subprotocol, not the path. A third-party client built to the documented URL gets 404. Line 61 (`WS /v1/projects/:id/chat`) is fine — that one really is under `/projects/{id}/`.

docs/concepts/policy-decision.mdx
```diff
@@ -3,18 +3,56 @@ title: "Policy decision"
...
-- **`allow`** — the original tool runs unchanged.
-- **`deny`** — the framework sees a `[policy_denied]` marker as the tool result; the model can recover.
-- **`approval_required`** — calls a dev-supplied handler, or returns an `[approval_required]` marker.
+## Where enforcement actually happens
+
+Walk through one tool call:
+
+1. The model emits a tool call. The framework's tool dispatcher invokes the tool.
+2. The tool is *not the dev's original* — it's a copy the adapter made, whose body
+   starts with `enforcer.decide(role, tool_name, args)`.
```
quanghexa94: Wrong signature. `PolicyEnforcer.decide` in `hexgate/security/enforcer.py:61` is `decide(self, tool_name, arguments)` — two args, no role. Role is read internally via `get_current_user()` from the User contextvar (`enforcer.py:71-72`).

A reader taking this literally and writing `enforcer.decide('billing', 'refund_order', {...})` in a debug script gets `TypeError: decide() takes 3 positional arguments but 4 were given`, and worse, is misled into thinking they must supply role at the call site.

docs/concepts/policy-decision.mdx
```diff
@@ -3,18 +3,56 @@ title: "Policy decision"
...
+3. `PolicyEnforcer.decide` reads `self.policy` — either a `PolicySet` (pydantic
+   engine, default fallback) or a `PolicyBundle` (WASM engine, what production
+   runs).
+4. Decision is `allow` → the original tool runs. `deny` → returns a
+   `[policy_denied]` marker the model sees as the tool result.
+   `approval_required` → either calls the dev-supplied approval handler or returns
+   an `[approval_required]` marker.
+5. **Before step 2, every turn:** `refresh_policy()` calls
+   `self._policy_source.fetch()`. If the source returns a new bundle instance,
```
quanghexa94: `_policy_source` attribute doesn't exist on the agent. The actual chain is:

- `HexgateAgent.refresh_policy()` → `self._binding.refresh()` (`hexgate/agents/factory.py:507`)
- `PolicyBinding.refresh()` → `self.source.fetch()`

So the attribute is `agent._binding.source`. Line 37 ("`_policy_source` is set automatically by the loader") is the same issue — someone grepping the SDK for `_policy_source` finds nothing and has to reverse-engineer the seam.

docs/policy/signing.mdx
```diff
@@ -3,25 +3,51 @@ title: "Signing & verification"
...
 
+`HEXGATE_BUNDLE_REQUIRE_SIGNATURE` controls strictness — warn-by-default keeps
+local dev frictionless; opt into refusal for CI/prod:
+
+| Bundle | `PUBKEY_PATH` set | `REQUIRE_SIGNATURE` | Outcome |
+|---|---|---|---|
+| signed | yes | either | verify; **refuse if it fails** |
+| signed | no | `false` | load with warning (can't verify) |
+| signed | no | `true` | **refuse** (no key to check against) |
+| unsigned | — | `false` (default) | load with warning |
```
quanghexa94: This row is wrong — unsigned bundles in permissive mode load **silently**, not with a warning.

`hexgate/security/source.py:505` docstring is explicit about the warn conditions:
```
require=false + signed bundle + no key  → warn_if_unverified()
require=false + unsigned              → silent OK
```

`warn_if_unverified` at `source.py:575` gates on `bundle.is_signed and self.verify_with is None` — unsigned bundles skip it. The only stderr line for unsigned+permissive is `_announce_local_override`'s `[hexgate] HEXGATE_LOCAL_POLICY active (…, unsigned)`, which is not a warning and fires regardless.

Ops operators reading this row will assume they'll be told if they forget to sign. They won't. Either fix the code to emit a real warning here, or fix the doc to say "silent load".

docs/sdk/index.mdx
```diff
@@ -3,24 +3,85 @@ title: "SDK reference"
...
+| `User(user_id, role=…, session_id=…, ttl_seconds=…)` | Per-request scope context manager — its `role` is the decision input. See [user scope](/concepts/user-scope). |
+
+## Built-in tools
+
+Filesystem tools are importable from `hexgate`: `read_file`, `write_file`,
+`edit_file`, `glob`, `grep`, `bash`. The shell-backed ones run inside the
+[workspace sandbox](/concepts/sandbox).
+
+<Note>
+`web_search`, `fetch`, and `refund_order` are **no longer built in** — they
```
quanghexa94, line 81: This Note is inverted from reality — the tools *are* still built in.

`hexgate/__init__.py:36-47` imports and re-exports all three:
```python
from hexgate.tools import (
    ...
    refund_order,
)
from hexgate.tools.fetch import fetch
from hexgate.tools.websearch import web_search
```
and `__all__` at lines 86/90 keeps them public. `examples/tools/` doesn't exist as a directory — the source lives under `hexgate/tools/`.

A reader following this Note either (a) expects `from hexgate import web_search` to raise `ImportError` (it succeeds), or (b) tries `from examples.tools.websearch import web_search` and gets `ModuleNotFoundError`. Either delete the Note or land the extraction it describes.

docs/reference/environment.mdx
```diff
@@ -0,0 +1,69 @@
...
+ones whose feature you use. A missing key surfaces when that feature runs: the
+built-in tools raise their own clear error at call time, and the model provider
+raises a key-named error when the agent is built.
+
+## Provider / tool keys
+
+| Env var | Purpose |
+|---|---|
+| `OPENAI_API_KEY` | Default `openai:gpt-5.4` model. Skip it if you pass your own model. |
+| `LINKUP_API_KEY` | The `web_search` example tool (`examples/tools/websearch.py`). |
```
quanghexa94, line 16: Both this row and line 17 (`TAVILY_API_KEY`) point at `examples/tools/…`. Actual paths:
- `hexgate/tools/websearch.py`
- `hexgate/tools/fetch.py`

`examples/tools/` doesn't exist. Same root cause as the sdk/index.mdx Note above.

docs/quickstart.mdx
```diff
@@ -18,47 +18,121 @@ description: "Install hexgate and run the example agent in 60 seconds."
...
-    LINKUP_API_KEY=...
-    TAVILY_API_KEY=...
+    hexgate policy test policy.yaml --role support \
+        --tool refund_order --args '{"amount": 400, "currency": "USD"}'
+    # ✗ DENY · support → refund_order({"amount": 400, "currency": "USD"})
+    #   reason: Policy on "refund_order" denied: constraint failed — args.amount <= 50
+
+    hexgate policy test policy.yaml --role billing \
+        --tool refund_order --args '{"amount": 400, "currency": "EUR"}'
+    # ✓ ALLOW · billing → refund_order({"amount": 400, "currency": "EUR"})
```
quanghexa94: Same policy example lives in both `README.md` (line ~82) and this quickstart, and they've **already diverged in this same PR**:
- Here: billing allowed for `{amount: 400, currency: "EUR"}`
- README: billing allowed for `{amount: 400, currency: "USD"}`
- Constraint layout differs too (multi-line here, inline `[ "…" ]` in README)

Pick one location and have the other link to it — otherwise a maintainer tweaking one output string will silently rot the other.

quanghexa94 (commented):
Nice restructure overall — nav is coherent and the slim-down doesn't lose anything the code still needs (I checked the old README's ripgrep-preflight section; the SDK no longer depends on `rg`, so dropping it is correct).

Most findings below are **factual accuracy against the SDK** — the docs describe APIs, routes, or file paths that don't match the code. The user-scope middleware example (#1) is the one worth pausing the merge for: it's a working pattern that will silently degrade auth in a reader's app.

CLAUDE.md rule invoked throughout: *"NEVER fabricate code examples… Verify third-party or config behavior against actual files in this repo before generating code."*

README.md
guillaume-hexamind: Good catch, and that would be a good idea. Deleting the link for now

README.md
guillaume-hexamind: Good point on avoiding duplication. I still think the part can be kept in the READBE rather than link to the notebook -> this is a zero friction two copy past commands, with no key or runtime, while the marimo notebook would need the reader to open notebook, paste an openai key...

I guess they can be kept as complementary -> README with static proof of role + args decide, and notebook for live agent execution.

README.md
guillaume-hexamind: Agree, modified

docs/sdk/index.mdx
guillaume-hexamind: In current version of main and current PR branch, examples/tools do exist, and the __init__ above is not up to date

docs/reference/environment.mdx
guillaume-hexamind: In current version of main and current PR branch, examples/tools do exist

docs/concepts/user-scope.mdx
guillaume-hexamind: Good catch thanks, user scope doc has been modified

docs/platform/control-plane.mdx
guillaume-hexamind: Good catch, fixed

docs/concepts/policy-decision.mdx
guillaume-hexamind: thanks, fixed

docs/concepts/policy-decision.mdx
guillaume-hexamind: leftover legacy logic from previous docs, fixed

docs/quickstart.mdx
guillaume-hexamind: You are right, they are now aligned

docs/policy/signing.mdx
guillaume-hexamind: fixed

victorludvig (commented):
LGTM

quanghexa94 (approved):

PR #64, opened 2026-07-02
refactor(platform/api): carve flat modules into core/deps/domains packages

guillaume-hexamind:
## What

Structural refactor of `platform/api/` into a proper package with per-domain vertical slices. **No behavior change** — all routes, paths, and response shapes are identical; it's a pure reorganization.

## Changes

- **Package-ified** the flat module dir into `hexgate_api/` (real imports, no CWD-on-`sys.path`).
- **`core/`** — shared infra: `db`, `keystore` (holds the signing singleton), `biscuits`, `clickhouse`, `relay`, `mailer`, `spa`, `ids`.
- **`deps/`** — FastAPI dependency gates (`identity`, `tokens`, `org`, `project`, `ws`, `clickhouse`).
- **`domains/<x>/`** — one slice per context (`tokens`, `projects`, `members`, `orgs`, `invitations`, `agents`, `audit`, `chat`, `auth`), each `router.py` + `service.py` (agents also `compiler.py`).
- **`constants.py`** (shared `ROLE_*` / `DEFAULT_*`) and **`bootstrap/seed.py`** (first-boot seeding).
- **`main.py` → `create_app()` factory** (2160 → ~185 lines); **`services.py` fully dissolved** (1470 → 0).
- **Tests** relocated to mirror the source (`tests/domains/**`, `tests/core/`); CI postgres-smoke path updated.

## Verification

`make platform-api-check` green: **376 passed, 2 skipped**, ruff + fmt clean, ~83% coverage. App boots (`/health` 200), no import cycles. Kept green at every intermediate step.

PR #71, opened 2026-07-06
fix(platform-api): patch clickhouse_ping at its actual import site in readiness test

victorludvig:
## Summary
- `test_readiness_reports_clickhouse` monkeypatched `hexgate_api.main.clickhouse_ping`, but the `#64` refactor (carve flat modules into core/deps/domains packages) moved `_readiness()` into `hexgate_api/health.py`, which resolves `clickhouse_ping` from its own module-level import — so the patch never took effect and the test silently exercised the real (healthy) ClickHouse instead of the unreachable case it's meant to cover.
- Repointed the patch to `hexgate_api.health.clickhouse_ping`, matching where `_readiness()` actually looks up the name.

## Test plan
- [x] `make platform-api-check`
- [x] `make check-all`

### DIFF
--- a/platform/api/tests/features/audit/test_audit.py
+++ b/platform/api/tests/features/audit/test_audit.py
@@ -727,7 +727,7 @@ def test_liveness_does_not_ping_clickhouse(
 def test_readiness_reports_clickhouse(
     monkeypatch: pytest.MonkeyPatch, path: str
 ) -> None:
-    monkeypatch.setattr("hexgate_api.main.clickhouse_ping", lambda: False)
+    monkeypatch.setattr("hexgate_api.health.clickhouse_ping", lambda: False)
     r = TestClient(app).get(path)
     assert r.status_code == 503
     assert r.json()["clickhouse"] == "unreachable"


victorludvig:
> *Posted by Claude Code*

Why this slipped through CI: the `platform-api` job has no ClickHouse service, and `ping()` swallows connection errors, returning `False` on failure. So in CI, `ping()` was always `False` regardless of whether the monkeypatch worked — the test passed for the right outcome, wrong reason. It only surfaced locally because a real, reachable ClickHouse made the actual ping return `True`, exposing that the patch target had gone stale since the `#64` refactor moved `_readiness()` into `health.py`.

Adding a ClickHouse CI service wouldn't fix this on its own — CI would then match local and fail the same way, since the mock is still broken. Useful for other integration tests, but orthogonal to this bug.

Longer-term note: `monkeypatch.setattr` on an internal name is fragile to the next refactor (has to target where the name is *called*, not where it's defined). A FastAPI `Depends`-based override would be more durable. Not proposing that change here — just flagging it.

guillaume-hexamind (approved):

PR #73, opened 2026-07-08
refactor(sdk): Generalize audit api keys registry

victorludvig:
## What is changing? 

The **AuditSender** class from hexgate/audit.py is moved to a new [hexgate/tracing/_senders.py](https://github.com/HexamindOrganisation/hexgate/compare/vl/refactor/generalize_audit_api_keys_registry?expand=1#diff-e16676d060a29f26683fab89d814a31b5b67b3c6839a35b7b39e623330b264a0) file. The class is nearly identical, except for **event** type parameter in **emit()**, **_spawn_send()**, **_send()**, which is changed from **AuditEvent** to a generalised **_PayloadEvent**.

To send the llm invocation reports, we need a class nearly identical to **AuditSender**, so it is generalised to be used to send both llm invocations and audits. This class includes an API registry which maps from **(api_key, path)** pairs to sender instances — the same **HEXGATE_API_KEY** covers both event types for a project, but each needs its own live connection to its own endpoint (`/v1/audit/decisions` vs. `/v1/audit/llm-invocations`). Keying by `api_key` alone would make configuring the second event type silently reuse the first one's sender and POST to the wrong endpoint.

The **HEXGATE_LOCAL_MODE** gate (suppresses sends during local iteration, e.g. `hexgate chat`) moves with it, so both event types share the same kill switch instead of duplicating the check.

`audit.py` keeps `AuditEvent` and the decision-specific redaction/truncation logic, and becomes a thin wrapper around the shared registry — `configure()` / `get_sender()` / `shutdown()` are unchanged from the caller's perspective. `usage.py` gains the equivalent `configure_usage_sender()` / `get_usage_sender()` / `shutdown()` as the first real consumer proving the registry generalises.

## Design choices
- Registry keyed by `(api_key, path)`, not `api_key` alone — see "Decision — sender registry for LLM-usage events: Design C" on [Notion](https://app.notion.com/p/Count-tokens-per-LLM-395fb45dbae2809abda5e148373b3096) for the alternatives considered (fully separate registry per event type vs. merging into audit.py).
- First of two PRs for LLM token usage capture — per-adapter hooks (`emit_llm_usage()`, LangChain/OpenAI/Google/Pydantic AI wiring) land separately.

## Tests
- make check
- pytest -m integration

### DIFF
--- a/hexgate/tracing/usage.py
+++ b/hexgate/tracing/usage.py
@@ -1,0 +1,70 @@
+from __future__ import annotations
+
+from dataclasses import dataclass, field
+from datetime import datetime, timezone
+from typing import Any
+from uuid import UUID, uuid4
+
+from hexgate.tracing._senders import AuditSender, get_or_create_sender
+from hexgate.tracing._senders import get_sender as _get_sender
+from hexgate.tracing._senders import shutdown as _shutdown_all
+
+_LLM_USAGE_PATH = "/v1/audit/llm-invocations"
+
+
+@dataclass(frozen=True, slots=True)
+class LlmUsageEvent:
+    agent_name: str
+    model: str
+    input_tokens: int
+    output_tokens: int
+    latency_ms: int
+    status: str  # e.g., "success", "error"
+    session_id: str = ""
+    user_id: str = ""
+    error_code: str | None = None  # optional error code if status is "error"
+    event_id: UUID = field(default_factory=uuid4)
+    occurred_at: datetime = field(default_factory=lambda: datetime.now(timezone.utc))
+
+    def as_payload(self) -> dict[str, Any]:
+        return {
+            "event_id": str(self.event_id),
+            "occurred_at": self.occurred_at.isoformat(),
+            "agent_name": self.agent_name,
+            "session_id": self.session_id,
+            "user_id": self.user_id,
+            "model": self.model,
+            "input_tokens": self.input_tokens,
+            "output_tokens": self.output_tokens,
+            "latency_ms": self.latency_ms,
+            "status": self.status,
+            "error_code": self.error_code or "",
+        }
+
+
+def configure_usage_sender(
+    api_key: str | None = None,
+    base_url: str | None = None,
+) -> AuditSender | None:
+    """Get-or-create the LLM-usage sender for ``api_key``. Idempotent per key.
+
+    Shares the registry (and the ``HEXGATE_LOCAL_MODE`` gate) with
+    :func:`hexgate.audit.configure` via ``hexgate.tracing._senders`` — the
+    same api_key configured for decisions gets its own sender here, keyed
+    separately by endpoint path so the two event types never collide.
+    """
+    return get_or_create_sender(_LLM_USAGE_PATH, api_key, base_url)
+
+
+def get_usage_sender(api_key: str | None = None) -> AuditSender | None:
+    """Return the LLM-usage sender for ``api_key`` (or ``HEXGATE_API_KEY``),
+    if configured. Never creates one."""
+    return _get_sender(_LLM_USAGE_PATH, api_key)
+
+
+async def shutdown() -> None:
+    """Drain in-flight emits and close every sender in the shared registry
+    (decisions and LLM usage alike). Safe to call multiple times;
+    equivalent to :func:`hexgate.audit.shutdown` — either name drains the
+    whole shared registry."""
+    await _shutdown_all()
--- a/docs/internals/audit-pipeline.md
+++ b/docs/internals/audit-pipeline.md
@@ -1,6 +1,6 @@
 # Audit Pipeline Specification
 
-> Status: living — kept in sync with the audit code. Last reviewed 2026-06.
+> Status: living — kept in sync with the audit code. Last reviewed 2026-07.
 
 Scope: the end-to-end path that records every policy decision a Hexgate-wrapped
 agent makes, from the SDK enforcement point to durable storage in ClickHouse and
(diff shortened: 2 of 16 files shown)


victorludvig:
### Code review

No issues found. Checked for bugs, CLAUDE.md compliance, and historical/regression context.

victorludvig:
### Code review (additional low-confidence findings)

Two more items were surfaced during review but scored below the reporting threshold (0 and 50 confidence, respectively). Flagging per request — treat as minor/optional:

1. Commit/PR title capitalization does not match the lowercase convention (CLAUDE.md says: "Commits: `type(scope): description` (lowercase, imperative, no period)"). Note: this is a widespread, unenforced pattern across many prior commits in this repo, so it's pre-existing style drift rather than something unique to this PR.

https://github.com/HexamindOrganisation/hexgate/commit/242ab70c18f76df2958117d7d2478a6b59ee0fdb

2. `docs/internals/audit-pipeline.md` still describes the old single-key registry ("Keyed by api_key. Senders live in a registry `dict[str, AuditSender]`"), but this PR changes the registry to a composite `(api_key, path)` key in `hexgate/tracing/_senders.py`. The doc isn't touched by this PR, so it's now stale. Not required by CLAUDE.md to update in the same PR, but worth a fast-follow.

https://github.com/HexamindOrganisation/hexgate/blob/242ab70c18f76df2958117d7d2478a6b59ee0fdb/docs/internals/audit-pipeline.md#L185-L189

victorludvig:
> ### Code review (additional low-confidence findings)
> Two more items were surfaced during review but scored below the reporting threshold (0 and 50 confidence, respectively). Flagging per request — treat as minor/optional:
> 
> 1. Commit/PR title capitalization does not match the lowercase convention (CLAUDE.md says: "Commits: `type(scope): description` (lowercase, imperative, no period)"). Note: this is a widespread, unenforced pattern across many prior commits in this repo, so it's pre-existing style drift rather than something unique to this PR.
> 
> [242ab70](https://github.com/HexamindOrganisation/hexgate/commit/242ab70c18f76df2958117d7d2478a6b59ee0fdb)
> 
> 2. `docs/internals/audit-pipeline.md` still describes the old single-key registry ("Keyed by api_key. Senders live in a registry `dict[str, AuditSender]`"), but this PR changes the registry to a composite `(api_key, path)` key in `hexgate/tracing/_senders.py`. The doc isn't touched by this PR, so it's now stale. Not required by CLAUDE.md to update in the same PR, but worth a fast-follow.
> 
> https://github.com/HexamindOrganisation/hexgate/blob/242ab70c18f76df2958117d7d2478a6b59ee0fdb/docs/internals/audit-pipeline.md#L185-L189
> 

1- nit, discarded
2- doc updated

hexgate/tracing/usage.py
```diff
@@ -0,0 +1,69 @@
...
+from hexgate.tracing._senders import AuditSender, get_or_create_sender
+from hexgate.tracing._senders import get_sender as _get_sender
+from hexgate.tracing._senders import shutdown as _shutdown_all
+
+_LLM_USAGE_PATH = "/v1/audit/llm-invocations"
+
+
+@dataclass(frozen=True, slots=True)
+class LlmUsageEvent:
+    event_id: str
```
guillaume-hexamind: AuditEvent uses event_id: UUID = field(default_factory=uuid4) -> better to do the same here to align both

hexgate/tracing/usage.py
```diff
@@ -0,0 +1,69 @@
...
+    agent_name: str
+    model: str
+    input_tokens: int
+    output_tokens: int
+    latency_ms: int
+    status: str  # e.g., "success", "error"
+    session_id: str = ""
+    user_id: str = ""
+    error_code: str | None = None  # optional error code if status is "error"
```
guillaume-hexamind, line 25: as_payload() always emits "error_code": self.error_code, and the field defaults to None. But the platform schema is:

error_code: str = Field(default="", max_length=64)

Platform receiving a None error code would thus be rejected at ingest -> either pass "" as default value, or  change the as_payload to emit self.error_code or ""

guillaume-hexamind (changes requested):
2 minor issues to check

But overall the sender refactor + LLM definition logic looks good.

You can also check this feedback from claude:

1. Unscoped shutdown() drains the shared sender registry — 72/100

hexgate/tracing/_senders.py:275 (root cause also at hexgate/audit.py:165, hexgate/tracing/usage.py:64) · correctness · CONFIRMED

shutdown() now drains the entire shared (api_key, path) registry. So hexgate.audit.shutdown() also closes LLM-usage senders (and vice versa) — the two event types are no longer independently teardown-able.

Failure: a process configures both audit.configure(k) and configure_usage_sender(k), then calls await hexgate.audit.shutdown() at the end of an audit-only phase. This clears _senders and sets _closing=True on the usage sender too. Any component still holding the injected usage sender then calls emit(), hits the if self._closing guard, and returns silently — every subsequent usage event is dropped with no error. Silent telemetry loss once usage tracing is wired in.

hexgate/tracing/usage.py
victorludvig: Good catch, fixed for the next revision.

hexgate/tracing/usage.py
victorludvig: Good catch, fixed for the next revision.

victorludvig:
> 2 minor issues to check
> 
> But overall the sender refactor + LLM definition logic looks good.
> 
> You can also check this feedback from claude:
> 
> 1. Unscoped shutdown() drains the shared sender registry — 72/100
> 
> hexgate/tracing/_senders.py:275 (root cause also at hexgate/audit.py:165, hexgate/tracing/usage.py:64) · correctness · CONFIRMED
> 
> shutdown() now drains the entire shared (api_key, path) registry. So hexgate.audit.shutdown() also closes LLM-usage senders (and vice versa) — the two event types are no longer independently teardown-able.
> 
> Failure: a process configures both audit.configure(k) and configure_usage_sender(k), then calls await hexgate.audit.shutdown() at the end of an audit-only phase. This clears _senders and sets _closing=True on the usage sender too. Any component still holding the injected usage sender then calls emit(), hits the if self._closing guard, and returns silently — every subsequent usage event is dropped with no error. Silent telemetry loss once usage tracing is wired in.

shutdown() draining the entire registry is the motivation behind design C which merges of both registry into a generalised class: we don't want to force the developer to call shutdown() several times. Draining both registry at once is by design, it's not a bug, and it is documented in the docstring:

```
async def shutdown() -> None:
    """Drain in-flight emits and close every sender in the shared registry
    (decisions and LLM usage alike). Safe to call multiple times."""
    await _shutdown_all()
````

guillaume-hexamind (approved):

PR #74, opened 2026-07-08
feat(platform/api): ban invokation api backend logic

guillaume-hexamind:
## Summary

Backend for the **kill switch** — operator-controlled hard blocks that override policy. A ban refuses execution outright for one **agent** (all runs of that agent) or one **user** (all runs by an end-user across every agent in the project), evaluated at a new invoke-time gate in the SDK *before* the LLM runs. This PR is the API/control-plane + storage; the SDK gate and dashboard page land in later phases.

## Changes

**Data model**
- New `Ban` SQLModel table (`platform/api/hexgate_api/models.py`), project-scoped, `ban_type` `"agent" | "user"`, one target set per row. *Active* = `revoked_at is null`; revoke is a soft delete keeping the who/when audit trail. `ban_` id prefix registered in `core/ids.py`; `BAN_TYPE_*` constants added.
- New ClickHouse `ban_enforcement` table (`platform/clickhouse/init/schema.sql`) — one row per refused execution, sibling of `policy_decision` on the shared envelope, kept separate (no tool/outcome, and per-attempt volume would swamp the decision feed). 90-day TTL. Also tidies stray blank lines / missing trailing newline in the `llm_invocation` DDL.

**API — `features/bans/`**
- Dashboard CRUD (cookie auth, admin/owner only): `POST` / `GET` / `DELETE /v1/projects/{id}/bans`. Create returns 409 on a duplicate active target; list is active-only unless `?include_revoked`; delete is a soft-revoke returning 204.
- SDK feed (bearer auth): `GET /v1/bans` returns the project's active bans with a `sha256` **ETag** so the invoke-time gate polls cheaply (`304` on no change). `BanFeedEntry` carries `ban_id` (so an enforcement event can link back to its ban) but no `created_by`/timestamps. Decoupled from the agent-policy ETag.
- At-most-one-active-ban-per-target enforced by a service-layer pre-check, mirroring the invitation dedup pattern. **Known v1 tradeoff:** no DB constraint backs it, so two concurrent creates can race into duplicate active bans (fail-safe — over-blocks, never under-blocks); a partial unique index is deferred hardening.

**Audit ingest — `features/audit/`**
- `POST /v1/audit/ban-enforcements` (bearer) ingests a refusal event into `ban_enforcement`; idempotent on `event_id` via ReplacingMergeTree. **Ingest-only in this PR** (`insert_ban_enforcement`); the read side that powers the dashboard's "recent blocked attempts" lands with its consumer in a later phase. `policy_decision` and the tool-decision Audit views are untouched.

**Schemas**
- `BanCreate` (with a `model_validator` enforcing target matches `ban_type`), `BanRead`, `BanFeedEntry`, `BanEnforcementEvent`, `BanEnforcementAccepted`.

**Wiring & deps**
- `bans_router` mounted under `/v1` in `main.py`.
- `uv.lock`: incidental re-sync to root `pyproject.toml` — bumps the editable `hexgate` dep `0.2.5 → 0.2.7` and lists the `mcp>=1.0` dependency (fixes pre-existing lock drift on `main`; no new package versions introduced).

## Tests
- `tests/features/bans/test_bans.py` (new) covering CRUD, auth gating, conflict/not-found, and the ETag feed.
- `tests/features/audit/test_audit.py` extended for ban-enforcement ingest.


victorludvig:
I had a first look and this looks good, here is 1 comment by Claude which seems nit but can you check it please:

2. Score: 60 — BAN_TYPE_AGENT/BAN_TYPE_USER added to constants.py, outside its documented scope
platform/api/CLAUDE.md describes constants.py as "shared seed identity (DEFAULT_*) + role names (ROLE_*)." This PR adds two feature-specific enum values there instead:
https://github.com/HexamindOrganisation/hexgate/blob/6ce6c435673957ddc824aaa396365bc4a25b3fe1/platform/api/hexgate_api/constants.py#L40-L45

victorludvig (approved):

guillaume-hexamind:
> I had a first look and this looks good, here is 1 comment by Claude which seems nit but can you check it please:
> 
> 2. Score: 60 — BAN_TYPE_AGENT/BAN_TYPE_USER added to constants.py, outside its documented scope
>    platform/api/CLAUDE.md describes constants.py as "shared seed identity (DEFAULT__) + role names (ROLE__)." This PR adds two feature-specific enum values there instead:
>    https://github.com/HexamindOrganisation/hexgate/blob/6ce6c435673957ddc824aaa396365bc4a25b3fe1/platform/api/hexgate_api/constants.py#L40-L45

Thanks, I moved the constants directly into bans/service.py!

PR #75, opened 2026-07-08
docs: marimo demo for gating an MCP server with policy

quanghexa94:
A little marimo notebook to show off MCP gating — how a third-party MCP server's
tools get run through the same policy engine + DSL that guards native tools.

**Run it:** `marimo edit deploy/mcp_gate_demo.py`

It's fully self-contained — spawns a tiny FastMCP server over stdio, no external
services, no LLM key.

**What it walks through**
- Connect to the server and auto-list its tools under `mcp-<server>-<tool>`
- Gate them with a `default-deny` policy + allowlist + per-argument constraints
- Fire calls and watch the gate decide — allowed calls actually run against the
  server, denied ones never reach it

The nice bit: `compute_tip(amount=42)` is allowed but `compute_tip(amount=5000)`
is denied — same tool, split by a DSL constraint on the argument. And `read_secret`
is blocked even though the server offers it.

**Files**
- `deploy/mcp_gate_demo.py` — the notebook
- `deploy/_mcp_demo_server.py` — the demo MCP server it spawns

### DIFF
--- a/deploy/_mcp_demo_server.py
+++ b/deploy/_mcp_demo_server.py
@@ -1,0 +1,38 @@
+"""Tiny FastMCP server for the MCP-gate marimo demo (deploy/mcp_gate_demo.py).
+
+Kept as its own file (not inline in the notebook) so marimo's file-format
+round-tripping can't strip it. The notebook spawns it over stdio:
+
+    python deploy/_mcp_demo_server.py
+
+Exposes three tools chosen to map cleanly onto policy outcomes:
+  * compute_tip   — pure math, safe                → allow (arg-gated)
+  * read_secret   — pretends to read a secret      → deny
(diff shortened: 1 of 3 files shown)


deploy/_mcp_demo_server.py
```diff
@@ -0,0 +1,38 @@
+"""Tiny FastMCP server for the MCP-gate marimo demo (deploy/mcp_gate_demo.py).
```
victorludvig, line 1: Should we add a link in the README or in the [Quickstart](https://docs.hexgate.ai/quickstart) to this notebook and the other one?

quanghexa94: i've reworked a bit the documentation page on mcp to add the instruction on how to launch the interactive mcp demo.

victorludvig (approved):

PR #76, opened 2026-07-09
feat(hexgate): add llm usage hook

victorludvig:
## What is changing
Adds a post-call hook to each of the four LangChain/OpenAI/Google/Pydantic AI
adapters (BYO-graph entry points) that sends input/output token counts to the
backend. Also wires the same hook into the manifest-driven `HexgateAgent`
(hexgate/agents/factory.py) — the class hexgate serve/the Playground actually
build agents with — since it's a separate integration point the four adapters
don't cover.

The LangChain hook's on_llm_end is async, not a plain sync method, so
LangChain awaits it inline instead of dispatching it to a thread pool with no
event loop — verified live in the playground that a sync callback here would
silently drop every event. Model name is read from response_metadata rather
than llm_output, since llm_output aggregates to None for streamed responses
(also confirmed live).

## Tests
- `make check`
- Ran the Hexgate agent in the playground and confirmed a row landed in the
  ClickHouse `llm_invocation` table (this is what caught the two bugs above)

## Note (not fixed in this PR)

Verifying this live surfaced a deeper issue in the shared sender
infrastructure both decisions and usage sit on: AuditSender.emit() silently
drops events (no exception, no retry) when called with no event loop
anywhere in the process — confirmed reproducible via HexgatePydanticAgent's
run_sync(). This isn't specific to usage tracking or to this PR's diff (it
predates it, in _senders.py from #73), and likely also affects policy
decisions under the same conditions. Written up here, needs a team decision
on the fix before it's actionable: https://app.notion.com/p/Bug-AuditSender-silently-drops-events-with-no-event-loop-shared-decisions-usage-infra-398fb45dbae281238de1ef0e664297b9?source=copy_link


victorludvig:
### Code review

Found 3 issues:

1. Sync `run()` closes its event loop before the fire-and-forget usage/audit send from the last event can complete, silently dropping it (confidence: 78/100)

This is a distinct trigger for the same disclosed-but-unfixed issue in the PR description ("AuditSender.emit() silently drops events... when called with no event loop anywhere in the process"). That note covers the *no-loop-at-all* case (reproduced via `run_sync()`); this is a different mechanism — a loop *does* exist and the send task *is* scheduled via `asyncio.create_task()` in `AuditSender._spawn_send()`, but `HexgateRunner.run()`'s per-call loop is torn down right after `agen.aclose()`, before the pending send has a chance to finish. `HexgateUsagePlugin` (wired into every `HexgateRunner` by this PR) now rides the same at-risk path for usage events, on top of the pre-existing audit/policy events.

https://github.com/HexamindOrganisation/hexgate/blob/b2ebd773789d6a58fd4ccf382d6784bd71fda632/hexgate/adapters/google/runner.py#L103-L114

2. `HexgateRunner` now unconditionally passes `plugins=` to Google ADK's `Runner`, which is deprecated in the installed `google-adk` and raises `ValueError` if a caller also passes `app=` via `runner_kwargs` (confidence: 68/100)

Before this PR, `plugins` was only forwarded if a caller explicitly supplied it. Now a `plugins` list (with `HexgateUsagePlugin` appended) is always built and passed, which triggers a `DeprecationWarning` on every `HexgateRunner` construction and would raise if a caller passes `app=` instead of `agent=` via `runner_kwargs` — a code path that isn't currently exercised in this repo but is a valid `Runner` usage pattern per its own docstring.

https://github.com/HexamindOrganisation/hexgate/blob/b2ebd773789d6a58fd4ccf382d6784bd71fda632/hexgate/adapters/google/runner.py#L48-L57

3. `iter()` emits a usage event unconditionally, even if the caller exits before the run completes, contradicting `emit_run_usage`'s own "completed run" contract (confidence: 77/100)

`emit_run_usage`'s docstring says it emits "one aggregate LlmUsageEvent for a **completed** pydantic_ai run," and the module docstring says usage is read "after the call completes." But `iter()` calls it right after `yield run` resumes, regardless of whether the caller actually drove the graph to completion. pydantic_ai's `AgentRun.usage()` is explicitly partial ("usage statistics for the run so far"), so an early exit from the `async for node in run:` loop (a legitimate, common pattern) results in an emitted event with `input_tokens=0, output_tokens=0` for a run that never finished. `run_stream()` doesn't have this problem since pydantic_ai auto-drains it on `__aexit__`; `iter()` has no equivalent.

https://github.com/HexamindOrganisation/hexgate/blob/b2ebd773789d6a58fd4ccf382d6784bd71fda632/hexgate/adapters/pydantic_ai/agent.py#L124-L138

Docstring being contradicted:
https://github.com/HexamindOrganisation/hexgate/blob/b2ebd773789d6a58fd4ccf382d6784bd71fda632/hexgate/adapters/pydantic_ai/usage.py#L31-L40

victorludvig:
> ### Code review
> Found 3 issues:
> 
> 1. Sync `run()` closes its event loop before the fire-and-forget usage/audit send from the last event can complete, silently dropping it (confidence: 78/100)
> 
> This is a distinct trigger for the same disclosed-but-unfixed issue in the PR description ("AuditSender.emit() silently drops events... when called with no event loop anywhere in the process"). That note covers the _no-loop-at-all_ case (reproduced via `run_sync()`); this is a different mechanism — a loop _does_ exist and the send task _is_ scheduled via `asyncio.create_task()` in `AuditSender._spawn_send()`, but `HexgateRunner.run()`'s per-call loop is torn down right after `agen.aclose()`, before the pending send has a chance to finish. `HexgateUsagePlugin` (wired into every `HexgateRunner` by this PR) now rides the same at-risk path for usage events, on top of the pre-existing audit/policy events.
> 
> https://github.com/HexamindOrganisation/hexgate/blob/b2ebd773789d6a58fd4ccf382d6784bd71fda632/hexgate/adapters/google/runner.py#L103-L114
> 
> 2. `HexgateRunner` now unconditionally passes `plugins=` to Google ADK's `Runner`, which is deprecated in the installed `google-adk` and raises `ValueError` if a caller also passes `app=` via `runner_kwargs` (confidence: 68/100)
> 
> Before this PR, `plugins` was only forwarded if a caller explicitly supplied it. Now a `plugins` list (with `HexgateUsagePlugin` appended) is always built and passed, which triggers a `DeprecationWarning` on every `HexgateRunner` construction and would raise if a caller passes `app=` instead of `agent=` via `runner_kwargs` — a code path that isn't currently exercised in this repo but is a valid `Runner` usage pattern per its own docstring.
> 
> https://github.com/HexamindOrganisation/hexgate/blob/b2ebd773789d6a58fd4ccf382d6784bd71fda632/hexgate/adapters/google/runner.py#L48-L57
> 
> 3. `iter()` emits a usage event unconditionally, even if the caller exits before the run completes, contradicting `emit_run_usage`'s own "completed run" contract (confidence: 77/100)
> 
> `emit_run_usage`'s docstring says it emits "one aggregate LlmUsageEvent for a **completed** pydantic_ai run," and the module docstring says usage is read "after the call completes." But `iter()` calls it right after `yield run` resumes, regardless of whether the caller actually drove the graph to completion. pydantic_ai's `AgentRun.usage()` is explicitly partial ("usage statistics for the run so far"), so an early exit from the `async for node in run:` loop (a legitimate, common pattern) results in an emitted event with `input_tokens=0, output_tokens=0` for a run that never finished. `run_stream()` doesn't have this problem since pydantic_ai auto-drains it on `__aexit__`; `iter()` has no equivalent.
> 
> https://github.com/HexamindOrganisation/hexgate/blob/b2ebd773789d6a58fd4ccf382d6784bd71fda632/hexgate/adapters/pydantic_ai/agent.py#L124-L138
> 
> Docstring being contradicted:
> 
> https://github.com/HexamindOrganisation/hexgate/blob/b2ebd773789d6a58fd4ccf382d6784bd71fda632/hexgate/adapters/pydantic_ai/usage.py#L31-L40
> 
> 
> - If this code review was useful, please react with 👍. Otherwise, react with 👎.

1) This potentially affects policy decisions as well, so it will be fixed in a follow-up PR

2) Fixed by creating an App object with the params and passing it to the Runner object.

3) Fixed

hexgate/adapters/pydantic_ai/usage.py
```diff
@@ -0,0 +1,50 @@
...
+from __future__ import annotations
+
+from typing import Any
+
+from pydantic_ai import Agent
+
+from hexgate.tracing.usage import emit_llm_usage
+
+
+def _agent_model_name(agent: Agent) -> str:
```
guillaume-hexamind: manifest/pydantic_ai.py already has an _extract_model that does the same thing -> we should find a way to only define the logic once and import it one way or the other

hexgate/agents/factory.py
```diff
@@ -350,6 +356,16 @@ def __init__(
         # load_hexgate_agent attaches it for lazy cloud-side attenuation.
         self._binding: PolicyBinding | None = binding
         self.hexgate_client: HexgateClient | None = hexgate_client
+        self._usage_handler = HexgateUsageCallbackHandler(agent_name=name or "default")
```
guillaume-hexamind, line 370: You need to resolve the api key (similar to what is being done on adapters/langchain/wrapper.py ) and instantiate the callback handler with it, so that events can be emmited

hexgate/adapters/pydantic_ai/agent.py
```diff
@@ -127,6 +132,9 @@ async def run_stream(
         async with self._abind(user, "run_stream"):
             async with self._agent.run_stream(*args, **kwargs) as result:
                 yield result
+                emit_run_usage(
```
guillaume-hexamind: run_stream emits usage unconditionally after the caller's async with block, with no completion guard. If a caller consumes only part of the stream (or breaks early), StreamedRunResult.usage() reflects only tokens streamed so far, output_tokens is typically 0 until the stream is fully drained.

pydantic_ai's StreamedRunResult exposes exactly the needed signal — is_complete (result.py:402, set true in _marked_completed once the stream is fully drained). So the following should probably work:

async with self._agent.run_stream(*args, **kwargs) as result:
    yield result
    if result.is_complete:
        emit_run_usage(self._agent_name, self._agent, result, api_key=self._api_key)

hexgate/adapters/pydantic_ai/usage.py
```diff
@@ -0,0 +1,50 @@
...
+    ruled out."""
+    model = agent.model
+    if model is None:
+        return ""
+    if isinstance(model, str):
+        return model
+    return model.model_name
+
+
+def emit_run_usage(agent_name: str, agent: Agent, result: Any, *, api_key: str) -> None:
```
guillaume-hexamind, line 17: Non blocking, but here, and actually most cases of llm usage emition; it is not guarded by a try / except block, so an error during info extraction or emission will actually make the agent fail. I am not sure what would be the best course of action, as making the fail silent also has some real cons; I guess we can keep it as, but we have to keep the design issue in mind

hexgate/adapters/google/runner.py
```diff
@@ -51,9 +53,11 @@ def __init__(
             approval_handler=approval_handler,
             client=client,
         )
+        plugins = list(runner_kwargs.pop("plugins", None) or [])
+        plugins.append(HexgateUsagePlugin(api_key=self.api_key))
+        app = App(name=app_name, root_agent=self._wrapped_agent, plugins=plugins)
```
guillaume-hexamind, line 58: Can you check which versions of google adk support App as Runner argument, and make sure that the current lowest pyproject supported version ( google-adk>=1.0 ) works with it, or bump the version to ensure compatibility

guillaume-hexamind (commented):
A couple of comments to check, but the logic seems good overall

hexgate/adapters/pydantic_ai/usage.py
victorludvig: Fixed by making _extract_model from manifest/pydantic_ai.py public and importing it in [hexgate/adapters/pydantic_ai/usage.py](https://github.com/HexamindOrganisation/hexgate/pull/76/files/1f6b96148c110a0050ffca0e0b4c2c7ac98a1ae4#diff-726d279a511e31f89b1fdbd5150a753a674c1f0981218a9a22490ff5e2518f8c).

hexgate/agents/factory.py
victorludvig: I was a bit confused because I tested the changes in the playground before publishing the PR and token counts were sent and saved in the llm invocation table. I doubled check and there is no need to pass an API key to the HexgateAgent, which resolves it from the environment variable:

HexgateUsageCallbackHandler(api_key=None) → emit_llm_usage(..., api_key=None) → configure_usage_sender(None) → get_or_create_sender(path, None, ...), which falls back to HEXGATE_API_KEY from env at emit time with _senders.py:221-238 which calls: https://github.com/HexamindOrganisation/hexgate/blob/4a27c3f1e9e80da3ef21704a974516d5592a349d/hexgate/config/env.py#L23-L30

So as long as HEXGATE_API_KEY is set, usage events still get sent. I added this comment before the handler instantiation:
```
# api_key intentionally omitted — create_agent has no explicit api_key param
# (only hexgate_client, attached post-init by _bind_policy). If one is added,
# thread it through here too, or usage events will keep silently resolving
# from HEXGATE_API_KEY instead of the caller's explicit key.
```

guillaume-hexamind: Indeed the fault is on me, only the adapters can pass an explicit api key that can override the env var as of now

hexgate/adapters/pydantic_ai/agent.py
victorludvig: I fixed it, and added a comment to exemplify a use-case:

```python
...
yield result
# Emit usage only if the run completed, not if the caller aborted mid-stream,
# because the usage counts from pydantic's side are 0 until the run completes.
# This can happen if a user cancels a LLM request mid-response: we will never
# know the total number of input / output tokens, however they are still charged by the LLM provider. 
# This is a known limitation of pydantic_ai's usage reporting, and we will not be able to report usage in this case.
if result.is_complete:
    emit_run_usage(
        self._agent_name, self._agent, result, api_key=self._api_key
    )
```

hexgate/adapters/pydantic_ai/usage.py
victorludvig: I think it is worth adding a try/except block with logs because we don't want the agent exception to fail because the token counts couldn't be extracted. I will make the change for each adapter in a follow-up PR.

hexgate/adapters/google/runner.py
victorludvig: Good catch, the minimum version is google-adk>=1.14, so I increased it in pyproject.toml.

guillaume-hexamind (approved):
Good to me!

PR #77, opened 2026-07-09
feat(hexgate): Agent or User Ban Gate

guillaume-hexamind:
## Summary

Adds the **SDK side of the kill-switch ban feature**: an invoke-time gate that refuses a banned agent or user *before the LLM runs*, independent of per-tool `PolicyEnforcer.decide()`. Bans are project-scoped and pulled from the platform `GET /v1/bans` feed; refusals are reported back via a ban-enforcement audit sender.

## What's new

**`hexgate/security/bans.py`** — the ban primitive:
- `BanEntry` / `BanSet` — immutable snapshot of active bans, indexed by agent name and user id for O(1) lookup. Entries with an unknown `ban_type` or blank target are dropped and logged (never enforced silently).
- `PlatformBanSource` — ETag/304-cached fetch of `/v1/bans` under a lock, shared **per api-key** (the feed is project-wide, so agents share one cache) via a module registry `get_ban_source`.
- `BanGate` — per-agent gate: refresh (**fail-soft**, degrades to last-good / `EMPTY` on error) → check agent ban, then user ban → emit enforcement event → raise `AgentBannedError`. Sync (`check`) and async (`check_async`, fetch off-loop / emit on-loop) paths.
- `BanEnforcementEvent` + `configure_ban_sink` — best-effort refusal reporting over the shared `AuditSender` registry (`/v1/audit/ban-enforcements`), drained by the existing `audit.shutdown`.
- `resolve_ban_gate` — builds the gate, or returns `None` in local mode / no key (caller skips the check).

**`hexgate/security/errors.py`** — `AgentBannedError`, enriched with `code` / `ban_type` / `target` / `reason` and a default localized `user_message`.

**`hexgate/cloud/client.py`** — `HexgateClient.get_bans()` fetches the `/v1/bans` array with ETag support.

**Adapter wiring** — the gate is checked before each run/stream entrypoint across all four adapters (Google ADK, LangChain, OpenAI Agents, Pydantic AI). Each wrapper now builds a **single shared `HexgateClient`** used by both the policy and ban resolvers, avoiding a second biscuit-verify + JWKS round-trip per wrapped agent.

## Design notes
- **Fail-soft by default** (fail-closed is a later opt-in): a control-plane blip must never crash a run. A ban that never fetched degrades to `EMPTY`, not to a prior restrictive state.
- Agent ban is checked before user ban so a coincident ban emits a deterministic `ban_type` / `ban_id`.
- Malformed `/v1/bans` bodies raise `BanContentError` (contract drift, logged loudly) rather than silently yielding no bans.

## Tests
New `tests/security/test_bans.py` plus per-adapter coverage for gate wiring, and `tests/cloud/test_client.py` for `get_bans`.


winlp4ever:
Can you explain where the ban mechanism intervenes in our gate architecture? A schema will help. Also if possible add a documentation page on explaning banning mechanism as a feature is very helpful

quanghexa94:
It's possible that I misunderstand a bit but the ban mechanism is a bit against the arguments of the library that there's no ping pong with our platform or at least the ping pong is in asynchronous (not blocking). Here the ban is a request to platform at every invocation (logical, but again, partially destroy the purpose of having local policy enforcement) so it violates a bit the architecture direction we wanted that is rules are enforced locally to avoid performance issues (hot path blocking)

guillaume-hexamind:
> Can you explain where the ban mechanism intervenes in our gate architecture? A schema will help. Also if possible add a documentation page on explaning banning mechanism as a feature is very helpful

This PR 2/3 on ban feature, the first one defining postgres + clickhouse schema, and platform endpoints is already merged.

Ban happens for a specific agent at runtime, either because of the agent being 'banned', or because of the user_id of enduser being banned; preventing the agent to run this specific query. Here are 2 schemas:

Here are two diagrams:

**1. Per-run flow — where the ban gate intervenes**
```
  Request handler:  async with User(user_id=…, role=…):
        │
        ▼
  ┌──────────────────────────────────┐
  │ binding.refresh[_async]()        │   policy pull – ETag/304, fail-soft
  └──────────────────────────────────┘
        │
        ▼
  ┌────────────────────────────────────────────────┐
  │  BanGate.check[_async](user)                   │  ← KILL SWITCH
  │   1. bans = source.fetch()  (fail-soft)        │     invoke-time gate,
  │   2. agent_ban(agent_name)?  ← checked first   │     BEFORE the LLM,
  │   3. user_ban(user.user_id)?                   │     independent of policy
  └────────────────────────────────────────────────┘
        │
   hit? ├─ yes ──►  emit ban_enforcement (fire-and-forget)
        │           └──► raise AgentBannedError
        │                  • invoke : raised to caller  → run never starts
        │                  • stream : raised before the first chunk
        │
        no
        ▼
  ┌──────────────────────────────────┐
  │  LLM / agent loop runs           │
  └──────────────────────────────────┘
        │  (on every tool call)
        ▼
  ┌────────────────────────────────────────────────┐
  │  PolicyEnforcer.decide(tool, args)  = GATE 1   │  ← per-tool policy
  │   → allow / deny / needs_approval              │     (UNCHANGED by bans)
  └────────────────────────────────────────────────┘
        │
        ▼
   tool result → LLM → response

  Seam is identical across all 5 run surfaces:
  LangChain · OpenAI · pydantic_ai · Google adapters + native HexgateAgent
  (ainvoke/astream_events).  No gate when there's no platform.
  HEXGATE_LOCAL_MODE / HEXGATE_LOCAL_POLICY / no key → BanGate is a no-op.
```

**2. Data flow — control plane → SDK**
```
  CONTROL PLANE (platform)                         SDK (per process)
  ─────────────────────────                        ─────────────────────────────
  ┌──────────────────────────┐
  │ Ban table (Postgres)     │  operator bans an
  │  ban_type: agent | user  │  agent or a user_id
  └─────────────┬────────────┘
                │  GET /v1/bans   (bearer, project-scoped, ETag)
                ▼
         list[BanFeedEntry]  (ban_id, ban_type, target_*, reason)
                │
                ▼
   ┌─────────────────────────────┐   one per (api_key, base_url)
   │ PlatformBanSource.fetch()   │   ETag/304 cache, shared across agents
   └─────────────┬───────────────┘
                 │  BanSet { _by_agent{…}, _by_user{…} }
                 ▼
   ┌─────────────────────────────┐   per agent (carries agent_name);
   │ BanGate.check(user)         │   fail-soft → last-good on error
   └─────────────┬───────────────┘
                 │
         hit ────┼───►  raise AgentBannedError  (agent_banned | user_banned)
                 │
                 └───►  POST /v1/audit/ban-enforcements  (fire-and-forget)
                                │
                                ▼
                       ban_enforcement table (ClickHouse)
```

guillaume-hexamind:
> It's possible that I misunderstand a bit but the ban mechanism is a bit against the arguments of the library that there's no ping pong with our platform or at least the ping pong is in asynchronous (not blocking). Here the ban is a request to platform at every invocation (logical, but again, partially destroy the purpose of having local policy enforcement) so it violates a bit the architecture direction we wanted that is rules are enforced locally to avoid performance issues (hot path blocking)

I am still working on my internal documentation, but would rather commit it after PR3 so that the full documentation ends up up to date if that is ok for you

quanghexa94:
Ok it's clearer thank you

guillaume-hexamind:
> It's possible that I misunderstand a bit but the ban mechanism is a bit against the arguments of the library that there's no ping pong with our platform or at least the ping pong is in asynchronous (not blocking). Here the ban is a request to platform at every invocation (logical, but again, partially destroy the purpose of having local policy enforcement) so it violates a bit the architecture direction we wanted that is rules are enforced locally to avoid performance issues (hot path blocking)

I guess it is not really a new pattern — it's the same one policy already uses, and local enforcement still holds.

The key distinction: the decision is local, only the refresh talks to the platform. A ban check is just a dict lookup on a cached set, same as a policy decision is a local WASM eval on a cached bundle — neither hits the network to decide. And the refresh is the exact hot-reload seam we already run every invocation: binding.refresh() does a conditional GET /v1/agents/{name}, the ban gate does a conditional GET /v1/bans right after. Both are ETag/304 (unchanged → tiny bodyless response, keep the cache), both fail-soft, and on async paths both run off the event loop. So bans are no worse than policy — actually lighter, since the feed is shared per key across all agents and the check is a dict get, not a WASM eval.

What bans do add is a second round-trip per run, on a separate feed. That's intentional: user bans are cross-agent so they can't ride on one agent's policy response, and a kill switch wants fresher propagation than policy anyway — an operator toggling it mid-incident expects it to bite fast.

winlp4ever (commented):
Nice PR overall. The primitive in bans.py is well-shaped (ETag/304, sync/async parity, fail-soft explicit). The findings below are mostly wiring: a few agent-invocation paths reach the LLM without going through the gate, so the mechanism can be sidestepped by picking a different entry point rather than by attacking the platform. Not blockers, but good to close before this scales.

One soft naming nit while we're here. "Kill switch" in the PR body and docstrings implies a hard cut, and fail-soft cannot really deliver that. Given the discussion elsewhere on this being closer to a platform ops feature than a security guarantee (and fail-soft is the right choice for a platform feature), might be worth easing the branding toward "ban feed" or "live refusal list" in customer-facing copy later. No urgency, current names in code (`bans.py`, `AgentBannedError`, `BanGate`) are already factual and good.

hexgate/agents/factory.py
```diff
@@ -350,6 +352,9 @@ def __init__(
         # load_hexgate_agent attaches it for lazy cloud-side attenuation.
         self._binding: PolicyBinding | None = binding
         self.hexgate_client: HexgateClient | None = hexgate_client
+        # Kill-switch gate; attached by load_hexgate_agent (platform path only —
+        # no gate without a control plane). Threaded through with_tools rebuilds.
+        self._ban_gate: BanGate | None = ban_gate
```
winlp4ever, line 359: Attaching the gate here is fine, but note that `self.graph` at line 333 is a public attribute. A caller can reach `agent.graph.ainvoke(...)` directly, which skips the wrapper on `ainvoke` at 359 and therefore skips both `_refresh_policy_safely` and this `_check_ban`. Renaming to `self._graph` and only exposing the wrapped methods would close the bypass. Same story for `astream_events`.

hexgate/adapters/openai/runner.py
```diff
@@ -105,6 +121,9 @@ async def run(
         self._setup_observability()
         binding = self._binding_for(agent)
         await binding.refresh_async()  # per-run policy pull; 304 when unchanged
+        ban_gate = self._ban_gate_for(agent)
+        if ban_gate is not None:
+            await ban_gate.check_async(user)
```
winlp4ever, line 126: The gate fires once for the agent passed to `run`, but OpenAI Agents SDK swaps `Agent` inside a single `Runner.run()` when handoffs kick in, and those sub-agents never get their own `_ban_gate_for()` lookup. If `orchestrator_agent` hands off to `finance_bot` and `finance_bot` is banned, it runs anyway inside the same call. Either attach the gate at the `Agent` level (each `Agent` carries its own gate that fires on entry), or hook into whatever handoff callback the SDK exposes so you can recheck.

hexgate/adapters/google/runner.py
```diff
@@ -52,6 +58,9 @@ def __init__(
             **runner_kwargs,
         )
         self._agent_name = getattr(agent, "name", "default")
+        self._ban_gate = resolve_ban_gate(
```
winlp4ever, line 61: Same shape as the OpenAI handoff issue. `self._agent_name` (line above) gets captured once at construction time from the top-level agent, and the gate resolved here is only for that one name. ADK routes into `BaseAgent.sub_agents` inside the same run without triggering another lookup. A banned sub-agent in the tree runs. Probably wants a per-agent gate resolved lazily as sub_agents fire.

hexgate/adapters/langchain/agent.py
```diff
@@ -142,6 +158,7 @@ async def astream_events(
     ) -> AsyncIterator[dict[str, Any]]:
         """Stream the agent events asynchronously inside a User scope."""
         await self._refresh_async()
+        await self._check_ban_async(user)
```
winlp4ever, line 161: You wrap five methods explicitly (invoke, ainvoke, stream, astream, astream_events) and place `_check_ban_async` at the top of each. But `HexgateLangchainAgent` also has `__getattr__` at line 172 delegating unknown attributes straight to the wrapped graph, so anything else `Runnable` exposes (batch, abatch, astream_log, batch_as_completed) reaches the model without a ban check. A `NoOpBanGate` with an unconditional check inside a base class would cover the `__getattr__` path in one shot, instead of enumerating methods, and the null-guard on the gate also disappears.

hexgate/adapters/pydantic_ai/agent.py
```diff
@@ -122,6 +137,7 @@ async def iter(
     ) -> AsyncIterator[AgentRun[Any, Any]]:
         """Iterate over the agent execution graph asynchronously."""
         await self._refresh_async()
+        await self._check_ban_async(user)
```
winlp4ever, line 140: Same shape as the LangChain agent. Five methods are wrapped explicitly and call `_check_ban_async` at the top, but `__getattr__` at line 145 delegates unknown attributes to the underlying `pydantic_ai.Agent`. So `to_a2a()`, `to_ag_ui()`, and anything else the wrapped Agent exposes go straight to the model without a ban check. If someone builds a FastAPI mount off `to_a2a().handle_request(...)`, the ban is not enforced.

hexgate/security/bans.py
```diff
@@ -0,0 +1,325 @@
...
+    def __init__(
+        self,
+        agent_name: str,
+        source: BanSource | None,
+        sink: AuditSender | None = None,
+    ) -> None:
+        self._agent_name = agent_name
+        self._source = source
+        self._sink = sink
+        self._last_good = EMPTY_BAN_SET
```
winlp4ever: `_last_good` starts at `EMPTY_BAN_SET`. If the very first fetch raises (platform down when the SDK boots), `_current` returns EMPTY and the caller allows. From ops' point of view there is no way to distinguish that from "no bans configured", since the audit stream is silent (no hit means no event). Design-consistent with fail-soft, so not a bug in itself. Given we agreed this is closer to a platform ops feature, might be worth emitting a `gate_degraded` event when `_current` falls back to EMPTY, so the dashboard can distinguish "zero refusals because nothing violated" from "zero refusals because SDKs never fetched".

hexgate/security/bans.py
```diff
@@ -0,0 +1,325 @@
...
+        if self._source is None:
+            return EMPTY_BAN_SET
+        try:
+            self._last_good = self._source.fetch()
+        except BanContentError as exc:
+            # Contract drift, not a blip — log loudly, like PolicyContentError.
+            logger.error("ban feed rejected; using last-good: %s", exc)
+        except Exception as exc:  # noqa: BLE001 — transient; keep last-good
+            logger.warning("ban refresh failed; using last-good: %s", exc)
+        return self._last_good
```
winlp4ever: The source itself (`PlatformBanSource._cached` at 136) already holds the last successful set for all gates on the same api_key. `BanGate._last_good` is a second cache one layer up. On fetch failure, `_current` returns the gate's copy, not the source's. So if gate A boots at 09:00 and successfully caches a ban on `evil-bot`, then gate B is constructed at 09:05 during a network blip and its very first fetch raises, gate B's `_last_good` stays at EMPTY even though the shared source already has the good data. Enforcement ends up depending on which gate initialized first. Dropping `BanGate._last_good` entirely and letting the source own last-good would fix it.

hexgate/security/bans.py
```diff
@@ -0,0 +1,325 @@
...
+            "agent_name": self.agent_name,
+            "user_id": self.user_id,
+            "session_id": self.session_id,
+            "ban_type": self.ban_type,
+            "ban_id": self.ban_id,
+            "reason": self.reason or "",
+        }
+
+
+def configure_ban_sink(
```
winlp4ever, line 213: The `AuditSender` is loop-bound. It latches on the loop that built it. `configure_ban_sink` gets called from `resolve_ban_gate`, which itself runs at wrap time from `wrap_langchain_agent`, `wrap_pydantic_agent`, `HexgateRunner.__init__`. If any of those runs at module import (no running loop), the sender's fallback path kicks in on later sync `emit()` calls and the event gets dropped with a one-shot warning. A pure-sync deployment (LangChain `.invoke`, OpenAI `run_sync`) that refuses a banned run would raise `AgentBannedError` fine but never audit the refusal. Might be moot if wrap-time is always inside an event loop in every real deployment, worth checking against the sync-only integration test path.

hexgate/adapters/openai/runner.py
guillaume-hexamind: Real gap, only entry agents are targeted right now. I guess we can simply document this behavior for now as the changes needed could be quite large; and create a specific PR to address the agent handoff logic altogether, as right now handoff is only be seen as a tool execution, or maybe even not be tracked at all: Needs handoff specific PR

hexgate/adapters/google/runner.py
guillaume-hexamind: Same thing -> we need a handoff specific PR

hexgate/security/bans.py
guillaume-hexamind: Confirmed and known -> it is a shared AuditSender no loop behavior, with enforcement itself unaffected. Already noted in _emit -> real fix is a sync-POST fallback in AuditSender, which should be shared-infra for a separate PR.

hexgate/security/bans.py
guillaume-hexamind: Indeed, I moved the last good responsability into the source

hexgate/security/bans.py
guillaume-hexamind: Good idea; but can't be pure sdk (/v1/audit/ban-enforcements validates a BanEnforcementEvent (ban_type/ban_id required)

I guess leaving it as is for now is fine as sdk already logs the degradation, but would be worth implementing later

hexgate/agents/factory.py
guillaume-hexamind: Renamed graph -> _graph and added a note that direct access skips the wrapper.

hexgate/adapters/langchain/agent.py
guillaume-hexamind: Good catch. Documented it -> __getattr__ now notes that only the wrapped run methods enforce the ban gate + User scope. The other methods are less standards entrypoints, which are not wrapped with a user argument as of now, so there is no identity on these methods regardless

hexgate/adapters/pydantic_ai/agent.py
guillaume-hexamind: Agreed, documented the same way. 
 On the fix: to_a2a() is build-once/serve-many, so gating the to_a2a() call would only check the ban at app-construction, not per request.
 We may want as a follow-up make the to_a2a and to_ag_ui fail loud instead rather than silently bypassing

quanghexa94 (approved):

winlp4ever (approved):
LGTM. The two structural fixes (graph → _graph, moving last-good ownership into the source) are clean, and the deferrals all have sound reasoning given the scope. Ship it.

Two things I'd track as follow-ups so they don't drift:

1. Sub-agent handoffs (the openai and google runner comments). Agreed a dedicated PR is the right vehicle here since it needs per-agent gate resolution plus framework-specific hooks into handoff callbacks. Worth an issue so we don't forget.

2. Pydantic AI `to_a2a` in particular. The docstring warning is fine for `batch` / `astream_log` / other rarely-used entry points, but `to_a2a` is the documented production pattern for shipping a pydantic_ai agent as an ASGI app. Anyone following pydantic_ai's own docs silently loses ban enforcement. The real fix (per-request check inside the ASGI seam, not at build time) deserves its own issue rather than only a doc line.

Minor observation on the source-owned last-good refactor: the broad `except Exception` in `fetch()` still swallows programmer-error TypeError / KeyError from a malformed payload and logs it as a transient blip. Same behavior as before, so not a regression, just something to keep in mind if audit rows ever start showing suspicious 'ban refresh failed' warns while the platform reports healthy.

Nice turnaround on the review.

PR #78, opened 2026-07-10
feat(demo): live demo on Daytona + Modal spawner (retire Codespaces)

quanghexa94:
## Why

Replaces the **GitHub Codespaces** live demo with a **one-click marimo demo** that shows off both sides of hexgate: the **library** (write agent code + policy) and the **platform interface** (chat in the dashboard, watch policy decisions stream). No GitHub login, no install — one link, ~1 min to a live sandbox.

## Architecture

```
 Visitor clicks "Launch the live demo"  (docs button)
              │
              ▼
 ┌─────────────────────────────────────────────┐
 │ MODAL  — serverless web endpoint (spawner)   │   always-on door, scales to zero
 │   GET /launch → daytona.create(snapshot)     │
 │   → boot the sandbox → redirect to notebook  │
 └───────────────────────┬─────────────────────┘
                         │
                         ▼
 ┌─────────────────────────────────────────────┐
 │ DAYTONA sandbox  (per visitor, from snapshot)│   deps + dashboard prebaked
 │   ├─ marimo notebook — define tools/agent    │   → the library
 │   └─ platform API + dashboard — chat + policy│   → the interface
 │   auto-stops after 15 min idle → self-deletes│   no standby cost (~6¢/session)
 └─────────────────────────────────────────────┘
```

**Modal** = the always-on door that mints a sandbox on demand. **Daytona** = the disposable room that runs the demo. (Marimo's live WebSocket works through Daytona's proxy — it didn't through Modal's, which is why the demo runs on Daytona and only the tiny HTTP spawner is on Modal.)

## Files

- `deploy/spawner/modal_app.py` — the Modal spawner: `/launch` creates a sandbox + redirects; `/status` polls readiness; global cap + per-IP cooldown.
- `deploy/daytona_snapshot.py` — build/register the Daytona snapshot (deps + dashboard baked); `--force` rebuilds.
- `.github/workflows/daytona-snapshot.yml` — rebuild the snapshot on release (needs a `DAYTONA_API_KEY` repo secret).
- `docs/platform/live-demo.mdx` — trimmed to a single "Launch the live demo" button.
- `deploy/README.md` — deploy steps for the snapshot + spawner.
- `.devcontainer/` — **removed** (Codespaces retired), plus Codespaces references cleaned from `boot.py` / `Makefile`.

## To go live (after merge)

1. Add repo secret `DAYTONA_API_KEY`.
2. `modal secret create daytona DAYTONA_API_KEY=…` → `modal deploy deploy/spawner/modal_app.py`.

Docs button already points at the deployed spawner (`quang-le--hexgate-spawner-web.modal.run/launch`).


deploy/README.md
```diff
@@ -41,19 +41,30 @@ and the loop restarts with the new object; the dashboard picks it up on its next
...
 - `platform/api/main.py` — wires `enable_demo(app)` when `HEXGATE_DEMO=1` (off by default)
 
-The `.devcontainer/` at the repo root makes this run per-visitor in Codespaces
-(`setup.sh` builds it, `start.sh` launches it, ports 2718/8000 are forwarded).
+## Public demo (deploy)
 
-## Run it in your browser (Codespaces)
+The docs "Launch" button hits a Modal spawner that creates a per-visitor Daytona
+sandbox from the snapshot (auto-stops + self-deletes, so no standby cost). Deploy
+from `asianf/`:
```
guillaume-hexamind: asianf -> exists here but also in many other parts of the repo. A specific asianf -> repo-root renaming PR should probably be done at some point?

deploy/spawner/modal_app.py
```diff
@@ -0,0 +1,171 @@
...
+body{{font-family:system-ui;background:#0b0b0f;color:#e5e7eb;display:grid;
+place-items:center;height:100vh;margin:0;text-align:center}}
+.card{{max-width:32rem;padding:2rem}}
+.spin{{width:2.5rem;height:2.5rem;border:3px solid #333;border-top-color:#3b82f6;
+border-radius:50%;animation:s 1s linear infinite;margin:1.5rem auto}}
+@keyframes s{{to{{transform:rotate(360deg)}}}}a{{color:#60a5fa}}</style></head>
+<body><div class=card>{body}</div>{redirect_js}</body></html>"""
+
+    @web_app.get("/launch")
+    def launch(request: Request):
```
guillaume-hexamind: Every sandbox is now created on Hexamind's Daytona account and billed to Hexamind.

The endpoint is discoverable: "Launch demo" button in public docs, and GET / redirects to /launch.

- Any scraper, crawler, or curl loop can create sandboxes.
- A GET request is trivially triggerable: prefetchers, link-preview bots (Slack/Discord/iMessage unfurling), and search crawlers will hit it without a human ever clicking.
- The PER_IP_COOLDOWN_SECONDS and MAX_LIVE_SANDBOXES guardrails are weak, and both live in per-container memory, so they don't hold under Modal autoscaling.

Not sure what the best course of action should be though; here are 4 potential solutions suggested by claude:

```
A. Put a lightweight proof-of-work / challenge in front (best UX-preserving option).
Add a Cloudflare Turnstile or hCaptcha challenge on the docs page. The docs button first solves the challenge, then calls /launch with the token; the spawner verifies the token server-side before daytona.create(). This keeps the "one click, no login" demo feel while blocking bots â which are the real cost driver. This is the standard pattern for public "try it live" demos.

B. Change /launch from GET to POST + a short-lived signed token minted by the docs origin.
The docs page (which you control) mints an HMAC token tied to a timestamp; /launch rejects requests without a valid, unexpired token. Stops naive crawlers/prefetchers immediately. Weaker than A against a determined attacker who scripts the docs page, so ideally combine with A.

C. Enforce a real, shared, account-scoped budget cap as the backstop.
Auth reduces abuse frequency; it does not bound cost. You still need a hard ceiling that actually holds 

D. Defense in depth:
- Move rate-limiting off per-container memory into something shared (Modal Dict, or Daytona's own state via labels â see IssueÂ 2), so autoscaling doesn't multiply your limits.
- Add a daily/hourly launch budget in addition to the concurrent cap, so a slow drip of launches can't rack up cost over time.

My recommendation: A + C for the merge blocker Turnstile on the button plus a cap that fails closed. B is a cheap incremental hardening on top.
```

victorludvig (approved):

deploy/spawner/modal_app.py
```diff
@@ -0,0 +1,171 @@
...
+              (request.client.host if request.client else "")).split(",")[0].strip()
+
+        now = time.time()
+        if ip and now - _last_launch.get(ip, 0) < PER_IP_COOLDOWN_SECONDS:
+            return HTMLResponse(_page(
+                "Slow down",
+                "<h2>One launch at a time 🙂</h2><p>Please wait a few seconds and retry.</p>",
+            ), status_code=429)
+
+        if _running_count() >= MAX_LIVE_SANDBOXES:
```
guillaume-hexamind: _running_count exception returns 0 -> if any issue is raised with daytona api, this condition will be False and launch will always execute

Probably better to do something along the lines of:

```
def _running_count() -> int | None:
    """Live sandbox count, or None if it can't be determined."""
    try:
        return len(daytona.list())
    except Exception:
        return None

count = _running_count()
if count is None or count >= MAX_LIVE_SANDBOXES:
    return HTMLResponse(_page(
        "Demo at capacity",
        "<h2>Demo is temporarily unavailable</h2><p>Please try again in a few minutes.</p>",
    ), status_code=503)
```

guillaume-hexamind: Since daytona.create() is not instant, there is a window where the daytona.list is not updated yet. If 20 new requests arrive at the same time with a 19/20 running_count, we could have 39 spawned sandboxes.

we could find a way to share an atomic counter instead

deploy/spawner/modal_app.py
```diff
@@ -0,0 +1,171 @@
...
+            poll,
+        ))
+
+    @web_app.get("/status")
+    def status(sb: str):
+        try:
+            sandbox = daytona.get(sb)
+            code = ""
+            for attr in ("result", "output", "stdout"):
+                r = sandbox.process.exec(
```
guillaume-hexamind: Should be moved out of the for loop? Here 3 curls will happen.

deploy/daytona_snapshot.py
```diff
@@ -0,0 +1,109 @@
...
+    image = (
+        Image.base("python:3.13-slim-bookworm")
+        .run_commands(
+            "apt-get update && apt-get install -y --no-install-recommends "
+            "git curl ca-certificates gnupg",
+            "curl -fsSL https://deb.nodesource.com/setup_20.x | bash - "
+            "&& apt-get install -y nodejs",
+            "npm install -g pnpm@9",
+            "pip install --no-cache-dir uv",
+            f"git clone --depth 1 {REPO} /app",
```
guillaume-hexamind: this bakes main, not the released code. Action/checkout grabs release ref, but that never will reach the image.

Potential solution below?

- Expose the release tag (e.g. github.event.release.tag_name, falling back to the current SHA for workflow_dispatch) as an env var to the script.
- In daytona_snapshot.py, read that ref and clone it pinned:
REF = os.environ.get("HEXGATE_SNAPSHOT_REF", "main")
...
f"git clone --depth 1 --branch {REF} {REPO} /app",   # tag or branch

guillaume-hexamind (changes requested):
A couple of comments to check

I like the auto-spawning feature on click though

winlp4ever:
Thanks for the review, all of it is addressed now (pushed 3 commits).

**Spawner abuse/cost (#2–#5)** — *harden the spawner*
- `/launch` is POST-only; `GET /` is a landing page with a Launch button, so crawlers/prefetchers/unfurlers that only issue GETs never spawn a sandbox. The docs button points at `<url>/` now.
- The concurrent cap fails **closed** — a Daytona `list()` error returns 503 instead of falling through to an unlimited launch.
- Per-IP cooldown + a new **daily launch budget** live in a shared `modal.Dict`, so the limits hold under Modal autoscaling rather than per-container memory.
- `/status` runs the readiness curl once, not once per attribute.
- Added an optional Cloudflare **Turnstile** hook (off unless `TURNSTILE_SECRET`/`TURNSTILE_SITEKEY` are set on the `daytona` secret) for when this needs to stop scripted abuse, not just bots.

**Snapshot bakes main (#6)** — *bake the release ref*
- The image now clones `HEXGATE_SNAPSHOT_REF` (default `main`), and the workflow pins it to the release tag. One tweak to your suggestion: I used `github.ref_name` rather than the SHA for the `workflow_dispatch` fallback, because `git clone --branch` doesn't accept a bare SHA.

**asianf → repo root (#1)**
- Folded in rather than a separate PR (only 8 files). References now say "the repo root" / plain `.env`, and the path constant is `REPO_ROOT`.

Verified: spawner routes smoke-tested (GET `/` = page, no spawn; GET `/launch` = 405; POST spawns; repeat = 429; `list()` error = 503; at cap = 503), and ruff + compile are clean across every touched file. Re-requesting review 🙏

guillaume-hexamind (approved):

PR #79, opened 2026-07-10
feat(dashboard): agent & user bans dashboard

guillaume-hexamind:
## Summary

Adds the operator-facing **Bans** page (agent + user kill-switch bans) and the control-plane read side it consumes. The SDK ban primitive, invoke-time gate, and per-adapter enforcement landed in earlier PRs (#74 API backend, #77 SDK ban gate); **this PR is the dashboard surface + the API endpoints/reads that back it.**

## Backend (`platform/api`)

- **Ban-enforcement read side** (the one missing API piece): `BanEnforcementRow` / `BanEnforcementPage` schemas, `list_ban_enforcements` (offset-paginated, single-scan `count() OVER ()` total), and `GET /v1/projects/{id}/audit/ban-enforcements` (cookie, project-admin). Reads the dedicated `ban_enforcement` table; `policy_decision` is untouched.
- **`BanRead.created_by_email`** — resolved server-side via a batch `User` lookup for display (null if the account was deleted); the id stays the stable key.
- Tests for the read endpoint (admin gating, pagination contract) and email resolution.

## Frontend (`platform/dashboard`)

- **New `/bans` route + "Bans" nav item.** Active-bans list, create dialog (Agent | User toggle, validation, 409/422/403 handling), revoke with confirm, and a **"Blocked attempts"** feed off the enforcement table with a window selector.
- **Blocked-attempt detail drawer** — click a row for full event detail, mirroring the Audit decision drawer.
- **Cross-page tie-ins** — "Ban user" on Audit anomalies and "Ban agent" on the Agents page, deep-linking into a prefilled create dialog (`?ban_user=` / `?ban_agent=`).
- **Data layer** — `lib/bans.ts` hooks (`useBans` / `useCreateBan` / `useRevokeBan` / `useCanManageBans`) delegating to ban methods on the shared `api` client, so calls go through `request()` (global 401 → sign-in redirect + central error handling).
- Admin-gated throughout (`useCanManageBans`); non-admins are hard-blocked with an "Admins only" notice.
- Component, hook, and tie-in tests.

## Docs

- `docs/internals/ban-enforcement.md` — end-to-end design & logic reference (data model, lifecycle, REST/SDK internals, failure semantics, rationale).

## Notable decisions

- **Offset pagination** on the blocked-attempts feed (the endpoint caps `limit` at 200) so every row is reachable and "Load more" clears at the end.
- Enforcement feed is a **dedicated table kept out of the Audit views**, so tool-decision analytics stay pure and coherent.
- User-facing name is **"Bans"**, not "Kill Switch" — the whole system already speaks *ban* (`Ban`, `/v1/bans`, `AgentBannedError`); "kill switch" is an internal codename only.

## Known limitations (follow-up PRs)

Push invalidation, per-ban TTL/expiry, fail-closed opt-in, signed feed, biscuit-bound `user_id`, and a DB-level uniqueness constraint are known, accepted gaps — see `ban-enforcement.md` §9.


platform/api/hexgate_api/features/audit/router.py
```diff
@@ -265,6 +268,41 @@ async def api_audit_decisions(
     return page
 
 
+# Ban enforcements read on their own admin-gated endpoint (like ban CRUD),
+# kept out of the tool-decision reads above so those views stay pure — a ban
+# is refused before any tool call and never lands in policy_decision.
+@router.get(
+    "/projects/{project_id}/audit/ban-enforcements",
+    response_model=BanEnforcementPage,
```
winlp4ever, line 278: Something to flag while you're surfacing the enforcement feed to operators. The ingest endpoint at line 104 above (`POST /v1/audit/ban-enforcements`) accepts arbitrary `ban_id` / `agent_name` / `user_id` / `reason` from any project bearer, with no cross-check that the referenced ban exists in the `Ban` table or that the reported event is real. Same trust model as the decisions ingest (predates this PR), so it's not a new vulnerability, but this UI is the first surface where forged rows would matter operationally. A leaked or misused SDK bearer could spray fake enforcement rows and drown out real refusals for 90 days (the ClickHouse TTL). Two possible follow-ups: join on `Ban` at ingest and drop rows whose `ban_id` isn't in the project, or badge self-reported rows in the UI so operators know the source is SDK-attested rather than platform-verified.

platform/api/hexgate_api/features/audit/service.py
```diff
@@ -517,6 +517,64 @@ def list_decisions(
...
+    )
+    where_sql = " AND ".join(where)
+
+    # Same one-scan page+total trick as list_decisions (count() OVER () is
+    # computed before LIMIT, so it carries the full match count per row).
+    page_params = {**params, "lim": limit, "off": offset}
+    result = client.query(
+        f"SELECT {_BAN_ENFORCEMENT_LIST_COLUMNS}, count() OVER () AS total_matches "
+        f"FROM ban_enforcement WHERE {where_sql} "
+        "ORDER BY occurred_at DESC LIMIT {lim:UInt32} OFFSET {off:UInt32}",
```
winlp4ever: `ORDER BY occurred_at DESC` alone isn't a total order. `DateTime64(3, 'UTC')` gives millisecond precision and the SDK gate can fire several enforcement events inside one iteration, so tied rows come back in an undefined order across two paginated queries. Offset pagination on the blocked-attempts feed will duplicate or skip rows on any burst. The storage sort key is already `(project_id, occurred_at, event_id)`, so adding `, event_id DESC` here is essentially free and gives a stable total order.

platform/dashboard/src/components/bans/BlockedAttemptsPanel.tsx
```diff
@@ -0,0 +1,144 @@
...
+      const loaded = allPages.reduce((n, p) => n + p.rows.length, 0);
+      if (lastPage.rows.length === 0 || loaded >= lastPage.total) {
+        return undefined;
+      }
+      return loaded;
+    },
+  });
+
+  const rows: BanEnforcementRow[] =
+    feed.data?.pages.flatMap((p) => p.rows) ?? [];
```
winlp4ever: Using accumulated `loaded` as the next offset is stable only if no new rows land between fetches. This is a live feed ordered by `occurred_at DESC`, so any new blocked attempt that arrives between clicking page 1 and page 2 shifts every subsequent offset by one row. React will warn about duplicate keys and the feed will look corrupted on any burst. Pair this with the backend `ORDER BY` tiebreaker (see the `service.py` comment) and either move to cursor pagination (`before_id`) or dedup by `event_id` when flattening pages.

platform/dashboard/src/routes/Bans.tsx
```diff
@@ -0,0 +1,164 @@
...
+    if (!canManage) {
+      clearPrefillParams();
+      return;
+    }
+    setInitial(
+      prefillUser
+        ? { ban_type: "user", target: prefillUser }
+        : { ban_type: "agent", target: prefillAgent as string },
+    );
+    setCreateOpen(true);
```
winlp4ever, line 80: This auto-opens CreateBanDialog with a target the user never chose. The target came from a URL query param, which is attacker-controlled input. An admin who follows a poisoned Slack or email link with `?ban_user=cus_important-customer` sees the dialog prefilled and one click on the destructive primary CTA bans the wrong target. The tie-ins from Audit and Agents are legitimate uses of the prefill; the concern is that URL params are untrusted. Worth showing a `prefilled from URL` banner that requires an explicit confirm, or opening the dialog empty and preselecting the target field so a user still has to type or click it to activate.

platform/dashboard/src/routes/Bans.tsx
```diff
@@ -0,0 +1,164 @@
...
+
+  useEffect(() => {
+    // Wait until the scope (and thus the org role) is resolved so we don't
+    // act on a still-loading `canManage`, then: admins get the prefilled
+    // dialog; a non-admin who followed the link sees AdminRequiredNotice, so
+    // just strip the stale params rather than dangling `createOpen` on a
+    // dialog that never mounts.
+    if (scope.status !== "ready" || (!prefillUser && !prefillAgent)) return;
+    if (!canManage) {
+      clearPrefillParams();
```
winlp4ever, line 72: Race with `useOrgs()`. `scope.status` flips to `ready` immediately, but `useCanManageBans` returns false while the orgs query is still loading. So a legitimate admin landing on `/bans?ban_user=alice` in a fresh tab can hit this branch, `clearPrefillParams()` strips the URL, and by the time `canManage` flips true the prefill is already gone and the dialog never auto-opens. Gate the clear-params branch on the orgs query being resolved (something like `orgs.status === 'ready' && !canManage`), not just on `!canManage`.

platform/dashboard/src/components/bans/CreateBanDialog.tsx
```diff
@@ -0,0 +1,205 @@
...
+
+  // Reset the form whenever the dialog (re)opens, seeding from any prefill.
+  useEffect(() => {
+    if (!open) return;
+    setBanType(initial?.ban_type ?? "agent");
+    setAgentName(initial?.ban_type === "agent" ? (initial.target ?? "") : "");
+    setUserId(initial?.ban_type === "user" ? (initial.target ?? "") : "");
+    setReason(initial?.reason ?? "");
+  }, [open, initial]);
+
```
winlp4ever, line 88: The reset effect keys on `[open, initial]`, but `initial` can change identity while `open` is true if the parent re-derives it (which `Bans.tsx` does when URL params change). Any parent re-render producing a fresh `initial` object wipes whatever the user typed into the dialog. Reset only on the false→true transition of `open`. Two options: track the previous `open` with a `useRef(false)`, or key the effect on `[open]` alone and read the current `initial` via a ref.

platform/api/hexgate_api/features/bans/router.py
```diff
@@ -83,7 +86,10 @@ async def api_list_bans(
     rows = await list_bans(
         session, project_id=project_id, include_revoked=include_revoked
     )
-    return [_ban_read(b) for b in rows]
+    emails = await emails_for_user_ids(session, {b.created_by_user_id for b in rows})
```
winlp4ever, line 89: `list_bans` above is unpaginated, so this line runs `SELECT ... WHERE id IN (<N>)` where N is `every distinct creator id across every ban in the project`. On SQLite (dev/test) N > 999 raises `OperationalError: too many SQL variables`. On Postgres it works but the payload grows unbounded. If bans grow into a bulk denylist (natural extension of the primitive), this endpoint becomes a footgun. Consider paginating `list_bans`, capping the include-revoked branch, or chunking the `IN` query.

platform/api/hexgate_api/schemas.py
```diff
@@ -199,6 +199,9 @@ class BanRead(BaseModel):
     target_user_id: Optional[str]
```
winlp4ever, line 204: `ban_type: str` on the wire is looser than the frontend union (`"agent" | "user"`). Correct in practice today because `BanCreate._check_target` and the ClickHouse Enum8 gate the write path, but any future DB value would slip past OpenAPI / Pydantic and reach the UI as `BanTypeBadge` silently rendering the `user` style (and `ActiveBansPanel`'s ternary picking the wrong id column). Tighten to `Literal["agent", "user"]` here and on `BanEnforcementRow`.

winlp4ever (commented):
Solid PR overall. Backend endpoints are cleanly gated (`require_project_admin` on write and on the enforcement list), Pydantic schemas closely mirror the TypeScript types, and the design doc (`docs/internals/ban-enforcement.md`) is one of the more accurate ones in the repo. I verified every code snippet, endpoint URL, and line reference against the actual code and they all match.

Inline comments cover the main correctness and security concerns. A few things that don't fit inline cleanly:

**Reason field leaks to the banned user via SDK.** `BanFeedEntry.reason` is included in the feed served on `/v1/bans` and propagates into `AgentBannedError.reason`. Naive integrators render that on the banned user's screen, so an admin writing something like `fraud investigation, ticket #1234, PII redacted at counsel@acme` in the reason box may unknowingly leak that string to the person being banned. Two options: strip `reason` from `BanFeedEntry` and keep it dashboard-side only, or add a clearly visible warning in CreateBanDialog that the reason field is user-visible.

**CSRF on write endpoints.** Cookie auth with SameSite=Lax, no CSRF token on `POST /bans` or `DELETE /bans/{id}`. Any XSS on a same-registrable-domain subdomain (docs subdomain, marketing preview host, etc.) compromises the kill switch primitive. Given this is now a destructive surface, a double-submit CSRF token or tightening the cookie to SameSite=Strict on the write routes is worth adding.

**SELECT-then-INSERT race in `_active_ban_for_target`** (`platform/api/hexgate_api/features/bans/service.py:59`). Two admins clicking Create for the same target within milliseconds both pass the `already banned?` check and both insert. Revoke lifts one, the other stays active. `ban-enforcement.md` §9 already lists `DB-level uniqueness constraint` as a known gap. A partial unique index (`WHERE revoked_at IS NULL`) on `(project_id, ban_type, target_agent_name, target_user_id)` closes it.

**422 error garble in `api.ts`.** `String(detail.detail)` on FastAPI's Pydantic error array `[{loc, msg, type}, ...]` yields `"[object Object]"`. CreateBanDialog masks it via status-code remapping, but any other page reading `err.message` on a 422 will show garbage. Not new to this PR but the ban UX will hit the 422 path more than most.

**Missing test cases** (each would catch a real regression):
- `DELETE /bans/{id}` returns 403 for non-admin org member (POST is covered, DELETE isn't).
- `GET /bans` returns 403 for non-admin org member (only non-member is tested).
- `created_by_email` is null after User deletion.
- `limit=500` clamps to 200.
- CreateBanDialog renders a user-facing error on 409/422 and keeps the dialog open (hook-level test only asserts the promise rejects).
- Revoke confirm cancel path does not fire DELETE.

**PR title scope.** `feat(platform):` uses a scope that isn't in CLAUDE.md's list (`platform-api`, `dashboard`, `sdk`, `cli`, `clickhouse`). Since this touches both `platform-api` and `dashboard`, either `feat(dashboard):` with a note about the API read side, or split into two commits with different scopes.

quanghexa94:
Looking back at the PR after a second pass, one thing worth calling out as a follow-up: the frontend test coverage is thin relative to the new surface. 15 frontend source files changed or added, 4 test files touched, and while the hook layer (`bans.test.tsx`) is thorough and `Bans.test.tsx` covers the page-level golden paths, five substantial new components ship without a dedicated test file:

- `CreateBanDialog.tsx` (~250 lines): the primary destructive action in the whole feature. Only tested indirectly via "opens the create dialog from the header CTA". Missing: form validation edges, 409/422 error rendering, the reset-effect race I flagged inline, dropdown agent selection, reason-field visibility.
- `BlockedAttemptsPanel.tsx`: one page test covers "Load more at end", but there's no pagination-stability test that would catch the offset drift I flagged.
- `BlockedAttemptDrawer.tsx`: only open/close is exercised. Escape key handling and focus trap are untested.
- `ActiveBansPanel.tsx`: revoke confirm-and-execute is tested, revoke cancel isn't.
- `BanTypeBadge.tsx`: small enum component, fine without a dedicated test.

All three of the frontend correctness bugs I flagged inline (CreateBanDialog reset-on-rerender, Bans.tsx useOrgs race, BlockedAttemptsPanel pagination drift) live in this gap. Not by coincidence: they're the class of bug that integration-style tests structurally can't catch, because they require:

- parent re-rendering while a child modal is open (reset effect)
- precise timing of a resolved scope + a still-loading orgs query (useOrgs race)
- new data arriving between two paginated fetches (pagination drift)

Minimum viable additions if you don't want to over-invest: `CreateBanDialog.test.tsx` (highest ROI, primary destructive action) and `BlockedAttemptsPanel.test.tsx` (catches the live-feed pagination bug and future regressions on the feed). Plus a couple of extra assertions in `Bans.test.tsx` for the revoke-cancel path and the orgs-loading race.

Not blocking, just flagging so it doesn't get lost under the inline threads.

guillaume-hexamind:
> Solid PR overall. Backend endpoints are cleanly gated (`require_project_admin` on write and on the enforcement list), Pydantic schemas closely mirror the TypeScript types, and the design doc (`docs/internals/ban-enforcement.md`) is one of the more accurate ones in the repo. I verified every code snippet, endpoint URL, and line reference against the actual code and they all match.
> 
> Inline comments cover the main correctness and security concerns. A few things that don't fit inline cleanly:
> 
> **Reason field leaks to the banned user via SDK.** `BanFeedEntry.reason` is included in the feed served on `/v1/bans` and propagates into `AgentBannedError.reason`. Naive integrators render that on the banned user's screen, so an admin writing something like `fraud investigation, ticket #1234, PII redacted at counsel@acme` in the reason box may unknowingly leak that string to the person being banned. Two options: strip `reason` from `BanFeedEntry` and keep it dashboard-side only, or add a clearly visible warning in CreateBanDialog that the reason field is user-visible.
> 
> **CSRF on write endpoints.** Cookie auth with SameSite=Lax, no CSRF token on `POST /bans` or `DELETE /bans/{id}`. Any XSS on a same-registrable-domain subdomain (docs subdomain, marketing preview host, etc.) compromises the kill switch primitive. Given this is now a destructive surface, a double-submit CSRF token or tightening the cookie to SameSite=Strict on the write routes is worth adding.
> 
> **SELECT-then-INSERT race in `_active_ban_for_target`** (`platform/api/hexgate_api/features/bans/service.py:59`). Two admins clicking Create for the same target within milliseconds both pass the `already banned?` check and both insert. Revoke lifts one, the other stays active. `ban-enforcement.md` §9 already lists `DB-level uniqueness constraint` as a known gap. A partial unique index (`WHERE revoked_at IS NULL`) on `(project_id, ban_type, target_agent_name, target_user_id)` closes it.
> 
> **422 error garble in `api.ts`.** `String(detail.detail)` on FastAPI's Pydantic error array `[{loc, msg, type}, ...]` yields `"[object Object]"`. CreateBanDialog masks it via status-code remapping, but any other page reading `err.message` on a 422 will show garbage. Not new to this PR but the ban UX will hit the 422 path more than most.
> 
> **Missing test cases** (each would catch a real regression):
> 
> * `DELETE /bans/{id}` returns 403 for non-admin org member (POST is covered, DELETE isn't).
> * `GET /bans` returns 403 for non-admin org member (only non-member is tested).
> * `created_by_email` is null after User deletion.
> * `limit=500` clamps to 200.
> * CreateBanDialog renders a user-facing error on 409/422 and keeps the dialog open (hook-level test only asserts the promise rejects).
> * Revoke confirm cancel path does not fire DELETE.
> 
> **PR title scope.** `feat(platform):` uses a scope that isn't in CLAUDE.md's list (`platform-api`, `dashboard`, `sdk`, `cli`, `clickhouse`). Since this touches both `platform-api` and `dashboard`, either `feat(dashboard):` with a note about the API read side, or split into two commits with different scopes.

## ✅ Fixed

- **`reason` leaks to the banned user (#1)** — kept the `AgentBannedError.reason` contract (it's a deliberate SDK field, distinct from the safe `user_message` default), but added a warning under the reason field in `CreateBanDialog`: *"Sent to the SDK with the refusal, so the banned user may see it — keep internal notes out."* If you'd prefer defense-by-default instead, I'm happy to strip `reason` from `BanFeedEntry` and keep it dashboard-only — that's a contract call, so flagging rather than doing it unilaterally.

- **422 garble in `api.ts` (#4)** — `request()` now formats FastAPI's two error shapes: a string `detail` verbatim, and the validation array `detail: [{loc, msg, type}, …]` joined into readable text (previously `String()`'d to `"[object Object]"`). Fixed in the shared helper so every page benefits, with `lib/api.test.ts` covering the array/string/no-detail cases.

- **Missing tests (#5)** — added all six:
  - `DELETE /bans/{id}` → 403 for a plain member
  - `GET /bans` → 403 for a plain member (previously only non-member)
  - `created_by_email` is null after the creating account is deleted
  - `limit=500` clamps to 200 on the enforcement read
  - `CreateBanDialog` shows a user-facing error on 409 and stays open
  - revoke-confirm **Cancel** does not fire `DELETE`

## ⏭️ Deferred (tracked, not in this PR)

- **CSRF on write endpoints (#2)** — real, but it's platform-wide: every cookie-auth write route (delete org, revoke token, remove member) is equally destructive, so patching only `/bans` is inconsistent and doesn't close the hole. This belongs in a dedicated security PR (global `SameSite=Strict` on writes or a double-submit token) rather than bolted onto the bans feature. Filing a follow-up.

- **SELECT-then-INSERT race → duplicate active ban (#3)** — already listed as a known gap in `ban-enforcement.md` §9. Low severity (needs two admins on the same target within milliseconds; fail-safe direction), and the fix (partial unique index `WHERE revoked_at IS NULL`) has friction in our no-Alembic / `create_all` setup — it wouldn't reach existing DBs without a manual index migration. Leaving documented for the hardening follow-up

guillaume-hexamind:
For your inline comments:

## ✅ Fixed

- **`ORDER BY occurred_at DESC` isn't a total order** (`features/audit/service.py`) — added `event_id DESC` as a tiebreaker, matching the storage sort key `(project_id, occurred_at, event_id)`. Offset pagination is now stable across millisecond-tied rows.
- **Offset pagination corrupts on a live feed** (`BlockedAttemptsPanel.tsx`) — `useInfiniteQuery` pages are now deduped by `event_id` on flatten, so a boundary row re-served after a new row lands mid-paging renders once instead of throwing duplicate React keys. (Pairs with the tiebreaker above.)
- **URL-param prefill = attacker-controlled target** (`routes/Bans.tsx`) — the create dialog now shows a *"target prefilled from a link — confirm before creating"* banner whenever it's opened prefilled; creating still requires the explicit destructive click. The Audit/Agents tie-in deep-links keep working.
- **Reset effect keys on `initial` identity** (`CreateBanDialog.tsx`) — the form now reseeds only on the closed→open transition (tracked via a ref), so the parent re-deriving `initial` while the dialog is open no longer wipes typed input.
- **`ban_type: str` looser than the union** (`schemas.py`) — tightened to `Literal["agent", "user"]` on `BanRead` and `BanEnforcementRow`. Left `BanFeedEntry` as `str` on purpose — the SDK intentionally tolerates/drops unknown ban types for forward-compat.

## ℹ️ Not reproducible (already handled)

- **Prefill race with `useOrgs()`** (`routes/Bans.tsx`) — the prefill effect already gates on `scope.status === "ready"` before touching `canManage`, and `useProjectScoped` only returns `ready` once the orgs query has resolved (the same `useOrgs()` cache `useCanManageBans` reads). So `canManage` can't still be loading at that point, and the premature `clearPrefillParams()` path doesn't occur. Added a test for the non-admin deep-link case.

## ⏭️ Deferred (noted, not in this PR)

- **Unbounded `IN (...)` in the creator-email lookup** (`features/bans/router.py`) — real, but N is the number of *distinct ban creators* per project (i.e. org admins), so it stays well under SQLite's 999-variable limit under the current one-active-ban-per-target design. It only becomes a footgun if bans grow into a bulk denylist, so I'm leaving `list_bans` unpaginated for now and tracking it as a follow-up rather than reworking it here.

All changes have tests; dashboard and platform-api suites are green.

guillaume-hexamind:
> Looking back at the PR after a second pass, one thing worth calling out as a follow-up: the frontend test coverage is thin relative to the new surface. 15 frontend source files changed or added, 4 test files touched, and while the hook layer (`bans.test.tsx`) is thorough and `Bans.test.tsx` covers the page-level golden paths, five substantial new components ship without a dedicated test file:
> 
> * `CreateBanDialog.tsx` (~250 lines): the primary destructive action in the whole feature. Only tested indirectly via "opens the create dialog from the header CTA". Missing: form validation edges, 409/422 error rendering, the reset-effect race I flagged inline, dropdown agent selection, reason-field visibility.
> * `BlockedAttemptsPanel.tsx`: one page test covers "Load more at end", but there's no pagination-stability test that would catch the offset drift I flagged.
> * `BlockedAttemptDrawer.tsx`: only open/close is exercised. Escape key handling and focus trap are untested.
> * `ActiveBansPanel.tsx`: revoke confirm-and-execute is tested, revoke cancel isn't.
> * `BanTypeBadge.tsx`: small enum component, fine without a dedicated test.
> 
> All three of the frontend correctness bugs I flagged inline (CreateBanDialog reset-on-rerender, Bans.tsx useOrgs race, BlockedAttemptsPanel pagination drift) live in this gap. Not by coincidence: they're the class of bug that integration-style tests structurally can't catch, because they require:
> 
> * parent re-rendering while a child modal is open (reset effect)
> * precise timing of a resolved scope + a still-loading orgs query (useOrgs race)
> * new data arriving between two paginated fetches (pagination drift)
> 
> Minimum viable additions if you don't want to over-invest: `CreateBanDialog.test.tsx` (highest ROI, primary destructive action) and `BlockedAttemptsPanel.test.tsx` (catches the live-feed pagination bug and future regressions on the feed). Plus a couple of extra assertions in `Bans.test.tsx` for the revoke-cancel path and the orgs-loading race.
> 
> Not blocking, just flagging so it doesn't get lost under the inline threads.

Thanks — most of this landed in the last two rounds, so the gap is small now:

**Already fixed + covered** (incl. the three correctness bugs you flagged):
- reset-on-rerender, pagination/offset drift (dedup by `event_id`), and revoke-cancel — all tested
- `useOrgs` race — confirmed not reproducible (prefill gates on `scope.status === "ready"`, which implies orgs resolved) + a non-admin phantom-dialog test

**Just added:** agent-ban submit (only the agent target), `Create ban` disabled until a target is set, reason-visibility warning, and drawer Escape-to-close.

**Skipping** a dedicated `BlockedAttemptsPanel.test.tsx` — its paging/dedup/drawer behavior is already exercised via `Bans.test.tsx`.

guillaume-hexamind:
Thanks you for your complete review @quanghexa94.

I fixed most of the raised issue, and documented in the comments above the defered ones we'll need to address in later PRs.

quanghexa94 (approved):
LGTM. Verified each fix on the new head, every correctness bug I flagged lands with tests.

**Standing corrected on the useOrgs race (`Bans.tsx:72`).** You're right, not reproducible. I read the ready-state check as gating only `activeProjectId`, but `useProjectScoped` in `lib/active.ts` explicitly waits on `orgsQuery.isLoading` before flipping to `"ready"`, so the "orgs still loading while scope is ready" state my finding described can't happen. My reasoning missed that coupling, sorry for the noise.

**Verified fixes** (spot-checked each):
- ORDER BY tiebreaker `, event_id DESC` at `audit/service.py:516`, matches the storage sort key.
- `dedupById` in `BlockedAttemptsPanel.tsx:13` applied on flatten.
- Prefill banner "Target prefilled from a link" at `CreateBanDialog.tsx:139`.
- Reset effect now gated on `open && !wasOpen.current` with a `useRef`, only fires on the false→true transition.
- `BanType = Literal["agent", "user"]` at `schemas.py:170`, applied on `BanRead` and `BanEnforcementRow`. Keeping `BanFeedEntry` as `str` for SDK forward-compat reads sensible.
- Reason-visibility warning at `CreateBanDialog.tsx:203`.
- 422 array formatting in `api.ts` + `api.test.ts` covering the string / array / no-detail shapes.
- Six missing tests all present across `test_bans.py`, `test_audit.py`, `Bans.test.tsx`, and the new `CreateBanDialog.test.tsx`.

**Deferrals are all reasonable:**
- CSRF is genuinely platform-wide; a dedicated security PR is the right vehicle.
- SELECT-then-INSERT duplicate-ban race is §9 known gap, and the partial unique index has friction with the `create_all` bootstrap.
- Unbounded `IN (...)` on the email lookup is bounded by distinct admin creators per project under the current one-active-ban-per-target model.
- Enforcement-feed forgery: you didn't respond to this one directly, but it's out of scope for the PR (predates it via #74) and I flagged it as a follow-up rather than a blocker. Worth tracking as an operator-visibility issue so it doesn't drift.

Skipping a dedicated `BlockedAttemptsPanel.test.tsx` is fine, since the paging + dedup + drawer flows are exercised through `Bans.test.tsx`.

One tiny nit: PR title is still `feat(platform): ...`, which isn't in the CLAUDE.md scope list. `feat(dashboard): ...` or splitting into two-scoped commits would match convention. Not blocking.

Ship it.

quanghexa94 (approved):

PR #81, opened 2026-07-10
Add the gates demo: marimo narrative, policy, and one-box launch

quanghexa94:
This is the hexgate half of a one-click, end to end demo.

The point of the demo is that hexgate plugs into whatever app you already run your agents in. Three surfaces:
- marimo is the narrative. The notebook shows the agent code, its MCP tools, a picture of the gate, and the policy, and it runs the gate live in the page with no key needed.
- hexkit is the user-facing interface where the same agent actually runs (companion PR: HexamindOrganisation/hexkit#16).
- hexgate is where you review the policy. The dashboard holds the `docs_agent` policy and edits hot-reload into hexkit.

What's in it:
- `deploy/gates-demo/`: the marimo notebook, a fake Google Docs MCP server, the policy (default deny, three roles, per-argument constraints), and `run-integrated.sh` to bring the whole box up
- `provision.py`: seeds the `docs_agent` policy so the dashboard shows it and hexkit binds it
- `boot.py`: can point marimo at any notebook and publishes the hexkit url for the notebook to link to
- `make demo-gates`: opens the notebook on its own, no key and no sandbox

Pairs with the hexkit PR above. Both are needed to run the full demo.

Stacked on #78 for now, since it builds on that branch.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -110,6 +110,10 @@ policy-test-wasm: ## Smoke a wasm-engine decision on the example policy
 demo-mcp: ## Run the MCP-proxy demo (self-contained, no external services or LLM key)
 	$(UV) python examples/mcp_demo.py
 
+.PHONY: demo-gates
+demo-gates: ## Open the gates showcase notebook (agent + MCP + policy; no LLM key). Local only — does NOT touch the Daytona snapshot.
+	$(UV) --with marimo marimo edit deploy/gates-demo/notebook.py
+
 .PHONY: demo-override
 demo-override: ## Build a deny-everything bundle + chat with HEXGATE_LOCAL_POLICY set
 	@echo "→ Writing a deny-everything override policy…"
--- a/deploy/boot.py
+++ b/deploy/boot.py
@@ -102,28 +102,28 @@ def start_services(dash_url: str | None = None) -> dict[str, str]:
     if hexkit_url:
         HEXKIT_URL_FILE.write_text(hexkit_url)
 
-    # --- Seed + mint the serve token BEFORE starting the API ------------
-    # provision talks to the shared SQLite + keystore directly (it doesn't
-    # need the API up), and it seeds docs_agent. Doing it first means the API's
-    # startup backfill_bundles compiles docs_agent's policy bundle — do it after
-    # the API boots and backfill has already run, and docs_agent is left
-    # bundle-less (usable only via the pydantic fallback, and unsignable).
-    sys.path.insert(0, str(DEPLOY_DIR))
-    sys.path.insert(0, str(API_DIR))
-    from provision import provision_serve_token  # noqa: E402  (path set above)
-
-    SERVE_KEY_FILE.write_text(provision_serve_token())
-    print("[boot] seeded demo agents + minted HEXGATE_API_KEY", flush=True)
-
     # --- Platform API (background) --------------------------------------
-    # Its lifespan re-runs the (idempotent) seed and compiles any missing
-    # bundles — including docs_agent's, now that it exists.
     _spawn(
         ["uvicorn", "hexgate_api.main:app", "--host", "0.0.0.0", "--port", str(API_PORT)],
         cwd=API_DIR,
         env=env,
     )
     _wait_healthy(f"{api_base}/v1/.well-known/keys")
+
+    # --- Mint the serve token + seed AFTER the API is healthy -----------
+    # Must come after _wait_healthy: run-integrated.sh treats SERVE_KEY_FILE
+    # appearing as "platform ready" and immediately binds the hexkit backend to
+    # the API, so the key must not exist before uvicorn answers. docs_agent is
+    # seeded without a compiled bundle — the SDK binds it via the pydantic
+    # policy_yaml fallback, which is fine here (the demo never requires signed
+    # bundles). The seed is best-effort (see provision) so a bad policy.yaml
+    # can't take down the BYOK notebook path.
+    sys.path.insert(0, str(DEPLOY_DIR))
+    sys.path.insert(0, str(API_DIR))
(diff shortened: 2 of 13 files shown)


guillaume-hexamind (commented):
I approve of the logic, and did not found any breaking issues within the PR.

For safety, if you can also check these auto-review comments:

1. Provision now runs before the API spawns — a seed failure kills the whole boot — 85/100

deploy/boot.py:115 · correctness · CONFIRMED
provision_serve_token() (now including the new docs_agent seed) runs before _spawn(uvicorn …). Previously the API was already up and past _wait_healthy, so even if minting failed the API stayed serving the notebook. Now any error in _seed_gdocs_agent (reading/parsing policy.yaml, DB/seed error) raises before uvicorn starts — taking down even the working BYOK notebook path.

2. SERVE_KEY_FILE written before API is healthy → backend binds against a dead API — 75/100

deploy/boot.py:115 · correctness · PLAUSIBLE
Same reorder. run-integrated.sh:50 treats the presence of SERVE_KEY_FILE as the "platform ready" signal and immediately launches the hexkit gdocs backend binding docs_agent to http://127.0.0.1:8000. The key used to appear only after _wait_healthy; now it appears before uvicorn even starts, so on any non-instant boot the backend's policy bind fails with a connection error — breaking the "run it in a different app" half of the demo.

3. Uncaught exception in the interactive "Try it yourself" cell — 70/100

deploy/gates-demo/notebook.py:426 · correctness · CONFIRMED
The cell only catches json.JSONDecodeError. If a user switches the tool dropdown without rewriting the default share_doc args, or enters valid-but-non-object JSON (5, []), json.loads succeeds, then wrapped[tool].ainvoke(args) raises a TypeError/validation error that escapes — rendering a raw traceback instead of an allow/deny result.

4. gdocs-backend readiness loop has no failure guard — 65/100

deploy/gates-demo/run-integrated.sh:67 · correctness · CONFIRMED
The loop breaks on success but never fails on timeout (unlike the serve-key wait). If the backend never answers /agents within 60s (crash, port conflict, bad venv), the loop just exhausts and set -e continues to launch the proxy + front-app. hexkit comes up pointing at a dead backend → empty/broken chat UI with no pointer to the real failure in /tmp/gates-agent.log.

quanghexa94:
Thanks Guillaume — all four addressed (3 commits appended).

**1 & 2 (provision reorder) — reverted.** You're right; that reorder traded a non-problem for two real ones. I'd moved provision before uvicorn only to get `docs_agent` a compiled WASM bundle, but the SDK binds it fine via the `policy_yaml` pydantic fallback (the demo never sets `HEXGATE_BUNDLE_REQUIRE_SIGNATURE`). So:
- Minting is back **after** `_wait_healthy`, so `SERVE_KEY_FILE` only appears once the API answers — `run-integrated.sh`'s "platform ready" signal is valid again and the hexkit backend won't bind against a dead API.
- The `docs_agent` seed is now **best-effort** (logged, non-fatal), so a bad `policy.yaml` can't take down the BYOK notebook path.

**3 (try-it cell) — fixed.** Now guards that args is a JSON object and catches any run error, rendering it as a warning callout instead of a traceback (covers `5`/`[]` and args that don't fit the tool).

**4 (backend readiness) — fixed.** The loop now tracks readiness and prints a clear pointer to `/tmp/gates-agent.log` on timeout. Kept it non-fatal (the notebook + dashboard are still worth bringing up), unlike the serve-key wait which is genuinely required.

Verified: ruff + compile clean, `bash -n` on the script, and the notebook still runs (19 cells, 7 allow / 6 deny).

guillaume-hexamind (approved):
Logic still looks good. Below two /code-review potential findings; you check them and ship then!

1. deploy/spawner/modal_app.py:232 — concurrent-cap flipped from fail-closed to fail-open.
The guard changed from count is None or count >= MAX to count is not None and count >= MAX. When _running_count() can't read Daytona (API outage, auth/SDK breakage), launches now proceed instead of being blocked. The old code explicitly guaranteed "an API blip can't turn the cap into unlimited launches" — that guarantee is gone. Up to the 200/day budget can spin up at once, blowing past the concurrent-sandbox quota and running up compute cost.

2. deploy/spawner/modal_app.py:45 — per-IP rate limit disabled entirely.
PER_IP_COOLDOWN_SECONDS lowered 20 → 0, and the cooldown check short-circuits on 0. A single IP can open sandboxes back-to-back with no throttle, draining the shared concurrent cap / daily budget by itself — legitimate visitors then hit "at capacity."

quanghexa94:
Thanks Guillaume, both handled:

**1 (fail-open cap) — fixed** (331835f). Restored the fail-closed guard (`count is None or count >= MAX`). It had only flipped to fail-open to work around the `daytona.list()` generator bug that made the count read *always* fail (→ permanent "at capacity"). That root cause is fixed by wrapping in `list()`, so fail-closed no longer false-bricks, and your "an API blip can't mean unlimited launches" guarantee is back. The logging + `/debug` probe stay so a real read failure is visible.

**2 (per-IP cooldown off) — intentional, keeping it off.** Two reasons:
- The live demo needs a presenter to open several sandboxes at once, so a per-IP timer is actively in the way.
- The per-IP check is unreliable behind Modal's proxy: `x-forwarded-for` is usually absent, so it falls back to the shared proxy IP and the cooldown becomes *global* (it actually blocked a second viewer on a different IP during a run).

Cost stays bounded without it: the concurrent cap (20, fail-closed) and the daily budget (200) are the hard guards. For a truly public/unattended endpoint the right protection is the Turnstile hook (already wired, just needs the secret), not the per-IP timer. Left `PER_IP_COOLDOWN_SECONDS` in place so it can be raised later if we fix real-client-IP extraction behind the proxy.

Thanks for the review!

PR #83, opened 2026-07-13
feat(dashboard): usage dashboard page

victorludvig:
## What is changing?

- New standalone Usage page: LLM call/token volume, a filterable breakdown card (by model/agent/user), and a time-range picker, mirroring the Audit page's UX.
- New api.getLlmUsageSummary client calling the already-merged /v1/projects/{id}/llm/summary endpoint (no backend changes in this PR).
- Fix: empty user_id values (untracked/system LLM calls) now map to a "(none)" display label instead of crashing the Users filter dropdown — applied to both Usage's user dimension and Audit's role/user dimensions.
- Refactor: extracted shared BarRow / BreakdownCardShell primitives so Audit's and Usage's breakdown cards stop duplicating the same Tabs/Select/grid/legend markup.

## Tests 
- make check-all 
- manual:

https://github.com/user-attachments/assets/0270a4b6-c308-44cb-aec1-3544877e98ab

### DIFF
--- a/platform/dashboard/src/components/audit/pieces.tsx
+++ b/platform/dashboard/src/components/audit/pieces.tsx
@@ -38,9 +38,9 @@ import {
   TableHeader,
   TableRow,
 } from "@/components/ui/table";
-import { Tabs, TabsList, TabsTrigger } from "@/components/ui/tabs";
 import { ToggleGroup, ToggleGroupItem } from "@/components/ui/toggle-group";
 import { Sparkline } from "@/components/ui/charts";
+import { BreakdownCardShell } from "@/components/ui/breakdown-card";
 import { BreakdownBar, type BreakdownDatum, DecisionBadge } from "./charts";
 import { OUTCOME_SERIES } from "./chart-tokens";
 import { fmtTs } from "./fmt";
@@ -314,62 +314,43 @@ export function BreakdownCard({
   const fkey = dim;
 
   return (
-    <Card className="flex flex-col p-6">
-      <div className="mb-4 flex items-center justify-between">
-        <Tabs value={dim} onValueChange={(v) => setDim(v as typeof dim)}>
-          <TabsList>
-            {DIMS.map((d) => (
-              <TabsTrigger key={d.id} value={d.id}>
-                {d.label}
-              </TabsTrigger>
-            ))}
-          </TabsList>
-        </Tabs>
-        <Select
-          value={sort}
-          onValueChange={(v) => setSort(v as "volume" | "denials")}
-        >
-          <SelectTrigger className="h-7 w-auto gap-1.5 text-xs">
-            <SelectValue />
-          </SelectTrigger>
-          <SelectContent>
-            <SelectItem value="volume">by volume</SelectItem>
-            <SelectItem value="denials">by denials</SelectItem>
-          </SelectContent>
-        </Select>
-      </div>
-      <div className="grid flex-1 grid-cols-2 gap-x-8">
-        {data.map((row) => (
-          <BreakdownBar
-            key={row.key}
-            label={row.key}
-            row={row}
-            max={max}
-            active={!f[fkey] || f[fkey] === row.key}
-            onClick={() =>
-              setF((p) => ({
-                ...p,
-                [fkey]: p[fkey] === row.key ? "" : row.key,
-              }))
-            }
-          />
-        ))}
-        {!data.length && (
-          <div className="py-2 text-[12.5px] text-muted-foreground">
-            No decisions match.
-          </div>
-        )}
-      </div>
-      <div className="mt-1 flex gap-3.5 border-t border-border pt-3 text-[11px] text-muted-foreground">
-        {OUTCOME_SERIES.map((s) => (
-          <span key={s.key} className="flex items-center gap-1.5">
-            <span className={`size-2 rounded-sm ${s.swatchClass}`} />
-            {s.label}
-          </span>
-        ))}
-        <span className="ml-auto">click a bar to filter →</span>
-      </div>
-    </Card>
+    <BreakdownCardShell
+      dims={DIMS}
+      dim={dim}
+      onDimChange={setDim}
+      metric={sort}
+      metricOptions={[
+        { value: "volume", label: "by volume" },
+        { value: "denials", label: "by denials" },
+      ]}
+      onMetricChange={setSort}
+      emptyMessage="No decisions match."
+      legend={
+        <>
+          {OUTCOME_SERIES.map((s) => (
+            <span key={s.key} className="flex items-center gap-1.5">
+              <span className={`size-2 rounded-sm ${s.swatchClass}`} />
+              {s.label}
+            </span>
+          ))}
+        </>
+      }
+      rows={data.map((row) => (
+        <BreakdownBar
+          key={row.key}
+          label={row.key}
+          row={row}
+          max={max}
+          active={!f[fkey] || f[fkey] === row.key}
+          onClick={() =>
+            setF((p) => ({
+              ...p,
+              [fkey]: p[fkey] === row.key ? "" : row.key,
+            }))
+          }
+        />
+      ))}
+    />
   );
 }
 
--- a/platform/dashboard/src/components/audit/pieces.tsx
+++ b/platform/dashboard/src/components/audit/pieces.tsx
@@ -57,8 +57,10 @@ const toDatum = (r: AuditBreakdownRow): BreakdownDatum => ({
 // Radix Select items can't carry value="" — UI-local stand-in for "all".
 const ALL = "__all__";
 
-// Module-level (not re-created per render).
-function FilterSelect({
+// Module-level (not re-created per render). Exported — reused by the
+// Usage page's filter bar (components/usage/pieces.tsx), which needs the
+// same "select with an all-option" shape for agent/model/user.
+export function FilterSelect({
   value,
   all,
   opts,
--- a/platform/dashboard/src/components/audit/pieces.tsx
+++ b/platform/dashboard/src/components/audit/pieces.tsx
@@ -23,13 +23,7 @@ import type {
 import { Badge } from "@/components/ui/badge";
 import { Button } from "@/components/ui/button";
 import { Card } from "@/components/ui/card";
-import {
-  Select,
-  SelectContent,
-  SelectItem,
-  SelectTrigger,
-  SelectValue,
-} from "@/components/ui/select";
+import { FilterSelect } from "@/components/ui/filter-select";
 import {
   Table,
   TableBody,
@@ -56,40 +50,6 @@ const toDatum = (r: AuditBreakdownRow): BreakdownDatum => ({
 
 // Radix Select items can't carry value="" — UI-local stand-in for "all".
 const ALL = "__all__";
-
-// Module-level (not re-created per render). Exported — reused by the
-// Usage page's filter bar (components/usage/pieces.tsx), which needs the
-// same "select with an all-option" shape for agent/model/user.
-export function FilterSelect({
-  value,
-  all,
-  opts,
-  onChange,
-}: {
-  value: string;
-  all: string;
-  opts: string[];
-  onChange: (v: string) => void;
-}) {
-  return (
-    <Select
-      value={value || ALL}
-      onValueChange={(v) => onChange(v === ALL ? "" : v)}
-    >
-      <SelectTrigger className="h-8 w-auto min-w-32 gap-1.5 text-[13px]">
-        <SelectValue />
-      </SelectTrigger>
-      <SelectContent>
-        <SelectItem value={ALL}>{all}</SelectItem>
-        {opts.map((o) => (
-          <SelectItem key={o} value={o} className="font-mono text-xs">
-            {o}
-          </SelectItem>
-        ))}
-      </SelectContent>
-    </Select>
-  );
-}
 
 // ————————————————————————————————————————————— Filter bar
 export function FilterBar({
--- a/platform/dashboard/src/components/audit/pieces.tsx
+++ b/platform/dashboard/src/components/audit/pieces.tsx
@@ -20,10 +20,11 @@ import type {
   AuditFilters as Filters,
   SetAuditFilters as SetFilters,
 } from "@/lib/audit-filters";
+import { ActiveFilterChips } from "@/components/ui/active-filter-chips";
 import { Badge } from "@/components/ui/badge";
 import { Button } from "@/components/ui/button";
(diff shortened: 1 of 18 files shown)


platform/dashboard/src/routes/Usage.tsx
```diff
@@ -0,0 +1,234 @@
...
+          )}
+        </div>
+      </header>
+
+      <UsageFilterBar
+        f={f}
+        setF={setF}
+        shown={totals.calls}
+        total={rangeTotal}
+        agents={options?.by_agent.map((r) => r.key) ?? []}
```
winlp4ever, line 114: Empty-string defense is applied to `by_user` on line 194 but not to `by_agent` here or `by_model` on line 193. If the backend ever emits an `LlmInvocationBreakdownRow` with `key=""` for agent or model (legacy events, unset agent name, model string not populated), `FilterSelect` hands `<SelectItem value="">` to Radix and the page throws the same crash the PR is fixing for users. `LlmInvocationBreakdownRow.key` is typed `string` on both sides, so the defense should be symmetric across dimensions.

winlp4ever (commented):
Solid PR overall. Backend contract is clean: I verified path, all query params (`window`, `agent`, `model`, `user`, `start_date`, `end_date`), the `window` literal union, and the full response shape (`LlmInvocationTotals` and `LlmInvocationBreakdownRow`) against the endpoint on `main`. Empty-user round-trip via `scopeNoValue` / `displayNoValue` is correct.

Inline comments cover two correctness / consistency concerns and four refactor-quality gaps. A few things that don't fit inline cleanly:

**Test files missing for the shared primitives.** The whole justification for extracting `BreakdownCardShell` / `BarRow` / `FilterSelect` is reuse across two dashboards, but no `breakdown-card.test.tsx` or `filter-select.test.tsx` ships. The `FilterSelect` `ALL ↔ ""` round-trip is the fix at the heart of the PR, yet it's only exercised via the specific integrations that send an empty value today. A third caller, or a change to the sentinel plumbing, would regress silently while both page tests stay green.

**Empty-string coverage only tests the user dimension.** `Usage.test.tsx` covers `by_user=[{key:'', ...}]` but there's no equivalent for `by_agent` or `by_model`. Pairs with the symmetric-fix concern flagged inline: if you make the fix symmetric, add fixtures for both.

**Calendar popover click flow is never driven.** All date-set assertions in Usage/Audit tests use `useUsageFilters.setState` / `useAuditFilters.setState` directly. If someone drops `endOfDay(range.to)`, the `end_date` filter silently drops the last day's data and CI stays green.

**Metric-select tab flip only tests `tokens → calls`.** The other three token metrics (input / output / total) go unverified, and no assertion checks that switching dim tabs preserves the selected metric across dimensions.

platform/dashboard/src/components/audit/pieces.tsx
```diff
@@ -57,38 +51,6 @@ const toDatum = (r: AuditBreakdownRow): BreakdownDatum => ({
 // Radix Select items can't carry value="" — UI-local stand-in for "all".
 const ALL = "__all__";
```
winlp4ever: The `ALL = "__all__"` sentinel is now declared in two places: here and `components/ui/filter-select.tsx:16`. Since `FilterSelect` was extracted to own the empty-string round-trip, the sentinel should be exported from there and imported here so the outcome `ToggleGroup` at line 97 uses the same constant. Otherwise if the value ever needs to change (Radix reserving `__all__`, a rogue tool literally named `__all__`), one definition updates and the other silently disagrees about what `all` means.

platform/dashboard/src/routes/Usage.tsx
```diff
@@ -0,0 +1,234 @@
...
+          icon={Sigma}
+          value={fmtTokens(totals.total_tokens)}
+          sub={avgLabel(totals.total_tokens)}
+        />
+      </div>
+
+      <UsageBreakdownCard
+        byModel={summaryQ.data?.by_model ?? []}
+        byAgent={summaryQ.data?.by_agent ?? []}
+        byUser={summaryQ.data?.by_user.map(displayNoValue) ?? []}
```
winlp4ever, line 150: `by_user.map(displayNoValue)` creates a new array reference on every render, so `UsageBreakdownCard`'s inner `useMemo(() => source.slice().sort(...).slice(0, 10), [source, metric])` never hits its cache when `dim === 'user'`. Only 10 rows so the cost is negligible, but the memo is effectively dead code. Either wrap the mapping in a `useMemo` at the call site, or move `displayNoValue` inside the breakdown card. Same pattern in `Audit.tsx` for `byRole`.

platform/dashboard/src/components/usage/pieces.tsx
```diff
@@ -0,0 +1,217 @@
...
+      />
+      <span className="ml-auto whitespace-nowrap text-xs text-muted-foreground">
+        <span className="text-foreground">{shown.toLocaleString()}</span> of{" "}
+        <span className="font-mono">{total.toLocaleString()}</span> LLM calls
+      </span>
+    </div>
+  );
+}
+
+export function UsageActiveChips({
```
winlp4ever, line 61: `UsageActiveChips` is a near-verbatim copy of `ActiveChips` in `audit/pieces.tsx` (~50 lines: Filters label, Badge map, X button, Clear-all row). Only the filter-key set and the reset payload differ. This is the second-biggest chunk of Audit ↔ Usage duplication and the extraction skipped it. Worth pulling into a generic `<ActiveFilterChips<F>>` alongside `BreakdownCardShell`, since the two dashboards will keep drifting on chip styling and Clear-all behavior otherwise.

platform/dashboard/src/routes/Usage.tsx
```diff
@@ -0,0 +1,234 @@
...
+      <header className="mb-6 flex items-end justify-between gap-4">
+        <div>
+          <h1 className="text-2xl font-semibold tracking-tight">Usage</h1>
+          <p className="mt-1 text-sm text-muted-foreground">
+            LLM token usage for project{" "}
+            <span className="font-mono text-foreground">{projectName}</span>
+          </p>
+        </div>
+        <div className="flex items-center gap-2">
+          <ToggleGroup
```
winlp4ever: The Range/Custom `<ToggleGroup>` + `<Popover>` + `<Calendar>` + `impliedRangeLabel` block from here through ~line 175 is copy-pasted between `Usage.tsx` and `Audit.tsx` (starts at `Audit.tsx:506`). Both operate on identical fields (`range`, `customMode`, `start_date`, `end_date`) that already share the `Range` type via `date-range.ts`. The next range-picker tweak (a `1h` preset, timezone handling, `endOfDay` behavior) will need to happen twice and will diverge. Good candidate for a `<DashboardRangePicker>` in `components/ui/` alongside `BreakdownCardShell`.

platform/dashboard/src/components/ui/breakdown-card.tsx
```diff
@@ -0,0 +1,122 @@
...
+  emptyMessage,
+  legend,
+}: {
+  dims: { id: Dim; label: string }[];
+  dim: Dim;
+  onDimChange: (d: Dim) => void;
+  metric: Metric;
+  metricOptions: { value: Metric; label: string }[];
+  onMetricChange: (m: Metric) => void;
+  rows: ReactNode[];
```
winlp4ever: Taking `rows: ReactNode[]` means every caller renders its own `<BarRow>` and duplicates the bar-click toggle (`active={!f[fkey] || f[fkey] === row.key}` and `onClick={() => setF((p) => ({ ...p, [fkey]: p[fkey] === row.key ? '' : row.key }))}`) verbatim in `audit/pieces.tsx` and `usage/pieces.tsx`. Passing pre-rendered rows in also blocks the shell from ever adding row-level concerns (virtualization, keyboard navigation, ARIA). If it took `rows: BreakdownRow[]` plus a `filterKey` / `onFilterToggle` prop, the click semantics would live in one place and the shell could grow into those concerns later.

platform/dashboard/src/lib/usage-filters.ts
```diff
@@ -0,0 +1,48 @@
...
+ * last week silently narrowing today's usage view would read as missing
+ * data. No ``tableLimit``/``loadMore`` here — Usage has no events table.
+ */
+
+import { create } from "zustand";
+
+import type { Range } from "./date-range";
+
+// '' = "all".
+export interface UsageFilters {
```
winlp4ever: `UsageFilters` and `AuditFilters` share six identically-shaped fields (`agent`, `user`, `range`, `customMode`, `start_date`, `end_date`) with matching `EMPTY_*` reset payloads. TypeScript has nothing enforcing they stay symmetric. If someone adds a new project-wide filter field (a `session_id`, an `env: 'prod' | 'staging'`) intended for both dashboards and updates one file, the other silently drops it on updates or TypeScript only complains at the callsite. A `BaseDashboardFilters` type extended by both would close that. Related: the curried `set = <K extends keyof Filters>(k, v) => setF((p) => ({ ...p, [k]: v }))` helper is redefined four times across `usage/pieces.tsx` and `audit/pieces.tsx`. Since the setter contract is identical, a single generic helper (or a hook) could replace all four.

platform/dashboard/src/routes/Usage.tsx
victorludvig: The agent and model fields can never be sent to the backend as empty string thanks to pydantic validation.
See:
- [model](https://github.com/HexamindOrganisation/hexgate/blob/a59887de663f7d24c0982e4d93fae8055bd1a244/platform/api/hexgate_api/schemas.py#L461-L462)
- [agent](https://github.com/HexamindOrganisation/hexgate/blob/a59887de663f7d24c0982e4d93fae8055bd1a244/platform/api/hexgate_api/schemas.py#L423-L424)

platform/dashboard/src/components/audit/pieces.tsx
victorludvig: ALL is now exported from filter-select.tsx:16 and imported in pieces.tsx, with the local duplicate declaration removed.

platform/dashboard/src/components/usage/pieces.tsx
victorludvig: Added platform/dashboard/src/components/ui/active-filter-chips.tsx, a generic ActiveFilterChips<F> component (chip row + "Clear all"), following the same shared-component pattern as BreakdownCardShell.

platform/dashboard/src/routes/Usage.tsx
victorludvig: Created a shared DashboardRangePicker to avoid duplicate code. Extracted the range-preset ToggleGroup + custom-date Popover/Calendar block (and impliedRangeLabel/DATE_FMT) into DashboardRangePicker in components/ui/dashboard-range-picker.tsx, same pattern as BreakdownCardShell/ActiveFilterChips. Both pages now just render <DashboardRangePicker f={f} setF={setF} />.

platform/dashboard/src/lib/usage-filters.ts
victorludvig: Added RangePickerFilters (4 shared range/custom-mode fields) and BaseDashboardFilters extends RangePickerFilters (adds agent, user) to platform/dashboard/src/lib/date-range.ts — the module both filter stores already treat as neutral shared ground.

platform/dashboard/src/components/ui/breakdown-card.tsx
victorludvig: Centralized the bar's filter-toggle click logic in BreakdownCardShell (via renderRow/activeFilterValue/onFilterChange), removing the duplicated active/onClick math from BreakdownCard and UsageBreakdownCard.

platform/dashboard/src/routes/Usage.tsx
victorludvig: Centralized the bar's filter-toggle click logic in BreakdownCardShell (via renderRow/activeFilterValue/onFilterChange), removing the duplicated active/onClick math from BreakdownCard and UsageBreakdownCard.

winlp4ever (approved):
LGTM. Verified every fix on the new head. The refactor commit is clean and the shared primitives now live where they should.

**Standing corrected on the empty-string defense for agent/model.** You're right, and I checked `platform/api/hexgate_api/schemas.py` on main: `LlmInvocationEvent.agent_name` and `LlmInvocationEvent.model` both carry `Field(min_length=1, max_length=256)`, so the ingest endpoint rejects empty strings and the `llm_invocation` table can never have those as empty. That means `by_agent` and `by_model` can never return `key=""`. Only `user_id: str = Field(default="", max_length=256)` permits empty, which is exactly where the `displayNoValue` mapping is applied. My finding was overly cautious.

**Verified fixes** (spot-checked each):
- `ALL` now exported from `filter-select.tsx:16` and imported in `audit/pieces.tsx:27`; the local duplicate is gone.
- `BreakdownCardShell` now takes `rows: Row[]` with `Row extends { key: string }` plus `renderRow` / `activeFilterValue` / `onFilterChange`. The bar-click toggle math lives in one place and both callers pass a `renderRow` callback.
- `ActiveFilterChips<F>` in `components/ui/active-filter-chips.tsx`. Generic on `F extends object`, takes `chipKeys` and `onClearAll`. Both dashboards call it.
- `DashboardRangePicker<F extends RangePickerFilters>` in `components/ui/dashboard-range-picker.tsx`. Uses the `RangePickerFilters` type to keep the field contract explicit.
- `RangePickerFilters` + `BaseDashboardFilters extends RangePickerFilters` in `date-range.ts`. Nice choice of placement in the neutral module both stores already depend on. `AuditFilters` and `UsageFilters` both `extends BaseDashboardFilters` now, so TypeScript enforces the shared six fields.

**One thing still open from the review body: no test files for the shared primitives.** `active-filter-chips.test.tsx`, `dashboard-range-picker.test.tsx`, `breakdown-card.test.tsx`, and `filter-select.test.tsx` all still absent. The whole justification for pulling these into `components/ui/` is reuse; a regression to the `ALL ↔ ""` round-trip or the shell's filter-toggle wiring would slip past the page tests today. Not blocking for this PR (matches the pattern established by prior dashboard PRs), but worth tracking as a follow-up if you want the extraction to hold up as more callers land.

**Minor observation, take it or leave it:** the fresh-array-ref memo invalidation on `byUser={by_user.map(displayNoValue) ?? []}` in `Usage.tsx:150` and the equivalent in `Audit.tsx` for `byRole` still stands. Cost is 10 rows of sort/slice per render, so nothing to lose sleep over. If it ever matters, wrap the `.map` in a `useMemo` at the call site or move `displayNoValue` inside `UsageBreakdownCard`.

Ship it.

quanghexa94 (approved):

PR #84, opened 2026-07-15
feat(platform-scripts): enhance seed_audit.py script

victorludvig:
## What is changing?
The seed_audit script is updated:
- Number of users can be passed as arguments, up to 40.
- Number of anomalies can be passed as arguments, up to 100. The date of the anomaly is now randomly generated.
- Project id / agent name can be passed as arguments
- Input validation / logging .
- Delete queries sql is parameterised to avoid SQL injection and user ids are prefixed with _test to avoid real data deletion.

## Tests
- ran the script:
`uv run python ../scripts/seed_audit.py --number_users 40 --number_anomalies 3 --project_id 00000000-0000-0000-0000-000000000003 --agent_name support-bot`

Screenshot 2026-07-16 at 10 57 05

### DIFF
--- a/platform/scripts/seed_audit.py
+++ b/platform/scripts/seed_audit.py
@@ -83,13 +83,14 @@ USER_IDS = [
         "Noah",
     ]
 ]
-ANOMALY_USERS = ["test_Bob", "test_Mallory"]
-
 # ── Traffic shape ─────────────────────────────────────────────────────────────
 # Normal: 300 rows per user over 30 days, outcomes weighted 80% allow / 10% deny / 10% needs_approval.
-# Anomaly: for each anomaly, one user (sampled from ANOMALY_USERS) spikes 20-50 denies
-# in a 5-minute window at a random point in the last 30 days,
-# probing restricted tools (refund_customer, create_ticket).
+# Anomaly: for each anomaly, one user (sampled from the seeded normal subset,
+# USER_IDS[:number_users]) spikes 20-50 denies in a 5-minute window at a random
+# point in the last 30 days, probing restricted tools (refund_customer, create_ticket).
+# Sampling from the seeded subset (not a fixed list) ensures the anomaly user
+# already has normal-behavior rows, so the anomaly reads as a deviation rather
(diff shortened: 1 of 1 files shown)


platform/scripts/seed_audit.py
```diff
@@ -22,77 +38,138 @@
...
+        "Hugo",
+        "Iris",
+        "Jack",
+        "Kira",
+        "Leo",
+        "Maya",
+        "Noah",
+    ]
+]
+ANOMALY_USERS = ["test_Bob", "test_Mallory"]
```
guillaume-hexamind: Anomaly users include test_Mallory (index 12 in USERS) -> with default --number_users 3, Mallory will only have some anomaly deny data ingested instead of some normal behavior + an anomaly deny spike during a time range. 

Probably better to sample the anomaly users from the seeded normal subset (USER_IDS[:number_users])

platform/scripts/seed_audit.py
```diff
@@ -116,67 +200,150 @@ def generate_normal_data(rng: random.Random, now: datetime) -> list[list]:
...
+        description="Seed ClickHouse with audit test data."
+    )
+    parser.add_argument(
+        "--clear",
+        action="store_true",
+        help="Delete seed rows instead of inserting.",
+    )
+    parser.add_argument(
+        "--agent_name",
+        default=DEFAULT_PROJECT_NAME,
```
guillaume-hexamind: Probably better to have a standalone default agent name (like support-agent) to differentiate from project name here

guillaume-hexamind (commented):
Agree with the logic; 2 small changes to check before merging

platform/scripts/seed_audit.py
victorludvig: Good point, changed this line to: default="support-agent",

platform/scripts/seed_audit.py
victorludvig: Good point, the anomaly users are now sampled from the active users.

guillaume-hexamind (approved):

PR #85, opened 2026-07-16
fix(sdk): resolve .env from consumer cwd, not installed package path

guillaume-hexamind:
## What

`bootstrap()` now resolves `.env` from the consumer's current working directory instead of relative to the installed package.

```python
# before — resolved to site-packages/hexgate/../.env once pip-installed
env_path = Path(__file__).parent.parent / env_file
# after — scoped to the consumer's cwd, no upward walk, no-op if missing
env_path = Path.cwd() / env_file
if env_path.is_file():
    load_dotenv(env_path, override=False)
```

## Why

For a pip-installed SDK, `__file__` lives in `site-packages/`, so `hexgate register` looked for `.env` there and never found the consumer's. Result: `HEXGATE_API_KEY must be set` even with a valid `.env` in the project — breaking the first-run quickstart for every external developer. It only "worked" from a source checkout, where that path happens to be the repo root.

Resolution is deliberately scoped to `Path.cwd()` with **no upward directory walk**: a stray `.env` in an ancestor directory (e.g. `~/.env` left over from another project) must never be loaded silently and point the SDK at the wrong credentials.

`override=False` is unchanged (shell still wins over `.env`); a missing `.env` degrades gracefully to a no-op, falling back to shell env vars.

## Tests

- `test_bootstrap_loads_env_file_from_cwd` — real-fs regression: `.env` in a temp cwd loads.
- `test_bootstrap_shell_var_wins_over_env_file` — asserts `override=False` (shell wins over `.env`).
- `test_bootstrap_ignores_ancestor_env` — guards against loading a `.env` from a parent directory (no upward walk).
- `test_bootstrap_no_env_file_is_noop` — a missing `.env` is a clean no-op.

### DIFF
--- a/hexgate/bootstrap.py
+++ b/hexgate/bootstrap.py
@@ -4,8 +4,9 @@ from __future__ import annotations
 
 import logging
 import os
+from pathlib import Path
 
-from dotenv import find_dotenv, load_dotenv
+from dotenv import load_dotenv
 
 from hexgate import audit
 from hexgate.config.env import resolve_api_key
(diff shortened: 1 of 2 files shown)


victorludvig:
I think the description is outdated because now load_dotenv is used and not find_dotenv. Can you update the description, and then we can merge.

hexgate/bootstrap.py
```diff
@@ -44,12 +44,11 @@ def bootstrap(env_file: str = ".env", *, local_only: bool = False) -> Settings:
...
-    # ``override=False``: shell wins over .env, matching the convention
-    # uvicorn / vite / cargo / npm follow. A pre-existing test
-    # (test_bootstrap_loads_requested_env_file) pinned this contract;
-    # the code had drifted to ``True``. Flipped back here.
-    load_dotenv(env_path, override=False)
+    # Resolve against the consumer's cwd, no upward walk — a stray ancestor
+    # .env must never load silently. Missing → no-op.
+    env_path = Path.cwd() / env_file
+    if env_path.is_file():
+        load_dotenv(env_path, override=False)
```
victorludvig, line 51: I think the PR description is outdated because now load_dotenv is used and not find_dotenv. Can you update the description, and then we can merge.

victorludvig (approved):
Check the description and then we can merge.

PR #88, opened 2026-07-16
build(claude): add per-file format-on-edit hook

victorludvig:
## Summary
- Adds a PostToolUse hook that runs on every Edit/Write and formats only the touched file, scoped by path (`ruff format` for `platform/api/*.py` and other `*.py`, `prettier --write` for dashboard `src/*.{ts,tsx,js,jsx,css,json}`)
- Wires it up in `.claude/settings.json` so it's shared across the team instead of living in local-only config

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -1,0 +1,8 @@
+{
+  "hooks": {
+    "PostToolUse": [{
+      "matcher": "Edit|Write",
+      "hooks": [{"type": "command", "command": "bash .claude/hooks/format-on-edit.sh 2>/dev/null || true"}]
+    }]
+  }
+}
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -2,7 +2,7 @@
   "hooks": {
     "PostToolUse": [{
       "matcher": "Edit|Write",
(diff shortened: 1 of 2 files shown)


.claude/settings.json
```diff
@@ -0,0 +1,8 @@
+{
+  "hooks": {
+    "PostToolUse": [{
+      "matcher": "Edit|Write",
+      "hooks": [{"type": "command", "command": "bash .claude/hooks/format-on-edit.sh 2>/dev/null || true"}]
```
guillaume-hexamind: 2>/dev/null || true discards and masks any potential failure (missing jq, uv , ruff, prettier...).

Would it be better to let real failure print a warning or hard fail to let the user setup its env properly?

guillaume-hexamind (commented):
A small question, but looks good, useful hook

guillaume-hexamind (approved):

.claude/settings.json
victorludvig: Good point, I added some logs and removed the Fail Silent 2>/dev/null || true, so that the logs are output in the claude console.

PR #91, opened 2026-07-17
tests(sdk): find the framework versions compatible with hexgate adapters

guillaume-hexamind:
## What

A test harness that determines **which released versions of each framework hexgate's adapters can wrap with working policy enforcement**, plus a driver to re-run it as a version matrix.

## Why

Each adapter splices policy enforcement in front of a framework's tool call by patching a *private* callable. That splice is what breaks across framework versions — and the dangerous failure is **silent bypass** (enforcement stops attaching, a denied tool runs anyway), not an exception. So the probes assert the **deny path**, not just that a run completes.

## What's in this PR

- **Probe suite** (`tests/framework_compat/`), per framework (pydantic_ai, openai-agents, google-adk, langchain, deepagents):
  - **Tier 0** — contract check (the private surfaces the adapter depends on still exist).
  - **Tier 1** — deterministic seam check: invoke the guarded tool directly and assert a denied call is blocked *and never runs* (catches silent bypass). No LLM.
  - **Tier 2** — full LLM e2e. All probes run on OpenAI `gpt-4o-mini` + `OPENAI_API_KEY` (google-adk via its LiteLLM wrapper), so the suite needs a single provider key.
- **Opt-in** — every module carries the `framework_compat` marker and is excluded from the default `pytest` / `make test` / `coverage` / CI (`addopts -m "not framework_compat"`). The per-commit regression gate for *pinned* versions stays `tests/adapters/`.
- **Matrix driver** (`scripts/framework_matrix.py`) — installs each `(framework, version)` cell in an isolated `uv` venv (versions auto-discovered from PyPI or passed explicitly), runs the probes, classifies per tier, and writes a compatibility table to `build/framework-matrix/` (gitignored).
- **Make targets** — `make framework-compat`, `make framework-matrix ARGS="…"`.

Local + offline by default (policy from a bundled YAML via `HEXGATE_LOCAL_POLICY`, opa-compiled to WASM), so a red cell means "the version broke the wrap," never "the platform 404'd."

## Findings — enforcement (Tier 1) supported ranges

| Framework | Works on | Notes |
|---|---|---|
| pydantic_ai | 1.88.0 → 2.12.0 | survives the 1.x→2.x bump |
| openai-agents | **0.8.0** → 0.18.3 | broken below 0.8.0 — adapter can't import `RunState` |
| google-adk | 1.14.0 → 2.5.0 | survives the 1.x→2.x bump |
| langchain | 1.0.0, 1.2.15 → 1.3.14 | 1.1.0 / 1.2.5 fail to import (transitive langgraph mismatch) |
| deepagents | 0.1.4 → 0.6.12 | 0.0.1 predates the `model=` kwarg |

Two source issues the matrix surfaced are addressed in a **follow-up PR**: the `openai-agents` floor was wrong (`>=0.0.10` → `>=0.8.0`, enforcement was silently impossible below 0.8.0) and pydantic-ai 2.x broke the usage adapter.

## Run

```bash
make framework-compat                     # Tier 0/1, offline, no key
OPENAI_API_KEY=… make framework-compat    # + Tier 2 e2e
make framework-matrix ARGS="--dry-run"    # preview the version grid
```

More detail in `tests/framework_compat/README.md`


scripts/framework_matrix.py
```diff
@@ -0,0 +1,592 @@
...
+
+    def pin(self, framework: Framework, version: str) -> subprocess.CompletedProcess:
+        """Re-resolve + install hexgate with ``<dist>==<version>`` pinned.
+
+        Installing ``-e .[dev]`` alongside the pin (rather than reinstalling
+        only the primary dist) lets uv move coupled companions to a coherent
+        set for this version.
+        """
+        python = self.python_for(framework)
+        specs = [f"{framework.dist}=={version}", *framework.pin_extra]
```
quanghexa94, line 258: Reusing one venv per framework across all its versions with `uv pip install` (not sync, no `--reinstall`) can leave stale transitive companions, so a cell may be tested against the wrong coupled versions. This is the exact failure the docstring says it avoids. Example: cell A tests `langchain==1.3.14`, pulling langgraph 1.3.x. Cell B then tests `langchain==1.2.15` in the same venv. `uv pip install` prefers already-installed versions and only moves a package when constraints force it, so if 1.2.15 permits langgraph 1.3.x the stale langgraph stays, and cell B reports a result for a pairing a fresh resolve would never pick. The docstring claims 'every cell re-resolves the full graph' and warns that reusing installs 'silently leaves stale companions', but the code reuses the venv without `--reinstall` or `uv sync`. Fix: fresh venv per cell, `uv pip install --reinstall`, or `uv sync` against a per-cell requirement set.

scripts/framework_matrix.py
```diff
@@ -0,0 +1,592 @@
...
+        else:
+            raw[tier].append(TIER_PASS)
+
+    tiers: dict[int, str] = {}
+    for tier, outcomes in raw.items():
+        if not outcomes:
+            tiers[tier] = TIER_NA
+        elif TIER_FAIL in outcomes:
+            tiers[tier] = TIER_FAIL
+        elif TIER_PASS in outcomes:
```
quanghexa94, line 318: parse_junit marks a tier PASS when any test passed and none failed, so a SKIPPED Tier-1 test alongside a passing one yields tier1=PASS. Tier 1 is two tests: `test_deny_path_*` (the enforcement signal) and `test_allow_decision`. If the deny-path test is ever skipped (a future provider gate, a conditional skipif, an env-dependent collection skip) while `test_allow_decision` passes, this returns PASS for tier 1 and classify() reports the cell Tier-1-green, even though the deny path never ran. The README guards against unmapped tests masking green but not against an intra-tier skip doing the same. Consider treating a SKIP inside an otherwise-expected tier as a distinct 'incomplete' verdict rather than folding it into PASS.

scripts/framework_matrix.py
```diff
@@ -0,0 +1,592 @@
...
+        "T2 = LLM e2e (blank when no provider key).",
+        "",
+        "| Framework | Version | T0 | T1 | T2 | Status |",
+        "| --- | --- | :-: | :-: | :-: | --- |",
+    ]
+    for r in results:
+        lines.append(
+            f"| {r.framework} | {r.version} | {_cell(r.tiers.get(0, TIER_NA))} | "
+            f"{_cell(r.tiers.get(1, TIER_NA))} | {_cell(r.tiers.get(2, TIER_NA))} | "
+            f"{r.status}{(' — ' + r.detail) if r.detail else ''} |"
```
quanghexa94: `r.detail` is interpolated into a markdown table cell without escaping `|`, so a detail string containing a pipe breaks the row's column structure. An install failure sets `detail = _last_line(pin.stderr)`, and uv dependency-conflict messages often print constraint expressions containing `|`. That row then renders with a spurious extra column and every downstream cell shifts. Escape `|` and newlines in `detail` before interpolating, or wrap it in backticks.

scripts/framework_matrix.py
```diff
@@ -0,0 +1,592 @@
...
+    latest_only: bool,
+) -> list[str]:
+    """Choose the versions to test for ``framework``.
+
+    Explicit versions win. Otherwise take the floor, the latest, and an
+    evenly-spaced sample in between up to ``limit`` — the installed version
+    folded in when asked.
+    """
+    if explicit:
+        return [str(v) for v in sorted({Version(v) for v in explicit})]
```
quanghexa94: The explicit `--versions` path calls `Version(v)` outside the RuntimeError guard, so a malformed version string aborts the whole run with an uncaught traceback instead of a clean per-framework error. `python scripts/framework_matrix.py --versions pydantic=1.88.0,typo` raises `packaging.version.InvalidVersion` here; main()'s `except RuntimeError` doesn't catch it (InvalidVersion is a ValueError subclass), so the driver dies with a raw traceback rather than reporting DISCOVERY-FAIL for just that framework. Validate or catch InvalidVersion in `_parse_explicit` or `select_versions` and surface it the same way discovery errors are.

scripts/framework_matrix.py
```diff
@@ -0,0 +1,592 @@
...
+
+# ---------------------------------------------------------------------------
+# CLI
+# ---------------------------------------------------------------------------
+
+
+def _parse_explicit(pairs: list[str] | None) -> dict[str, list[str]]:
+    out: dict[str, list[str]] = {}
+    for pair in pairs or []:
+        key, _, versions = pair.partition("=")
```
quanghexa94: `--versions FW` with no `=` silently discards the request and falls back to version discovery. `pair.partition('=')` yields versions='' for an entry like `--versions pydantic`, so `out['pydantic']` becomes `[]`, and in select_versions `if explicit:` is falsy for the empty list, triggering full discovery. A user who mistypes the `FW=v1,v2` form gets 6 discovered cells with no warning that their explicit request was dropped. Either reject an entry with no `=` or no versions, or log that discovery is being used instead.

quanghexa94 (commented):
Ran /code-review at high effort. Nice suite. The probe design (assert the deny path and that the denied tool body never ran, not just that a run completes) is the right call, and `asyncio_mode = auto` in pytest.ini covers the un-decorated async deny-path probes in both `make framework-compat` and the matrix subprocess (`.[dev]` installs pytest-asyncio into each venv). Findings are all in the matrix driver's methodology and robustness, most severe first.

One cross-PR note tied to #92: once #92 raises the pyproject floors (langchain 1.2.15, openai-agents 0.8.0), the `FRAMEWORKS` floors here (langchain 1.0.0, openai 0.0.10) will make those low-version cells permanent INSTALL-FAIL/UNUSABLE, since each cell installs `-e .[dev]` which carries the pyproject floor. That's arguably intended discovery, but worth expecting the noise.

scripts/framework_matrix.py
guillaume-hexamind: Good catch: this was a real correctness hole, not just noise. Fixed: pin() now passes --reinstall

scripts/framework_matrix.py
guillaume-hexamind: Fixed. Added a partial tier outcome: a tier with some tests passing and others skipped no longer folds into PASS. Tier 1 partial now classifies as T1 INCOMPLETE, so a skipped deny-path test alongside a passing allow-decision can never read as green.

scripts/framework_matrix.py
guillaume-hexamind: Fixed. detail is now run through _md_inline() before interpolation

scripts/framework_matrix.py
guillaume-hexamind: Fixed. The explicit --versions path now catches InvalidVersion and re-raises as RuntimeError

scripts/framework_matrix.py
guillaume-hexamind: Fixed. _parse_explicit now rejects a bare FW (no =) or FW= (no versions) with a clear error

guillaume-hexamind:
Thanks @quanghexa94 for the review, all 5 comments and issue raised are now fixed.

On the #92 cross-PR note: you're right, and it's intended behavior rather than a bug once #92 raises the pyproject floors, a below-floor cell should fail to install, because -e .[dev] legitimately can't coexist with that version. That's the matrix correctly reporting "this pairing is impossible," which is signal, not noise. The FRAMEWORKS floors here intentionally sit lower so we still probe the boundary and see exactly where it moves. I'll keep them decoupled from the pyproject floors for that reason, and add a one-line note to the results README (when #92 lands) so those INSTALL-FAIL rows read as "below supported floor" rather than a wrap break.

quanghexa94 (approved):
LGTM

PR #92, opened 2026-07-17
fix(sdk): correct framework version support and publish compatible ranges

guillaume-hexamind:
Follow-up to #91 (the framework-version test matrix), which surfaced compatibility
gaps and established the supported ranges. This PR ships the fixes and documents
the ranges.

## Changes

- **fix(sdk): pydantic-ai 2.x usage.** `result.usage` is a property in pydantic-ai
  2.x (a callable method in 1.x); the usage adapter now resolves both, so a 2.x run
  no longer raises `TypeError: 'RunUsage' object is not callable`. Enforcement
  already attached on 2.x — this fixes the run path.
- **build: raise `openai-agents` floor to `>=0.8.0`.** Below 0.8.0 the openai
  adapter can't import (`RunState` was added in 0.8.0), so policy enforcement was
  silently impossible while the old `>=0.0.10` pin advertised support.
- **build: floor `langchain` to `>=1.2.15` (and `langgraph` to `>=1.1.10`).**
  langchain 1.1.0/1.2.5 fail to import via a transitive
  `langgraph.runtime.ExecutionInfo` mismatch; the floor keeps the resolver on the
  verified range (1.2.15+ constrains its own langgraph/langchain-core).
- **docs: compatible framework versions, single-sourced.** A dedicated adapters
  overview page carries the compatibility table; each per-adapter page links to it
  instead of repeating the ranges, so a matrix re-run updates one place. The
  overview also lists which floors are enforced by hexgate's requirements.
- `uv.lock` updated to match.

## Verified compatible ranges (from #91's matrix)

| Framework | Package | Verified compatible |
|---|---|---|
| OpenAI Agents SDK | `openai-agents` | 0.8.0 – 0.18.3 |
| LangChain / LangGraph | `langchain` | 1.0.0, 1.2.15 – 1.3.14 |
| Google ADK | `google-adk` | 1.14.0 – 2.5.0 |
| Pydantic AI | `pydantic-ai-slim` | 1.88.0 – 2.12.0 |
| deepagents | `deepagents` | 0.1.4 – 0.6.12 |

Upper bounds are the latest tested, not hard ceilings. deepagents has no adapter of
its own — it rides the LangChain adapter.


docs/adapters/overview.mdx
```diff
@@ -0,0 +1,53 @@
...
+
+## Compatible framework versions
+
+hexgate's adapters are exercised against a framework-version matrix. Policy
+enforcement is verified across these ranges:
+
+| Framework | Package | Verified compatible | Notes |
+| --- | --- | --- | --- |
+| OpenAI Agents SDK | `openai-agents` | **0.8.0 – 0.18.3** | below 0.8.0 the adapter cannot import (`RunState` was added in 0.8.0) |
+| LangChain / LangGraph | `langchain` | **1.0.0, 1.2.15 – 1.3.14** | 1.1.0 and 1.2.5 have a known import break — use 1.2.15+ |
```
quanghexa94: This row lists langchain `1.0.0` as verified-compatible, but the pyproject floor is `langchain>=1.2.15`, which makes 1.0.0 uninstallable. The two contradict each other in the same PR. A user who reads 'Verified compatible: 1.0.0, 1.2.15 – 1.3.14' and pins `langchain==1.0.0` to match gets a resolver failure, and the closing note on line 52 ('Minimums ... are enforced by hexgate's requirements') reinforces the floor. Either drop 1.0.0 from the table, or mark it 'compatible but below the enforced floor'.

docs/adapters/langchain.mdx
```diff
@@ -5,9 +5,15 @@ description: "wrap_langchain_agent — policy-gated tools, in-place installation
...
 <Note>
 New to the adapters? Start with the [wrapping overview and comparison
-table](/adapters/openai#overview-how-framework-wrapping-works).
+table](/adapters/overview#how-framework-wrapping-works).
 </Note>
 
+<Info>
+**Compatible versions:** see the [compatibility
+table](/adapters/overview#compatible-framework-versions) for the verified
+`langchain` range (it also covers deepagents, which rides this adapter).
```
quanghexa94: The Info block says the compatibility table covers deepagents via 'this adapter's `langchain` range', but the table gives deepagents its own distinct row and range (0.1.4 – 0.6.12), not the langchain range. A deepagents user following this pointer lands on the langchain row and reads the wrong supported range. Point the note at the deepagents row specifically.

pyproject.toml
```diff
@@ -29,8 +31,10 @@ dependencies = [
     "pyyaml>=6.0.2",
     "rich>=13.9.4",
     "websockets>=13.0",
-    "openai-agents>=0.0.10",
-    "langgraph>=0.2",
+    # Floor 0.8.0: the adapter imports `RunState` (added in 0.8.0); below that it
+    # fails to import. Verified by the framework-version matrix (tests/framework_compat/).
+    "openai-agents>=0.8.0",
+    "langgraph>=1.1.10",
```
quanghexa94, line 40: langgraph floor jumps `>=0.2` to `>=1.1.10`, a large bump, but langgraph has no row in the compatibility table and no note explaining the floor. Only langchain is documented. A user hitting a langgraph resolution conflict checks the table for guidance and finds nothing. The langchain floor comment (line 23) explains 1.2.15, but the paired langgraph 1.1.10 minimum is undocumented. Add a one-line note or a langgraph sub-entry so the bump is traceable.

quanghexa94 (commented):
Ran /code-review at high effort. The pydantic_ai `usage` property-vs-method fix is correct and tested (both the 1.x callable form and the new 2.x property form, and RunUsage instances aren't callable so the property path is safe). All internal links repointed from `/adapters/openai#overview-...` to `/adapters/overview#how-framework-wrapping-works` with no stragglers, and the new heading anchors resolve. The three findings below are consistency issues between the docs table and the dependency floors, not code bugs.

guillaume-hexamind:
@quanghexa94 Thanks for the review.

All three consistency findings fixed:

1. overview table vs floor: dropped langchain 1.0.0 from the verified range (now 1.2.15 - 1.3.14); noted separately as tests-clean-but-below-the-enforced-floor.
2. deepagents pointer: langchain note now sends deepagents users to their own table row, not the langchain range.
3. langgraph floor: documented langgraph>=1.1.10 in pyproject (companion to langchain>=1.2.15) and added it to the overview's enforced-minimums note.

Confirmed the usage property/method fix and the link repointing, no action needed there.

quanghexa94 (approved):
LGTM

PR #93, opened 2026-07-20
fix(hexgate): pydantic ai run sync no loop drop

victorludvig:
## What is changing

`AuditSender.emit()` (`hexgate/tracing/_senders.py`) now falls back to a bounded, non-daemon background thread + a plain `httpx.Client` when no asyncio event loop exists anywhere in the process, instead of warning once and silently dropping the event. `close()` now tracks and joins those fallback threads before tearing down the shared sync client, and the diagnostic drop counter is now lock-guarded against concurrent OS threads.

To read more about daemon / non-daemon threads, see [this Notion doc](https://app.notion.com/p/How-CPython-shuts-down-threads-at-exit-daemon-vs-non-daemon-3a3fb45dbae281e1afabc5bf16fa1fb4).

For a caller built purely on `pydantic_ai`'s `run_sync()` (no event loop ever present), every decision/LLM-usage event was being dropped deterministically — `AuditSender` had no live loop to schedule the send on. The two follow-up commits close two races the initial fix introduced: `close()` could tear down the sync client while a send was still in flight (a real case, since `run_sync()` scripts commonly call `asyncio.run(hexgate.shutdown())` as their last action), and the saturation counter could lose updates under concurrent callers.

## Tests

- make check-all
- `tests/adapters/pydantic_ai/test_integration.py` — opt-in (`pytest -m integration`) end-to-end test against a real platform-api + ClickHouse, asserting the event actually lands (not just "no exception").
- Manually verified live: reverted the fix, reproduced the silent drop against real ClickHouse; restored it, confirmed delivery; reproduced the `run_sync()` → `asyncio.run(shutdown())` race live, confirmed no error and the event still lands.

### DIFF
--- a/hexgate/tracing/_senders.py
+++ b/hexgate/tracing/_senders.py
@@ -108,8 +108,16 @@ class AuditSender:
         )
 
     def _new_sync_client(self) -> httpx.Client:
+        # Connect gets its own short budget, separate from the read/write/pool
+        # timeout: an unreachable host (dead IP, packet black hole) fails at
+        # the connect stage, and this thread is non-daemon, so a slow connect
+        # timeout directly extends how long process exit can hang on it
+        # (see close()'s finalizer-thread comment for why that thread can
+        # outlive an explicit close() entirely). A live host's TCP handshake
+        # is sub-second, so 2s is generous for the happy path while still
+        # capping the unreachable-host case well below self._http_timeout.
         return httpx.Client(
-            timeout=self._http_timeout,
+            timeout=httpx.Timeout(self._http_timeout, connect=2.0),
             headers={"Authorization": f"Bearer {self._api_key}"},
         )
 
@@ -249,15 +257,18 @@ class AuditSender:
         Runs entirely on its own thread; always releases
         ``self._sync_semaphore`` so the next caller past the cap can
         proceed, and discards itself from ``self._sync_threads`` so close()
-        isn't left waiting on threads that already finished."""
+        isn't left waiting on threads that already finished.
+
+        Unlike ``_send``, does not retry on a 503: this thread is
+        non-daemon (see ``_spawn_sync_send``), so on a run_sync()-only
+        script that never calls ``hexgate.shutdown()``, whatever it's
+        blocked on directly extends how long process exit hangs. A retry
+        would add a full extra ``self._http_timeout`` plus jitter sleep to
+        that; one attempt keeps the bound to a single request."""
         try:
             payload = event.as_payload()
             try:
                 response = self._sync_client.post(self._endpoint, json=payload)
-                if response.status_code == 503:
-                    delay = min(self._http_timeout, 2.0)
-                    time.sleep(random.uniform(delay / 2, delay))
-                    response = self._sync_client.post(self._endpoint, json=payload)
                 if response.status_code >= 400:
                     _log.error(
                         "audit ingest failed: %s %s",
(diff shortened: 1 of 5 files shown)


hexgate/tracing/_senders.py
```diff
@@ -171,6 +294,47 @@ async def close(self, drain_timeout: float = 5.0) -> None:
...
+            )
+            if timed_out:
+                _log.warning(
+                    "audit close: sync drain timed out with %d thread(s) still "
+                    "running; leaving the sync client open rather than closing "
+                    "it out from under them",
+                    timed_out,
+                )
+                return
+        self._sync_client.close()
```
guillaume-hexamind, line 351: close() leaks _sync_client when the sync drain times out. On the timeout branch the comment counts on "a later close() call" to clean up. But shutdown() clears _senders and awaits close() once, then drops the last reference so that later call could never come and the httpx.Client (pool + sockets) leaks for the rest of the process. 

Could we hand the close off to a small finalizer thread that joins the stragglers and then closes the client? They're non-daemon so the interpreter waits for them anyway, and close() stays bounded.

hexgate/tracing/_senders.py
```diff
@@ -155,9 +192,95 @@ async def _send(self, event: _PayloadEvent) -> None:
...
+                dropped = self._sync_dropped
+            if dropped % 100 == 1:
+                _log.warning(
+                    "audit sender saturated (no event loop path); %d events "
+                    "dropped (platform slow, or too many concurrent "
+                    "synchronous callers?)",
+                    dropped,
+                )
+            return
+        thread = threading.Thread(
```
guillaume-hexamind, line 238: The non-daemon send thread stalls process exit when the platform is unreachable. daemon=False is the right call for delivery, but it means CPython joins the thread at shutdown, bounded only by http_timeout (5s) — plus ~2s sleep and a second POST on a 503. So a run_sync()-only script that used to exit instantly now hangs several seconds per outstanding send when the platform is down — which is exactly what the now-deleted test_no_running_loop_skips_silently guarded against. A short dedicated connect timeout plus skipping the 503 retry on this path would cap most of it; if we want the old "exit never hangs" guarantee back, daemon threads + an atexit bounded drain gets both. We could either way add a test for bounded exit against a dead endpoint.

guillaume-hexamind (commented):
Thanks for the PR and documentation, I approve of the global fix logic.

You can check the two comments below on some 'unhappy' paths issues; I guess the one on the sync client close leak would be the one to fix before merge

hexgate/tracing/_senders.py
victorludvig: Good observation, I fixed it by spawning a non-daemon thread in close() when the sync drain times out. That thread waits for the still-running straggler(s), the emit-decision/llm-usage sync-fallback threads that didn't finish within drain_timeout, and only then closes the shared httpx.Client, instead of close() giving up and counting on a "later close() call" that never actually comes.

Summary of the fix (hexgate/tracing/_senders.py):

When close()'s sync drain times out with a straggler thread still running, it now spawns a non-daemon finalizer thread (_finalize_sync_close) that blocks on the straggler(s) and closes self._sync_client once they actually finish — instead of just returning and counting on a "later close() call" that shutdown()'s one-shot semantics never provide.
_join_sync_threads now returns the list of still-alive threads (was just a count) so the finalizer knows what to join.
Updated tests/tracing/test_sender.py: renamed/rewrote the test that used to manually call close() a second time to instead verify the finalizer closes the client on its own once the straggler completes.

hexgate/tracing/_senders.py
victorludvig: Fixed — sync client now gets a dedicated 2s connect timeout instead of sharing the full 5s http_timeout, and the 503 retry is dropped on this path only, since it's a non-daemon thread and a retry would double how long process exit hangs when shutdown() is never called.

victorludvig (commented):
Thank you for the review, I addressed both comments.

guillaume-hexamind (approved):

PR #95, opened 2026-07-23
feat(sdk): optional egress proxy — gate outbound network by host

quanghexa94:
## What & why

Today Hexgate authorizes the tool calls a model asks for. It can't see where a tool actually goes on the network. If a tool shells out to `curl`, or an SDK makes its own HTTP request, it can reach any host and policy never notices.

This adds an optional **egress proxy**. Point the agent's process at it and every outbound HTTP(S) request runs through the same policy engine, checked on the destination host. You can say "this agent may only reach `api.github.com`" and the rest is refused before anything leaves the process.

The nice part: egress is treated as one more gated tool, `net.http_request`. It reuses the existing `Decision`, the constraint DSL, and the audit stream. There's no new engine and no change to how policy is evaluated.

## How it works

```
 agent code                 egress proxy                    destination
 httpx / curl / SDK   ──►  decide("net.http_request",  ──┬── allow ──►  api.github.com
   (unchanged)               {host, scheme, port})       │
                                                         └── deny  ──►  403 refused
                                    │
                                    └──►  same audit log as the tool-call decisions
```

For HTTPS the proxy reads the host from the plaintext `CONNECT` line, decides, then tunnels the encrypted bytes without reading them. So it gates the host, not the path or body. Plain HTTP exposes the full URL. No TLS is decrypted.

Routing uses the standard `HTTP_PROXY` / `HTTPS_PROXY` env vars that `egress_guard` sets, so the agent's request code doesn't change.

## Where to look (in order)

| File | What it is | Priority |
|---|---|---|
| `hexgate/egress/gate.py` | the one place egress talks to the enforcer; also identity + approval | start here |
| `hexgate/egress/proxy.py` | the asyncio server + `egress_guard()` env wiring | high |
| `hexgate/egress/transport.py` | CONNECT tunnel (host-level) + plain-HTTP forwarding; the seam a future TLS-intercepting transport slots into | high |
| `hexgate/security/network.py` + `net_allow` in `hexgate/security/builder.py` | the policy sugar (`net_allow` / `net_approve`) | important |
| `hexgate/egress/model.py`, `hexgate/egress/wire.py` | small plumbing (request parsing, byte relay) | skim |
| `docs/concepts/egress.mdx` | the reader-facing explanation | context |

## Using it

```python
policy = PolicyBuilder(default="deny").net_allow(hosts=["api.github.com"]).build()
enforcer = build_enforcer(load_policy_set(policy), agent_name="support-agent")

async with egress_guard(enforcer, User(user_id="alice", role="agent")):
    ...   # every outbound HTTP(S) request here is gated
```

Two runnable demos: `examples/egress_demo.py` (plain script) and `deploy/egress_gate_demo.py` (marimo notebook, code-only, no platform or API key).

## Limits (called out on purpose)

- Env-proxy based, so it's intent-shaping in a plain SDK deployment. A tool that unsets the vars, or sets `trust_env=False`, bypasses it. Making it unbypassable is a sandbox job.
- Host-level only on HTTPS. Seeing the path or body would need TLS interception, which fits a controlled runtime rather than a library.
- One guard per process (it sets global proxy env vars; a nested guard raises instead of clobbering).

## Tests

46 egress + network tests, including a pydantic↔WASM parity check so the sugar behaves the same on the compiled bundle. The full security suite stays green (788 passing).

## Commits

- `feat(sdk)`: the module, the `net_allow` sugar, tests, and the two demos.
- `docs(sdk)`: new `concepts/egress` page, `net_allow` docs under "Define policy in code", and a cross-link from the sandbox page.

### DIFF
--- a/deploy/egress_gate_demo.py
+++ b/deploy/egress_gate_demo.py
@@ -1,0 +1,241 @@
+"""Hexgate egress-gate demo (marimo) — gate outbound network by policy, in code.
+
+A code-only tour of the optional egress enforcement plane: no platform, no API
+key, no YAML on disk. The policy is written with `PolicyBuilder` + `C`, and the
+same policy both (1) decides hosts offline and (2) drives a *live* in-process
+forward proxy that this notebook's own HTTP client is routed through — so you
+watch a real request get allowed or refused at the network layer.
+
+Run with `uv run --with marimo marimo edit deploy/egress_gate_demo.py`.
+"""
+
+import marimo
+
+__generated_with = "0.23.10"
+app = marimo.App(width="medium")
+
+
+@app.cell
+def _():
+    import inspect
+
+    import httpx
+    import marimo as mo
+
+    from hexgate import PolicyBuilder, User
+    from hexgate.egress import NET_TOOL, egress_guard
+    from hexgate.security.enforcer import build_enforcer
+    from hexgate.security.policy_set import load_policy_set
+
+    return (
+        NET_TOOL,
+        PolicyBuilder,
+        User,
+        build_enforcer,
+        egress_guard,
+        httpx,
+        inspect,
+        load_policy_set,
+        mo,
+    )
+
+
+@app.cell
+def _(mo):
+    mo.md(
+        """
+        # 🌐 Hexgate — gating network egress
+
+        Hexgate normally gates the **tool call** the model proposes. But a tool
+        that runs `curl`, or an SDK that makes its own HTTP calls, can reach the
+        network *without* the tool-argument policy seeing the real destination.
+
+        The **egress proxy** closes that gap: it routes the process's outbound
+        HTTP(S) through the *same* policy engine, mapped to a synthetic
+        `net.http_request` tool. Network egress becomes **just another gated tool**.
+
+        This demo is **code-only** — no platform, no API key, no YAML file. The
+        policy is written in Python and drives a live in-process proxy.
+
+        > **Tier 1 (this demo):** HTTPS is gated on the `CONNECT` host — visible in
+        > plaintext before TLS begins. The tunnel relays ciphertext untouched; we
+        > never decrypt. So constraints are on `args.host` / `args.scheme` /
+        > `args.port`, not the path or body.
+        """
+    )
+    return
+
+
+@app.cell
+def _(mo):
+    mo.md("""## 1 · The policy, written in code""")
+    return
+
+
+@app.cell
+def _(PolicyBuilder, load_policy_set):
+    # Deny by default; allow HTTPS to two GitHub hosts — via the net_allow sugar,
+    # which renders the host + scheme allowlist into ordinary constraints (shown
+    # below as the YAML it produces). schemes defaults to HTTPS-only.
+    def build_policy():
+        return (
+            PolicyBuilder(default="deny")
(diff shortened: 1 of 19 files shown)


victorludvig:
Hello @quanghexa94 thank you for this solid PR. No real issue overall, mostly nits, but can you check these findings:

#: 1
Issue: Dangling doc reference — three docstrings cite egress-proxy-plan.md, which was never added to the repo
hexgate/egress/__init__.py, hexgate/egress/model.py, hexgate/egress/proxy.py
Score: 75
Why: Verified real (confirmed via git log --all, file never existed). Docked below 75→80 because it's a docs-only defect with no functional/security impact.
────────────────────────────────────────
#: 2
Issue: EgressProxy.stop() doesn't cancel in-flight connection tasks — a CONNECT tunnel open when the guard exits keeps relaying after env vars are restored
hexgate/egress/proxy.py (stop(), ~L850-859)
Score: 72
Why: Verified real via code inspection (no task tracking exists). Docked for being a lifecycle/cleanup gap rather than a security bypass — traffic was already authorized
before the tunnel opened.
────────────────────────────────────────
#: 3
Issue: http_to_args() hardcodes port default to 80 regardless of scheme, so an https:// absolute-URI request with no explicit port yields inconsistent scheme=https,
port=80 args
hexgate/egress/model.py (~L714-723)
Score: 70
Why: Verified reachable in code, but requires a non-standard client (RFC 7230 says HTTPS proxying should use CONNECT, not absolute-form GET) to trigger.
────────────────────────────────────────
#: 4
Issue: read_headers() treats client EOF (b"") identically to a blank-lintly accepting truncated header blocks
hexgate/egress/wire.py (~L1131-1139)
Score: 60
Why: Confirmed as a genuine "fail-closed" design inconsistency, but no actual bypass — the truncated request just gets rejected by the upstream server instead of by
hexgate.
────────────────────────────────────────
#: 5
Issue: hexgate/egress/__init__.py docstring claims the package "depends only on hexgate.security/hexgate.runtime/hexgate.audit," but omits the real direct import of
hexgate.approvals and lists hexgate.audit (only a transitive dependency
hexgate/egress/__init__.py (~L8-9)
Score: 50
Why: Verified as a real inaccuracy, but a minor documentation nitpick with no functional consequence.

winlp4ever:
Thanks @victorludvig, good catches. All five are addressed in `4b8c9dd` (pushed):

1. Dropped the `egress-proxy-plan.md` references. That's a local planning doc that was never committed, so the docstrings now point at `docs/concepts/egress.mdx` instead.
2. `stop()` now tracks the connection-handler tasks and cancels them, so an open CONNECT tunnel can't keep relaying after the guard exits. One sting worth flagging: `Server.wait_closed()` on Python 3.12+ waits for active connections, so the cancel has to happen *before* it, otherwise an open tunnel deadlocks `stop()`. Added a test that opens a tunnel and asserts it's torn down on stop.
3. `http_to_args` now takes the default port from the scheme (443 for `https`, else 80).
4. `read_headers` raises on a client EOF mid-block instead of accepting a truncated header block.
5. Fixed the `__init__` dependency note: `hexgate.approvals` is a direct import, and `hexgate.audit` is transitive via the enforcer.

There are tests for each, and egress patch coverage is up to ~92%. Ready for another look whenever you have a minute.

victorludvig (approved):

PR #96, opened 2026-07-24
feat(sdk): raw-TCP egress reachability gate (gate DB connections by host:port)

quanghexa94:
## What & why

The egress proxy (#95) gates HTTP(S) by host. But a database driver opens a raw TCP socket and ignores `HTTP_PROXY`, so DB traffic isn't covered by it at all.

This adds a raw-TCP **reachability gate**. Point a connection string at the proxy and it decides `net.tcp_connect{host, port}` before any bytes flow (so before TLS), then tunnels the connection or drops it. Same `PolicyEnforcer`, same `Decision`, same audit stream as tool calls and HTTP egress.

Scope worth stating up front: it gates the *destination* ("may this agent reach `db.internal:5432`"), not the SQL sent over the connection. Query-level inspection would need a protocol-aware, connection-terminating proxy per database, which this is not.

## How it works

```
 db driver / client           tcp egress proxy               target
 conn string -> proxy   ──►  decide("net.tcp_connect",  ──┬── allow ──►  db.internal:5432
   (proxy host:port)           {host, port})              │              (raw tunnel, TLS untouched)
                                                          └── deny  ──►  socket dropped
                                    │
                                    └──►  same audit log as the tool + HTTP decisions
```

Because the decision happens before the TLS handshake, a TLS'd database connection is gated without being decrypted.

## Where to look (in order)

| File | What it is | Priority |
|---|---|---|
| `hexgate/egress/tcp.py` | `TcpEgressProxy` + `tcp_egress_guard(target=(host,port))`: gate then tunnel-or-drop | start here |
| `hexgate/egress/server.py` | **New** shared `ProxyServer` base for the socket lifecycle. The cancel-in-flight-before-`wait_closed()` `stop()` now lives once; both proxies subclass it | high — this refactor also touches the HTTP proxy |
| `hexgate/security/builder.py` + `network.py` | `net_tcp_allow` / `net_tcp_approve` and the `net.tcp_connect` tool | important |
| `hexgate/egress/gate.py`, `wire.py` | `Gate` gains a `tool` param (one seam, two planes); `open_upstream` bounds the upstream connect with a timeout | skim |
| `examples/tcp_egress_demo.py` | mock-DB demo, no real database needed | context |

## Using it

```python
policy = PolicyBuilder(default="deny").net_tcp_allow(hosts=["db.internal"], ports=[5432]).build()

async with tcp_egress_guard(enforcer, User(user_id="alice", role="agent"),
                            target=("db.internal", 5432)) as proxy:
    dsn = f"postgresql://app@{proxy.host}:{proxy.port}/db"
    ...   # connections opened through this DSN are gated
```

## Notes and limits

- **Routing is opt-in.** DB drivers don't read `HTTP_PROXY`, so the app points its connection string at `proxy.host:proxy.port` (or a sandbox redirect forces traffic through it). Intent-shaping unless a sandbox enforces it, same as the HTTP proxy.
- **Reachability only.** Gates host + port, not the SQL/commands on the connection.
- Builds on #95, and refactors the shared server lifecycle out of `EgressProxy` into `ProxyServer` so the deadlock-prone teardown isn't duplicated.

## Tests

TCP allow / deny (host and port) / approval / identity / upstream-unreachable / stop-cancels-open-tunnel / guard, a pydantic↔WASM parity check for `net.tcp_connect`, an `open_upstream` timeout test, and the full HTTP-egress + security suites still pass.

### DIFF
--- a/deploy/egress_gate_demo.py
+++ b/deploy/egress_gate_demo.py
@@ -17,13 +17,14 @@ app = marimo.App(width="medium")
 
 @app.cell
 def _():
+    import asyncio
     import inspect
 
     import httpx
     import marimo as mo
 
     from hexgate import PolicyBuilder, User
-    from hexgate.egress import NET_TOOL, egress_guard
+    from hexgate.egress import NET_TOOL, egress_guard, tcp_egress_guard
     from hexgate.security.enforcer import build_enforcer
     from hexgate.security.policy_set import load_policy_set
 
@@ -31,12 +32,14 @@ def _():
         NET_TOOL,
         PolicyBuilder,
         User,
+        asyncio,
         build_enforcer,
         egress_guard,
         httpx,
         inspect,
         load_policy_set,
         mo,
+        tcp_egress_guard,
     )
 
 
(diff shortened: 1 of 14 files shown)


victorludvig (approved):
LGTM

PR #97, opened 2026-07-24
feat(sdk): multi-module policy — modules + linker (1/3)

quanghexa94:
First of 3 PRs adding **multi-module policy**: instead of one policy file per agent, an agent's policy becomes a stack of independently-authored modules that compose into one effective policy.

- **PR 1 (this one)** — SDK core: the module model + the linker.
- **PR 2** — analyzer: lints (dead / shadowed / conflicting rules), all file-attributed.
- **PR 3** — dashboard editor + resolved-policy view.

Everything here is SDK-side and fully local (no platform, no key).

## What it does

A bundle of boundary + capability modules is folded into a single `AgentPolicy`, which then compiles to Rego → WASM exactly as today. The composition happens at the model layer, so the engines only ever see one policy and the pydantic-vs-WASM parity gate is untouched.

```
many *.yaml modules            link()              one effective policy
  boundaries  (caps + denies) ──────▶  fold  ──────▶  AgentPolicy  ──▶ rego ──▶ wasm ──▶ signed bundle
  capabilities (grants)          (the new step)                        (all unchanged)
```

Two tiers, one rule set: **fences intersect, grants union, denies win.**

- **Boundaries** cap and hard-deny (cf. AWS permissions boundary). A boundary `allow` is a ceiling (a cap), not a grant.
- **Capabilities** only grant. A capability that tries to `deny` is a hard error, so a lower-authority layer can never loosen a boundary.

You can drive it end to end:

```
hexgate policy resolve --dir deploy/demo_policies    # prints the effective policy + provenance
marimo edit deploy/policy_modules_demo.py            # interactive tour
```

## Important files (suggested review order)

| File | What to look at |
| --- | --- |
| `hexgate/security/linker.py` | The fold. Start here — `_fold_tool` is the whole thing. |
| `hexgate/security/modules.py` | Data model: `ModuleContent`, provenance, `LinkResult`. |
| `hexgate/security/module_loader.py` | Local-files loader (the seam the platform swaps later). |
| `hexgate/cli/policy/main.py` | `hexgate policy resolve` — the inspection command. |
| `tests/security/test_linker.py` | Fold behavior, ceiling/floor, and the pydantic-vs-WASM parity check. |
| `docs/internals/policy-modules.md` | Architecture + reconciliation rules. |
| `deploy/policy_modules_demo.py` | Runnable marimo demo + sample bundle. |

## Notes

- "boundary" (not "guardrail") for the security-owned tier, to avoid collision with LLM content/safety guardrails and with the ceiling/floor posture.
- Not wired into the platform or versioning yet — that's later. `default_policy` "floor" is expressed as `mode: allow` for now (no dedicated sentinel).
- Full security suite green (909 tests, incl. an opa-gated parity test on resolved policies).

### DIFF
--- a/deploy/demo_policies/policies/boundaries/org_core.yaml
+++ b/deploy/demo_policies/policies/boundaries/org_core.yaml
@@ -1,4 +1,4 @@
-# Org guardrail (security-owned). Floor posture: default_policy allow means it
+# Org boundary (security-owned). Floor posture: default_policy allow means it
 # only SUBTRACTS what it names — unlisted tools pass through to capabilities.
 # Its `allow` on refund_order is a CEILING (a cap), not a grant.
 default_policy: { mode: allow }
--- a/deploy/demo_policies/policies/capabilities/payments.yaml
+++ b/deploy/demo_policies/policies/capabilities/payments.yaml
@@ -1,0 +1,4 @@
+# Payments capability pack (team-owned). Grants only.
+tools:
+  refund_order: { mode: allow, constraints: ['args.currency in ["USD", "EUR"]'] }
+  lookup_order: { mode: allow }
--- a/deploy/demo_policies/policies/capabilities/support_leaf.yaml
+++ b/deploy/demo_policies/policies/capabilities/support_leaf.yaml
@@ -1,0 +1,4 @@
+# The support agent's own inline leaf (capability). Grants only.
+tools:
+  send_email: { mode: allow }
+  escalate: { mode: approval_required }
--- a/deploy/demo_policies/policies/guardrails/org_core.yaml
+++ b/deploy/demo_policies/policies/guardrails/org_core.yaml
@@ -1,0 +1,7 @@
+# Org guardrail (security-owned). Floor posture: default_policy allow means it
+# only SUBTRACTS what it names — unlisted tools pass through to capabilities.
+# Its `allow` on refund_order is a CEILING (a cap), not a grant.
+default_policy: { mode: allow }
+tools:
+  delete_database: { mode: deny }                       # absolute — no layer can undo
+  refund_order: { mode: allow, constraints: ["args.amount <= 1000"] }   # cap
--- a/deploy/policy_modules_demo.py
+++ b/deploy/policy_modules_demo.py
@@ -5,7 +5,7 @@ local — no platform, no API key):
 
   many module .yaml files  →  link()  →  one effective policy  →  (rego → wasm)
 
-You'll see the two tiers compose — **guardrails** (caps + hard denies) over
+You'll see the two tiers compose — **boundaries** (caps + hard denies) over
 **capabilities** (grants) — under the rule *fences intersect, grants union,
 denies win*. Edit a decision's args and watch allow / deny / approval resolve
 live. If `opa` is on PATH, each decision is also checked against the compiled
@@ -79,7 +79,7 @@ def _(mo):
         Instead of one policy file per agent, an agent's policy is a **stack of
         modules** composed into one effective policy:
 
-        - **Guardrails** (security-owned) — caps + hard denies. A guardrail
+        - **Boundaries** (security-owned) — caps + hard denies. A boundary
           `allow` is a *ceiling*, not a grant.
         - **Capabilities** (team-owned) — additive grants only.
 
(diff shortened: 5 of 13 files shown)


guillaume-hexamind:
Claude auto-review:

1. Ceiling guardrail via conditional-deny fails to gate eligibility — 92/100
hexgate/security/linker.py:177

A ceiling guardrail with default_policy: deny that mentions a tool only via a conditional deny (mode: deny with constraints) never gates eligibility. Step-2 shadowing only fires when tp is None, so the tool is neither shadowed nor capped, and a capability grant for it survives.

Failure: Org ceiling has default_policy: deny and lists refund_order: {mode: deny, constraints: ["args.amount > 10000"]} but never grants it; a capability grants refund_order: allow. Expected → refund_order denied (not in ceiling's allow set). Actual → not shadowed (tp is not None), gets no cap, grant applies minus the deny region → refund_order ALLOWED for any amount ≤ 10000 → a tool the org ceiling never granted becomes broadly permitted.

2. _merge_consts last-wins on colliding const names across guardrails — 90/100
hexgate/security/linker.py:106 (same root cause at :116)

_merge_consts hard-errors when a capability redefines a guardrail constant, but silently last-wins (guard_consts.update(...)) when two guardrails define the same const name with different values. A lower-precedence guardrail can overwrite a higher one's cap constant.

Failure: Org guardrail sets consts.max_refund = 100 with cap args.amount <= consts.max_refund; a team guardrail in the bundle sets consts.max_refund = 100000. After merge, effective policy carries 100000 → the org's hard cap is loosened and a $100k refund the org meant to block is authorized, with no error or warning.

🟠 High — valid-input crashes

3. _canonical_hash called outside the parse/validate try/except — 68/100
hexgate/security/module_loader.py:81

_canonical_hash(payload) runs outside the try/except wrapping parse+validate, and json.dumps can't serialize non-JSON scalars YAML produces (e.g. datetime.date), raising an uncaught TypeError.

Failure: A module sets consts: {window_start: 2026-01-01} (unquoted date). YAML → datetime.date, AgentPolicy (consts: dict[str, Any]) validates fine, then _canonical_hash calls json.dumps(date) → TypeError escapes the loader's except Exception. Crash with a traceback that names no offending file — the opposite of the loader's stated intent.

4. _main_resolve loader call is outside the error-handling try — 60/100
hexgate/cli/policy/main.py:452 (reported 3×, one root cause)

_main_resolve calls load_local_modules(args.dir) before its try/except; the loader raises a bare ValueError on malformed/schema-invalid modules, and main() dispatches args.func(args) unwrapped, so the exception is uncaught.

Failure: hexgate policy resolve --dir . on a bundle with one bad module YAML prints a raw traceback + crashes, instead of the intended link error: ... + exit 1 that the try block gives for link failures — inconsistent with every other policy verb's UX.

guillaume-hexamind (changes requested):
I agree with the logic overall.
I let you check these issues found by claude, the first two (ceiling eligibility and const collision) are real issues that can be fixed pretty easily.

The one thing I really want to push back is the 'guardrail' naming. It collides with the actual concept of LLM guardrails, that we actually wanna implement as plugin for the sdk in the future; and can be easily replaced by something with similar clarity.

Changing it to kind: boundary (AWS has some similar "permissions boundary"), or kind: ceiling seems like good alternatives.

winlp4ever:
Thanks - all addressed, 909/909 green.

**Fixed** (`a038ad9`):

- **Ceiling eligibility** - a boundary that mentions a tool only via a conditional deny (or not at all) no longer lets a capability grant slip through. It must explicitly allow/approve the tool to make it eligible; otherwise the tool is ineligible and the grant is shadowed. Test added.
- **Const collision** - two boundaries defining the same constant with different values is now a `LinkError`, matching the capability-override guard (no more silent last-wins). Test added.
- **Loader hash** - the content hash is computed inside the try (with `json.dumps(default=str)`), so a YAML date scalar can't raise a `TypeError` that names no file.
- **`policy resolve`** - loader errors are caught and printed as `load error: ...` instead of a raw traceback, consistent with the other verbs.

**Naming** (`2d72bf6`): renamed the tier `guardrail` to `boundary` (the `kind:` value and the `policies/boundaries/` dir). Agree on the collision with LLM content/safety guardrails. Went with `boundary` over `ceiling` because "ceiling" is already the name of the posture a boundary takes (ceiling vs floor), so it would just move the collision; `boundary` also maps cleanly to AWS permissions boundaries.

Note: I don't have edit rights on the PR description from my account, so its wording still says "guardrail" in places - stale until someone with write access updates it. The reconciliation rules are written up in `docs/internals/policy-modules.md`.

guillaume-hexamind (approved):
Good to me, thanks for the changes. Would be a plus if you could update the PR's description before merge!

winlp4ever:
description already updated along with the changes

PR #98, opened 2026-07-29
docs(sdk): replace readme decision-flow ascii diagram with svg

victorludvig:
## What is changing?
The ASCII README diagram is replaced with a SVG diagram.

Ships as light/dark variants via <picture> so it renders correctly in both GitHub themes.


winlp4ever (approved):
LCTM (look cool to me) !

PR #99, opened 2026-07-30
test(sdk): add live integration tests for all four framework adapters

victorludvig:
## What is changing

Adds an opt-in (`-m integration`) live test for each of the four SDK framework adapters (openai, google, langchain, pydantic_ai), extending the existing pydantic_ai sync/no-loop regression test into full coverage across all four.

Running these live surfaced three real bugs in the adapters themselves (not the tests), all fixed here:
- **google**: the last turn's fire-and-forget audit-send task could still be pending when the run's event loop closed, silently dropping it.
- **openai**: same class of bug on its persistent per-thread default loop, plus a case where an empty model name reached the audit payload.
- **audit sender**: cached loop reuse checked `is_closed()` instead of `is_running()`, which could hand back a loop mid-teardown.

Also silences Langfuse's background span export in tests when no real key is configured (`pytest_configure` sets `LANGFUSE_TRACING_ENABLED=false`) — `Langfuse()` defaults `tracing_enabled=True` even with empty keys, so any `@observe`-wrapped call was failing with an unrelated 401 during the openai integration test.

Once both loop-drain fixes existed independently, the duplication between them (and the near-identical Langfuse tag-building across all four adapters) became obvious — `hexgate/adapters/_common.py` now holds `drain_pending_tasks`/`langfuse_propagate_kwargs`, used by all four adapters. No behavior change there.

`tests/adapters/conftest.py` (the `hexgate_platform_env` fixture) and `tests/adapters/helpers.py` (`poll_until`, `assert_policy_and_usage_events_landed`) let the four adapter integration tests share one implementation instead of four copies.

Each adapter's wrapper mechanics were already covered by mocked unit tests, but nothing proved the actual live plumbing works: that policy is really pulled from the platform API at wrap time, and that a real tool-call decision and its LLM-usage event actually reach ClickHouse. That gap is exactly what let the three loop/draining bugs above ship unnoticed — they only show up when a real agent is wrapped and run against a live platform + ClickHouse, not against mocks.

## Tests

- Four new opt-in integration tests (one per adapter), each wrapping a real minimal agent with one real tool, driven by a free/deterministic fake model (no real paid LLM calls), asserting a `policy_decision` row and an `llm_invocation` row land correctly in ClickHouse.
- Two new regression tests for the loop-drain fixes (`test_run_drains_pending_tasks_before_closing_loop` for google, `test_run_sync_drains_pending_tasks_on_the_default_loop` for openai) — each verified by temporarily reverting its fix, confirming the test fails, then restoring the real code.
- Confirmed live: `pytest -m integration` against a real local platform-api + ClickHouse now passes all 6 selected tests (all four adapters + both `tests/audit/test_integration.py` cases) — this was the failing state a few iterations ago before the fixes above landed.
- Each adapter test's `agent_name` now embeds the framework (`integration_test_agent_openai_...`, `..._google_...`, etc.) so the audit dashboard's Decisions view shows which framework produced which row at a glance, instead of an indistinguishable random suffix.
- Full existing SDK test suite (1656 tests) still passes after the `_common.py` refactor — no regressions.

### DIFF
--- a/hexgate/adapters/_common.py
+++ b/hexgate/adapters/_common.py
@@ -10,20 +10,41 @@ import asyncio
 from typing import Any
 
 from hexgate.runtime import User
+from hexgate.tracing._senders import DEFAULT_DRAIN_TIMEOUT, pending_send_tasks
 
 
-def drain_pending_tasks(loop: asyncio.AbstractEventLoop) -> None:
-    """Give any still-pending tasks on ``loop`` one last chance to finish.
+def drain_pending_tasks(
+    loop: asyncio.AbstractEventLoop, *, drain_timeout: float = DEFAULT_DRAIN_TIMEOUT
+) -> None:
+    """Give hexgate's own still-pending audit-send tasks on ``loop`` one
+    last chance to finish.
 
     A fire-and-forget audit-send (policy decision / LLM usage) from the
     last turn can still be scheduled and pending once the top-level run
     settles, with nothing left to pump the loop for it — silently
-    abandoned otherwise. ``return_exceptions=True`` so a failed send can't
-    raise out of here and crash the caller's run.
+    abandoned otherwise. Scoped to ``pending_send_tasks(loop)`` rather than
+    every task on the loop — ``loop`` is a thread's shared default loop,
+    not one hexgate owns exclusively, and awaiting (or, on timeout,
+    cancelling) a caller's unrelated task would be a surprising, unrelated
+    side effect. ``return_exceptions=True`` so a failed send can't raise
+    out of here and crash the caller's run.
+
+    Bounded by ``drain_timeout``: an unreachable platform can hold a send
+    open for well over 10s (POST timeout, then a 503 jitter sleep, then a
+    retry POST), and this runs synchronously inside the caller's
+    ``run_sync()`` — a dropped audit event beats hanging its return.
     """
-    pending = asyncio.all_tasks(loop)
-    if pending:
-        loop.run_until_complete(asyncio.gather(*pending, return_exceptions=True))
+    pending = pending_send_tasks(loop)
+    if not pending:
+        return
+    try:
+        loop.run_until_complete(
+            asyncio.wait_for(
+                asyncio.gather(*pending, return_exceptions=True), drain_timeout
+            )
+        )
+    except TimeoutError:
+        pass
 
 
 def langfuse_propagate_kwargs(user: User, tag: str) -> dict[str, Any]:
--- a/hexgate/adapters/_common.py
+++ b/hexgate/adapters/_common.py
@@ -1,0 +1,37 @@
+"""Internal helpers shared across all four framework adapters.
+
+Not part of the public API — each adapter's own module is the supported
+import surface.
+"""
+
+from __future__ import annotations
+
+import asyncio
+from typing import Any
+
+from hexgate.runtime import User
+
+
+def drain_pending_tasks(loop: asyncio.AbstractEventLoop) -> None:
+    """Give any still-pending tasks on ``loop`` one last chance to finish.
+
+    A fire-and-forget audit-send (policy decision / LLM usage) from the
+    last turn can still be scheduled and pending once the top-level run
+    settles, with nothing left to pump the loop for it — silently
+    abandoned otherwise. ``return_exceptions=True`` so a failed send can't
+    raise out of here and crash the caller's run.
+    """
+    pending = asyncio.all_tasks(loop)
+    if pending:
+        loop.run_until_complete(asyncio.gather(*pending, return_exceptions=True))
+
+
+def langfuse_propagate_kwargs(user: User, tag: str) -> dict[str, Any]:
+    """Build the ``propagate_attributes(**kwargs)`` mapping for a Langfuse
+    span tagged ``tag``, carrying the active User's identity."""
+    return {
+        "tags": [tag],
+        "user_id": user.user_id,
+        "session_id": user.session_id,
+        "metadata": {"user_role": user.role},
+    }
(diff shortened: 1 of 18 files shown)


victorludvig:
### Code review

Found 1 issue:

1. `run_sync()`'s new pending-task drain isn't exception-safe, unlike the equivalent fix in the google runner

`_drain_default_loop()` runs as plain sequential code after the `with` blocks, not inside a `try/finally`. If `Runner.run_sync(...)` raises (tool error, policy/approval denial, API error), the drain is skipped entirely and the last turn's fire-and-forget audit-send task is left pending on the persistent default loop — exactly the class of bug this PR sets out to fix. The google runner's fix for the same bug class does drain inside `finally`, so this is only fixed on the success path.

https://github.com/HexamindOrganisation/hexgate/blob/52033663ba58461cc9be255ad449831407ebeb9b/hexgate/adapters/openai/runner.py#L215-L227

Compare to the exception-safe google fix:

https://github.com/HexamindOrganisation/hexgate/blob/52033663ba58461cc9be255ad449831407ebeb9b/hexgate/adapters/google/runner.py#L114-L130

victorludvig:
> ### Code review
> Found 1 issue:
> 
> 1. `run_sync()`'s new pending-task drain isn't exception-safe, unlike the equivalent fix in the google runner
> 
> `_drain_default_loop()` runs as plain sequential code after the `with` blocks, not inside a `try/finally`. If `Runner.run_sync(...)` raises (tool error, policy/approval denial, API error), the drain is skipped entirely and the last turn's fire-and-forget audit-send task is left pending on the persistent default loop — exactly the class of bug this PR sets out to fix. The google runner's fix for the same bug class does drain inside `finally`, so this is only fixed on the success path.
> 
> https://github.com/HexamindOrganisation/hexgate/blob/52033663ba58461cc9be255ad449831407ebeb9b/hexgate/adapters/openai/runner.py#L215-L227
> 
> Compare to the exception-safe google fix:
> 
> https://github.com/HexamindOrganisation/hexgate/blob/52033663ba58461cc9be255ad449831407ebeb9b/hexgate/adapters/google/runner.py#L114-L130
> 
> 
> - If this code review was useful, please react with 👍. Otherwise, react with 👎.

Fixed, added the finally clause in OpenAI's run_sync method.

hexgate/adapters/openai/usage.py
```diff
@@ -28,8 +28,14 @@ async def on_llm_end(
...
     ) -> None:
         # agent.model is `str | Model | None` — only the str case gives a
-        # clean name; a Model implementation has no guaranteed name field.
-        model = agent.model if isinstance(agent.model, str) else ""
+        # clean name; a Model implementation has no guaranteed name field
+        # (agents.models.interface.Model exposes none). Fall back to the
+        # instance's class name rather than "" — the platform rejects an
+        # empty `model` outright (min_length=1), silently dropping the
+        # event.
+        model = (
```
guillaume-hexamind: Good finding for the base error.

The type(agent.model).__name__ is probably not the best fallback though -> agent.model is None by default, which will give a 'NoneType' model in the manifest, while for a Model instance we should get OpenAIResponsesModel instead of the model id.

Fix idea:
if isinstance(agent.model, str):
    model = agent.model
elif agent.model is None:
    model = "default"  # agent uses the runner/SDK default model
else:
    # Standard Model impls expose the real id in `.model`; class name is a
    # last resort for an exotic Model that doesn't.
    model = getattr(agent.model, "model", None) or type(agent.model).__name__

hexgate/adapters/_common.py
```diff
@@ -0,0 +1,37 @@
...
+def drain_pending_tasks(loop: asyncio.AbstractEventLoop) -> None:
+    """Give any still-pending tasks on ``loop`` one last chance to finish.
+
+    A fire-and-forget audit-send (policy decision / LLM usage) from the
+    last turn can still be scheduled and pending once the top-level run
+    settles, with nothing left to pump the loop for it — silently
+    abandoned otherwise. ``return_exceptions=True`` so a failed send can't
+    raise out of here and crash the caller's run.
+    """
+    pending = asyncio.all_tasks(loop)
```
guillaume-hexamind: Should be fine in the majority of our cases, but could still be kinda dangerous to drain all pending tasks instead of only the audit ones. Do you think targeting only these tasks could be possible?

They can be accessed through AuditSender._tasks -> 
def pending_send_tasks() -> set[asyncio.Task[None]]:
    return {t for s in _senders.values() for t in AuditSender._tasks if not t.done()}

hexgate/adapters/_common.py
```diff
@@ -0,0 +1,37 @@
...
+
+from __future__ import annotations
+
+import asyncio
+from typing import Any
+
+from hexgate.runtime import User
+
+
+def drain_pending_tasks(loop: asyncio.AbstractEventLoop) -> None:
```
guillaume-hexamind: No timeout here -> unreachable platform can take 12s (5s POST + 503 jitter + 5s retry).

Fix idea: wait for pattern

tasks = [t for t in pending_send_tasks() if not t.done()]  # only Hexgate sends
if not tasks:
        return
    try:
        loop.run_until_complete(
            asyncio.wait_for(
                asyncio.gather(*tasks, return_exceptions=True), drain_timeout
            )
        )
    except asyncio.TimeoutError:
        pass  # bounded: a dropped audit beats a hung run() return

guillaume-hexamind (commented):
A couple of small comments to check, but I approve the integration and fixes logic

hexgate/adapters/openai/usage.py
victorludvig: Good catch, I implemented the proposed fix, along with a new unit test.

hexgate/adapters/_common.py
victorludvig: Good catch, I added a DRAIN_TIMEOUT=5 (s) in _sender (since we also have a 5s timeout when closing the AuditSender), and re-used it here.

hexgate/adapters/_common.py
victorludvig: Good catch, I fixed it.

guillaume-hexamind (approved):
Good to me!

PR #100, opened 2026-07-30
test(sdk): add enforcement + policy-pull latency benchmarks

guillaume-hexamind:
## Summary

Adds `benchmarks/` — a standalone, stdlib-only latency benchmark suite that isolates the deterministic costs the SDK adds around an agent's tool calls, kept separate from LLM and network variance. Not part of the shipped package.

## What's included

- **Bench A — enforcement (`bench_enforce.py`)** — local, no secrets, no network. Compiles `examples/devops_policy.yaml` to WASM via `opa` at startup, then times `wasm_instantiate`, `wasm_cache_hit`, raw `engine:<case>` eval, and full `enforcer:<case>` `decide()`. A four-case workload (fast allow / constrained allow / constraint deny / default deny) exercises every branch of the compiled decision tree, and each case's outcome is asserted before timing so a policy drift fails loudly instead of benchmarking the wrong branch.
- **Bench B — policy pull (`bench_pull.py`)** — targets `app.hexgate.ai` (needs `HEXGATE_API_KEY`). Separates deterministic crypto (`verify_only`) from variable network (`pull_cold_200`, `pull_warm_304`, `key_verify+jwks`).
- **Shared harness (`_report.py`)** — `perf_counter_ns` timing, nearest-rank percentiles, a p50/p95/p99/max table, and `--json PATH` to dump raw stats for diffing later runs.
- **README** with run instructions and per-segment explanations.

## Review hardening

- **Percentile** — nearest-rank now uses `ceil`, not `round`; `round` understated the tail at small N (e.g. p95 over 30 samples).
- **`pull_warm_304`** — skipped on the no-bundle (pydantic-fallback) branch, which never returns a 304, instead of mislabeling a full 200 pull as the 304 hot path.
- **Error handling** — `bench_pull` catches `RuntimeError`/`OSError` (invalid/tampered bundle, socket read-timeout) so it exits cleanly with a message instead of crashing.
- **Enforcer wrapper tax** — benched with a no-op `decision_observer` so `decide()` takes the same `copy.deepcopy(args)` branch every audited production enforcer takes; it was previously understated via the shallow-copy branch.
- Extracted magic literals to named constants.

## Notes

- The README "Indicative numbers" are illustrative from a single dev run — not a contract; re-run on the target host.
- Bench A measures `engine:` and `enforcer:` in separate passes; interleaving them per-case would give a more trustworthy engine-vs-enforcer delta (possible follow-up).

### DIFF
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -51,8 +51,13 @@ Segments:
   compiled bundle (pydantic-fallback shape).
 - `pull_cold_200` — full `fetch()`: round-trip + decode + verify (fresh source
   each sample forces a 200).
-- `pull_warm_304` — conditional `fetch()` with `If-None-Match` → 304. This is
-  the actual per-chat-turn refresh cost.
+- `pull_warm_304` — conditional policy `fetch()` (`If-None-Match` → 304), in
+  isolation.
+- `ban_warm_304` — the same conditional GET against `/v1/bans`, in isolation.
+- `policy+ban_b2b` — both awaited back to back, the way every invoke entrypoint
+  runs them (`_refresh_policy_safely` then `_check_ban`, sequential). **This is
+  the actual per-chat-turn refresh cost** — the two round-trips are additive,
+  not overlapped.
 
 `pull_*` rows carry real network variance — read min vs p99, not p50 alone, as
 the SDK's fixed overhead. `verify_only` is the part that stays constant across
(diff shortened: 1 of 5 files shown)


benchmarks/README.md
```diff
@@ -0,0 +1,79 @@
...
+From one dev-machine run (Apple Silicon, `app.hexgate.ai`) — **illustrative,
+not a contract**; re-run on the target host:
+
+| segment | p50 |
+|---|---|
+| `wasm_instantiate` (one-off) | ~135 ms |
+| `wasm_cache_hit` | < 1 µs |
+| enforcement `decide()` (per tool call) | ~200–300 µs |
+| `verify_only` (per pull, deterministic) | ~1.8 ms |
+| `pull_warm_304` (per chat turn) | ~50 ms (network-bound) |
```
victorludvig: Can we benchmark the policy+ban pull, executed back to back?

victorludvig (commented):
Thank you for the PR, I suggest to add the ban pull to the benchmark to have a more accurate latency estimation.

benchmarks/README.md
guillaume-hexamind: Thanks, very good point, the ban pull was omitted in this benchmark. I added new tests to benchmark the back to back policy + ban tool. 

Right now, these two pull are done sequentially -> doing them together within an asyncio.gather would thus almost halve the pull time, to do in a specific PR.

victorludvig (approved):

PR #101, opened 2026-07-31
feat(sdk): policy analyzer + hexgate policy check (2/3)

quanghexa94:
Second of the 3 multi-module policy PRs (after #97, modules + linker). This one is the **analyzer**: it turns the linker's provenance into file-attributed lints a dev sees in the terminal / CI, and that the PR-3 editor will render inline.

Pure SDK, fully local, no platform. Mostly *surfacing* the provenance PR 1 already produces (`RuleTrace.contributors` / `shadowed`), not new engine work.

## What it does

The link step already *raises* `LinkError` for the unfixable cases (a capability that denies, a conflicting const, `file_scope`). The analyzer runs over a successfully-linked bundle and reports the soft problems that don't stop composition but are almost always mistakes:

| Lint | Severity | From |
| --- | --- | --- |
| `dead-grant` | warning | a capability grants a tool a boundary ceiling excludes (never fires) |
| `redundant-grant` | info | two capabilities grant the same tool identically |
| `unknown-tool` | boundary → error, capability → warning | a rule references a tool the agent's manifest doesn't declare (drift) |
| `unknown-arg` | warning | a constraint `args.x` the tool doesn't accept |

`PolicyLint(code, severity, message, source, tier, tool)` is the shared contract the CLI and the PR-3 editor both consume.

Drift (`unknown-tool` / `unknown-arg`) needs the agent's `AgentManifest`, so it only runs when one is supplied; the two provenance-only lints always run. Boundary drift is an **error** because a stale deny is fail-open (protects nothing); capability drift is a warning (fail-safe dead grant).

**Deferred (needs a solver):** semantic conflicts - empty intersection, always-true/false, subsumption.

## Try it

```
hexgate policy check --dir deploy/demo_policies                 # provenance lints
hexgate policy check --dir . --manifest agent.json             # + drift
hexgate policy check --dir . --max-severity warning            # gate CI harder
```

Exits non-zero at or above `--max-severity` (default `error`).

## Important files

| File | What to look at |
| --- | --- |
| `hexgate/security/analyzer.py` | `PolicyLint`, `analyze()`, `check()`, and the four detectors. |
| `hexgate/security/constraints.py` | `iter_arg_refs` (mirrors `iter_const_refs`), for the arg-drift check. |
| `hexgate/cli/policy/main.py` | the `check` subcommand + exit-code gating. |
| `tests/security/test_analyzer.py` | one fixture per lint class + link-error-as-lint + severity order. |
| `tests/cli/test_policy_check.py` | CLI exit codes (clean / warning threshold / link error). |

## Notes

- No `platform/api` changes. Full suite green (937 tests).
- `line` on `PolicyLint` is `None` for now (file-level attribution); line-level lands with YAML position tracking in the loader.
- Design: `policy-modules-plan.md` (Phase 3) and `policy-modules-implementation.md` (PR 2 section).

### DIFF
--- a/hexgate/cli/policy/main.py
+++ b/hexgate/cli/policy/main.py
@@ -1,10 +1,11 @@
 """`hexgate policy` subcommand — author + inspect + dry-run policy documents.
 
-Wraps the compiler library and both enforcement engines in a five-verb
-CLI: ``build``, ``validate``, ``show-rego``, ``test``, ``keygen``. Every
-verb is a thin wrapper — the heavy lifting lives in
-:mod:`hexgate.security`. That symmetry lets the platform's save flow use
-the same code without duplication.
+Wraps the compiler library and both enforcement engines in a set of thin
+verbs: ``build``, ``validate``, ``show-rego``, ``test``, ``keygen``,
+``resolve`` (compose a module bundle into one effective policy), and
+``check`` (lint that bundle). Every verb is a thin wrapper — the heavy
+lifting lives in :mod:`hexgate.security`. That symmetry lets the platform's
+save flow use the same code without duplication.
 
 ``build`` compiles the policy to a signed WASM bundle (yaml + rego +
 wasm + manifest, ``--sign-key`` to sign); ``test`` evaluates a decision
@@ -554,30 +555,34 @@ def _main_check(args: argparse.Namespace) -> int:
             print(f"manifest error: {exc}", file=sys.stderr)
             return 1
 
+    from hexgate.security.analyzer import SEVERITY_RANK
+
     lints = check_bundle(boundaries, capabilities, manifest=manifest)
 
-    if not lints:
-        print("✓ No policy lints.")
-        if manifest is None:
+    # A link error short-circuits before drift/soft lints run, so the
+    # "supply a manifest" hint only makes sense when linking succeeded.
+    linked = not (len(lints) == 1 and lints[0].code == "link-error")
+
+    def _drift_hint() -> None:
+        if manifest is None and linked:
             print(
                 "  (drift checks skipped — pass --manifest to enable them)",
                 file=sys.stderr,
             )
+
+    if not lints:
+        print("✓ No policy lints.")
+        _drift_hint()
         return 0
 
     icon = {"error": "✗", "warning": "!", "info": "·"}
-    rank = {"error": 0, "warning": 1, "info": 2}
     for lint in lints:
         where = f" ({lint.source})" if lint.source else ""
         print(f"{icon.get(lint.severity, '·')} [{lint.code}] {lint.message}{where}")
-    if manifest is None:
-        print(
-            "  (drift checks skipped — pass --manifest to enable them)",
-            file=sys.stderr,
-        )
+    _drift_hint()
 
-    threshold = rank[args.max_severity]
-    worst = min(rank[lint.severity] for lint in lints)
+    threshold = SEVERITY_RANK[args.max_severity]
+    worst = min(SEVERITY_RANK[lint.severity] for lint in lints)
     return 1 if worst <= threshold else 0
 
 
--- a/hexgate/cli/policy/main.py
+++ b/hexgate/cli/policy/main.py
@@ -213,6 +213,41 @@ def add_parser(subparsers: argparse._SubParsersAction) -> None:
     )
     p_resolve.set_defaults(func=_main_resolve)
 
+    # ---- check ----
+    p_check = sub.add_parser(
+        "check",
+        help="Lint a bundle of policy modules (dead / redundant / drift).",
+        description=(
+            "Links the boundary + capability modules under <dir>/policies/ and "
+            "reports authoring problems that don't stop composition but are "
+            "almost always mistakes: a capability grant a boundary ceiling makes "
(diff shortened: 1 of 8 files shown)


guillaume-hexamind:
1. check() only catches LinkError — 85/100 · blocker

analyzer.py:74 guards LinkError, but link_policy_set() also raises PolicySetError and ConstraintParseError (via PolicySet.__init__ → _validate_const_refs). A linkable-but-invalid bundle exits with a raw traceback instead of the documented [link-error] lint.

Repro: capability with constraints: ["args.amount <= consts.max_refund"] and no consts block → traceback through main.py:560.
Fix: _main_resolve (main.py:500) already catches (LinkError, PolicySetError, ConstraintParseError, ValidationError) for the identical call — match it.

2. unknown-arg hardcoded to warning, but the case is fail-open — 80/100 · blocker

_drift computes a per-tier severity at analyzer.py:193, but _unknown_args never receives it — line 235 is a literal "warning". A typo'd arg in a boundary conditional deny folds to not((args.amoun > 1000)), which evaluates true, so the deny inverts to an allow.

Repro: refund: {mode: deny, constraints: ["args.amoun > 1000"]} → effective mode='allow' constraints=['not ((args.amoun > 1000))']; evaluate_tool_call(amount=5000) → ALLOW. CLI prints ! [unknown-arg] and exits 0 under the default --max-severity error.
Fix: thread severity into _unknown_args so boundary arg drift is an error, consistent with the rule the docstring already states.

3. default_policy constraints silently dropped, no lint — 72/100

_fold_tool reads default_policy.mode (linker.py:188) and never .constraints. Nothing in modules.py or the loader rejects the field, and _drift/_unknown_args only walk module.policy.tools (analyzer.py:195, :211), so the analyzer stays silent.

Why it matters: the pydantic engine does enforce default_policy.constraints in a single-file policy (confirmed: amount=50 → ALLOW, amount=5000 → DENY). Migrating a working single-file policy to modules silently loses the fence.
Fix: either reject default_policy.constraints in a module at load time, or lint it — silently discarding a fence is the worst option.

4. Union erases a sibling capability's constraint, no lint — 70/100 · analyzer gap, not a linker bug

_union returning None on any unconditional grant (linker.py:230) is documented, intended semantics. The problem is that the analyzer doesn't flag the consequence.

Repro: tight.yaml grants refund with args.amount <= 100; adding loose.yaml with an unconditional grant yields effective constraints=[]; evaluate(amount=999999) → ALLOW; check() → [], CLI exits 0.
Fix: new lint (suggest constraint-erased, warning) when a grant's constraints are dropped because a sibling grant is unconditional. Same class as dead-grant, but the security-relevant direction.

5. Defensive boundary denies rated error — 60/100

analyzer.py:193 picks severity by tier, not by mode, so a boundary deny on a tool the agent doesn't expose is an error. The fail-open rationale in the docstring applies to a boundary allow/ceiling naming a missing tool — a deny on an absent tool protects nothing but breaks nothing.

This fires on the repo's own bundle: deploy/demo_policies/policies/boundaries/org_core.yaml declares delete_database: { mode: deny } (comment: "absolute — no layer can undo"). policy check --dir deploy/demo_policies --manifest <support manifest> → ✗ [unknown-tool] 'org_core' references tool 'delete_database', exit 1. Confirmed intent at module_loader.py:33: "the platform loader will use it to scope-attach org boundaries."
Fix: rate boundary deny on an unknown tool info (or skip it); keep error for boundary allow/ceiling drift.

guillaume-hexamind (commented):
Ran a code review high, that found a couple of medium and minor issues above.

Otherwise, I approve of the whole bounday and capabilities abstraction with linker + this implementation of policy analyzer; PR can be merged after the issues are addressed

quanghexa94:
Thanks for the thorough pass. All five addressed in `f15ed2f`, 942/942 green.

1. **`check()` only caught LinkError** now catches `PolicySetError` and `ConstraintParseError` too, matching `_main_resolve`'s tuple. An undefined-const bundle returns a `[link-error]` lint instead of a traceback.
2. **`unknown-arg` hardcoded to warning** severity is threaded through `_unknown_args` now. I scoped the error to the genuinely fail-open case: a boundary `deny` with a typo'd arg, where `not(<missing> ...)` folds to an allow. A boundary ceiling `allow` with a typo'd arg fails closed (over-restrictive), so that stays a warning. Repro from your comment now reports `error` and exits 1.
3. **`default_policy` constraints silently dropped** now a `LinkError` at compose time (surfaced as a `[link-error]` lint), with a message pointing the author to put the rule on a named tool. Rejecting beats linting here since the effective default is always fail-closed deny.
4. **Union erasing a sibling's constraint** new `constraint-erased` lint (warning): fires when a capability grants a tool unconditionally and drops a tighter sibling grant's constraint. Same class as `dead-grant`, the security-relevant direction.
5. **Defensive boundary denies rated error** `unknown-tool` severity now follows the failure direction, not the tier: a boundary `deny` on an absent tool is `info` (protects nothing, breaks nothing), a boundary ceiling `allow` naming an absent tool stays `error` (fail-open), capability drift stays `warning`. `policy check --dir deploy/demo_policies --manifest <support>` now passes clean.

Tests added for each. Ready for another look.

guillaume-hexamind (approved):

PR #104, opened 2026-08-03
build(claude): add integration test skill

victorludvig:
## What is changing
Add an integration test skill so that when you prompt Claude 'Run the sdk integration tests', it does.

## Tests
Prompted Claude to run both sdk and platform integration tests. It was successful.

### DIFF
--- a/.claude/skills/integration-tests/SKILL.md
+++ b/.claude/skills/integration-tests/SKILL.md
@@ -1,0 +1,29 @@
+# Run Integration Tests
+
+## SDK integration tests (repo root — `tests/adapters/*/test_integration.py`, `tests/audit/test_integration.py`)
+Hits a live platform server + ClickHouse end-to-end. Needs ClickHouse running, `platform-api` running, and a real `HEXGATE_API_KEY` minted via the dashboard (repo-root `.env`).
+```bash
+make clickhouse-up   # idempotent — safe to run even if already up, no-ops if unchanged
+
+# Check platform-api first — it's NOT idempotent. uvicorn binds :8000 directly,
+# so running `make platform-api` while one's already up fails with
+# "Address already in use" instead of reusing the running instance.
+curl -sf http://localhost:8000/health
+# ^ if this 200s, platform-api is already up — skip starting it.
+# If it fails, start `make platform-api` in a separate terminal (it's a
+# foreground/blocking dev server — don't chain it into this script).
+
+set -a && source .env && set +a   # loads HEXGATE_API_KEY etc. from .env into the shell
+pytest -m integration
+```
+Unset afterward because some unit tests rely on not having any API key defined:
+```bash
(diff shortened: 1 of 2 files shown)


.claude/skills/integration-tests/SKILL.md
```diff
@@ -0,0 +1,29 @@
+# Run Integration Tests
```
guillaume-hexamind, line 6: Claude SKILL.md should follow a header pattern:

---\nname: ...\ndescription: ...\n---

I don't think as is, the skill is reachable by claude

guillaume-hexamind (commented):
Good idea to keep the loose logic as a claude skill. One comment on the format below.

Do you think also adding make commands would make sense as a complement, to have both the skill and the make commands available for a repo user?

Something like 
`.PHONY: platform-api-integration-test
platform-api-integration-test: clickhouse-up ## Platform API integration tests (needs ClickHouse)
      cd platform/api && uv run pytest -m integration

.PHONY: integration-test
integration-test: clickhouse-up ## SDK integration tests (also needs platform-api up + HEXGATE_API_KEY)
      $(UV) pytest -m integration`

victorludvig:
> Good idea to keep the loose logic as a claude skill. One comment on the format below.
> 
> Do you think also adding make commands would make sense as a complement, to have both the skill and the make commands available for a repo user?
> 
> Something like `.PHONY: platform-api-integration-test platform-api-integration-test: clickhouse-up ## Platform API integration tests (needs ClickHouse) cd platform/api && uv run pytest -m integration
> 
> .PHONY: integration-test integration-test: clickhouse-up ## SDK integration tests (also needs platform-api up + HEXGATE_API_KEY) $(UV) pytest -m integration`

While it is a good idea for the platform integration tests which communicate with the clickhouse instance directly, this would be more complex for the sdk integration tests.
These go through the api server, which requires setting the api server, then setting the api keys, and then unsetting them because some unit tests specifically test the no-key-configured behavior. This sequence logic makes a skill better than a make target which would become convoluted.

So I am keeping the sdk integration tests as a sequence of commands in the skill, and moving the platform integration tests into a make target.

.claude/skills/integration-tests/SKILL.md
victorludvig: It is reachable without the header pattern (and I tested it), because the name defaults to the directory name (integration-tests here), and the description defaults to the first line of the markdown body (in our case, the H1 heading: Run Integration Tests).

However this can be a good practice to add a description frontmatter field to improve auto-invocation.
I added it to define the standard for our next skills.

Test: it is indeed better with the frontmatter fields. It now asks if we want to run the sdk or api integration tests:
Screenshot 2026-08-04 at 10 28 49

guillaume-hexamind (approved):

PR #105, opened 2026-08-03
fix(sdk): fail open when emit_llm_usage raises in adapter usage hooks

victorludvig:
## What is changing
Wraps the LLM-usage side channel in a single fail-open guard so it can never crash the agent run it's reporting on.

- `emit_llm_usage` (`hexgate/tracing/usage.py`) now wraps its own body in try/except and logs on failure — it's already the single entry point every adapter's usage hook calls into, so the guard lives there instead of being duplicated per adapter.
- All four adapters (LangChain, OpenAI Agents, Google ADK, Pydantic AI) call it as a plain, unguarded function call — no per-adapter try/except needed, and any future adapter inherits the safety property for free.

Before this change, a failure inside usage emission (a bug in the sender registry, a malformed event, etc.) could take down the real agent run, not just drop a usage event:

- **Google ADK**: `PluginManager._run_callbacks` re-raises an unhandled plugin exception as a `RuntimeError`, failing the whole run.
- **OpenAI Agents SDK**: the run loop (`agents/run_internal/run_loop.py`) calls `hooks.on_llm_end` directly via `asyncio.gather` with no try/except — an exception propagates straight out of `Runner.run*`.
- **Pydantic AI**: `emit_run_usage` is called inline in `hexgate/adapters/pydantic_ai/agent.py`, right before returning the run result to the caller — no framework layer in between at all.
- **LangChain** was the only one incidentally safe, because LangChain's own `AsyncCallbackManager` wraps every handler callback in try/except and swallows it (unless `raise_error=True`).

## Tests
- `tests/tracing/test_usage_emit.py::test_when_sender_emit_fails_then_emit_llm_usage_does_not_raise` — a fake sender whose `emit()` raises confirms `emit_llm_usage` itself never propagates.
- Existing per-adapter usage tests (LangChain, OpenAI, Google, Pydantic AI) continue to pass unmodified — they cover extraction/emission logic, now backed by the same guarded entry point.
- `make lint && make fmt-check && make coverage` all pass (1686 passed, 0 failed).

### DIFF
--- a/hexgate/adapters/google/usage.py
+++ b/hexgate/adapters/google/usage.py
@@ -9,15 +9,11 @@ named sub-agents.
 
 from __future__ import annotations
 
-import logging
-
 from google.adk.agents.callback_context import CallbackContext
 from google.adk.models.llm_response import LlmResponse
 from google.adk.plugins.base_plugin import BasePlugin
 
 from hexgate.tracing.usage import emit_llm_usage
-
-_log = logging.getLogger(__name__)
 
 
 class HexgateUsagePlugin(BasePlugin):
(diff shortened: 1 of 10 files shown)


guillaume-hexamind (approved):
Good to me!

PR #106, opened 2026-08-04
feat(sdk): replace User scope with HexgateContext (user_roles + attributes)

guillaume-hexamind:
## What

First PR of the `User → HexgateContext` generalization. Renames the request-scoped `User` model to **`HexgateContext`** and replaces the single `role: str | None` with **`user_roles: list[str]`**, adding an (inert) `attributes` bag for the upcoming ABAC work.

Everything downstream of the model still thinks in a single role, reduced from the list at the enforcer + token boundaries via one `primary_role` helper.

```python
# before
async with User(user_id="alice", role="billing"): ...
# after
async with HexgateContext(user_id="alice", user_roles=["billing"]): ...
```

### Breaking changes (vs `main`)

1. **`User` is removed** → `HexgateContext`. No compatibility alias.
2. **`get_current_user` is removed** → `get_current_context`. No compatibility alias; all call-sites migrated.
3. **`role: str | None` is removed** → `user_roles: list[str]`. The `role=` kwarg no longer exists; `user_roles[0]` (`primary_role`) selects the policy bucket.

Model contract for clarity: **`user_id` is required** (unchanged from `main`), `user_roles` is optional and defaults to `[]` (empty → the `default` policy bucket, parity with the old `role=None`).

## Review priority

**Read these ~5 functional sites first — this is where behavior could change:**

- `hexgate/runtime/context.py` — the new model, `primary_role`, removed `User`/`role`, inert `attributes`. The heart of the PR.
- `hexgate/security/enforcer.py` — role reduction: `role = context.primary_role`. Empty `user_roles` forwards `role=None` → PolicySet still falls back to the `default` bucket (parity with the old `role=None`).
- `hexgate/agents/factory.py` — attenuation now signs `role=context.primary_role` (token format unchanged).
- `hexgate/cli/serve.py` — `_context_from_payload` wire mapping. Accepts a new `roles` list **and** the legacy singular `role` key, so **no platform-side change is required**.
- `hexgate/adapters/_common.py` + `hexgate/tracing/usage.py` — Langfuse metadata keeps the `user_role` key (value = `primary_role`); usage getter swap.

**Everything else is mechanical and safe to skim:** ~18 files of pure `User → HexgateContext` type/import/docstring renames (adapters, egress, security docstrings), the test migration (`test_user.py → test_context.py` + new `user_roles`/`primary_role` coverage), and the in-repo caller updates below.

## In-repo callers migrated (outside `hexgate/` + `tests/`)

[...]

PR #107, opened 2026-08-06
feat(sdk): thread ctx.* attributes through both policy engines

guillaume-hexamind:
## What

PR 2 of the HexgateContext generalization (follows #106). Threads the
`HexgateContext.attributes` bag — landed inert in #106 — through the engine
contract into **both** policy engines, so a policy constraint can now filter on
caller attributes via a `ctx.<key>` namespace (RBAC → RBAC + ABAC):

```yaml
constraints:
  - ctx.department == "finance"
  - ctx.region in ["EU", "UK"]
  - ctx.clearance_level >= 3
```

Attributes are **advisory / untrusted** in this PR — read straight off the
contextvar, never token-verified. The signed trusted tier (attributes that may
gate a production `deny`) is the next PR.

## Why one PR (P2 + P3 fused)

The pydantic engine (dev / `hexgate chat` / fallback) and the compiled WASM
engine (production bundle) must decide a `ctx.*` constraint **identically**, or a
rule would filter on one engine and no-op on the other. So the contract change,
both engine implementations, the golden fixture, and the parity matrix all ship
together.

## Changes

Every hop takes a keyword-only `attributes` (WASM: `ctx`) arg defaulted to
`None`, so existing callers are source-compatible:

- **`decision.py`** — `PolicyEngine.evaluate` protocol gains `attributes`.
- **`constraints.py`** — `check_constraints` injects `"ctx": dict(attributes or {})`
  into the eval context; `ctx.<key>` resolves via the existing `_walk` with the
  same missing-key → fail-closed semantics as any other ref.
- **`policy.py`** — forwarded through the four evaluators (`evaluate_tool_call`,
  `authorize_tool_call`, and the two `_wasm` variants).
- **`policy_set.py` / `bundle.py`** — both `PolicyEngine.evaluate` entry points forward it.
- **`wasm_engine.py`** — `WasmPolicy.decide` adds `ctx` to the input document
  (`input.ctx`), mirroring the pydantic context dict key-for-key.
- **`enforcer.py`** — reads `context.attributes` off the contextvar and forwards it.

**The Rego compiler is unchanged.** `_render` already prefixes any `Ref` with
`input.`, so `ctx.department → input.ctx.department` for free — including the
ordered-comparison cross-type guard (`is_number(...)`). Proven by tests rather
than edited.

## Tests

- **Parity matrix** (`test_engine_parity.py`) — new `ctx.*` cases across eq /
  `in` / `not in` / ordered / bool, plus fail-closed on missing attribute and on
  cross-type (`"3" >= 3`). Pydantic and real `opa`→`wasmtime` must agree.
- **Constraints unit tests** — `ctx.*` resolution, missing-key deny, bare `ctx`
  rejected at parse (prefix required).
- **Enforcer** — end-to-end through the real pydantic engine + a forwarding
  assertion that context attributes reach the engine.
- **Rego codegen** + regenerated golden fixture (diff is only `input.ctx.*` lines).

## Notes for review

- **Membership against a list-valued attribute** (`"x" in ctx.tags`) is not
  expressible — the grammar restricts `in`'s RHS to a list literal or
  `consts.<name>`. Not a regression; noted as a known ABAC limitation.
- Four hand-rolled fake engines in the test suite gained the `attributes` kwarg
  to satisfy the updated protocol.

## Checks

`make lint` + `make fmt-check` clean; security + runtime + cli + audit suites
green (**1088 passed / 5 skipped**); full `make coverage` (**1729 passed /
14 skipped**) and the platform-api suite (**441 passed**) green. Parity + CLI
wasm suites run with `opa` on PATH (not skipped).

### DIFF
--- a/hexgate/cli/policy/main.py
+++ b/hexgate/cli/policy/main.py
@@ -178,6 +178,16 @@ def add_parser(subparsers: argparse._SubParsersAction) -> None:
         help='Tool arguments as a JSON object (e.g. \'{"amount": 30, "currency": "USD"}\'). Defaults to {}.',
     )
     p_test.add_argument(
+        "--attributes",
+        default="{}",
+        help=(
+            "Caller ABAC attributes as a JSON object, exposed to ctx.* "
+            'constraints (e.g. \'{"department": "finance", "clearance_level": 3}\'). '
+            "JSON (not key=value) so numbers/bools keep their type and match "
+            "production. Defaults to {}."
+        ),
+    )
+    p_test.add_argument(
         "--engine",
         choices=("pydantic", "wasm"),
         default="pydantic",
@@ -507,6 +517,15 @@ def _main_test(args: argparse.Namespace) -> int:
         return 1
 
     try:
+        attributes: dict[str, Any] = json.loads(getattr(args, "attributes", "{}"))
+    except json.JSONDecodeError as exc:
+        print(f"--attributes is not valid JSON: {exc}", file=sys.stderr)
+        return 1
+    if not isinstance(attributes, dict):
+        print("--attributes must be a JSON object (dict).", file=sys.stderr)
+        return 1
+
+    try:
         policy_set = load_policy_set_from_dict(payload)
     except (PolicySetError, ValidationError) as exc:
         print(f"policy schema: {exc}", file=sys.stderr)
@@ -523,8 +542,12 @@ def _main_test(args: argparse.Namespace) -> int:
     engine = getattr(args, "engine", "pydantic")
 
     if engine == "wasm":
-        return _test_via_wasm(payload, args.role, args.tool, tool_args, label)
-    return _test_via_pydantic(policy_set, args.role, args.tool, tool_args, label)
+        return _test_via_wasm(
+            payload, args.role, args.tool, tool_args, attributes, label
+        )
+    return _test_via_pydantic(
+        policy_set, args.role, args.tool, tool_args, attributes, label
+    )
 
 
 def _render_verdict(verdict: Verdict, label: str) -> int:
@@ -551,20 +574,31 @@ def _render_verdict(verdict: Verdict, label: str) -> int:
 
 
 def _test_via_pydantic(
-    policy_set: Any, role: str, tool: str, tool_args: dict, label: str
+    policy_set: Any,
+    role: str,
+    tool: str,
+    tool_args: dict,
+    attributes: dict,
+    label: str,
 ) -> int:
     """Run the decision through the in-process constraint evaluator."""
     policy: AgentPolicy = policy_set.policy_for(role)
-    # Forward role so role-scoped constraints (role == "admin") decide the same
-    # as the wasm path and production — omitting it here made `policy test
-    # --engine pydantic` diverge from --engine wasm on exactly those rules.
+    # Forward role AND attributes so role-scoped (role == "admin") and ctx.*
+    # constraints decide the same as the wasm path and production — omitting
+    # either here made `policy test` fail closed on rules production allows.
     return _render_verdict(
-        evaluate_tool_call(policy, tool, tool_args, role=role), label
+        evaluate_tool_call(policy, tool, tool_args, role=role, attributes=attributes),
+        label,
     )
 
 
 def _test_via_wasm(
-    payload: dict, role: str, tool: str, tool_args: dict, label: str
+    payload: dict,
+    role: str,
+    tool: str,
+    tool_args: dict,
+    attributes: dict,
+    label: str,
 ) -> int:
     """Compile to wasm on the fly + evaluate — matches production semantics."""
     try:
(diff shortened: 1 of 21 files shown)


hexgate/security/enforcer.py
```diff
@@ -81,13 +88,16 @@ def decide(self, tool_name: str, arguments: Mapping[str, Any]) -> Decision:
...
+        verdict = self.policy.evaluate(
+            role=role, tool=tool_name, args=args_snapshot, attributes=attributes
+        )
         decision = Decision.from_verdict(
             verdict,
             agent_name=self.agent_name,
             tool_name=tool_name,
             role=role,
             arguments=args_snapshot,
+            attributes=dict(attributes) if attributes is not None else None,
```
quanghexa94: **Inconsistent snapshot depth (low-medium).** The `attributes` snapshot stamped on the `Decision` is only shallow-copied (`dict(attributes)`), while `args` is `copy.deepcopy`'d for the same observer/audit-safety reason.

If an `AttrValue` holds a `list[str]`, a later mutation of `context.attributes['tags']` (the contextvar persists across tool calls within a scope) would be reflected in a retained `decision.attributes` — the exact corruption the `args` deep-copy defends against, applied inconsistently. Suggest `copy.deepcopy(attributes)` for parity.

hexgate/cli/policy/main.py
```diff
@@ -506,6 +516,15 @@ def _main_test(args: argparse.Namespace) -> int:
         print("--args must be a JSON object (dict).", file=sys.stderr)
         return 1
```
quanghexa94, line 525: **`policy test --attributes` bypasses the `AttrValue` schema (low-medium).** This does a raw `json.loads` and forwards the result unvalidated, bypassing the `AttrValue` schema (`str | int | bool | list[str]`) that `HexgateContext` enforces at runtime.

So `--attributes '{"clearance_level": 3.5}'` (float) or a nested object is accepted by the simulator and fed to both engines, though such a shape can never reach the real engine via `HexgateContext`. Since the flag is billed as making values "match production," this is a fidelity gap — consider validating against `AttrValue` before forwarding.

tests/security/test_engine_parity.py
```diff
@@ -411,6 +425,76 @@ def test_role_tool_fact_parity(role: str, tool: str, args: dict, expect: str) ->
...
+        # string equality
+        ("eq", {"department": "finance"}, "allow"),
+        ("eq", {"department": "sales"}, "deny"),
+        ("eq", {}, "deny"),  # missing attr → fail closed on BOTH engines
+        # membership against a literal list
+        ("in_list", {"region": "EU"}, "allow"),
+        ("in_list", {"region": "US"}, "deny"),
+        ("in_list", {}, "deny"),  # missing → fail closed
+        # negated membership
+        ("not_in_list", {"region": "EU"}, "allow"),
```
quanghexa94, line 477: **Parity matrix omits the missing-attribute case for `not in` (low).** `eq` / `in_list` / `ordered` all include a `{}` (missing-attr) parity case, but `not_in_list` only tests region present (EU/US).

`not input.ctx.region in [...]` on an undefined ref is exactly the subtle case (verified here to correctly deny on both engines), so leaving it uncovered means a future compiler change to negated-membership rendering could silently break pydantic/wasm parity with no failing test. Worth adding a `{}` case.

quanghexa94 (commented):
Reviewed the threading of `ctx.*` / `attributes` through both policy engines. The PR is clean and well-tested on correctness — I chased the most promising lead (Rego `not X in Y` on an undefined ref possibly ALLOWing while pydantic DENIES) and confirmed with `opa eval` that Rego returns `undefined` (deny) for both inline and helper-rule forms, so **the engines agree — not a bug**. I also verified every `PolicyEngine.evaluate` impl (incl. fakes) was updated for the new `attributes` kwarg, and that `Decision.attributes` is not leaked (`as_error_payload` / `AuditEvent.as_payload` enumerate fields explicitly and omit it).

Three low / low-medium findings inline below.

guillaume-hexamind:
@quanghexa94 Thanks for the review, all three fixed.

1. Snapshot depth — extracted _snapshot(values, *, deep) in enforcer.py; args and attributes now deep-copy on the same retained condition, and attrs_snapshot is also what's passed to policy.evaluate. New observer test mutates a list[str] attribute on the live contextvar after decide() and asserts the retained Decision is unchanged (verified it fails without the fix).
2. --attributes fidelity — validated through TypeAdapter(dict[str, AttrValue]), so the simulator accepts exactly what HexgateContext accepts. Rejects 3.5 / nested / [1, 2] / null; still accepts 3.0 → 3, matching production's lax coercion. Existing JSON/non-object errors and their tests unchanged.
3. Parity gap — added ("not_in_list", {}, "deny"). I re-confirmed your opa eval result end-to-end (opa 1.17 → wasmtime): pydantic resolves _MISSING → False, Rego's inline not <undefined> in [...] is undefined so allow never fires — both deny, now pinned.

winlp4ever (approved):

quanghexa94 (approved):

quanghexa94 (approved):

PR #108, opened 2026-08-06
refactor(platform-api): rename DevToken to ApiKey

victorludvig:
## What is Changing
- `DevToken` → `ApiKey` (SQLModel table `devtoken` → `apikey`)
- `mint_dev_token` → `mint_api_key`, `list_dev_tokens` → `list_api_keys`, `delete_dev_token` → `delete_api_key`, `optional_dev_token` → `optional_api_key`
- Updated call sites and docstrings in `core/ids.py`, `deps/tokens.py`, `features/tokens/{router,service}.py`, `deploy/provision.py`, and the two tests that mint tokens directly (`test_agents.py`, `test_ws_serve.py`)

`DevToken` never actually scoped to a "dev" environment — it's the general-purpose API-key model used for both `fty_test` and `fty_live` keys, minted and sent as the `Authorization: Bearer` credential on every SDK request. The name was misleading; this renames it to match what it actually is, ahead of the OTel migration work that references this model directly.

## Tests
- `make platform-api-check` — 441 passed, 4 skipped, lint clean
- `make test` (SDK) — 1704 passed, 14 skipped
- SDK integration suite (`pytest -m integration`) — 6 passed
- Platform API integration suite (`make platform-api-test-integration`) — 3 passed against real ClickHouse

**Note:** this rename stops at the model/service layer. The router endpoints (`list_tokens`/`mint_token`/`revoke_token`), the Pydantic schemas (`TokenListItem`/`TokenMintRequest`/`TokenMintResponse`), the URL paths (`/projects/{id}/tokens`), the `features/tokens/` directory name, and the dashboard UI still say "token(s)" in several places. Left as a known, deliberate scope boundary to keep this PR small — worth a follow-up PR if/when it's worth the larger blast radius (router + URL contract + dashboard copy).

### DIFF
--- a/deploy/provision.py
+++ b/deploy/provision.py
@@ -37,7 +37,10 @@ async def _seed_gdocs_agent() -> None:
     """
     # NOTE: stdout is the token channel (see __main__) — log to stderr only.
     if not _GDOCS_POLICY.is_file():
-        print(f"[provision] {_GDOCS_POLICY} missing — skipping docs_agent seed", file=sys.stderr)
+        print(
+            f"[provision] {_GDOCS_POLICY} missing — skipping docs_agent seed",
+            file=sys.stderr,
+        )
         return
 
     from hexgate_api.constants import DEFAULT_PROJECT_ID
@@ -67,21 +70,24 @@ async def _seed_gdocs_agent() -> None:
             )
         )
         await session.commit()
-    print(f"[provision] seeded {_GDOCS_AGENT_NAME} policy for the dashboard", file=sys.stderr)
+    print(
+        f"[provision] seeded {_GDOCS_AGENT_NAME} policy for the dashboard",
+        file=sys.stderr,
+    )
 
 
 async def _mint() -> str:
     from hexgate_api.constants import DEFAULT_PROJECT_ID
     from hexgate_api.core.db import async_session_factory, init_db
     from hexgate_api.core.keystore import keystore  # same singleton the API uses
-    from hexgate_api.features.tokens.service import mint_dev_token
+    from hexgate_api.features.tokens.service import mint_api_key
     from hexgate_api.seeds.defaults import ensure_default_seed
 
     await init_db()
     keystore.ensure_keypair()
     async with async_session_factory() as session:
         await ensure_default_seed(session)
-        _, full_token = await mint_dev_token(
+        _, full_token = await mint_api_key(
             session,
             DEFAULT_PROJECT_ID,
             name="demo-serve",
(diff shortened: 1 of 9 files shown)


guillaume-hexamind:
Nice cleanup, and the new tokens tests are a welcome addition. One blocker before this lands:

**`ApiKey` doesn't pin `__tablename__`, so the table renames too.** SQLModel derives it from the class name:

```
main:   DevToken.__tablename__ == 'devtoken'
here:   ApiKey.__tablename__   == 'apikey'
```

`init_db()` is bare `create_all`, and there's no Alembic in `platform/api`. I simulated the upgrade against a populated DB:

```
AFTER  devtoken rows (orphaned): 1
AFTER  apikey rows (what the app reads): 0
AFTER  find_token_by_secret('fty_live_...') -> None
```

Since the keystore is a separate volume, the signature gate still passes and only the revocation gate fails — so every existing key returns `401 "unknown or revoked hexgate key"` when it's neither. WS `serve` just closes `4401` with no detail, and the dashboard key list silently renders empty. Nothing re-mints (`seeds/` has no token refs, prod is `HEXGATE_SEED=skip`).

This matters for prod/staging specifically, per our own `platform/DEPLOY.md`:
- L105 — API runs `init_db()`, "no migration step"
- L127 — upgrade is `git pull && make platform-up`, reusing the existing `pg-data` volume
- L156 — "Schema changes apply only on an empty volume; changing one after first boot needs a manual migration"

Also worth noting the orphaned `devtoken` keeps full plaintext secrets in its `secret` column, with no code left that knows it exists.

**Suggested fix** — one line in `models.py`, keeps the rename purely at the Python level, which is what the PR description is after:

```python
class ApiKey(SQLModel, table=True):
    __tablename__ = "devtoken"  # historical name; renaming needs a migration
```

The alternative is a guarded `ALTER TABLE devtoken RENAME TO apikey` in `init_db` — that would also clean up the plaintext table, but it's migration code on a codebase with no migration harness, so I'd take the one-liner now and do it properly when Alembic lands.

guillaume-hexamind (commented):
One issue found by claude above that could be major;

Otherwise it looks good, will approve once the issue is checked!

victorludvig:
Good catch — pushed `f1307db` pinning `ApiKey.__tablename__ = "devtoken"`. Confirmed the class-name-derived table name via SQLModel's `__tablename__` declared_attr (`cls.__name__.lower()`), and that `init_db()` is bare `create_all` with no Alembic in this project — matches what you described. Full test suite (44 tests incl. the new tokens tests) passes with the pin in place.

Worth noting: we have no real users/prod keys yet, so the orphaned-`devtoken`-rows scenario is hypothetical for now rather than an active incident. Keeping the fix anyway since it's a one-liner and this exact landmine goes live the moment someone mints a real key.

guillaume-hexamind (approved):

PR #110, opened 2026-08-06
build(redpanda): stand up local Redpanda infra for OTLP ingestion

victorludvig:
Design doc: [OpenTelemetry migration design](https://app.notion.com/p/3b2fb45dbae2803ca64adb866704d69b) — Design 3, row 1 ("Kafka infra" — now Redpanda, see below) from the "PR Plan (Design 2)" table.

**Updated 2026-08-07**: this PR started as Kafka infra and was swapped to Redpanda mid-flight (see the doc's new "Design 3" section for why — same architecture, different broker, chosen for operational simplicity while staying Kafka-wire-protocol-compatible for potential enterprise clients who may already run Kafka themselves).

## What is Changing
- Add a `redpanda` service to `platform/docker-compose.yml`: single-binary Redpanda broker (`--mode dev-container`, no JVM/ZooKeeper), local dev only, no auth.
- Two client-facing listeners, same shape as the original Kafka version: `PLAINTEXT` (host-machine clients, `localhost:9092`) and `DOCKER` (sibling containers on the compose network, `redpanda:29092`) — matches how `docker-compose.deploy.yml` already points other services at each other by service name instead of `localhost`.
- `platform/redpanda/init/create-topics.sh` (moved from `platform/kafka/`): creates `hexgate.otlp.raw` (3-day retention) and `hexgate.otlp.dlq` (30-day retention), 3 partitions each. Also disables `auto_create_topics_enabled` and reconciles retention on every run, not just on first creation — see the review-fixes section below.
- Makefile targets renamed: `redpanda-up`, `redpanda-stop`, `redpanda-topics`, `redpanda-list-topics`, `redpanda-reset`.
- `CLAUDE.md`: Tech Stack line + commit-scope list updated (`kafka` → `redpanda`).
- `clickhouse-down`/`clickhouse-reset` scoped to the `clickhouse` service only (previously ran `down`/`down -v` on the whole compose project — see below).
- Topic names, retention, and partition-by-`project_id` (via the future Collector's `kafkaexporter` config — unaffected, since Redpanda speaks the same Kafka wire protocol) are unchanged from the original Kafka version.

First infra piece of the OTLP-ingestion pipeline: the Go Collector needs a real topic to write to and test against. Redpanda over Kafka specifically: single binary, no JVM/ZooKeeper to configure — `--mode dev-container` bundles the tuning flags Kafka needed several separate `KAFKA_*` env vars for. Still Kafka-wire-protocol-compatible, so nothing downstream (the Collector's `kafkaexporter` config) needed to change.

## Review fixes (4 issues, all reproduced and verified fixed)
1. **`create --if-not-exists` never applied retention on an existing topic.** Reproduced: pre-created `hexgate.otlp.raw` with 1 partition/default 7-day retention, ran the script — still 7-day retention afterward, exit 0, no error. Fixed by reconciling retention via `rpk topic alter-config` every run, regardless of whether create just made the topic. Partition count is deliberately *not* reconciled the same way — rpk's only knob (`add-partitions --num N`) adds rather than sets a target, so it can't be idempotent; a wrong partition count needs a manual fix.
2. **`auto_create_topics_enabled` was left at its default `true`.** Confirmed rpk applies this (and other dev-container defaults) regardless of whether `--mode dev-container` was actually passed — a stale/wrong topic name from any producer would silently fabricate a topic with cluster defaults instead of erroring. Now disabled at the top of `create-topics.sh`, applied live via `rpk cluster config set`.
3. **`redpanda-reset` wiped the volume but never recreated topics.** Unlike ClickHouse (which re-runs its init scripts for free via `docker-entrypoint-initdb.d`), Redpanda has no such hook. Confirmed empty topic list after a reset. Fixed by appending `$(MAKE) redpanda-topics` to the recipe (not a prerequisite — Make would run that *before* the recipe, creating the topics and then wiping them).
4. **`clickhouse-down`/`clickhouse-reset` also destroyed Redpanda.** Both ran `down`/`down -v` on the whole compose project; confirmed via `--dry-run` that `platform_redpanda-data` gets removed alongside ClickHouse's own volume. Scoped both to the `clickhouse` service only, matching the `postgres-*`/`redpanda-*` pattern already used elsewhere, and dropped the old NOTE comment warning about this blast radius (previously only mentioned Postgres) since it's no longer true for either service.

## Tests
Re-verified end-to-end against the real committed config:
- Broker starts and reports healthy on `make redpanda-up` (cold start, fresh volume, ~6s).
- `make redpanda-topics` creates both topics; re-running it is a no-op that also re-applies retention (see fix #1).
- Retention confirmed via `rpk topic describe -c`: `retention.ms=259200000` (raw) / `2592000000` (DLQ), including after simulating drift.
- Produce → consume round trip succeeded on `hexgate.otlp.raw` from **both** a host-side client (`localhost:9092`) and a real sibling container on the compose network (`redpanda:29092`) — both messages readable back from either path.
- `redpanda-stop`/`redpanda-reset` are scoped to the `redpanda` service only — confirmed `clickhouse` stayed running throughout; `clickhouse-down`/`clickhouse-reset` are now scoped the same way — confirmed `redpanda` stays running/intact through both.
- Caught and fixed one real bug while re-verifying the healthcheck itself: `rpk cluster health --brokers ...` doesn't accept `--brokers` at all (only `rpk topic` subcommands do) — would have left the container stuck in "starting" forever and never satisfied `--wait`. Fixed to bare `rpk cluster health`, which defaults to the local node.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -21,5 +21,5 @@ tests/                # hexgate package tests (agents, cli, security, tracing, s
 
 ## Rules & Constants
 - **Branches:** `{initials}/{type}/{short_description}` (e.g., `vl/feat/web_search`)
-- **Commits:** `type(scope): description` (lowercase, imperative, no period). Scopes: `platform-api`, `dashboard`, `sdk`, `cli`, `clickhouse`. Types: `feat`, `fix`, `docs`, `build`, `refactor`, `test`.
+- **Commits:** `type(scope): description` (lowercase, imperative, no period). Scopes: `platform-api`, `dashboard`, `sdk`, `cli`, `clickhouse`, `kafka`. Types: `feat`, `fix`, `docs`, `build`, `refactor`, `test`.
 - **Envs:** All prefixed with `HEXGATE_`. Never commit private keys.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,7 +1,7 @@
 # Hexgate — Claude Code Instructions
 
 ## Tech Stack
-- Python ≥ 3.13 (`uv`), FastAPI (`platform/api/`), ClickHouse (Docker), Kafka (Docker, dev-infra only for now — see `platform/docker-compose.yml`)
+- Python ≥ 3.13 (`uv`), FastAPI (`platform/api/`), ClickHouse (Docker), Redpanda (Docker, Kafka-protocol-compatible, dev-infra only for now — see `platform/docker-compose.yml`)
 - React, Vite, pnpm (`platform/dashboard/`)
 - Ruff (Python). WASM via `wasmtime`.
 
@@ -21,5 +21,5 @@ tests/                # hexgate package tests (agents, cli, security, tracing, s
 
 ## Rules & Constants
 - **Branches:** `{initials}/{type}/{short_description}` (e.g., `vl/feat/web_search`)
-- **Commits:** `type(scope): description` (lowercase, imperative, no period). Scopes: `platform-api`, `dashboard`, `sdk`, `cli`, `clickhouse`, `kafka`. Types: `feat`, `fix`, `docs`, `build`, `refactor`, `test`.
+- **Commits:** `type(scope): description` (lowercase, imperative, no period). Scopes: `platform-api`, `dashboard`, `sdk`, `cli`, `clickhouse`, `redpanda`. Types: `feat`, `fix`, `docs`, `build`, `refactor`, `test`.
 - **Envs:** All prefixed with `HEXGATE_`. Never commit private keys.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,7 +1,7 @@
 # Hexgate — Claude Code Instructions
 
 ## Tech Stack
-- Python ≥ 3.13 (`uv`), FastAPI (`platform/api/`), ClickHouse (Docker)
+- Python ≥ 3.13 (`uv`), FastAPI (`platform/api/`), ClickHouse (Docker), Kafka (Docker, dev-infra only for now — see `platform/docker-compose.yml`)
 - React, Vite, pnpm (`platform/dashboard/`)
 - Ruff (Python). WASM via `wasmtime`.
 
--- a/Makefile
+++ b/Makefile
@@ -192,6 +192,36 @@ postgres-reset: ## Wipe ONLY the Postgres data volume and restart
 	-docker volume rm platform_postgres-data
 	$(COMPOSE) up -d --wait postgres
 
+# -------- Platform infra (Kafka — OTLP ingestion buffer) --------
+#
+# Single-node KRaft dev broker (no ZooKeeper). Topics aren't auto-created
+# on first boot (Kafka has no docker-entrypoint-initdb.d equivalent) — run
+# `make kafka-topics` once after `make kafka-up`.
+
+.PHONY: kafka-up
+kafka-up: ## Start local Kafka and wait until healthy
+	$(COMPOSE) up -d --wait kafka
+
+.PHONY: kafka-stop
+kafka-stop: ## Stop Kafka (keeps the data volume)
+	$(COMPOSE) stop kafka
+
+.PHONY: kafka-topics
+kafka-topics: kafka-up ## Create the hexgate.otlp.raw / hexgate.otlp.dlq topics (idempotent)
+	docker exec hexgate-kafka /opt/kafka/bin/kafka-topics.sh --version >/dev/null
+	docker cp platform/kafka/init/create-topics.sh hexgate-kafka:/tmp/create-topics.sh
+	docker exec hexgate-kafka bash /tmp/create-topics.sh
+
+.PHONY: kafka-list-topics
+kafka-list-topics: ## List topics on the local Kafka broker
+	docker exec hexgate-kafka /opt/kafka/bin/kafka-topics.sh --bootstrap-server localhost:9092 --list
+
+.PHONY: kafka-reset
+kafka-reset: ## Wipe ONLY the Kafka data volume and restart
+	$(COMPOSE) rm -sf kafka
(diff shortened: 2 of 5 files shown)


guillaume-hexamind:
Ran code review high on this PR, here are its findings:

### 1. `--if-not-exists` never applies retention or partition count — 70/100

`platform/redpanda/init/create-topics.sh:16,28`

`rpk topic create --if-not-exists` skips an existing topic **entirely**, so `-c retention.ms=...` and `--partitions 3` are never applied. The script still prints `OK (topic already exists)` and exits 0.

Reproduced: pre-created `hexgate.otlp.raw` with 1 partition, ran the script verbatim — still `retention.ms 604800000 DEFAULT_CONFIG` (7 days, not the intended 3) and 1 partition. Control confirms the bug is specific to the existing-topic path: `hexgate.otlp.dlq`, created fresh in the same run, correctly got `2592000000 DYNAMIC_TOPIC_CONFIG`.

This is not hypothetical once the Collector lands — see #2.

**Fix** — reconcile after create, instead of trusting `--if-not-exists`:

```bash
rpk topic alter-config hexgate.otlp.raw --set retention.ms=259200000 --no-confirm -X brokers="$BOOTSTRAP"
```

Verified: flips a drifted topic from `604800000 DEFAULT_CONFIG` to `259200000 DYNAMIC_TOPIC_CONFIG`.

Partitions are messier — `rpk topic add-partitions --num N` **adds** N rather than setting a target, so it isn't idempotent and can't be dropped into a re-runnable script as-is. Recommend not trying to reconcile partitions; just prevent the drift via #2.

---

### 2. Broker-side auto topic creation is left enabled — 58/100

`platform/docker-compose.yml:63-70`

`auto_create_topics_enabled` is `true`, so a producer using a wrong/stale topic name — or hitting a broker where `make redpanda-topics` was never run — gets the topic fabricated with cluster defaults (**1 partition, 7-day retention**) instead of a clear error. Proven end-to-end: a Kafka Metadata request with `allow_auto_topic_creation=true` materialized `hexgate.otlp.raw` at 1 partition / `retention.ms 604800000`; the same request with the flag `false` left the broker empty. That flag defaults to **true** in Sarama, which the OTel Kafka exporter uses.

**Important:** this is *not* caused by `--mode dev-container`, and removing that flag will not help. I started a broker without it and rpk still wrote `auto_create_topics_enabled: true` into `/etc/redpanda/redpanda.yaml` (and silently applied `--unsafe-bypass-fsync`) — rpk auto-detects a container environment and applies dev-container defaults regardless. The image's shipped config contains no such key.

**Fix** — set it explicitly. Applies live, no restart needed (`needs_restart: false`):

```bash
rpk cluster config set auto_create_topics_enabled false --no-confirm
```

Verified. Best placed at the top of `create-topics.sh` so it's applied on every provisioning run.

---

### 3. `redpanda-reset` wipes the volume but never recreates the topics

`Makefile:218-222`

Ran the target's exact recipe: exits 0, prints `Container hexgate-redpanda Healthy`, but `rpk topic list` is empty and a produce fails with `UNKNOWN_TOPIC_OR_PARTITION`. `clickhouse-reset` doesn't have this problem because the ClickHouse entrypoint re-runs its init scripts for free.

**Fix** — append to the recipe. Note that a `redpanda-reset: redpanda-topics` prerequisite would be wrong: Make runs prerequisites *before* the recipe, so the topics would be created and then wiped.

```make
redpanda-reset: ## Wipe ONLY the Redpanda data volume and restart
	$(COMPOSE) rm -sf redpanda
	-docker volume rm platform_redpanda-data
	$(COMPOSE) up -d --wait redpanda
	$(MAKE) redpanda-topics
```

---

### 4. `clickhouse-reset` / `clickhouse-down` now also destroy the Redpanda volume

`Makefile:152,165` and the NOTE at `Makefile:170-172`

`clickhouse-reset` runs `$(COMPOSE) down -v`. Confirmed via `--dry-run`: it reports `Volume platform_redpanda-data Removing / Removed` alongside `clickhouse-data`.

`clickhouse-down`'s plain `down` also stops and removes `hexgate-redpanda`. The NOTE warning about this blast radius was extended for Postgres but still names only Postgres.

Scored moderate because this is pre-existing whole-project `down` behavior extended by one more volume, and the lost data is a dev buffer recoverable via `make redpanda-topics`.

**Fix** — either update the NOTE to name Redpanda too, or (better) make the target scoped, matching the shape `postgres-reset` and `redpanda-reset` already use:

```make
clickhouse-reset: ## Wipe ONLY the ClickHouse data volume and re-run init scripts
	$(COMPOSE) rm -sf clickhouse
	-docker volume rm platform_clickhouse-data
	$(COMPOSE) up -d clickhouse
```

guillaume-hexamind (commented):
Above are a couple of issues found by claude.

Otherwise logic looks good for local dev redpanda; will approve once the issues are addressed

victorludvig:
Addressed all 4 findings from the code review, all reproduced before fixing:

1. **`--if-not-exists` never applied retention on an existing topic** — fixed by reconciling retention via `rpk topic alter-config` on every run, not just on first create. Partition count deliberately left unreconciled (rpk's `add-partitions` adds rather than sets a target, so it isn't idempotent).
2. **`auto_create_topics_enabled` left at its default `true`** — now disabled at the top of `create-topics.sh` via `rpk cluster config set`, applied live.
3. **`redpanda-reset` wiped the volume but never recreated topics** — fixed by appending `$(MAKE) redpanda-topics` to the recipe (as a recipe step, not a prerequisite, since prerequisites run before the recipe).
4. **`clickhouse-down`/`clickhouse-reset` also destroyed Redpanda** — scoped both to the `clickhouse` service only, matching the existing `postgres-*`/`redpanda-*` pattern.

Commit `5b583a1`, description updated with a "Review fixes" section detailing each one.

guillaume-hexamind (approved):

PR #112, opened 2026-08-07
refactor(sdk): user -> context/hexgate_context rename + ctx.* docs

guillaume-hexamind:
Naming cleanup + `ctx.*` docs closeout of the `User → HexgateContext` work. **No engine or security-logic change.**

**Eliminates the standalone `user` identifier** (confusing now that the scope type is `HexgateContext`):
- Consumer-facing kwarg **`user=` → `hexgate_context=`** on every adapter method (langchain `ainvoke`/`invoke`/`astream`/`stream`/`astream_events`; google/openai/pydantic_ai runners) and the egress guards. **Breaking**: callers passing `user=` must switch.
- Internal scope variables/params **`user` → `context`** (adapter helpers, `Gate`, `BanGate`, `tracing/usage`, tests, examples).
- Kept as-is: the `HexgateContext.user_id`/`user_roles` fields, `attenuate_for_user(user=<id>)`, biscuit `user(...)` facts, plain-English "user", and `role="user"` message fields.

**Type rename**: the attribute-bag value type **`AttrValue` → `ContextAttributeValue`** (full-word, house-style; avoids colliding with OpenTelemetry's `AttributeValue`), now **exported** from `hexgate` and `hexgate.runtime` so consumers can type their `attributes` bag.

**Adapter metadata**: Langfuse span metadata now stamps `user_roles` (comma-joined string, since the propagation layer drops non-string values) instead of the lossy single `user_role`.

**`ctx.*` docs**: documented the attribute namespace as a first-class advisory fact (peer of `args.*`/`role`/`tool`) in `constraints.py`, `docs/policy/constraints.mdx`, and `docs/cli/policy.mdx`, with the trust contract (populate from trusted server data, same as `user_roles`) — no `trusted_attributes`/signing (that tier is parked in #109).

**Docs migration**: the whole `docs/` tree + `examples/` moved off the removed `User` API to `HexgateContext`; `docs/concepts/user-scope.mdx` rewritten ("Request context") with a worked advisory `ctx.*` example.

Verification: full suite green (1761 passed, 87.6% cov), ruff clean, repo-wide grep gate confirms zero standalone `user` scope identifiers and zero `AttrValue` references remain.

### DIFF
--- a/docs/adapters/google-adk.mdx
+++ b/docs/adapters/google-adk.mdx
@@ -28,7 +28,7 @@ from google.adk.models.lite_llm import LiteLlm
 from google.adk.sessions import InMemorySessionService
 from google.genai import types
 
-from hexgate.runtime import User
+from hexgate.runtime import HexgateContext
 from hexgate.adapters.google import HexgateRunner
 
 INSTRUCTION = (
@@ -64,13 +64,13 @@ async def main():
         tools=[read_logs, restart_service, scale_deployment],
     )
 
-    user = User(user_id="engineer_1", session_id="session_1", role="operator")
+    hexgate_context = HexgateContext(user_id="engineer_1", session_id="session_1", user_roles=["operator"])
 
     session_service = InMemorySessionService()
     await session_service.create_session(
         app_name="devops_demo",
-        user_id=user.user_id,
-        session_id=user.session_id,
+        user_id=hexgate_context.user_id,
+        session_id=hexgate_context.session_id,
     )
 
     runner = HexgateRunner(
@@ -84,7 +84,7 @@ async def main():
         parts=[types.Part(text="Scale the checkout service to 5 replicas in staging.")],
     )
 
-    async for event in runner.run_async(new_message=user_msg, user=user):
+    async for event in runner.run_async(new_message=user_msg, hexgate_context=hexgate_context):
         if event.is_final_response():
             print(event.content.parts[0].text)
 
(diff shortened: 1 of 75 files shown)


victorludvig (changes requested):
Looks good to me, one more file to modify (1) and 1 nit Claude comments to check:

1. (75) benchmarks/bench_enforce.py wasn't migrated — leftover user variable
benchmarks/bench_enforce.py:160-161 still has user = HexgateContext(user_id="bench", user_roles=[case.role]) / with user.sync_scope():. Not a functional bug (the APIs are unchanged), but it contradicts the PR's own claim of renaming internal user variables across "adapter helpers, Gate, BanGate, tracing/usage, tests, examples." Notably, this is the same file the maintainer (Victor) already flagged as a missed call site on the predecessor PR #106 — this is a recurring gap in the same refactor series, not a one-off.

3. (50) OpenAI adapter's hexgate_context param isn't keyword-only, unlike the other three adapters
hexgate/adapters/openai/runner.py's run/run_sync/run_streamed declare hexgate_context as the 3rd positional-or-keyword parameter (no *, guard), while langchain/google/pydantic_ai all enforce it keyword-only. The PR touches this exact line (renaming user→hexgate_context in place) without adding the *, that would make it consistent. Pre-existing shape (same positional-ness existed under the old user name), so arguably out of scope — but worth a "fix while you're here" nudge for consistency.

victorludvig:
Non blocking:
Nit: ", ".join(context.user_roles) has no length cap, and Langfuse's propagate_attributes metadata values are limited to ≤200 chars (silently dropped over that, per propagate_attributes's docstring). Unlikely to bite with typical role counts (1-5), but worth a truncation guard if user_roles can grow large.

guillaume-hexamind:
@victorludvig Thanks for the review and your findings. 
All three issues addressed:

1. bench_enforce.py — migrated the leftover user → context (benchmarks/ was outside
   the rename sweep's scope, hence the recurrence; noted for next time).
2. OpenAI adapter — added *, so hexgate_context is keyword-only in run/run_sync/
   run_streamed, matching the other three adapters. Verified no caller passed it
   (or run_config/hooks) positionally.
3. Langfuse metadata nit — the joined user_roles is now truncated to ≤200 chars with
   an ASCII ellipsis (ASCII since Langfuse drops non-ASCII too). Added tests for the
   join + truncation.

Full suite green, lint/fmt clean. Ready for another look.

victorludvig (approved):
LGTM, approved!

PR #113, opened 2026-08-07
feat(sdk, platform): send attributes in audit events, update clickhouse schema

guillaume-hexamind:
Persists the ABAC context bag (`ctx.*`) that a policy decision was evaluated against, so a `ctx.*`-driven deny can be explained after the fact. `Decision.attributes` already existed in-process but was never sent on the wire.

## Changes

**SDK**
- `audit.py`: `as_payload()` now emits `attributes`, redacted through its own whole-key seatbelt (`_SENSITIVE_ATTR_KEY_RE`) and capped at its own `MAX_ATTRIBUTES_BYTES` (4 KiB — half the args cap; the bag holds caller facts, not tool payloads). Redaction is deliberately narrower-matching than for `arguments`: the bag holds policy facts, so `authorization_tier` / `access_token_scope` survive — blanking them would leave the `ctx.*`-driven deny they caused unexplainable, which is the whole point of persisting the bag — while a key named exactly `token` is still blanked. `arguments` keeps the substring rule, since tool inputs are arbitrary caller data.
- `_truncate_args` generalized to `_truncate_json(payload, *, cap)`, now also applied to `hint` (`MAX_HINT_BYTES`, 4 KiB). `hint` is capped but not redacted — it is policy config (glob lists), not caller data — and only the audit copy is trimmed, so `as_error_payload` keeps the intact hint for the host's error message. Under-cap payloads are shallow-copied on the way out, so no wire payload aliases a live `Decision` field (`hint` had no copy boundary of its own, and it is the same object `as_error_payload` hands the host).
- An empty bag normalizes to `None`: an active context with no attributes yields `{}`, which is indistinguishable from "no bag" downstream (both store `''` and read back as `None`).
- `security/decision.py`: comment only — `attributes` is now persisted by the audit sender, and remains deliberately absent from `as_error_payload` (the model must never see it).

**Platform API**
- `DecisionEvent.attributes` and `AuditDecisionRow.attributes`. Advisory and client-assertable, same trust level as `role` / `user_id`.
- `insert_decision` enforces `MAX_ATTRIBUTES_BYTES` post-serialization (413 on overflow), independently of the arguments cap; `list_decisions` decodes the column.
- The `{}` → `''` normalization is enforced platform-side too, so direct API and non-Python callers hold the invariant rather than storing `"{}"` and making the dashboard render an empty box.
- `test_sdk_caps_do_not_exceed_the_platform_caps` pins the invariant the SDK's truncation rests on: the caps are hand-mirrored across the package boundary, and lowering one side alone would silently reintroduce the 413-drops the truncation exists to prevent.

**ClickHouse**
- New `attributes String CODEC(ZSTD(3))` column on `hexgate_audit.policy_decision`.
- `migrations/0001_add_policy_decision_attributes.sql` — `init/` only runs on an empty volume, so existing volumes need this applied **by hand via `make clickhouse-cli`, before deploying the API that references the column** (skipping it 422s ingest and 503s the whole Audit page, not just the new section). The `ALTER` is additive, metadata-only, and idempotent. Interim convention until a real migration runner exists: every DDL edit in `init/schema.sql` also lands a counterpart here.

**Dashboard**
- New "Context attributes" section in the audit detail drawer, omitted entirely when there is no bag.
- `seed_audit.py` seeds non-PII bags (populated ~2/3 of the time so the omit-when-absent path shows up too), plus a row-width guard that fails before the insert when a hand-written row builder lags `_DECISION_COLUMNS`.

**Docs**
- `concepts/user-scope.mdx`: warns that the bag is persisted for 90 days and rendered verbatim to anyone with project read access — redaction is key-name-only (whole-key for attributes, substring for `arguments`), so callers should pass the coarsest value the policy needs (`{"department": "finance"}`, not an email).
- `internals/audit-pipeline.md`: schema, the 413 cap table, the truncation section, and the known-gaps list updated for `attributes` and the migration convention, plus a privacy-section note on why attribute redaction matches on the whole key where `arguments` matches on substrings.

## Tests

25 added, covering: empty/absent normalization on both sides of the wire, whole-key attribute redaction vs substring argument redaction, non-mutation of the source `Decision`, no wire payload aliasing a live `Decision` field, `attributes` and `hint` truncation staying under the platform caps, each cap being independent, SDK caps ≤ platform caps, snapshot immutability after `decide()`, platform-side rejection and round-trip decoding, and the drawer's render/omit paths.

`platform/api/uv.lock` is an incidental refresh (editable `hexgate` version + dependency specifiers), no functional change.


hexgate/audit.py
```diff
@@ -99,11 +115,30 @@ class AuditEvent:
...
+        # for every denial on that tool.
+        hint = (
+            _truncate_json(d.hint, cap=MAX_HINT_BYTES) if d.hint is not None else None
+        )
+        # Falsy, not ``is not None``: an active context with no attributes
+        # yields ``{}`` (HexgateContext.attributes defaults to an empty dict),
+        # and an empty bag is indistinguishable from no bag downstream — both
+        # store '' and read back as None.
+        attributes = (
+            _truncate_json(_redact(d.attributes), cap=MAX_ATTRIBUTES_BYTES)
```
quanghexa94: **Over-redaction of policy attributes (medium — the one live bug).**

The tool-argument secret redactor (`_SENSITIVE_KEY_RE`) is now reused verbatim on the ABAC attribute bag. Any attribute key whose *name* contains `token`/`authorization`/`secret`/`credential`/`password`/`api_key` is silently blanked to `[REDACTED]`.

**Failure scenario:** a policy that denies on `ctx.authorization_tier == "restricted"` (or a key like `access_token_scope`) stores `{"authorization_tier": "[REDACTED]"}`. The audit record can then no longer explain the `ctx.*`-driven deny — defeating this feature's stated purpose — and it happens silently, because attribute keys are policy *facts*, not secrets like argument keys.

Attribute values are already documented as coming from a trusted source (auth/session/IdP, not raw client input), so key-name redaction arguably shouldn't apply to them at all. Options: a separate (no-op, or value-content-only) redaction path for `attributes`, or — if key-name redaction on `ctx.*` is intended — document explicitly that legitimately-named attribute keys will be blanked.

hexgate/audit.py
```diff
@@ -31,9 +31,21 @@
...
+# Mirrors the platform's MAX_ATTRIBUTES_BYTES. Same reject-don't-truncate
+# semantics as arguments; smaller because the ABAC bag holds caller facts
+# (department, clearance level), not tool payloads.
+MAX_ATTRIBUTES_BYTES = 4 * 1024
+
+# Mirrors the platform's MAX_HINT_BYTES. Same reject-don't-truncate semantics.
+# Only the audit copy is trimmed — ``Decision.as_error_payload`` still carries
+# the intact hint, so the host's file-scope error message keeps its full
+# allowed/denied path lists.
+MAX_HINT_BYTES = 4 * 1024
```
quanghexa94, line 43: **Duplicated byte caps across the package boundary (low — latent).**

`MAX_ATTRIBUTES_BYTES` / `MAX_HINT_BYTES` hand-mirror the platform's caps, kept in sync only by the `# Mirrors the platform's ...` comments. No test asserts equality, and the "truncate so the platform won't 413" invariant depends on SDK cap ≤ platform cap.

**Failure scenario:** someone lowers the platform-side cap and forgets this file. The SDK truncates to the larger size, the platform 413-rejects the payload as non-retryable, and the sender drops the event — exactly the event-loss the truncation exists to prevent, now with nothing to catch the drift. Consider a shared constant, or a contract test asserting SDK caps ≤ platform caps.

hexgate/audit.py
```diff
@@ -99,11 +115,30 @@ class AuditEvent:
...
+            _truncate_json(_redact(d.arguments), cap=MAX_ARGS_BYTES)
+            if d.arguments is not None
+            else None
+        )
+        # Not redacted — a file-scope hint is policy config (glob lists), not
+        # caller data — but still capped: a policy enumerating enough paths
+        # would 413 and take the whole event down with it, deterministically,
+        # for every denial on that tool.
+        hint = (
+            _truncate_json(d.hint, cap=MAX_HINT_BYTES) if d.hint is not None else None
```
quanghexa94, line 154: **`hint` payload aliases the live `Decision.hint` (low — latent).**

When the hint fits under the cap, `_truncate_json` returns the same object it was given, so the wire payload's `hint` key aliases the live `Decision.hint` rather than a copy. Unlike `arguments`/`attributes` (which pass through `_redact` and get a fresh dict), `hint` has no copy boundary.

Harmless today — `as_payload`'s result is JSON-serialized immediately and never mutated — but a future in-place mutation of the payload's `hint` would corrupt the host-held `Decision.hint` that `as_error_payload` hands back to the model. Cheap to make robust with a shallow copy at the choke point.

quanghexa94 (commented):
Automated code review — 3 findings on the audit envelope change (attribute bag → ClickHouse/dashboard). Implementation is clean and well-tested; findings below are low-to-medium severity, only the first is a live bug.

guillaume-hexamind:
Thanks @quanghexa94 for the comments.
All three issues fixed, ready for another review

1. **Over-redaction (medium).** Attributes now use their own whole-key pattern
   (`_SENSITIVE_ATTR_KEY_RE`); `arguments` keeps the substring rule. So
   `authorization_tier` / `access_token_scope` survive and the deny stays
   explainable, while a key named exactly `token` is still blanked. You were
   right that the bag is policy facts, not payloads — I kept a seatbelt rather
   than dropping redaction entirely, since the drawer renders it verbatim.
2. **Duplicated caps (low).** Added `test_sdk_caps_do_not_exceed_the_platform_caps`
   asserting SDK ≤ platform for all three fields. I skipped the shared constant:
   it would tie server-side validation to whichever SDK version is installed and
   still wouldn't cover an old SDK in the wild against a lowered platform cap —
   that one needs the platform to truncate instead of 413, noted as a known gap.
3. **`hint` aliasing (low).** `_truncate_json` now returns `dict(payload)` on the
   under-cap path, which covers all three call sites. Worth noting it was
   slightly broader than described: `as_error_payload` hands the model-facing
   payload the same `self.hint` object, so the aliasing spanned three holders.
   The copy is shallow — enough for a rebind, not a nested in-place mutation.

winlp4ever (approved):
LGTM

quanghexa94 (approved):

PR #115, opened 2026-08-11
feat(sdk): role-aware policy assembly (roles as capability bindings)

quanghexa94:
Make policy resolution role-aware: a project with many roles now resolves each role to its own effective policy, instead of the linker folding everything into a single `default`. This is the SDK groundwork for phase 3 (bringing multi-module policy to the platform), after #97 (modules + linker) and #101 (analyzer). Pure SDK, fully local, no platform change.

The real policies are multi-role, but doing that by mapping roles to authored inheritance is exactly what gets dangerous at scale (role bundles duplicated per agent, an `inherits` DAG whose effective permissions you have to compute). This PR avoids that.

## Design

A role stops being a policy block with `inherits`. It becomes a thin binding, `role -> [capability names]`, in a `roles.yaml` that lives outside `policies/` on purpose (a role is a binding, not a policy module). Three properties fall out:

- Capabilities are the only authored, reusable unit. A role selects them, flat union, no inheritance.
- Boundaries are role-independent ceilings. Every role folds against the same boundaries, so a role can only narrow, never widen, its ceiling. Escalation past a boundary is structurally impossible.
- Multi-role is compiled, not authored: the role-keyed `PolicySet` the WASM engine already expects is a derived expansion of the binding map.

`resolve_for_project(boundaries, library, roles, *, agent_leaf, agent_boundaries)` loops the existing fold once per role and returns `ProjectLinkResult(policy_set, by_role)`. No `roles.yaml` means one `default` role importing every capability, the all-compose behaviour `--dir` had before, so existing bundles resolve unchanged. A missing `default` is synthesised as an empty selection (unroled callers get fail-closed deny). An unknown capability name in a role is a `LinkError`.

The analyzer runs per role and tags each lint with the role it fired in (a grant dead under one role's ceiling can be alive under another), plus two project-level lints:

| Lint | Severity | From |
| --- | --- | --- |
| `unused-capability` | info | a library capability no role imports |
| `no-default-role` | info | roles defined but no `default`, so unroled callers deny |

## Important files

| File | What to look at |
| --- | --- |
| `hexgate/security/linker.py` | `resolve_for_project`: the per-role loop over the unchanged `link`. The fold is untouched, this only picks each role's inputs. |
| `hexgate/security/module_loader.py` | `load_roles`: reads `<root>/roles.yaml`, and why it sits outside `policies/`. |
| `hexgate/security/analyzer.py` | `check_project` / `analyze_project`: per-role attribution and the two new lints. |
| `hexgate/security/modules.py` | `ProjectLinkResult`. |
| `hexgate/cli/policy/main.py` | `--role` on `resolve` and `check`, and the single-role vs multi-role output. |
| `tests/security/test_linker.py` | role folding, boundary-applies-to-all, unknown capability, no-roles back-compat, synthesised default. |

## Try it

```
hexgate policy resolve --dir deploy/demo_policies                 # every role
hexgate policy resolve --dir deploy/demo_policies --role billing  # one role
hexgate policy check   --dir deploy/demo_policies                 # lints, per role
```

`deploy/demo_policies` now ships a `roles.yaml` (default / support / billing) and a shared `read_only` capability. billing gets refunds capped at 1000 by the boundary and to USD/EUR by the payments pack; default is read-only; support adds email and escalate.

## Notes

- Full SDK suite green (1789 passed, 6 opt-in skips). ruff clean. No `platform/api` changes.
- Untouched on purpose: the single-role `link` / `link_policy_set`, the compile path, the WASM runtime, and signing.
- Design and the remaining phase-3 plan (platform `policy_module` + `role_binding` store, dashboard, converter) are in `policy-modules-phase3-design.md`.

### DIFF
--- a/deploy/demo_policies/policies/capabilities/read_only.yaml
+++ b/deploy/demo_policies/policies/capabilities/read_only.yaml
@@ -1,0 +1,5 @@
+# Read-only capability pack (team-owned). Grants only. Shared by every role,
+# so it lives once here instead of being copied into each role.
+tools:
+  view_orders: { mode: allow }
+  lookup_order: { mode: allow }
--- a/deploy/demo_policies/roles.yaml
+++ b/deploy/demo_policies/roles.yaml
@@ -1,0 +1,11 @@
+# Role bindings. A role is a named set of capabilities, not a policy — it lives
+# outside policies/ on purpose. Boundaries (policies/boundaries/) apply to every
+# role automatically and are never listed here.
+#
+#   hexgate policy resolve --dir deploy/demo_policies              # all roles
+#   hexgate policy resolve --dir deploy/demo_policies --role billing
+version: 1
+roles:
+  default: [read_only]                        # fallback: read-only
+  support: [read_only, support_leaf]          # + email / escalate
+  billing: [read_only, payments]              # + refunds (capped by the boundary)
--- a/deploy/policy_modules_demo.py
+++ b/deploy/policy_modules_demo.py
@@ -98,7 +98,8 @@ def _(mo):
 
 @app.cell
 def _(dedent):
-    # Three modules — the same files that live in deploy/demo_policies/. Edit
+    # A boundary + two capabilities, mirroring deploy/demo_policies/ (which also
+    # ships a read_only capability and a roles.yaml for the multi-role view). Edit
     # them here and the whole notebook re-resolves.
     BOUNDARIES = {
         "org_core": dedent(
--- a/hexgate/cli/policy/main.py
+++ b/hexgate/cli/policy/main.py
@@ -226,6 +226,15 @@ def add_parser(subparsers: argparse._SubParsersAction) -> None:
         help="Repo root containing a policies/ tree (default: current dir).",
     )
     p_resolve.add_argument(
+        "--role",
+        default=None,
+        help=(
+            "Print just this role's effective policy. Without it, a multi-role "
+            "project (a roles.yaml) prints every role; a single-role project "
+            "prints the one effective policy."
+        ),
+    )
+    p_resolve.add_argument(
         "-o",
         "--output",
         help="Write the effective policy YAML here (default: stdout).",
@@ -249,6 +258,11 @@ def add_parser(subparsers: argparse._SubParsersAction) -> None:
         "--dir",
         default=".",
         help="Repo root containing a policies/ tree (default: current dir).",
+    )
+    p_check.add_argument(
+        "--role",
+        default=None,
+        help="Restrict lints to this role (plus project-level ones). Default: all roles.",
     )
     p_check.add_argument(
         "--manifest",
@@ -496,15 +510,17 @@ def _main_show_rego(args: argparse.Namespace) -> int:
 
 
 def _main_resolve(args: argparse.Namespace) -> int:
-    """Link the local module bundle into one effective policy and print it."""
+    """Resolve the local project into effective policy per role and print it."""
     from hexgate.security import (
         LinkError,
-        link_policy_set,
         load_local_modules,
+        load_roles,
+        resolve_for_project,
     )
 
     try:
         boundaries, capabilities = load_local_modules(args.dir)
+        roles = load_roles(args.dir)
     except (ValueError, OSError) as exc:
         print(f"load error: {exc}", file=sys.stderr)
         return 1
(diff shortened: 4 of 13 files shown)


guillaume-hexamind (commented):
I approve of the logic and your design docs. A couple of issues found by claude below; I will approve once addressed.


---
1. policy resolve -o on a multi-role project produces a deny-everything bundle — 94/100

hexgate/cli/policy/main.py:565-570

The multi-role branch of _main_resolve dumps a bare role-keyed mapping:

payload = {
    role: lr.effective[DEFAULT_ROLE_NAME].model_dump(mode="json")
    for role, lr in sorted(result.by_role.items())
}

The resulting document has billing: / default: at the top level, with no roles: wrapper. But load_policy_set_from_dict (policy_set.py:214) dispatches on isinstance(payload.get("roles"), dict) and, failing that, valis a single flat AgentPolicy. AgentPolicy is not extra=forbid,so every role key is silently discarded.

Repro (two roles, default: [ro], billing: [ro, pay]):

hexgate policy resolve --dir . -o eff.yaml     # exit 0
hexgate policy build eff.yaml --no-wasm        # exit 0, writeon
grep -c 'view\|refund' eff.rego                # -> 0
hexgate policy validate eff.yaml               # "✓ Policy par

The compiled Rego contains no tool rules at all — tools={}, deg in the pipeline warns. build also overwrites the sourceeff.yaml with the flattened result. resolve --help describes this file as "the intermediate artifact between many module files and the signed WASM
bundle", so resolve -o → build is the advertised path. test_reo asserts only the exit code, so nothing catches it.

Suggested fix — emit the shape load_policy_set_from_dict alrea

payload = {
    "roles": {
        role: lr.effective[DEFAULT_ROLE_NAME].model_dump(mode=
        for role, lr in sorted(result.by_role.items())
    }
}

Worth adding a test that round-trips the resolved file back through load_policy_set_from_dict (or through policy build) and asserts the tools survive —
an exit-code assertion can't catch this class of bug. The --ros can stay flat (that's the legacy shapeload_policy_set_from_dict handles), but wrapping them too would make the output shape uniform.

---
2. An empty or misspelled roles: key fails open to every capab

hexgate/security/linker.py:87-92, hexgate/security/module_load
                                                                                                                                                  load_roles returns {} for all of: missing file, missing roles ), roles: null, and roles: {}. resolve_role_map reads any falsymap as the all-compose back-compat signal:                                                                                                        
names_by_role = (                                                                                                                                     {name: list(sel) for name, sel in roles.items()}
    if roles                                                                                                                                          else {DEFAULT_ROLE_NAME: [cap.name for cap in library]}
)                                                                                                                                                 
So "no roles.yaml" and "a roles.yaml I typo'd" are indistinguishable, and the typo resolves to the most permissive outcome.                       
Repro: with roles: {default: [ro]}, resolve --dir . gives default → view only. Change the key to role: (or set roles: {}) and the same tree gives default → view and refund. Contrast with a roles.yaml that's vhat path correctly synthesises an empty, fail-closed default.

Bounded by the boundary ceiling (a role can never exceed policover-grants inside the fence rather than outside it — but it'sstill a silent widening from a one-character typo.

Suggested fix — key the all-compose fallback off the file's absence, not off an empty parsed map. E.g. have load_roles return None when roles.yaml
doesn't exist and {} when it exists but binds nothing, then mae_for_project take Mapping[...] | None and use the all-composebranch only on None. Callers to update: _main_resolve, _main_check, check_project/analyze_project.

Complementary hardening: reject unknown top-level keys in roles.yaml so role: is a loud ValueError rather than a silent empty binding.

---
3. Capabilities bound to no role skip link validation, so poliroken module — 68/100

hexgate/security/analyzer.py:174-183, hexgate/security/linker.

resolve_role_map only materialises capabilities a role names,  resolved.items(), and the capability-may-only-grant check lives inside _fold_tool — so it never runs for a module no role imports. _unused_capabilities reports it at info, below the default --max-severity=error.

Repro — policies/capabilities/bad.yaml with tools: {refund: {mode: deny}}:

# no roles.yaml
✗ [link-error] capability 'bad' denies 'refund'; capabilities

# add roles.yaml with `roles: {default: [ro]}` (does not bind
· [unused-capability] capability 'bad' is imported by no role                     exit 0

Scope: schema validation still runs on every file (_read_dir → AgentPolicy.model_validate), and an unbound module isn't enforced anywhere, so this is a
CI safety-net regression rather than a runtime hole — a malforthe repo green until someone binds it and production resolutionraises LinkError.

Suggested fix — validate the whole library once, independent of role binding, before the per-role folds in resolve_for_project. The capability-deny rule
is a per-module property (a capability tool with mode: deny isch tools the fold reaches), so it can be hoisted out of_fold_tool into a module-level check run over the full library, alongside _reject_file_scope / _reject_default_policy_constraints on the same full list.
Cheaper alternative if you'd rather not restructure: raise unu which at least makes it visible — though it still won't catchthe structural error itself.

---
4. Two capability files sharing a stem (x.yaml + x.yml) shadowrop grants — 64/100

hexgate/security/linker.py:86, hexgate/security/module_loader.

_read_dir globs both extensions into one list and derives the (directory).with_suffix("").as_posix(), so dup.yaml and dup.ymlboth load as dup. resolve_role_map then builds index = {cap.name: cap for cap in library} and one silently wins.

This is a regression, and it hits projects with no roles.yaml at all — the back-compat path. Same fixture, dup.yaml allows view, dup.yml allows refund:

origin/main : tools: view + refund
327c08d     : tools: refund only
              provenance prints "[capability] dup (…/dup.yml)" twice

On the no-roles path names_by_role becomes ["dup", "dup"], so the survivor is folded twice while the other file's grants vanish — no diagnostic, and
_unused_capabilities stays quiet because the name is imported.e fine (there's already a test for that); it's only thetwo-extension case.

Low likelihood — it needs an unusual layout — but the failure mode is a silently denied tool the operator believes is granted.

Suggested fix — make duplicate module names a load error in _read_dir, naming both files:

by_name: dict[str, Path] = {}
...
if name in by_name:
    raise ValueError(
        f"duplicate module name {name!r}: {by_name[name]} and {file}"
    )
by_name[name] = file

Deduping the name list in resolve_role_map is worth doing too (it also fixes the repeated-name lint noise), but it won't recover the lost module — the
collision has to be rejected at load.

---
5. A non-mapping roles.yaml crashes with an uncaught AttributeError — 58/100

hexgate/security/module_loader.py:83-87

The try wraps only yaml.safe_load; raw = payload.get("roles") sits just below it. Both CLI handlers catch only (ValueError, OSError), so a non-mapping
document escapes as a traceback.

roles.yaml = "- default\n- billing"
  hexgate policy resolve --dir .  -> AttributeError: 'list' object has no attribute 'get'
  hexgate policy check  --dir .   -> same traceback

roles.yaml = "just a string"
  -> AttributeError: 'str' object has no attribute 'get'

Every other malformed-roles.yaml shape in this function is deliberately converted to a clean ValueError with a roles.yaml is invalid / roles.yaml: role
'x' must map to … message, so this is the one gap in an otherw

Suggested fix — guard the top-level type right after the parse

if not isinstance(payload, dict):
    raise ValueError(f"{ROLES_FILE}: top level must be a mapping (got {type(payload).__name__})")

quanghexa94:
Thanks for the thorough pass. All five addressed and folded into the feature commit (`aa3bbb4`); full SDK suite green (1800), each fix with a regression test.

1. **`resolve -o` on a multi-role project produced a deny-everything bundle.** The multi-role branch now wraps the output in a `roles:` map, so it round-trips through `load_policy_set_from_dict` / `policy build` instead of flattening to a single AgentPolicy with the role keys silently dropped. Added a test that loads the resolved file back and asserts the tools survive, not just the exit code.

2. **An empty or misspelled `roles:` failed open to all-compose.** `load_roles` now returns `None` only when the file is absent (the all-compose signal) and a dict, possibly empty, when it exists (a fail-closed default). `resolve_role_map` / `resolve_for_project` / `check_project` / `analyze_project` take `Mapping[...] | None` and only all-compose on `None`. Complementary hardening: unknown top-level keys in `roles.yaml` are now a loud `ValueError`, so a typo'd `role:` errors instead of binding nothing.

3. **Capabilities bound to no role skipped link validation.** Hoisted the capability-may-only-grant check out of `_fold_tool` into `_reject_capability_denies`, run over the whole library in `resolve_for_project` (bound or not), alongside the existing `file_scope` / `default_policy` rejections. An unbound `deny` now raises at resolve and surfaces as a `link-error` lint on check.

4. **Two files sharing a stem (`dup.yaml` + `dup.yml`) shadowed each other.** `_read_dir` now rejects duplicate module names, naming both files, so the collision is a load error rather than a silently dropped grant.

5. **A non-mapping `roles.yaml` crashed with an uncaught `AttributeError`.** Guarded the top-level type right after the parse, so a list or scalar document gives the same clean `ValueError` as every other malformed shape.

Tests added for each: resolve round-trip, absent-vs-empty roles, unbound capability-deny, duplicate-stem, non-mapping and typo'd `roles.yaml`. Ready for another look.

guillaume-hexamind (approved):
Good to me

PR #116, opened 2026-08-12
feat(platform-api): project policy-module store + resolve/check API

quanghexa94:
Add the platform policy-module store and resolve/check API: the backend that lets a project hold boundary + capability modules and role bindings (not one policy_yaml per agent), and inspect the composed result. This is the groundwork behind the dashboard's no-code policy display.

Rebased onto `main` now that #115 (role-aware assembly) has merged, so the diff here is platform-only.

Additive by design: two new tables and one feature slice, no change to `Agent` or the compile/serve path, so no live agent is touched.

## Design

Two new SQLModel tables, both project-scoped (see `docs/adr/R-POL-001`):

- `PolicyModule` — one boundary or capability module, keyed `(project_id, tier, path)`. No `agent_id`: agent-scoped self-restrictions are deferred until a compile path consumes them, so shipping the column now would be unused.
- `RoleBinding` — `(project_id, role) -> [capability names]`. A role is a binding, not a policy; boundaries apply to every role and are never listed.

Both are new tables, so `SQLModel.metadata.create_all` picks them up with no migration (the platform has no Alembic yet, and this deliberately avoids needing it).

The `policy_modules` feature slice mirrors the existing `features/projects` shape: module CRUD, role bindings, and `resolve` / `check`. The resolve/check handlers build `ModuleContent` from the stored rows and call the SDK's `resolve_for_project` / `check_project` — the fold itself is not reimplemented. Reads gate on org membership; writes on project admin (policy editing is a management action). `check` returns diagnostics-as-data (always 200, mirroring `validate`); `resolve` 422s when the set can't be composed (e.g. a capability that denies).

## Important files

| File | What to look at |
| --- | --- |
| `platform/api/hexgate_api/models.py` | `PolicyModule` + `RoleBinding` — the store shape, and why it's project-scoped with no `agent_id`. |
| `platform/api/hexgate_api/features/policy_modules/service.py` | building `ModuleContent` from rows + the resolve/check wrappers over the SDK. |
| `platform/api/hexgate_api/features/policy_modules/router.py` | the endpoints, auth split (member reads / admin writes), and the 200-diagnostics vs 422 convention. |
| `docs/adr/R-POL-001-multi-module-policy.md` | the architecture decision this implements. |
| `platform/api/tests/features/policy_modules/test_policy_modules.py` | CRUD, per-role resolve, and link-error-as-lint. |

## Test plan

Full platform-api suite green (457 passed, 4 skipped) against the post-#115 `main`. Manually exercised the slice end to end: seed a boundary + two capabilities + a `roles.yaml`, `resolve` returns the boundary cap intersected with the payments grant on `refund_order` for the billing role and a read-only default; `check` on a capability that denies returns a `link-error` lint and `resolve` 422s. Regression tests added: `set_roles` re-set with a reused role name (was a 500 before the flush), and module replace-not-duplicate.

## Try it

New endpoints, all under `/v1/projects/{project_id}`:

```
GET    /policy-modules                          list modules
PUT    /policy-modules/{tier}/{path}            upsert a module (body: {content})
DELETE /policy-modules/{tier}/{path}            remove a module
GET    /policy-roles      PUT /policy-roles     read / set role bindings
GET    /policy/resolve[?role=]                  composed effective policy per role
GET    /policy/check                            lints (diagnostics-as-data)
```

```
cd platform/api && pytest tests/features/policy_modules/ -q
```

## Notes

- No `platform/api` behaviour outside this slice changes.
- Deferred on purpose: `policy_mode` on Project + compiling an agent's bundle *from* the modules (the piece that needs the one column `create_all` can't add, and touches the serve path); `check?agent=` manifest drift lints; tier-split write auth (currently admin for all writes).

### DIFF
--- a/docs/adr/R-POL-001-multi-module-policy.md
+++ b/docs/adr/R-POL-001-multi-module-policy.md
@@ -1,0 +1,52 @@
+# R-POL-001: Policies compose from project-scoped modules; roles are bindings
+
+**Status:** Accepted · 2026-08-11
+**Applies to:** `hexgate/security/**`, `platform/api/hexgate_api/features/policy_modules/**`, policy authoring under `policies/boundaries/**`, `policies/capabilities/**`, `roles.yaml`
+
+## Decision
+
+An effective policy MUST be composed from small modules by the linker, not authored as one file per agent.
+
+- Modules MUST live in two tiers, decided by the top-level folder, not a field: `policies/boundaries/` (security ceilings and hard denies) and `policies/capabilities/` (grants only). A capability MUST NOT `deny`.
+- The tier folders MUST be at `<root>/policies/boundaries/` and `<root>/policies/capabilities/`. Files MAY nest to any depth inside them; the subpath becomes the module name. `roles.yaml` MUST be at the repo root, outside `policies/`.
+- A role MUST be a binding (`role -> [capability names]`), not a policy block with `inherits`. Boundaries MUST apply to every role and MUST NOT be listed in a role.
+- On the platform, modules MUST be scoped to the project. Agent-scoped modules MUST NOT be added until a compile path consumes them.
+- The reconciliation fold MUST stay unchanged: fences intersect, grants union, denies win. The resolved policy compiles to Rego then WASM then a signed bundle exactly as a single policy does.
+
+## Why
+
+An agent's policy grows two ways at once. Security wants org-wide ceilings that every agent inherits and no team can loosen; teams want reusable tool grants. One file per agent forces both into one place, so the org ceiling is copied into every agent and drifts, and a shared capability has no single home. Scoping modules to the project fixes both: a boundary exists once and is inherited, a capability exists once and is imported.
+
+Roles are bindings, not inheritance, for two reasons that bite at scale. Authored role blocks duplicate across agents, so "what can billing do" has no single definition. And `inherits` is a DAG whose effective permission set you have to compute, which is where multi-role permission bugs hide. A flat `role -> [capabilities]` list has neither problem.
+
+Boundaries are role-independent on purpose: every role folds against the same boundaries, so a misconfigured role binding can never exceed a ceiling. Role escalation past a boundary is structurally impossible, not merely discouraged.
+
+Full narrative (the worked examples, the store shape, the alternatives) lives in the design doc and the authoring guide; this ADR is the citable decision, not a copy. See `../../policy-modules-phase3-design.md` and `../../how-to-structure-policy-modules.md`, and the Notion "Multi-module policy: structure and design decisions".
+
+## Consequences
+
+- Capabilities become the reusable, reviewable unit; a role is a one-line list. Change a capability once and every role that imports it updates.
+- The manifest keeps its separate job: it is the drift cross-check (`unknown-tool` / `unknown-arg` lints), not the grant source. Presence is not permission.
+- Migration is coexistence, not a big-bang: classic single-file and modular resolve to the same compiled bundle, so agents move over one at a time.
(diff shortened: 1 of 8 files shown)


victorludvig (commented):
Hi @quanghexa94, can you check these findings:

1. get_roles() returning {} breaks the PR's own fail-open default — service.py:427
When a project has capability modules but no roles ever set, get_roles() returns {}, and resolve/check treat empty-dict roles as fail-closed (no tools granted) rather than the documented "no roles.yaml → single default role imports everything" behavior from this PR's own ADR (docs/adr/R-POL-001-multi-module-policy.md:37). Verified live: after PUTting a capability and never touching /policy-roles, resolve returns tools:{} and check reports ok:true with only an info-level lint — looks healthy while access is actually all denied. No test exercises this path since every existing test calls PUT /policy-roles first.

2. set_roles has no race protection on concurrent wholesale replace — service.py:445
Unlike upsert_module's explicit retry-on-IntegrityError, concurrent PUT /projects/{id}/policy-roles calls can both delete+reinsert and collide on the unique (project_id, role) constraint, propagating an uncaught IntegrityError → 500. Reproduced with two concurrent set_roles calls via asyncio.gather.

3. upsert_module's IntegrityError handler assumes only one cause — service.py:402
It assumes the only possible IntegrityError is a create-race on (project_id, tier, path), re-queries, and asserts the row exists. An FK violation (bogus project_id) instead trips the assert as an unhandled AssertionError. Currently unreachable via HTTP (the route 404s on missing project first), but latent for any future direct caller.

4. content_hash uses a different hashing scheme than the SDK's loader — service.py:315
Platform store hashes raw YAML text; the SDK's module_loader.py hashes canonical JSON of the parsed payload. Two semantically-identical YAML docs with different formatting get different hashes here but the same hash in the SDK loader, breaking the "stable identity" invariant content_hash is meant to provide once platform- and SDK-authored modules are compared/deduped.

5. No caching of parsed policy on resolve/check — service.py:467
Every resolve/check call re-parses and re-validates every module's YAML from scratch, with no cache keyed on the already-computed content_hash. Not a correctness bug, but wasted CPU scaling with module count per project on every request (e.g. dashboard polling).

quanghexa94:
Thanks for these (relaying them here since GitHub was down when you reviewed). All five addressed on `74c9f11` — four fixed, one deferred as perf. Full platform suite green (459).

1. **`{}` fail-closed vs the ADR.** Fixed. `_sdk_inputs` now passes `roles or None`, so a project with capabilities but no bindings resolves as one `default` importing everything, matching R-POL-001 and `hexgate policy resolve --dir`. The `{}`-vs-`None` split stays for the SDK (an empty or typo'd `roles.yaml` file still fail-closes), but the platform has no typo-able file, so "no bindings" is the no-roles case (`None`), not the empty one. Added the no-bindings test that nothing covered.

2. **`set_roles` race.** Fixed. Retries once on `IntegrityError`, the same posture as `upsert_module`'s create-race handling, so two concurrent wholesale replaces no longer collide into a 500.

3. **`upsert_module` assert on a non-race cause.** Fixed. If the post-rollback re-query is still `None` it re-raises the original error instead of asserting, so an FK violation surfaces as itself rather than an `AssertionError`.

4. **`content_hash` diverges from the SDK loader.** Fixed. Now hashes canonical JSON of the parsed payload (`default=str`), byte-identical to `module_loader._canonical_hash`. Added a test pinning the match plus format-stability.

5. **No parsed-policy cache on resolve/check.** Deferred. Perf only, as you noted. I'll add a `content_hash`-keyed parse cache when dashboard polling makes it a real hot path.

Good catches, especially 1 — it contradicted our own ADR and looked healthy while denying everything. Ready for another look.

victorludvig (approved):

PR #117, opened 2026-08-12
feat(sdk): multi-role policy enforcement

guillaume-hexamind:
Enforce **every** role in `HexgateContext.user_roles` instead of only the first.

Access is granted iff *any* of the caller's roles grants it: the enforcer asks the engine for one `Verdict` per role and folds them with a permissive union — `ALLOW > NEEDS_APPROVAL > DENY` — short-circuiting on the first allow.

SDK only. The audit wire payload is byte-identical to today, so this needs nothing from the platform. Persisting the role set + deciding role (ClickHouse columns, ingest, dashboard) is the follow-up PR.

**Two accepted gaps until follow-ups land:**

1. **The audit trail records *who was calling*, not *which role granted the call*.** The `role` column keeps its literal current meaning — the caller's first role — so nothing stored is false; for genuinely multi-role callers it is incomplete. The deciding role and the full set are on the `Decision` in-process (decision observer, `hexgate chat`, the serve approval frame), just not yet on the wire. `docs/concepts/user-scope.mdx` says so explicitly rather than implying otherwise.
2. **The token still attests only the caller's first role** while the enforcer evaluates all of them. Multi-role attenuation was dropped from this PR (`945650f`) to keep the diff on enforcement: closing the enforce-on-N / attest-on-1 gap also means teaching the enforcer to *prefer* the verified facts over the contextvar, which belongs with the signed-attribute-tier work. `attenuate_for_user`'s docstring states the limitation at the seam.

## Changes

- **`security/decision.py`** — `combine_role_verdicts`: pure, engine-agnostic fold. Merged denials union their violations and bound the composed reason to 4096 chars (the platform's `reason` cap — an unbounded merge would 422 and lose the audit event). Identical denials collapse verbatim, so single-role messages are unchanged.
- **`security/enforcer.py`** — resolves the caller's distinct roles in order and folds. Capped at 32 roles (fail-closed: dropping roles can only narrow a permissive union), warned once per process. Resolution lives in a shared **`runtime/roles.py`** so the enforcer, the `policy test` dry-run and token attenuation cannot drift — a dry-run that normalised differently from production would be worse than none. A bare `str` is rejected there rather than iterated per character: `user_roles="admin"` type-checks against `Sequence[str]` but would have become five one-letter roles and never `admin`.
- **`Decision`** — gains `user_roles` and `deciding_role`; `role` becomes a property returning the caller's first role, so the audit wire field is unchanged; every renderer moved to `deciding_role`. The LLM-facing payload names the deciding role on an approval and **no** role on a deny (a deny had none, and the caller's other roles are withheld like `attributes` are).
- **`HexgateContext.primary_role` deleted** — enforcement no longer reduces the set at all, and a new call site would silently re-introduce single-role enforcement. The one remaining reduction is explicit and local: `factory.py` passes `user_roles[0]` to the token attenuation, with a comment pointing at gap 2 above.
- **CLI** — `hexgate chat` and the approval panel show `roles:` + `granted by:`; the serve approval frame keeps its scalar `role` (now the granting role) and adds `roles`, additively; `hexgate policy test --roles a,b` dry-runs the union through the same fold *and* the same 32-role cap. An empty `--roles` now fails instead of silently dry-running `default` — scripted as `--roles "$ROLES"` with an unset variable, that used to exit 0 while asserting nothing.
- **Two new lints, both `warning`** — `permissive-default` fires when the `default` role grants a tool no named role grants; `implicit-default` fires when a document declares roles but names none of them `default`, so the loader silently aliases one and hands its whole grant set to every undefined name. Both wired into `hexgate policy validate --max-severity`, which now folds on each lint's own severity (like `policy check` already did) and prints its success line only when the gate passes. See the note below.

## Engines are untouched

No Rego, WASM, DSL or `policy_for` change; no golden-fixture regen. The fold operates on `Verdict`s both engines already produce, so they stay byte-for-byte comparable one role at a time. The `role` fact still binds to the role being evaluated, so `role == "billing"` keeps working for a multi-role caller.

`test_engine_parity.py` gains 14 cases running the same role sets through the pydantic engine and real `opa`→`wasmtime`: outcome **and** deciding role must agree, plus a check that wrapping a single role in the union is identical to evaluating it directly.

## Two semantics worth reviewing

**Approval is not sticky.** If one role allows a tool outright and another would gate it on approval, the call runs with no prompt. This is the permissive reading; the alternative is restrictive and contradicts the union. Pinned by a named test.

**`default` is a fallback, not a floor.** A caller whose role is defined never inherits `default`. But every *unrecognised* name in `user_roles` still resolves to `default` and contributes its permissions — so a caller who controls their own role list can always reach whatever `default` grants. That is why the lints ship in this PR rather than a docs paragraph. The implicit case is the sharper one: with no `default` role declared at all, the loader aliases the first concrete role as the fallback, so *every* grant that role has becomes reachable by any undefined name. Verified clean under **both** lints against all six example policies and the three platform seed policies.

## Compatibility

Non-breaking for single-role callers: one role in `user_roles` produces the verdict that role produces alone (pinned on both engines). **It does widen access for anyone already passing multiple roles** — they now get the union rather than the first role's permissions.

`Decision.role` moving from field to property means `Decision(role=...)` no longer constructs — reading `.role` is unchanged. `Decision` is internal (`hexgate.security.decision`) and not exported for construction, but any in-repo or out-of-repo constructor needs `user_roles=("x",)` instead.

## Diff shape

37 files, +2089 / −195. Tests **+1269 / −61** (17 files) · SDK **+630 / −125** (14 files, of which ~411 added lines are code and ~163 comments/docstrings) · docs **+119 / −9** (5 files) · benchmarks **+71** (1 file). Three lines of test per line of SDK code.

## Perf

10,000 iterations/case (the `bench_enforce.py` default), WASM engine, p50: 1 role 54µs · 2 roles allowing on the first 54µs · allowing on the second 107µs · 5 roles all-denying 260µs. Linear in roles *actually* evaluated (~53µs per invocation), with the short-circuit holding the common case at single-role cost. `benchmarks/bench_enforce.py` covers all four.

Parallel evaluation was considered and rejected: `wasmtime.Store` isn't thread-safe so `evaluate` already serialises under a per-instance lock, the pydantic engine is a GIL-bound microsecond walk, and a fan-out would discard the short-circuit and make the deciding role a race.


hexgate/security/decision.py
```diff
@@ -65,6 +70,106 @@ def evaluate(
...
+    stay within, so merging two different scopes would state something false.
+    """
+    if not denials:  # pragma: no cover - callers only merge non-empty denials
+        raise ValueError("_merge_denials needs at least one denial")
+    verdicts = [verdict for _, verdict in denials]
+    first = verdicts[0]
+    if all(other == first for other in verdicts[1:]):
+        return first
+
+    violations: list[str] = []
```
quanghexa94, line 130: **Audit event can be dropped on wide multi-role denials — `violations` is uncapped.**

`_merge_denials` unions violations across roles with no cap, but the platform's `DecisionEvent.violations` is bounded (`max_length=64` items, 1024 chars each). A caller carrying ~9+ roles whose policies each contribute several distinct violations (e.g. 9 × 8 = 72 unique) produces a `Verdict` whose payload the platform rejects with **422 — losing the audit record for a *denied* call**, which is exactly the failure the `reason` cap (`_MAX_REASON_CHARS`) on this same function was added to prevent.

`reason` was bounded via `_bounded_reason`; `violations` was not. Consider capping the unioned `violations` the same way (drop-and-disclose, e.g. a trailing `"(+N more)"` marker) so a wide multi-role deny still produces a persistable event.

hexgate/security/analyzer.py
```diff
@@ -420,3 +422,99 @@ def _unknown_args(
...
+    named = [role for role in policy_set.roles if role != DEFAULT_ROLE_NAME]
+    if not named:
+        return []
+
+    default_policy = policy_set.policy_for(DEFAULT_ROLE_NAME)
+    # Drop the role ``default`` merely aliases: it resolves to the *same policy
+    # object*, so leaving it in answers "does a named role grant this too?" with
+    # yes for every one of its own grants — silencing the check on the very
+    # shape that most needs it.
+    others = [
```
quanghexa94, line 480: **False `permissive-default` warnings when a policy declares roles but no explicit `default`.**

`others` excludes the implicitly-aliased default role via the identity check `policy is not default_policy`. For a document that declares roles but no `default`, the loader aliases one named role as the fallback (`aliased_default`) — so that role's policy *is* `default_policy` and gets dropped from `others`.

Result: every tool the aliased role grants fires a `permissive-default` warning whose message ("the `default` role grants X … and no named role does") is **false** — the aliased role is a named role that grants X. It's also redundant with the `implicit-default` lint that already covers this shape.

Suggest short-circuiting the per-tool `permissive-default` scan when `aliased_default is not None` (the `implicit-default` lint already speaks to that case), or comparing against named roles by identity-excluding only the alias rather than dropping it from the grant comparison.

hexgate/security/policy_set.py
```diff
@@ -176,6 +199,9 @@ def load_policy_map(
     resolved = {name: pol for name, pol in fully_resolved.items() if not pol.is_mixin}
     if not resolved:
         raise PolicySetError("policy_map contains only mixins; need a concrete role")
+    # An explicit ``default=`` is the caller's deliberate choice; only the
+    # fallback we *infer* here is worth warning about downstream.
+    inferred = default is None and DEFAULT_ROLE_NAME not in resolved
     default_name = default or (
         DEFAULT_ROLE_NAME if DEFAULT_ROLE_NAME in resolved else next(iter(resolved))
```
quanghexa94: **Default-alias fallback is chosen inconsistently across the two load paths.**

Here (`load_policy_map`, inline / dict load) the inferred fallback is `next(iter(resolved))` — insertion order. But `_load_from_directory` uses `sorted(concrete)[0]` — alphabetical. The same role set resolves undefined role names to a *different* fallback policy depending on how the policy was loaded.

For roles `{support, billing}` with no explicit default: loaded from a directory, `billing` (alphabetically first) becomes the fallback; loaded from equivalent inline YAML where `support` was written first, `support` becomes the fallback. An undefined role name (which resolves to `default`) then gets different effective grants — and the `implicit-default` lint names a different role — for identical policy content.

Suggest picking the alias the same way in both paths (e.g. `sorted(...)[0]` here too) so the inferred default is deterministic regardless of load path.

guillaume-hexamind:
Thanks @quanghexa94 for the review, all 3 issues adressed, ready for another look.

1. **Uncapped `violations`** — capped in `audit.py` (`MAX_VIOLATIONS=64`,
   `MAX_VIOLATION_CHARS=1024`) alongside the existing byte caps, rather than in
   `_merge_denials`: a single role with >64 violations hit the same wall, so the
   merge wasn't the only path. Drop-and-disclose (`"(+N more)"`);
   `Decision.violations` stays intact for the model-facing payload. The 9×8
   payload that `DecisionEvent` rejected now validates.

2. **False `permissive-default`** — fixed the wording, kept the scan. Messages
   now name the aliased role ("`billing` is the inferred fallback and grants
   `refund`…") when the default is inferred. Didn't short-circuit after
   `implicit-default`: `test_implicit_default_still_reports_the_grants_it_exposes`
   deliberately requires the per-tool lints, and `implicit-default` alone doesn't
   say *which* grants are exposed. The identity-drop stays — an explicit
   `default=` to `load_policy_map` aliases too, so it's needed even when
   `aliased_default is None`.

3. **Inconsistent fallback pick** — `load_policy_map` now uses
   `sorted(resolved)[0]`, matching `_load_from_directory`. ⚠️ Behaviour change:
   an inline policy with no explicit `default` falls back to its
   alphabetically-first role, not the first written.

quanghexa94 (approved):

PR #118, opened 2026-08-12
feat(sdk): pre/post tool-hook pipeline with observe, halt, and arg rewrite

quanghexa94:
Add tool guards: functions that run before and after a tool call, alongside `PolicyEnforcer.decide`. `decide` answers allow / deny / approval and nothing else; guards are the open surface for the rest (strip a secret out of args, rate-limit per user, watch a result before the model sees it) without growing `decide` into a kitchen sink. This is PR1 of the series. It is pure SDK, no platform change, no migration. It lands the runtime and the LangChain seam only; the other three adapters (PR2) and the official plugins (PR3) follow.

## Design

Author a guard with the `@before_tool` / `@after_tool` decorators and register them as one flat `hooks=[...]` list on `enforce_policy`. The list is split into a pre list and a post list internally, preserving order within each. The whole thing reduces to one shared sequence, `run_guarded` (`hexgate/hooks/runner.py`): before-guards, then `decide`, then invoke, then after-guards. The load-bearing decision is ordering. Before-guards run **before** `decide`, so `decide` always authorizes the exact args that execute. That keeps arg rewrite from being a policy bypass and gives the invariant that guards can only narrow, never force an allow.

```python
from hexgate import before_tool, after_tool
from hexgate.hooks import Halt, Proceed

@before_tool                                   # every tool
def scan_secrets(call):
    if has_secret(call.args):
        return Halt(reason="Refused: remove the credential.")

@before_tool(tool_names=["refund_order"])      # scoped
def check_refund(call): ...

@after_tool(observe=True)                       # fail-open watcher
def log_result(call, result): ...

agent = base_agent.enforce_policy(policy, hooks=[scan_secrets, check_refund, log_result])
```

Guards attach at every entry point: `create_agent(..., hooks=[...])` (the constructor, which wraps the tools whether or not a policy binds), `agent.enforce_policy(policy, hooks=[...])`, and `wrap_langchain_agent(..., hooks=[...])` (the BYO-graph seam).

v1 capability matrix:

| | observe | halt | rewrite |
| --- | --- | --- | --- |
| before | yes | yes | args (always JSON, clean recursive walk) |
| after  | yes | yes | result deferred (objects need a projection rule) |

- A guard returns `Proceed()` / `Proceed(args=...)` / `Halt(reason, ...)` or `None`. A `Halt` renders through `Decision.as_error_payload`, so a block reaches the model as `{"ok": False, ...}` with a model-safe `reason` and no arg echo; `detail` stays on the observer/audit channel.
- A rewrite records a `Modification` reported to an optional `hook_observer` (local, fire-and-forget, like `decision_observer`), which doubles as the coercion-detection signal.
- Fail-closed by default (a raising guard denies); `observe=True` opts into fail-open for pure side effects.
- After-guards also run when a tool raises (`ToolOutcome(ok=False, error=...)`), so a watcher sees failures; if none halts, the exception propagates unchanged.
- Public types: `Halt`, `Proceed`, `ToolCall`, `ToolOutcome`, `Modification`, and the `before_tool` / `after_tool` decorators. No `Hook` / `ToolPipeline` type to construct by hand; position is the decorator, reach is `tool_names`.

The enforcer stays pure (guards ride `enforce_policy` -> `GuardedTool.wrap`, not the enforcer), and `PolicyBinding` is untouched (refresh swaps `enforcer.policy` in place, so guards on the tool survive hot reload for free). Decisions and deferred boundaries live as ADRs in `docs/adr/` (R-HOOK-001..003); the intuition is in `tool-hooks-explained.md`, the roadmap in `ROADMAP.md`.

## Important files

| File | What to look at |
| --- | --- |
| `hexgate/hooks/runner.py` | the core. The before/decide/invoke/after order, fail-closed vs observe-open, halt rendering, provenance, and the try/except that lets after-guards see a raised tool. Review this first. |
| `hexgate/hooks/types.py` | the surface: the `@before_tool` / `@after_tool` decorators, `build_pipeline` (the split), `Halt` / `Proceed` / `ToolCall` / `ToolOutcome` / `Modification`, and the `_UNSET` sentinel reserving result rewrite. |
| `hexgate/adapters/langchain/tools.py` | how `GuardedTool` and `install_enforcer_on_tool` route through `run_guarded` (behavior-preserving refactor). |
| `hexgate/agents/factory.py` | `enforce_policy(..., hooks=..., hook_observer=...)`, the split via `build_pipeline`, and the guards-only `policy=None` path. |
| `tests/hooks/test_runner.py` | the cases that pin the ordering invariant, halt safety, and the error tiers. |
| `docs/adr/R-HOOK-001..003` | the durable decisions (ordering before decide, fail-closed tiers, deferred result rewrite) as ADRs. |

## Test plan

- Full unit suite green: **1847 passed**, 6 skipped (opt-in integration), 20 deselected (framework_compat). Ran `pytest` from the worktree against the SDK.
- The **67 existing LangChain adapter tests pass unchanged**, the check that the `GuardedTool` refactor is behavior-preserving.
- `tests/hooks/` (45 tests) plus `tests/adapters/langchain/test_guarded_tool_hooks.py` and `tests/agents/test_enforce_policy_hooks.py` cover: the decorators and the pre/post split; `decide` sees the rewritten args (the security invariant, `test_pre_rewrite_is_what_decide_and_the_tool_see`); the halt message carries the reason but never echoes the input or the operator `detail`; decide-deny, post-halt, and after-guards seeing a tool that raised; fail-closed on a raising guard; observe fail-open (incl. an async-on-sync guard); guards-only `enforce_policy(None, hooks=...)`; `tool_names` scoping; ordering; NEEDS_APPROVAL; and the sync path.
- Each ADR carries a runnable `Verify` command; all three pass.
- Ruff lint + format clean on all new and changed files.

## Try it

```python
from hexgate import before_tool, after_tool
from hexgate.hooks import Halt, Proceed

@before_tool
def block_secrets(call):
    if "secret" in call.args:
        return Halt(reason="Refused: a credential was found. Remove it and resend.")

@before_tool(tool_names=["refund_order"])
def drop_token(call):
    return Proceed(args={k: v for k, v in call.args.items() if k != "token"})

agent = base_agent.enforce_policy(policy, hooks=[block_secrets, drop_token])
```

Or run the tests:

```
pytest tests/hooks tests/adapters/langchain
```

## Notes

- Deferred to later PRs (see `ROADMAP.md`): result rewrite (after `Proceed(result=...)` plus a projection rule for objects); a hard `stop_run` run-abort; the OpenAI / Google / pydantic_ai adapters onto `run_guarded` (PR2); the official `hexgate.plugins` secret detector (PR3); egress post-guards; tool-local co-located guards.
- Provenance ships as the observer channel only; durable persistence of `Modification`s into the platform audit / ClickHouse is a later increment, and the enforcer's own `AuditEvent` is unchanged.
- No platform change, no migration, no breaking change to shipped code: `hooks` defaults to `None`, so every existing `enforce_policy` call is unaffected. Revert by dropping the branch.


guillaume-hexamind:
Claude auto review:

1 — Pre-guards can rewrite authorized args with no provenance — 55/100

hexgate/hooks/types.py:63

ToolCall is frozen=True, which blocks attribute rebinding but not mutation of the dict it holds. _new_call stores args=dict(args) and that same object is handed to enforcer.decide and to invoke. Reproduced: an @before_tool(observe=True) guard doing call.args["path"] = "/etc/passwd" on a call for /safe/x gave decide saw {'path': '/etc/passwd'}, tool saw {'path': '/etc/passwd'}, hook events: [].

Two documented invariants are convention-only: R-HOOK-001's "coercion is visible via Modification" (no Modification is synthesized) and R-HOOK-002's "observe cannot rewrite" (_normalize discards an observe guard's return value, which is the only channel it polices). Mitigating: decide re-authorizes, and _new_call/_apply_pre both copy, so the caller's own dict is never touched.

Worth noting for internal consistency: enforcer.py:32 already deep-copies args precisely because "a shallow copy would let the caller mutate a nested value first and make the captured record lie about what was decided." The hooks layer doesn't apply its own reasoning.

Fix — make Proceed(args=...) the only rewrite channel. Type ToolCall.args as Mapping[str, Any] and wrap in MappingProxyType at the two construction sites (_new_call, and _apply_pre's replace(...)). I verified compatibility: a proxy unpacks under **final for invoke, and _snapshot does dict(values), so decide needs no change. Residual to document: a proxy is shallow, so nested dicts stay mutable — full protection needs a deep freeze, which probably isn't worth the cost. Cheaper alternative if you want to keep mutation ergonomic: snapshot before each guard and synthesize a Modification on diff — but that doesn't restore the observe tier.

2 — After-guards can rewrite the result R-HOOK-003 says they can't — 58/100

hexgate/hooks/types.py:80

Same frozen-shell problem, but ToolOutcome(ok=True, value=raw) does not copy raw. Reproduced: an @after_tool(observe=True) guard setting outcome.value["ssn"] = "REDACTED" and injecting a key made the caller receive {'ssn': 'REDACTED', 'name': 'bob', 'injected': True} with hook events: []. _reject_post_proceed only catches the honest Proceed(result=...) path.

I score this slightly above issue 1 for two reasons the original review missed. First, no compensating control exists at all — nothing re-checks the result, unlike decide re-checking args. Second, the blast radius is wider: raw is the tool's actual return object, so mutation escapes into anything the tool retains (a cached record, an ORM instance), not just a hexgate-internal copy. The practical risk is that operators will discover in-place redaction works and build on it, making the deferred result-rewrite phase a breaking change.

Fix — enforce the contract where it's enforceable and be honest where it isn't. When pipeline.post is non-empty and raw is JSON-ish (dict/list), pass a deepcopy to the post-hooks and return the original raw; for opaque objects pass through and document that v1 can't enforce it there. Gate on pipeline.post so the no-hooks path stays zero-cost. If the deepcopy cost is unacceptable, downgrade R-HOOK-003's "MUST NOT" to "is not enforced in v1" — the current wording promises something the code doesn't do.

3 — Guards-only path drops approval_handler, killing Halt(NEEDS_APPROVAL) — [score missing in source]

hexgate/agents/factory.py:545, same drop at :711

GuardedTool.wrap(t, pipeline=pipeline) omits approval_handler, and [source garbled] short-circuits on approval_handler is None. Reproduced side by side: the guards-only wrap returned {'type': 'approval_required', 'message': 'needs sign-off', ...} — unapprovable — where wrap(base, pipeline=p, approval_handler=True) returned refunded 5. create_agent(appro…) [source garbled] with no policy binding hits the same drop via enforce_policy(None, hooks=hooks, hook_observer=hook_observer).

Downscored from 72 because it fails closed: the call is blocked, not allowed. It's an advertised feature dead on one of two paths, not a hole. Note that the model is told approval is required and then can never obtain it.

Fix — pass approval_handler=approval_handler at both sites. [Verb garbled in source; likely "Drop"] the separate guards-only branch and run the single wrap loop with enforcer=None, so the paths can't diverge again — that also fixes the non-BaseTool skip (original issue 12) in one place instead of two.

4 — Guard halts bypass the audit trail, and post-halts record ALLOW — 72/100

hexgate/hooks/runner.py:157

Now the highest-scoring of the five. AuditSender.emit and decision_observer both fire inside enforcer.decide (enforcer.py:113-128). _halt_to_decision calls Decision.from_verdict directly, so neither runs. Two distinct failures, reproduced with an enforcer carrying a sender and an observer:

Pre-guard halt — decide never runs. Model got policy_denied; audit events: 0, observer calls: 0. The block is absent from the trail.
Post-guard halt — decide already ran and emitted ALLOW. Model got policy_denied / 'pii in result' while the trail recorded [DecisionOutcome.ALLOW] and the observer saw [ALLOW].

That second case is why I raised the score. The original review missed it. An actively wrong trail entry is worse than a missing one, and this happens on the correct, documented usage path — no operator misbehavior required, unlike issues 1 and 2. For a product whose value proposition is the audit trail, that's the most serious thing [sentence truncated in source].

Fix — split emission out of decide into a PolicyEnforcer.record (or similar), have decide build-then-record, and call record from run_guarded_* after _halt_to_decision. For the post-halt case, emit the halt as an additional event rather than suppressing the ALLOW — the allow genuinely happened, the tool executed, and the side effect [garbled; likely "landed — only the result"] was blocked. Also add provenance to the halt Decision: it currently renders type: "policy_denied", indistinguishable from a real policy denial to both the model and any trail consumer.

5 — @before_tool("tool_name") builds silently, then blocks every tool — 58/100

hexgate/hooks/types.py:166

_decorator's wrap never checks callable(f). Since tool_names documents a bare string as valid, putting it positionally is a natural mistake. Reproduced:

before_tool("refund_order") → Hook(fn='refund_order', position=…, observe=False), no complaint. tool_names=None means it reaches every tool, so at call time each one raises TypeError: 'str' object is not callable, _fail_closed swallows it, and an unrelated tool returned {'type': 'policy_denied', 'message': 'Blocked by a policy guard.'}. It takes down every tool with only a log line — and the model-facing message points the operator at their policy, not at the decorator.

Fix — in wrap, after the existing isinstance(f, Hook) unwrap, raise TypeError when not callable(inner), special-casing str with the actual remedy in the message (did you mean @before_tool(tool_names="refund_order")?). That converts a total runtime outage into an import-time error that names its own fix.

guillaume-hexamind (commented):
I let you check these 5 potential issues found by code review high + auto fact check.

Otherwise I approve of these first logic of pre and post tool hook, it seems like a pretty good abstraction for now.

We are getting to a point where the audit events are starting to be numerous and of different forms (decisions, tokens, pre / post hooks; we'll need too result and agent input / output too, agent bans...); a more global refactor of audit events will probably be needed at some point.

quanghexa94:
Thanks for the thorough review — all five held up, fixed in `0d2a416` (plus a follow-up merge, see the bottom).

**1 — before-guards mutate args in place, no provenance.** Fixed: `ToolCall.args` is now a read-only `MappingProxyType`, so `Proceed(args=...)` (which records a `Modification`) is the only channel that changes the args a guard passes downstream — an in-place write raises. Shallow, so a nested dict is the documented residual. R-HOOK-001 updated to say this is enforced, not convention.

**2 — after-guards rewrite the result R-HOOK-003 says they can't.** Fixed: when post-hooks exist, the value handed to them is a `deepcopy` (gated on `pipeline.post`, so the no-hooks path stays zero-cost) and the runner returns the original object, so an in-place mutation can't escape into the tool's real return. Opaque (non-JSON) objects pass through unenforced in v1, same boundary as the deferred projection rule. R-HOOK-003 updated.

**3 — guards-only path drops `approval_handler`.** Fixed by removing the divergence you pointed at: the guards-only `enforce_policy(None, hooks=...)` branch is folded into the single wrap loop with `enforcer=None`, so it can't silently drop `approval_handler` (or the non-BaseTool skip) again. `create_agent`'s branch threads it too.

**4 — guard halts bypass the audit trail; post-halts record ALLOW.** This was the good one. Emission is factored out of `decide` into `PolicyEnforcer.record`, which the runner now calls for a guard halt: a before-halt is recorded (decide never ran), and an after-halt is recorded *in addition to* the tool's genuine ALLOW — the tool did run, only the result was withheld, so the trail reads "ran, then withheld" rather than lying either way. A DENY halt carries a distinct `guard_denied` marker so it isn't indistinguishable from a real policy denial, and the halt Decision carries args for the trail (still never rendered to the model). New ADR R-HOOK-004.

**5 — `@before_tool("tool_name")` builds silently, then blocks every tool.** Fixed: the decorator raises at decoration time on a non-callable, special-casing a string with `did you mean @before_tool(tool_names=[...])?`, so a positional-name mistake is an import-time error instead of a run-time outage.

Each fix has a test; the ADR `Verify` commands cover the load-bearing ones.

Separately, the CI red you may have seen was not one of these — it was staleness: multi-role (#115/#117) landed on `main` after this branched and renamed the `Decision` API (`from_verdict(role=)` → `user_roles`/`deciding_role`, dropped `primary_role`), which the merged tree tripped over. I merged `main` in and adapted the guard-halt `Decision`; SDK CI is green again.

On the audit-events note — fully agree the trail is drifting toward many shapes (decisions, tokens, hooks, soon results + agent I/O + bans) and will want a consolidation pass. This PR only did the targeted piece (guard halts now emit through the same `record` path), which nudges toward that without attempting it. Happy to open a tracking issue for the broader refactor.

guillaume-hexamind (approved):
thanks, looks good to me

PR #119, opened 2026-08-12
feat(sdk): route OpenAI/Google/Pydantic AI adapters through run_guarded (guards on all frameworks)

quanghexa94:
Point the OpenAI Agents, Google ADK, and Pydantic AI adapters at the shared `run_guarded_async`, so before/after tool guards (`@before_tool` / `@after_tool`) work on every framework, not just LangChain. PR2 of the tool-guard series.

**Stacked on PR1 (#118).** Base is `lhq/feat/tool-hooks`, so this diff shows only the adapter changes. Retarget to `main` once #118 merges. Pure SDK, no platform change, no migration.

## Design

Each adapter already wrapped its tool's single async callable and inlined `decision = enforcer.decide(...)` → branch → render or invoke. This replaces that inline block with a call to the shared runner, passing two closures:

- **`invoke`** — runs the real tool on the (possibly rewritten) args,
- **`render_error`** — shapes a blocked `Decision` in that framework's idiom.

Then each public entry point gains `hooks=` / `hook_observer=` and builds the pipeline (`build_pipeline`) where it builds the enforcer. Exactly the pattern PR1 used for LangChain; mostly deleting duplicated decide-then-invoke.

| Adapter | Renders a block via | Pipeline built in | The wrinkle |
| --- | --- | --- | --- |
| Google ADK | returns `as_error_message()` string | `wrap_google_agent` | none — `args` is already a dict |
| OpenAI Agents | returns `as_error_message()` string | `HexgateRunner.__init__` → its 3 wrap sites | `input` is a JSON string; the `invoke` closure re-serializes rewritten args and keeps the original raw `input` when no guard rewrote it |
| Pydantic AI | **raises** `ModelRetry(...)` | `wrap_pydantic_agent` | `render_error` *raises* rather than returns; `run_guarded` returns the closure's value, so a raiser propagates, which is the pydantic_ai idiom |

All three wrap a single async callable, so only `run_guarded_async` is used (no sync runner).

### One shared change: `Decision.as_error_message()` now includes `reason`

LangChain renders a block as a dict (`as_error_payload`, which carries `reason`), but these three render via the **string** `as_error_message()`, which was generic and dropped `reason`. A guard `Halt`'s actionable message ("remove the credential") lives in `reason`, so it would have been lost on these adapters. `as_error_message()` now appends `reason` when present. Existing decision tests use substring asserts, so they pass unchanged.

## Important files

| File | What to look at |
| --- | --- |
| `hexgate/security/decision.py` | `as_error_message()` now surfaces `reason` (the one shared change). |
| `hexgate/adapters/openai/tools.py` | the `invoke` closure's JSON re-serialize / original-input preservation. |
| `hexgate/adapters/pydantic_ai/tools.py` | `_raise_model_retry` as `render_error` (raise, not return). |
| `hexgate/adapters/{openai,google,pydantic_ai}/{wrapper,runner}.py` | `hooks=` / `hook_observer=` threading + where the pipeline is built. |
| `tests/adapters/*/test_tools_hooks.py` | the parity tests per framework. |

## Test plan

- Full unit suite green: **1856 passed**, 6 skipped (opt-in integration), 20 deselected (framework_compat). Ran `pytest` from the worktree against the SDK.
- The existing OpenAI / Google / pydantic_ai adapter suites pass unchanged, which is the check that routing the policy path through `run_guarded` is behavior-preserving.
- New `tests/adapters/{openai,google,pydantic_ai}/test_tools_hooks.py` (3 each): a before-guard halt blocks and its reason reaches the model; a before-guard rewrite reaches the tool; an after-guard halt withholds a result the tool already produced (raises `ModelRetry` for pydantic_ai).
- Ruff lint + format clean.

## Try it

```python
from hexgate import before_tool
from hexgate.hooks import Halt
from hexgate.adapters.openai.runner import HexgateRunner  # or google's; or wrap_pydantic_agent

@before_tool
def block_secrets(call):
    if "secret" in call.args:
        return Halt(reason="Refused: remove the credential and resend.")

runner = HexgateRunner(hooks=[block_secrets])   # guards now run on OpenAI too
```

## Notes

- No new public API beyond `hooks=` / `hook_observer=` on each adapter's entry point, mirroring PR1.
- Deferred (see `ROADMAP.md`): result rewrite; `hexgate.plugins` (PR3); egress post-guards; tool-local co-located guards.
- Revert by dropping the branch.


guillaume-hexamind:
2 Claude auto review feedbacks:

- hexgate/adapters/openai/tools.py:68 — the raw-input passthrough never fires for dict payloads — 55/100

The invoke closure only forwards the model's original input string when parsed is None and not final, i.e. for non-object payloads. Every JSON-object payload is unconditionally json.loads → json.dumps round-tripped, including when pipeline=None and no hooks are configured at all — which is every current user, since guards are new. Models emit compact JSON, so the string changes on essentially every call:

'{"a":1}'                      -> '{"a": 1}'
'{"a": 1e400}'                 -> '{"a": Infinity}'     # invalid JSON downstream
'{"n":1.0000000000000000001}'  -> '{"n": 1.0}'
'{"s":"café"}'                 -> '{"s": "caf\u00e9"}'
'{"a":1, "a":2}'               -> '{"a": 2}'            # duplicate key collapsed

Typed @function_tool tools reparse input and are unaffected, but a hand-built FunctionTool that forwards, logs, hashes, or signs its raw input now sees a different string, Infinity breaks any strict parser downstream, and every call pays an extra serialize.

Fix. Note that if final == parsed: return original_invoke(ctx, input) does not work: _new_call builds MappingProxyType(dict(args)), a shallow copy that aliases the nested dicts, so a guard mutating call.args["meta"]["secret"] mutates parsed too and the comparison is True even though a rewrite happened — it would silently drop the nested in-place rewrite the PR deliberately supports. Decide at wrap time whether guards exist at all instead:

```python
# in wrap_tool, next to `original_invoke`
has_guards = pipeline is not None and not pipeline.is_empty

def invoke(final: dict[str, Any]) -> Any:
    # No guards: nothing can have rewritten the args, so forward the
    # model's payload byte-for-byte instead of round-tripping it.
    if not has_guards:
        return original_invoke(ctx, input)
    ...
```

Verified: with this applied, tests/adapters/openai tests/hooks is 124 passed / 1 skipped (including test_before_guard_nested_in_place_rewrite_reaches_the_tool), and the no-guards path forwards the raw string identically.




-  hexgate/adapters/openai/runner.py:96 — hooks= on __init__ collides with hooks= on run/run_sync/run_streamed — 50/100

HexgateRunner.__init__(hooks=...) now means Sequence[Hook] (Hexgate guards), while run/run_sync/run_streamed (lines 182/216/277) already have hooks= meaning the agents SDK's RunHooks, forwarded through _merge_hooks. A guard list passed to the wrong one is accepted silently:

runner.run(agent, input, hexgate_context=ctx, hooks=[my_guard])
→ _merge_hooks builds _CompositeRunHooks([[my_guard], usage_hooks])
→ AttributeError: 'list' object has no attribute 'on_agent_start'

The crash lands at the first lifecycle callback, far from the mistake. A type checker catches it, but nothing does at runtime.

Fix. Rename the constructor param on the OpenAI runner to tool_hooks= (or guards=). Keep hooks= as-is on every wrap_*_agent — there's no collision there — since the OpenAI runner is the one class that also forwards the SDK's hooks. Nothing in this stack is released yet, so the rename is free now and won't be later. Worth adding the guard at the crash site too:

python
def _merge_hooks(self, hooks: RunHooks | None) -> RunHooks:
    usage_hooks = HexgateUsageHooks(api_key=self.api_key)
    if hooks is None:
        return usage_hooks
    if not isinstance(hooks, RunHooksBase):
        raise TypeError(
            f"run(hooks=...) takes an agents RunHooks object, got {type(hooks).__name__}. "
            "Hexgate guards go on the constructor: HexgateRunner(tool_hooks=[...])."
        )
    return _CompositeRunHooks([hooks, usage_hooks])

guillaume-hexamind (commented):
2 Issues found by claude, see above.

Otherwise good PR, re-using the run_guarded_async this way seems like the right way to adapt the adapters logic, will approve after the 2 issues are checked!

quanghexa94:
Both addressed, thanks — good catches.

**1. `openai/tools.py` — the raw-input passthrough never fired for dict payloads (`beac6bf`)**

Right, and worse than just a rewrite edge: with `pipeline=None` (every current call, since guards are new) each object payload was still `json.loads → json.dumps` round-tripped, so `'{"a":1}'` → `'{"a": 1}'`, `1e400` → `Infinity`, unicode escaped, duplicate keys collapsed — a silent change on essentially every call.

Took your fix: decide at wrap time whether guards exist at all, instead of comparing `final == parsed` (which, as you noted, is unreliable — `MappingProxyType(dict(args))` aliases the nested dicts, so a nested in-place rewrite leaves `final == parsed` True and would be dropped).

```python
has_guards = pipeline is not None and not pipeline.is_empty

def invoke(final):
    if not has_guards:
        return original_invoke(ctx, input)   # forward the raw payload byte-for-byte
    ...
```

No guards → the model's string is forwarded verbatim. Guards present → serialize `final` (still honoring the nested in-place rewrite the PR supports). Added a test that a guard-free call forwards `'{"a":1, "big":1e400}'` unchanged.

**2. `openai/runner.py` — `hooks=` on `__init__` collides with `hooks=` on `run*` (`788107b`)**

Renamed the constructor's guard param to `guards=` (`HexgateRunner(guards=[...])`), leaving `run(hooks=...)` as the agents SDK's `RunHooks`. Did the same on the Google runner for cross-runner symmetry; the `wrap_*_agent` functions keep `hooks=` since they have no `run` to collide with. Also added the guard you suggested at the crash site:

```python
if not isinstance(hooks, RunHooksBase):
    raise TypeError(
        f"run(hooks=...) takes an agents RunHooks object, got {type(hooks).__name__}. "
        "Hexgate guards go on the constructor: HexgateRunner(guards=[...])."
    )
```

Full suite green (1995 passed), lint + format clean. Ready for another look.

guillaume-hexamind (approved):

quanghexa94:
Heads-up on a larger change since your review, so the growing diff doesn't surprise you.

Your second finding flagged that `hooks=` on the runner constructor shadowed the SDK's `run(hooks=)`. Rather than special-case one constructor, we unified the whole guard vocabulary on **`guards`**, so Hexgate's word and the framework's word never collide:

- `guards=` is now the argument everywhere: `create_agent`, `enforce_policy`, all four `wrap_*_agent`, and both `HexgateRunner` constructors (`guard_observer=` likewise).
- The module is `hexgate.guards` (was `hexgate.hooks`); the type is `Guard` (plus `GuardEvent` / `GuardObserver`); the ADRs are `R-GUARD-00x`.
- The **only** surviving `hooks=` is `HexgateRunner.run(hooks=...)`, which is the agents SDK's own `RunHooks` — we mirror `agents.Runner`, so that name is fixed. Rule of thumb: if it says `guards`, it's ours; if `hooks`, it's the framework's. A guard list passed to `run(hooks=)` is still rejected with the clear error from the previous round.

It's a mechanical rename layered on top of the two fixes you already reviewed — nothing behavioral changed. `hexgate.hooks` → `hexgate.guards`, `tests/hooks` → `tests/guards`, and the SDK's `RunHooks` / `HexgateUsageHooks` deliberately keep `hook` since they aren't the guard list.

Both your original findings remain fixed: the no-guards byte-for-byte payload passthrough, and the `run(hooks=)` type-guard. Full suite green (1995 passed), lint + format clean. Ready for another look whenever.

PR #120, opened 2026-08-14
feat(sdk, platform): persist multi-role decisions in the audit pipeline

guillaume-hexamind:
Persists multi-role decisions end to end. #117 shipped the enforcement semantic (every role evaluated, most-permissive wins) but left the audit wire untouched, so the trail recorded *who was calling* but not *which role granted the call*.

Roles are now a **set everywhere** — the legacy scalar `role` is gone from the SDK wire, the ClickHouse table, the API read models and the dashboard.

| Area | Change |
|---|---|
| clickhouse | `user_roles` + `deciding_role`; no `role` column |
| platform-api | ingest both; role filter/breakdown become membership-based; startup schema guard |
| sdk | send both; drop `Decision.role` and the `role` wire key |
| dashboard | role sets + deciding role in table/drawer/playground; policy-lint warnings in the editor |

## ⚠️ Deploy: recreate the ClickHouse volume

**Recreate the volume → API → dashboard → SDK release.**

No migration — the schema ships in `init/schema.sql`. **This discards existing audit rows**, which is fine in beta (the table already caps them at 90 days). Locally: `make clickhouse-reset`. If you skip it, `verify_schema` refuses to boot rather than silently dropping events.

## Reviewer notes

- **Old SDKs still work — that direction only.** `DecisionEvent.role` survives as an ingest-only shim: an SDK ≤ 0.2.11 sends `role`, `insert_decision` folds it into `user_roles`, and nothing downstream ever sees the scalar. Verified end-to-end by replaying a byte-exact 0.2.11 payload against a live ClickHouse. The reverse (new SDK → pre-multi-role platform) is **deliberately broken** — nobody is on it, and supporting it was the only thing keeping the scalar alive.
- **`by_role` counts membership**, so it can sum above `totals` (a call by two roles lands in both). Role moved out of the `GROUPING SETS` scan into its own query — an `arrayJoin` there would have inflated every other breakdown. Stated next to the panel.
- **Drive-by:** `tests/audit/test_integration.py` built `Decision(role=…)`, broken since #117 — latent because it's `-m integration`.

## Verification

`make check-all` green (SDK 1909 · API 474 · dashboard 204). Integration suites and the membership / no-role / `arrayJoin` reads re-checked on a freshly recreated volume (ClickHouse 24.10.4.191).


platform/api/hexgate_api/features/audit/service.py
```diff
@@ -324,18 +409,27 @@ def _add(store: dict[str, dict[str, int]], key: str, outcome: str, n: int) -> No
...
-            _add(by_role, role, outcome, n)
         elif not g_tool:  # (tool_name, outcome)
             _add(by_tool, tool, outcome, n)
         else:  # (user_id, outcome)
             _add(by_user, user, outcome, n)
 
+    # Second scan: role membership. Same where_sql AND params — the window is a
+    # bound instant, not ``now()``, so both scans see one slice. Sharing only
+    # the SQL text would not: ClickHouse re-evaluates now() per query.
+    role_result = client.query(
```
quanghexa94: **`by_role` breakdown double-counts co-occurring roles under a single-role filter**

When the Audit page is filtered to `role="billing"`, `_scope` adds `has(user_roles, 'billing')`, so only billing-carrying calls match. But this second scan then runs `arrayJoin(if(empty(user_roles),[''],user_roles))` over those same rows and unnests **every** role each caller carried. Result: with the filter chip showing "billing", the breakdown also renders bars for `support` (and any co-role) with non-zero counts. The panel note only warns that sums can exceed totals — it never explains why unrelated roles appear under a single-role filter.

platform/api/hexgate_api/features/audit/service.py
```diff
@@ -176,14 +210,47 @@ def insert_ban_enforcement(
...
+    for table, columns in (
+        (DECISION_TABLE, _DECISION_COLUMNS),
+        (BAN_ENFORCEMENT_TABLE, _BAN_ENFORCEMENT_COLUMNS),
+    ):
+        try:
+            present = table_columns(client, table)
+        except OperationalError:
+            _log.warning("ClickHouse unreachable during %s schema check", table)
+            return
+        except DatabaseError:
```
quanghexa94: **`verify_schema` treats any `DatabaseError` as a schema gap and refuses to boot**

`except DatabaseError: present = set()` interprets every `DESCRIBE`/`table_columns` failure as "table absent → all columns missing", raising `AuditSchemaOutOfDate` and aborting lifespan. A recoverable server-side error (permissions/quota/temporary metadata error that isn't an `OperationalError`) then crash-loops the whole control plane and tells the operator to reset the volume — discarding audit data — when the schema is actually fine. Consider narrowing the caught exception, or distinguishing "table missing" from "DESCRIBE failed".

platform/dashboard/src/components/audit/pieces.tsx
```diff
@@ -348,8 +372,10 @@ export function EventsTable({
                 <TableCell className="font-mono text-[12.5px]">
                   {e.agent_name}
                 </TableCell>
-                <TableCell className={e.role ? "" : "text-muted-foreground"}>
-                  {e.role || "—"}
+                <TableCell
+                  className={e.user_roles.length ? "" : "text-muted-foreground"}
```
quanghexa94, line 378: **Same unguarded `e.user_roles.length` in the events table**

Companion to the drawer finding in `Audit.tsx`: this cell dereferences `e.user_roles.length` with no optional-chaining fallback after `callerRoles` was removed. A row lacking `user_roles` (rolling deploy / cached older response) throws a `TypeError` and blanks the table.

platform/dashboard/src/routes/Audit.tsx
```diff
@@ -209,8 +209,30 @@ function DetailDrawer({
             <KV k="tool_name" mono>
               {e.tool_name}
             </KV>
-            <KV k="role">
-              {e.role || <span className="text-muted-foreground">∅ none</span>}
+            {/* Two rows, not three: the legacy scalar `role` would say almost
+                the same thing as "roles evaluated". */}
+            <KV k="roles evaluated">
+              {e.user_roles.length ? (
```
quanghexa94, line 215: **Drawer/table dereference `e.user_roles.length` with no fallback**

The "drop legacy role scalar" change deleted the guarded `callerRoles(e)` helper (`e.user_roles?.length ? ... : e.role ? ...`) and now dereferences `e.user_roles.length` directly (here and in `pieces.tsx`). During a rolling deploy where the dashboard updates before the API — or from any cached/older API response omitting `user_roles` — this throws a `TypeError` and blanks the audit table/drawer, the exact case the removed guard covered. Consider `e.user_roles?.length`.

platform/api/hexgate_api/features/agents/router.py
```diff
@@ -288,7 +292,43 @@ def _check_policy(policy: AgentPolicy, role_name: str | None) -> None:
...
+        policy_set = load_policy_set_from_dict(parsed)
+    except (PolicySetError, ValidationError):
+        # Inheritance/mixin failures the per-role checks don't cover. Surfacing
+        # them here would widen this endpoint's error contract, and dropping the
+        # lint can't make a broken document look clean — `ok` comes from
+        # `errors`, and the CLI and platform build both still reject these.
+        return []
+
+    return [
+        PolicyValidationError(role=lint.tool, message=f"{lint.code}: {lint.message}")
```
quanghexa94: **Lint→validation mapping puts a tool name in the `role` field**

`PolicyValidationError(role=lint.tool, message=...)` stuffs `lint.tool` into `role` and discards `lint.role`. For a permissive-default lint, `lint.tool` is the tool (e.g. `refund_order`) and `lint.role` is `None`. The dashboard's `DiagnosticList` renders `d.role` in the slot used for real role loci, so the warning shows `refund_order` where every other diagnostic shows a role name — mislabeling the locus and dropping the lint's own `role`.

guillaume-hexamind:
@quanghexa94 Thanks for the different catches.

- verify_schema treats any DatabaseError as a schema gap

Right, and worse than describedReproduced it: a user granted only on policy_decision gets code 497 ACCESS_DENIED (a plain DatabaseError, not OperationalError) when describing ban_enforcement, so the API refused to boot and told the operator to recreate the volume, destroying audit data over a missing GRANT.

Fixed: added _server_error_code() (the driver only exposes the code in the message text) and now only code 60 UNKNOWN_TABLE counts as a gap; everything else logs and degrades. Safe to degrade — a genuinely broken table makes inserts 503, which the SDK retries rather than dropping. Test added for the 497 path.

- by_role double-counts co-occurring roles under a single-role filter

Right. Reproduced: with rows [billing,support]/allow, [support]/deny, [billing]/allow, filtering to billing charted ('billing',2) and ('support',1). The framing I'd add is that role is the only dimension that behaves this way — filtering by agent collapses by_agent, because the GROUPING SETS scan groups on the same filtered column, but the arrayJoin re-expands what the WHERE already narrowed.

Fixed with HAVING role = {role:String} when a non-empty role filter is active. Unfiltered multi-counting and the "" "(none)" drill-down are unchanged (both covered by tests). Also gated the panel's membership caveat on there being no active role filter — with the fix it would now mislead.

- Lint→validation mapping puts a tool name in the role field

Right on the main point. Fixed: PolicyValidationError gained a tool field, the router passes role=lint.role, tool=lint.tool, and DiagnosticList renders tool refund_order → … so it can't read as a role name.

One correction: the mapping isn't discarding lint.role. check_default_role_exposure has three construction sites (analyzer.py:489, :512, :524) and none sets role — it's always None here. Passing it through anyway future-proofs the field.

- Unguarded e.user_roles.length in the drawer and the events table

Deferring both — the dereference is real but the trigger isn't reachable:

platform/api/Dockerfile:4-6 bakes platform/dashboard/dist into the API image and serves it same-origin. The bundle and the API are one artifact; there's no rolling deploy that can skew them, and a "cached older response" would have come from an API serving an older bundle.
schemas.py:599 is user_roles: list[str] = Field(default_factory=list), and FastAPI serializes response models with exclude_unset=False, so the key is always present. api.ts:262 types it non-optional to match.

quanghexa94 (approved):

PR #121, opened 2026-08-17
feat(platform-api): compile modular agents' bundles from resolved modules

quanghexa94:
Compile a modular project's agent bundles from its resolved modules, so multi-module policy actually enforces. Until now the store + resolve/check (#116) were inspection-only; this wires the resolved role-keyed policy into the signed WASM bundle each agent serves.

**Stacked on #116** (the policy-module store). The base of this PR is that branch, so the diff here is just the enforcement wiring; it retargets to `main` once #116 merges. Review/merge #116 first.

Additive: no schema change, no migration, and classic single-file agents are untouched.

## Design (see `docs/adr/R-POL-002`)

- **Modular = the project has a role binding.** Inferred, not a `policy_mode` column — so no migration (the platform has no Alembic, and `create_all` can't add a column to a live DB). Binding a role is the deliberate opt-in; uploading a module alone doesn't flip enforcement.
- **One chokepoint.** `bundle_for_agent(session, agent, sign)` compiles from the resolved role-keyed policy for a modular project, or from `agent.policy_yaml` for a classic one. The three existing compile sites (`update_agent`, `register_manifest`, `backfill_bundles`) all route through it, so the whole lifecycle is mode-aware in one place.
- **Resolve → the YAML `build` already accepts.** `resolved_policy_yaml` serializes the `PolicySet` to the inline-`roles:` shape, so `build_signed_bundle` (the same helper the CLI and #116 use) is unchanged.
- **Fan-out.** Editing a module or role binding recompiles every agent in the project (`recompile_project`), resolving once and reusing the single bundle for all of them.
- **Fail-safe.** An unresolvable edit, or a resolve that can't compile (opa absent), leaves live bundles untouched — a work-in-progress or typo never blanks running agents. `check` still surfaces the problem.
- **Serve path unchanged.** The bundle stays on the `Agent` row; the SDK fetches it as today.

## Important files

| File | What to look at |
| --- | --- |
| `platform/api/hexgate_api/features/agents/service.py` | `bundle_for_agent` (the chokepoint) + `recompile_project` (fan-out); the 3 compile sites now route through it |
| `platform/api/hexgate_api/features/policy_modules/service.py` | `is_modular` + `resolved_policy_yaml` (serialize the resolved set to inline-roles YAML) |
| `platform/api/hexgate_api/features/policy_modules/router.py` | `_recompile_modular` wired into the module/role write endpoints |
| `docs/adr/R-POL-002-compile-agents-from-modules.md` | the decision + the fail-safe / infer-modular rationale |

## Test plan

Full platform-api suite green (464 passed, 4 skipped) against the post-#118 `main` SDK. New service-level tests cover: `is_modular` flips on the first role binding; `bundle_for_agent` compiles from resolved YAML when modular and from `policy_yaml` when classic; `recompile_project` fans out to every agent; an unresolvable edit (role importing an unknown capability) returns 0 and leaves the last-good bundle byte-for-byte unchanged. One `opa`-gated end-to-end asserts a modular agent's compiled WASM equals compiling the resolved YAML directly (identical bytes → same enforced policy). Logic tests monkeypatch the compiler so they don't need `opa`; the byte-match test skips without it.

## Notes

- For a modular agent, `agent.policy_yaml` is no longer what's enforced (the resolved modules are); the dashboard (3b) edits modules, not per-agent `policy_yaml`, for modular projects.
- Deferred: agent-scoped self-restriction modules; deduping the identical per-agent bundles into one per-project bundle; the dashboard itself (3b).

### DIFF
--- a/docs/adr/R-POL-002-compile-agents-from-modules.md
+++ b/docs/adr/R-POL-002-compile-agents-from-modules.md
@@ -1,0 +1,45 @@
+# R-POL-002: Compile modular agents' bundles from resolved modules
+
+**Status:** Accepted · 2026-08-17
+**Applies to:** `platform/api/hexgate_api/features/policy_modules/**`, `platform/api/hexgate_api/features/agents/**`
+
+## Decision
+
+A project is **modular** iff it has at least one role binding. Follows from R-POL-001; this ADR covers how a modular project's agents are enforced.
+
+- A modular agent's signed bundle MUST be compiled from the project's *resolved* role-keyed policy (`resolve` → inline-`roles:` YAML → `build_signed_bundle`), not from `agent.policy_yaml`. A classic project (no bindings) MUST keep compiling from `policy_yaml`, unchanged.
+- Editing a module or a role binding MUST recompile every agent in the project (fan-out). The project resolves once; the single bundle is reused for all its agents (they are identical until agent-scoped modules exist).
+- A recompile that can't resolve, or resolves but can't compile (e.g. `opa` absent), MUST leave existing bundles untouched. A broken or work-in-progress edit MUST NOT blank agents that were working.
+- Modular MUST be inferred from the presence of a role binding. There MUST NOT be a `policy_mode` column in this phase.
+- The serve path MUST NOT change: the bundle stays on the `Agent` row and the SDK fetches it as today.
+
+## Why
+
+Enforcement has to come from the composed modules, or the whole module system is inert (it only fed inspection before this). The one real question is *when* a project switches from `policy_yaml` to modules, and how not to break live agents doing it.
+
+Inferring modular from a role binding, rather than adding a `policy_mode` column, is deliberate on two counts. First, it needs no migration: the platform has no Alembic, and `create_all` can add tables but not columns to a populated DB, so a column would force that decision now for no functional gain. Second, binding a role is the meaningful opt-in — you have decided which roles map to which capabilities. Uploading a capability or boundary module alone must *not* flip enforcement, or dropping in a single boundary would resolve every agent to a deny-all default and brick them.
+
+Fan-out is required because the policy is project-level: one module edit changes every agent's effective policy, so every agent must recompile. Reusing one resolved bundle for all of them keeps that to a single `opa` call per edit.
+
+Leaving bundles untouched on an unresolvable edit is the safety property. The alternative — clearing bundles when the project doesn't resolve — turns a typo in the dashboard into an outage across every agent in the project. Failing safe means the last good policy keeps enforcing while `check` surfaces the problem. This is why the write endpoints do not reject an unresolvable edit: the store can hold a work-in-progress state, but only a clean resolve replaces what is enforced.
+
+## Consequences
+
+- No schema change, no migration. Additive code only.
+- All modular agents in a project share one bundle (redundant bytes on each row). Deduping into a per-project bundle is a later optimization, not needed for correctness.
+- For a modular agent, `agent.policy_yaml` is no longer what's enforced (the resolved modules are). The dashboard must edit modules, not per-agent `policy_yaml`, for modular projects.
+- A saved-but-unresolvable project keeps enforcing its last good bundle; the divergence is visible through `check`, never silent.
+
+## Rejected alternatives
+
+- **A `policy_mode` column on `Project`.** Forces Alembic now (the column can't be added to a live DB by `create_all`) for no behavior a role binding doesn't already imply.
+- **A per-project compiled-bundle table.** Removes the redundant per-agent bytes but changes the serve path (agents would fetch a project bundle). Deferred; not worth the serve-path risk yet.
+- **Clearing bundles when a project won't resolve.** A one-character typo would blank every agent in the project. Fail-safe (leave last good) is mandatory for a product whose value is uninterrupted enforcement.
+
+## Verify
+
+```
+cd platform/api && pytest tests/features/policy_modules/ -q
+```
+
+Covers: `is_modular` flips on the first binding; `bundle_for_agent` routes classic vs modular; `recompile_project` fans out to every agent; an unresolvable edit leaves bundles untouched; and (with `opa`) a modular agent's bundle equals compiling the resolved YAML.
--- a/docs/adr/R-POL-002-compile-agents-from-modules.md
+++ b/docs/adr/R-POL-002-compile-agents-from-modules.md
@@ -8,7 +8,7 @@
 A project is **modular** iff it has at least one role binding. Follows from R-POL-001; this ADR covers how a modular project's agents are enforced.
 
 - A modular agent's signed bundle MUST be compiled from the project's *resolved* role-keyed policy (`resolve` → inline-`roles:` YAML → `build_signed_bundle`), not from `agent.policy_yaml`. A classic project (no bindings) MUST keep compiling from `policy_yaml`, unchanged.
-- Editing a module or a role binding MUST recompile every agent in the project (fan-out). The project resolves once; the single bundle is reused for all its agents (they are identical until agent-scoped modules exist).
+- Editing a module or a role binding MUST recompile every agent in the project (fan-out). Each agent is resolved for its own column of the `(role, agent)` matrix and compiled to its **own** bundle; agents that resolve to identical policy share a single `opa` compile (memoized). *(This superseded the original single-shared-bundle design once the `(role, agent)` matrix landed — see `agent-policy-dimension-design.md`.)*
 - A recompile that can't resolve, or resolves but can't compile (e.g. `opa` absent), MUST leave existing bundles untouched. A broken or work-in-progress edit MUST NOT blank agents that were working.
 - Modular MUST be inferred from the presence of a role binding. There MUST NOT be a `policy_mode` column in this phase.
 - The serve path MUST NOT change: the bundle stays on the `Agent` row and the SDK fetches it as today.
@@ -19,14 +19,16 @@ Enforcement has to come from the composed modules, or the whole module system is
 
 Inferring modular from a role binding, rather than adding a `policy_mode` column, is deliberate on two counts. First, it needs no migration: the platform has no Alembic, and `create_all` can add tables but not columns to a populated DB, so a column would force that decision now for no functional gain. Second, binding a role is the meaningful opt-in — you have decided which roles map to which capabilities. Uploading a capability or boundary module alone must *not* flip enforcement, or dropping in a single boundary would resolve every agent to a deny-all default and brick them.
 
-Fan-out is required because the policy is project-level: one module edit changes every agent's effective policy, so every agent must recompile. Reusing one resolved bundle for all of them keeps that to a single `opa` call per edit.
+Fan-out is required because the policy is project-level: one module edit changes every agent's effective policy, so every agent must recompile. Each agent resolves its own `(role, agent)` column into its own bundle; agents whose columns resolve identically are compiled once (memoized), so a project whose agents all share the generic `*` column still pays a single `opa` call per edit.
 
-Leaving bundles untouched on an unresolvable edit is the safety property. The alternative — clearing bundles when the project doesn't resolve — turns a typo in the dashboard into an outage across every agent in the project. Failing safe means the last good policy keeps enforcing while `check` surfaces the problem. This is why the write endpoints do not reject an unresolvable edit: the store can hold a work-in-progress state, but only a clean resolve replaces what is enforced.
+Leaving *compiled bundles* untouched when an edit can't **build** (e.g. `opa` absent) is the safety property: clearing bundles on a transient build failure would turn it into an enforcement outage across every agent. Failing safe means the last good bundle keeps enforcing while `check` surfaces the problem.
+
+A later refinement tightened this: a write whose result is **semantically unresolvable** (a role importing an unknown capability, a capability with a `deny`) is now **rejected at write time** (409), not silently stored. The roles PUT validates the proposed bindings *before* writing; the module PUT/DELETE reject and roll back an edit that breaks resolution. So the store no longer holds an uncomposable state that keeps the old bundle with no signal — only a *transient* build failure (opa down) leaves the last-good bundle in place.
 
 ## Consequences
 
 - No schema change, no migration. Additive code only.
-- All modular agents in a project share one bundle (redundant bytes on each row). Deduping into a per-project bundle is a later optimization, not needed for correctness.
+- Each modular agent compiles to its own bundle from its `(role, agent)` column; agents that resolve to the same policy dedupe the `opa` compile (memoized). *(Originally all agents shared one bundle; the `(role, agent)` matrix made bundles per-agent so each carries only its own authority.)*
 - For a modular agent, `agent.policy_yaml` is no longer what's enforced (the resolved modules are). The dashboard must edit modules, not per-agent `policy_yaml`, for modular projects.
 - A saved-but-unresolvable project keeps enforcing its last good bundle; the divergence is visible through `check`, never silent.
 
--- a/platform/api/hexgate_api/core/locks.py
+++ b/platform/api/hexgate_api/core/locks.py
@@ -1,0 +1,26 @@
+"""Per-project in-process locks serializing a policy write + its recompile.
+
+A policy write (module or role edit) resolves the project, compiles a bundle,
+then commits. Two overlapping writes on the same project can otherwise interleave
+those steps and commit bundles out of order: writer A resolves R1 and yields on
+the opa compile, B commits R2 and its bundles, then A commits the R1 bundle over
+them — DB and ``/policy/resolve`` say R2 while agents enforce R1, with no error.
+Holding this lock across the whole read→write→recompile makes them serial.
+
+Scope: one event loop / process only. Under multiple uvicorn workers it does not
+serialize across processes; the durable fix is a DB advisory lock (Postgres
+``pg_advisory_xact_lock``), deferred here because policy writes are admin-only
+and infrequent.
+"""
+
+from __future__ import annotations
+
+import asyncio
+from collections import defaultdict
+
+_locks: defaultdict[str, asyncio.Lock] = defaultdict(asyncio.Lock)
+
+
+def project_lock(project_id: str) -> asyncio.Lock:
+    """The lock for one project — the same object per id within this process."""
+    return _locks[project_id]
--- a/platform/api/hexgate_api/features/agents/router.py
+++ b/platform/api/hexgate_api/features/agents/router.py
@@ -179,19 +179,24 @@ async def api_update_agent(
     session: AsyncSession = Depends(get_session),
 ) -> AgentRead:
     from hexgate_api.core.keystore import keystore
+    from hexgate_api.core.locks import project_lock
 
     await ensure_default_project(session)
-    agent = await update_agent(
-        session,
-        project_id,
-        name,
-        agent_yaml=body.agent_yaml,
-        policy_yaml=body.policy_yaml,
-        system_md=body.system_md,
-        # Compile + sign the policy into a WASM bundle at save time, using
-        # the platform's root key (same key that signs biscuits).
-        sign=keystore.sign,
-    )
+    # Hold the project lock: this compiles + writes a bundle, so it must not
+    # interleave with recompile_project (a concurrent policy write) and commit
+    # out of order. See core.locks.
+    async with project_lock(project_id):
+        agent = await update_agent(
+            session,
+            project_id,
+            name,
+            agent_yaml=body.agent_yaml,
+            policy_yaml=body.policy_yaml,
+            system_md=body.system_md,
+            # Compile + sign the policy into a WASM bundle at save time, using
+            # the platform's root key (same key that signs biscuits).
+            sign=keystore.sign,
+        )
     if agent is None:
         raise HTTPException(status_code=404, detail="agent not found")
     return _agent_read(agent)
@@ -348,10 +353,14 @@ async def api_register_agent(
     edits are preserved.
     """
     from hexgate_api.core.keystore import keystore
+    from hexgate_api.core.locks import project_lock
 
-    version, created = await register_manifest(
-        session, project_id, body.manifest, sign=keystore.sign
-    )
+    # Hold the project lock: a first-time register compiles + writes a bundle,
+    # so it must not interleave with recompile_project and commit out of order.
+    async with project_lock(project_id):
+        version, created = await register_manifest(
+            session, project_id, body.manifest, sign=keystore.sign
+        )
     response.status_code = 201 if created else 200
     return RegisterAgentResponse(
         agent_id=version.agent_id,
--- a/platform/api/hexgate_api/features/agents/router.py
+++ b/platform/api/hexgate_api/features/agents/router.py
@@ -354,10 +354,19 @@ async def api_register_agent(
     """
     from hexgate_api.core.keystore import keystore
     from hexgate_api.core.locks import project_lock
+    from hexgate_api.features.agents.service import get_agent
 
-    # Hold the project lock: a first-time register compiles + writes a bundle,
-    # so it must not interleave with recompile_project and commit out of order.
-    async with project_lock(project_id):
+    # Only a FIRST registration compiles + writes a bundle; re-registering an
+    # existing agent just snapshots the manifest (no compile). Take the project
+    # lock only in that case — otherwise a fleet redeploy's no-op re-registers
+    # (`hexgate serve` auto-registers on startup) would all queue behind the lock
+    # and any in-flight recompile for nothing.
+    if await get_agent(session, project_id, body.manifest.name) is None:
+        async with project_lock(project_id):
+            version, created = await register_manifest(
+                session, project_id, body.manifest, sign=keystore.sign
+            )
+    else:
         version, created = await register_manifest(
             session, project_id, body.manifest, sign=keystore.sign
         )
--- a/platform/api/hexgate_api/features/agents/service.py
+++ b/platform/api/hexgate_api/features/agents/service.py
@@ -91,16 +91,34 @@ def _apply_bundle(agent: Agent, bundle: tuple[bytes, str, bytes] | None) -> None
         agent.compiled_wasm, agent.bundle_manifest, agent.bundle_signature = bundle
 
 
+# The generic/default agent key — the "*" column of the (role, agent) matrix.
+# Mirrors hexgate.security.DEFAULT_AGENT; kept local so this module doesn't import
+# the SDK at load (the SDK is imported lazily on the modular path only).
+_DEFAULT_AGENT = "*"
+
+
 async def _modular_bundle(
-    session: AsyncSession, project_id: str, sign: Callable[[bytes], bytes]
+    session: AsyncSession,
+    project_id: str,
+    sign: Callable[[bytes], bytes],
+    *,
+    agent: str = _DEFAULT_AGENT,
+    compiled: dict[str, tuple[bytes, str, bytes] | None] | None = None,
 ) -> tuple[bytes, str, bytes] | None:
-    """Resolve a modular project once and compile its single shared bundle.
+    """Resolve + compile the bundle for ONE agent's column of the matrix (Path A).
+
+    ``agent`` selects the executing agent's ``(role, agent)`` column (default the
+    generic ``"*"``). Each agent resolves to its own role-keyed bundle, so a
+    signed artifact carries only that agent's authority.
 
     Returns ``None`` (never raises) when the project doesn't resolve or the
     resolved policy can't compile (e.g. ``opa`` absent), so callers can leave
     live bundles untouched — the fail-safe in docs/adr/R-POL-002. The SDK
     resolve exceptions are imported here, on the modular path only, so a classic
     save never touches the SDK.
+
+    ``compiled`` is an optional sha256-keyed compile cache: a fan-out over agents
+    that resolve to identical policy then shells out to ``opa`` only once.
 
     The except tuple also covers ``yaml.YAMLError`` / pydantic ``ValidationError``:
     ``resolved_policy_yaml`` re-parses every stored module, and a row valid at
@@ -115,7 +133,9 @@ async def _modular_bundle(
     from hexgate_api.features.policy_modules import service as modules
 
     try:
-        policy_yaml = await modules.resolved_policy_yaml(session, project_id)
+        policy_yaml = await modules.resolved_policy_yaml(
+            session, project_id, agent=agent
+        )
     except (
         LinkError,
         PolicySetError,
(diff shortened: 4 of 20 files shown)


victorludvig (commented):
Solid PR overall, I approve the logic.
Can you check these Claude findings before we can merge:

1. [Score 75 — confirmed] recompile_project's classic branch keeps a stale bundle when policy_yaml fails to compile, contradicting the established "never serve a wrong bundle" invariant.
platform/api/hexgate_api/features/agents/service.py:133 does if bundle is None: continue, keeping the old compiled_wasm. But update_agent (same file, line 189 onward) handles the identical situation by nulling all three bundle fields so the SDK falls back to the pydantic engine — an invariant documented in the Agent.compiled_wasm docstring in models.py ("Null when… the policy fails to compile") and in commit 875e3c4 ("a newly-broken policy drops its stale bundle so we never serve a wrong one"). The worst case is the modular→classic transition: drop the last role binding, and any agent whose own policy_yaml no longer compiles silently keeps enforcing the old shared modular WASM bundle indefinitely, with no fallback warning. The new tests stub compile_bundle to always succeed, so this path is untested. The ADR's "never blank bundles" fail-safe is scoped to resolve failures, not classic compile failures. Suggested fix: mirror update_agent — null the bundle fields on bundle is None instead of continue.

2. [Score 65 — confirmed, perf/availability] Every module/role write now runs an uncached resolve() plus a synchronous opa subprocess inside the async request handler.
The chain _recompile_project_agents (policy_modules/router.py:48, called at lines 102, 124, 150) → recompile_project → _modular_bundle (agents/service.py:62) → resolved_policy_yaml (policy_modules/service.py:289) re-parses every module from scratch — the exact concern you flagged and the author deferred in PR #116 ("I'll add a content_hash-keyed parse cache when dashboard polling makes it a real hot path"); this PR is what makes it a hot path, without adding the cache. Verification also confirmed a second facet: compile_bundle invokes the opa subprocess synchronously (no asyncio.to_thread), blocking the event loop for the duration of the compile on every policy write. Score is tempered because writes are far less frequent than the polling scenario the deferral targeted, and real impact depends on module count and policy size — but on a live production API, the event-loop blocking facet is worth addressing even if the cache stays deferred.

3. [Score 50 — confirmed nitpick] api_set_policy_roles recompiles the whole project even when the role write changes nothing.
policy_modules/router.py:150 calls _recompile_project_agents unconditionally — deliberately, since a role write can flip modular↔classic. But an idempotent PUT (re-saving identical roles) still pays a full resolve + opa compile. A cheap guard is possible (compare get_roles() before/after; recompile only on change) without losing the transition handling. Role edits are infrequent and admin-only, so this is minor.

quanghexa94:
Thanks — all three addressed in `768b790`, full platform suite green (470).

1. **Classic recompile kept a stale bundle on compile failure.** Fixed. `recompile_project`'s classic branch now **nulls** the three bundle fields when an agent's `policy_yaml` no longer compiles, mirroring `update_agent` and the `Agent.compiled_wasm` invariant (drop the stale bundle, fall back to pydantic, never serve a wrong one). You were right that this was the modular->classic-unbind worst case and that the existing tests stubbed a successful compile, so I added a regression test that stubs `compile_bundle -> None` and asserts the bundle is nulled. The ADR fail-safe (don't blank) stays scoped to the modular resolve-failure path, which is a different case.

2. **opa subprocess blocking the event loop on every write.** Fixed. Every compile in an async path (`_modular_bundle`, `bundle_for_agent`, `recompile_project`, `update_agent`, `backfill_bundles`) now runs via `asyncio.to_thread(compile_bundle, ...)`, so a policy write no longer blocks other in-flight requests for the compile. The content-hash-keyed parse cache stays deferred, as you noted — the polling read path (`resolve`/`check`) is where it earns its keep, and writes are infrequent and admin-only.

3. **`set_roles` recompiling on a no-op PUT.** Fixed. The role endpoint now compares bindings before/after and only recompiles when they changed; a real change (including a modular<->classic flip) still triggers it, since `recompile_project` handles both directions. Added a test that an identical re-save skips the recompile.

Ready for another look.

platform/api/hexgate_api/features/agents/service.py
```diff
@@ -59,6 +60,103 @@ async def get_agent(session: AsyncSession, project_id: str, name: str) -> Agent
...
+
+    agents = await list_agents(session, project_id)
+    if not agents:
+        return 0
+
+    count = 0
+    if await modules.is_modular(session, project_id):
+        bundle = await _modular_bundle(session, project_id, sign)
+        if bundle is None:
+            return 0  # fail-safe: leave live bundles as-is
```
victorludvig: On the classic→modular flip, when _modular_bundle returns None the early return 0 keeps agents' old policy_yaml-compiled WASM while is_modular is now True. Since backfill_bundles skips agents that already have a bundle and re-saving identical roles skips recompile (roles != before), nothing ever replaces it.

Consider rejecting PUT /v1/projects/{id}/policy-roles when the first modular compile fails.

platform/api/hexgate_api/features/agents/service.py
```diff
@@ -233,7 +362,7 @@ async def register_manifest(
         # Policies editor has something to render and ``hexgate serve``
         # has a signed bundle to ship.
         agent.policy_yaml = _default_policy_for_manifest(manifest)
-        bundle = compile_bundle(agent.policy_yaml, sign)
+        bundle = await bundle_for_agent(session, agent, sign)
```
victorludvig: Let's say I am in a modular project. There is a bundle policy for all agents. I create a new agent. Then it should get the bundle policy. But for some reason (e.g. opa unavailable) the bundle policy is not retrieved, then the agent gets the default policy (line above), which is potentially more permissive than the bundle policy.

To prevent this we can set a deny-all default policy for modular projects:

```
if agent_created:
    if await modules.is_modular(session, project_id):
        agent.policy_yaml = DENY_ALL_POLICY_YAML   # fallback fails closed
    else:
        agent.policy_yaml = _default_policy_for_manifest(manifest)
    bundle = await bundle_for_agent(session, agent, sign)
    ...
```

platform/api/hexgate_api/features/policy_modules/router.py
```diff
@@ -91,6 +120,8 @@ async def api_delete_policy_module(
     )
     if not deleted:
         raise HTTPException(status_code=404, detail="module not found")
+    if await service.is_modular(session, project_id):
+        await _recompile_project_agents(session, project_id)
```
victorludvig: If we delete a capability that a role binding still imports, the delete returns 204 but enforcement doesn't change. Chain: _recompile_project_agents → recompile_project → _modular_bundle → resolved_policy_yaml → the SDK linker raises LinkError("role imports unknown capability") because the role still lists the deleted module; _modular_bundle catches it and returns None, and recompile_project hits return 0 (fail-safe: keep live bundles). So the modules can't be composed into a policy ("doesn't resolve"), no new bundle is built, and every agent keeps the old bundle — still granting the deleted capability. We should probably return 409 while a role still references the module, or at least surface the resolve failure in the response.

platform/api/hexgate_api/features/agents/service.py
```diff
@@ -59,6 +60,103 @@ async def get_agent(session: AsyncSession, project_id: str, name: str) -> Agent
...
+    resolve exceptions are imported here, on the modular path only, so a classic
+    save never touches the SDK.
+    """
+    from hexgate.security import LinkError, PolicySetError
+    from hexgate.security.constraints import ConstraintParseError
+    from hexgate_api.features.policy_modules import service as modules
+
+    try:
+        policy_yaml = await modules.resolved_policy_yaml(session, project_id)
+    except (LinkError, PolicySetError, ConstraintParseError) as exc:
```
victorludvig: _modular_bundle (agents/service.py:63-87) is documented as "never raises" and only catches LinkError, PolicySetError, ConstraintParseError. But resolved_policy_yaml re-parses every stored module with yaml.safe_load + AgentPolicy.model_validate, which can raise yaml.YAMLError / pydantic ValidationError — neither is in the except tuple. Module content is only validated at write time, so a module that was valid when stored can later fail to parse (SDK upgrade tightens the schema, or a row edited directly in the DB).

Where it hurts: backfill_bundles calls _modular_bundle with no try/except (service.py:282) and is awaited in the app lifespan (main.py:151) → the platform fails to boot while that module exists. register_manifest → bundle_for_agent → same exception → hexgate register returns 500 for that project. Only the write endpoints are shielded, by the blanket except Exception in _recompile_project_agents.

Fix: catch (yaml.YAMLError, pydantic.ValidationError) in _modular_bundle too (or Exception, given the "never raises" contract), and/or wrap backfill_bundles in the lifespan so a bad module can't block startup.

platform/api/hexgate_api/features/agents/service.py
```diff
@@ -59,6 +60,103 @@ async def get_agent(session: AsyncSession, project_id: str, name: str) -> Agent
...
+    """
+    from hexgate_api.features.policy_modules import service as modules
+
+    agents = await list_agents(session, project_id)
+    if not agents:
+        return 0
+
+    count = 0
+    if await modules.is_modular(session, project_id):
+        bundle = await _modular_bundle(session, project_id, sign)
```
victorludvig: recompile_project has no lock across the to_thread(compile_bundle) step, so two overlapping policy writes on the same project (two admins, a UI autosave, a CI push of several modules) can commit out of order: A resolves R1 and yields during the opa compile, B commits R2 and writes its bundles, then A commits the R1 bundle over it. DB and /policy/resolve then say R2 while agents enforce R1 until the next write, with no error anywhere. Not a blocker, but we should probably add a per-project lock or re-check the bindings before writing.

platform/api/hexgate_api/features/policy_modules/router.py
```diff
@@ -113,7 +144,14 @@ async def api_set_policy_roles(
...
     session: AsyncSession = Depends(get_session),
 ) -> RoleBindingsRead:
+    before = await service.get_roles(session, project_id)
     roles = await service.set_roles(session, project_id=project_id, roles=body.roles)
+    # Recompile only when the bindings actually changed — an idempotent re-save
+    # shouldn't pay a resolve + opa compile. A real change (including a
+    # modular<->classic flip) still triggers it, and recompile_project handles
+    # both directions (resolved bundle vs each agent's policy_yaml).
+    if roles != before:
+        await _recompile_project_agents(session, project_id)
```
victorludvig: Race condition: roles != before compares a snapshot read before the write, so under a concurrent PUT it can skip a needed recompile (bindings say A, agents enforce B's bundle, and re-saving the same roles won't fix it). It also treats a reordered capability list as a change and triggers a full recompile for nothing. Suggest comparing as sets, and not skipping the recompile based on a pre-write read.

victorludvig (changes requested):
The new revision surfaced other issues, please check them. 
Here are 3 nit comments, up to you to address them or not:

Cleanup (3)

1. policy_modules/service.py:282 — roles_json/resolved_policy_yaml duplicate the CLI's ProjectLinkResult→roles: YAML serializer (cli/policy/main.py:620-630) and already drift: CLI iterates sorted(by_role.items()), service uses insertion order, contradicting the "byte-for-byte parity" docstring. Lift a shared serializer into security/linker.py.
2. agents/service.py:136 — the three-field bundle assign/null is hand-written 8 times (136, 149, 238, 286, 298, 367, plus = None triples at 145-147, 242-244) and the is_modular dispatch 4 times; bundle_for_agent is only called from register_manifest; recompile_project/backfill_bundles share a near-identical 20-line loop. An _apply_bundle helper + one loop body would remove the mismatch risk when adding a column.
3. policy_modules/router.py:101 — module PUT recompiles the whole project even for byte-identical content (upsert_module already computes _content_hash but doesn't return a changed flag), and classic recompiles spawn one opa subprocess per agent serially with no memo by sha256(policy_yaml) — including during startup backfill, delaying readiness.

platform/api/hexgate_api/features/agents/service.py
quanghexa94: Fixed in 16ccd87. \`PUT /policy-roles\` now rejects and rolls back the binding when a classic→modular flip can’t build a bundle: \`recompile_project\` returns \`None\` (modular-but-unbuildable) vs \`0\` (nothing to do), so \`is_modular\` never flips True while agents keep the stale classic WASM. Test: \`test_flip_to_modular_rejected_when_bundle_cannot_build\` (and \`..._with_unresolvable_roles_is_rejected\`).

platform/api/hexgate_api/features/agents/service.py
quanghexa94: Fixed in 16ccd87. A brand-new agent in a modular project now gets a deny-all \`policy_yaml\` fallback (\`DENY_ALL_POLICY_YAML\`) instead of the permissive tool-derived starter, so if the shared bundle can’t be served the pydantic fallback fails closed. Test: \`test_register_into_modular_project_seeds_deny_all_fallback\`.

platform/api/hexgate_api/features/policy_modules/router.py
quanghexa94: Fixed in 16ccd87. Deleting a capability a role binding still imports now returns 409 (\`roles_importing\` guard) rather than 204 with enforcement silently unchanged; the caller must remove it from the binding first. Test: \`test_delete_capability_still_imported_is_409\`.

platform/api/hexgate_api/features/agents/service.py
quanghexa94: Fixed in 16ccd87. \`_modular_bundle\`’s except now also catches \`yaml.YAMLError\` and pydantic \`ValidationError\`, so a stored module that later fails to parse folds to no-bundle (the R-POL-002 fail-safe) instead of escaping as a 500.

platform/api/hexgate_api/features/agents/service.py
quanghexa94: Fixed in 259f4f8. Added a per-project lock (\`core.locks\`) held across the whole read→write→recompile in all three write handlers, so overlapping writes can’t resolve/compile/commit out of order. In-process only — I documented the Postgres advisory-lock path as the durable multi-worker fix.

platform/api/hexgate_api/features/policy_modules/router.py
quanghexa94: Fixed in 259f4f8. Role bindings are now compared order-insensitively (a reorder is not a change → no needless recompile), and the pre-write snapshot is read under the project lock so it can’t go stale and wrongly skip a recompile. Test: \`test_reordered_role_put_skips_recompile\`.

quanghexa94:
All found issues are addressed — pushed as 5 commits on top of the reviewed revision.

**Blockers (fail closed):** `16ccd87`
- classic→modular flip that can't build a bundle now rejects + rolls back (no stale classic WASM left enforced)
- new agent in a modular project falls back to deny-all, not the permissive starter
- deleting a capability a role still imports → 409, not a silent 204
- `_modular_bundle` also catches `yaml.YAMLError` / pydantic `ValidationError`

**Concurrency:** `259f4f8`
- per-project lock across the whole write→recompile; role bindings compared order-insensitively under the lock

**Cleanup (the 3 nits):**
- `9459cdb` — one shared role-keyed serializer for CLI + platform (fixes the sorted-vs-insertion-order drift)
- `9067887` — `_apply_bundle` helper for the bundle triple
- `69f50b0` — skip no-op recompiles + memoize classic compiles

Validation: platform-api 477 passed / 4 skipped (opa-gated), SDK 1092 passed, ruff lint + format clean. Added 8 tests, updated 1. Per-thread details are in the review replies above.

One deliberate scope note: the per-project lock is in-process (fixes the single-worker race); the durable multi-worker fix is a Postgres advisory lock, documented in `core/locks.py` but not built here since policy writes are admin-only and infrequent — happy to add it if you'd prefer.

platform/api/hexgate_api/features/policy_modules/router.py
```diff
@@ -65,12 +101,27 @@ async def api_put_policy_module(
...
-            session, project_id=project_id, tier=tier, path=path, content=body.content
-        )
-    except service.InvalidModuleError as exc:
-        raise HTTPException(status_code=422, detail=str(exc)) from exc
+    # Serialize the write + recompile per project so overlapping edits can't
+    # commit bundles out of order (see core.locks).
+    async with project_lock(project_id):
+        prev_hash = await service.get_module_hash(session, project_id, tier, path)
+        try:
+            row = await service.upsert_module(
```
victorludvig, line 136: There is an edge case with this PUT method. If a capability containing mode: deny is PUT, upsert_module succeeds but the linker rejects the module at recompile. Result: the store holds a module that can't compose, and the agents keep enforcing the old (allow) bundle, with HTTP 200 and no error.

1. Admin PUTs a capability module containing a mode: deny tool → api_put_policy_module calls upsert_module:
   https://github.com/HexamindOrganisation/hexgate/blob/eeb3d9a12ed00f60d06c2e966c0d5d4b33b024a8/platform/api/hexgate_api/features/policy_modules/router.py#L106-L116
2. upsert_module validates only that the content parses as an AgentPolicy (deny is legal there), then commits the row:
   https://github.com/HexamindOrganisation/hexgate/blob/eeb3d9a12ed00f60d06c2e966c0d5d4b33b024a8/platform/api/hexgate_api/features/policy_modules/service.py#L126-L159
3. Router sees the hash changed and the project is modular → calls _recompile_project_agents, which wraps everything in except Exception: log:
   https://github.com/HexamindOrganisation/hexgate/blob/eeb3d9a12ed00f60d06c2e966c0d5d4b33b024a8/platform/api/hexgate_api/features/policy_modules/router.py#L120-L124
   https://github.com/HexamindOrganisation/hexgate/blob/eeb3d9a12ed00f60d06c2e966c0d5d4b33b024a8/platform/api/hexgate_api/features/policy_modules/router.py#L59-L76
4. recompile_project → _resolved_yaml_or_none → resolved_yaml_by_agent → resolve_for_project (the linker):
   https://github.com/HexamindOrganisation/hexgate/blob/eeb3d9a12ed00f60d06c2e966c0d5d4b33b024a8/platform/api/hexgate_api/features/agents/service.py#L238-L245
   https://github.com/HexamindOrganisation/hexgate/blob/eeb3d9a12ed00f60d06c2e966c0d5d4b33b024a8/platform/api/hexgate_api/features/policy_modules/service.py#L444-L449
5. Linker's _reject_capability_denies raises LinkError:
   https://github.com/HexamindOrganisation/hexgate/blob/eeb3d9a12ed00f60d06c2e966c0d5d4b33b024a8/hexgate/security/linker.py#L184-L201
6. _resolved_yaml_or_none catches it, logs a warning, returns None; recompile_project returns None, leaving all live bundles untouched:
   https://github.com/HexamindOrganisation/hexgate/blob/eeb3d9a12ed00f60d06c2e966c0d5d4b33b024a8/platform/api/hexgate_api/features/agents/service.py#L175-L187
7. Router returns HTTP 200 with the new module content. Store holds a module the linker rejects; agents still enforce the old allow.
   https://github.com/HexamindOrganisation/hexgate/blob/eeb3d9a12ed00f60datform/api/hexgate_api/features/policy_modules/router.py#L125

Same applies to the other per-module linker checks (consts conflicts, defile_scope, unsupported fields).

Contrast: the roles PUT checks service.resolves() and 409s + rolls back:
https://github.com/HexamindOrganisation/hexgate/blob/eeb3d9a12ed00f60d06c2e966c0d5d4b33b024a8/platform/api/hexgate_api/features/policy_modules/router.py#L207-L215    

Suggest: Before upsert_module writes, run the linker (resolve_for_project) on the current modules with the proposed module swapped in, and return 409 if it raises — so a module that can't compose is never stored.

docs/adr/R-POL-002-compile-agents-from-modules.md
```diff
@@ -0,0 +1,45 @@
...
+
+Fan-out is required because the policy is project-level: one module edit changes every agent's effective policy, so every agent must recompile. Reusing one resolved bundle for all of them keeps that to a single `opa` call per edit.
+
+Leaving bundles untouched on an unresolvable edit is the safety property. The alternative — clearing bundles when the project doesn't resolve — turns a typo in the dashboard into an outage across every agent in the project. Failing safe means the last good policy keeps enforcing while `check` surfaces the problem. This is why the write endpoints do not reject an unresolvable edit: the store can hold a work-in-progress state, but only a clean resolve replaces what is enforced.
```
victorludvig: ADR R-POL-002 contradicts the shipped code in two places. Line 24: "the write endpoints do not reject an unresolvable edit" — roles PUT and capability DELETE now 409. Lines 11/22/29: "single bundle reused for all agents / a single opa call per edit / all modular agents share one bundle" — the (role, agent) matrix in this same PR compiles one bundle per agent (test_recompile_builds_distinct_bundles_per_agent proves they differ). Update the ADR to what shipped.

platform/api/hexgate_api/features/policy_modules/router.py
```diff
@@ -113,7 +183,67 @@ async def api_set_policy_roles(
     _membership: tuple[User, OrganizationMember] = Depends(require_project_admin),
     session: AsyncSession = Depends(get_session),
 ) -> RoleBindingsRead:
-    roles = await service.set_roles(session, project_id=project_id, roles=body.roles)
+    # Hold the project lock across read-before → write → recompile so `before`
+    # can't go stale under a concurrent PUT (comparing a pre-write snapshot
+    # outside the lock could wrongly skip a needed recompile). See core.locks.
+    async with project_lock(project_id):
+        before = await service.get_roles(session, project_id)
+        roles = await service.set_roles(
```
victorludvig: We set the roles before validating them. So if the roles are invalid, they are briefly available to a GET request before validation.
Suggestion: validate before setting the roles.

platform/api/hexgate_api/core/locks.py
```diff
@@ -0,0 +1,26 @@
...
+
+from __future__ import annotations
+
+import asyncio
+from collections import defaultdict
+
+_locks: defaultdict[str, asyncio.Lock] = defaultdict(asyncio.Lock)
+
+
+def project_lock(project_id: str) -> asyncio.Lock:
```
victorludvig, line 24: update_agent and register_manifest (agents/router.py) also compile + write bundles for a project but don't take this lock, so they can still interleave with recompile_project. Wrap those two calls in project_lock too.

platform/api/hexgate_api/features/agents/service.py
```diff
@@ -232,10 +474,20 @@ async def register_manifest(
...
+        # built/served. Seed it deny-all (fail closed) rather than the
+        # permissive tool-derived starter, so a transient modular-compile
+        # failure can't hand the new agent more than the modules would allow.
+        from hexgate_api.features.agents.compiler import DENY_ALL_POLICY_YAML
+        from hexgate_api.features.policy_modules import service as modules
+
+        if await modules.is_modular(session, project_id):
+            agent.policy_yaml = DENY_ALL_POLICY_YAML
+        else:
+            agent.policy_yaml = _default_policy_for_manifest(manifest)
```
victorludvig, line 481: This branch only runs at registration. An agent registered while the project was modular keeps DENY_ALL as its policy_yaml, so if the project later reverts to classic (which test_modular_to_classic_transition_recompiles_from_policy_yaml allows), that agent is compiled deny-all with no signal. Worth either regenerating the tool-derived starter on the revert, or documenting this in the ADR.

platform/api/hexgate_api/models.py
```diff
@@ -370,10 +370,16 @@ class PolicyModule(SQLModel, table=True):
...
+    """A project role: the capabilities it imports, per executing agent.
 
     A role is a binding, not a policy (see R-POL-001) — boundaries apply to
     every role and are never listed here. One row per ``(project_id, role)``.
+
+    ``capabilities`` carries the ``(role, agent)`` matrix in its JSON value:
+    a mapping ``{agent-or-"*": [capability names]}``. A legacy flat ``[names]``
+    list is read as the generic ``{"*": [names]}`` agent — so the agent axis is
+    a value-shape evolution with no schema change (the column stays JSON, the
+    key stays ``(project_id, role)``). See ``agent-policy-dimension-design.md``.
```
victorludvig, line 382: File agent-policy-dimension-design.md is nowhere to be found.

victorludvig (changes requested):
Hi @quanghexa94, thank you for the revision. A couple of rare edge cases to check.

platform/api/hexgate_api/features/policy_modules/router.py
quanghexa94: Fixed in 61965d7. The module PUT now validates resolvability after upserting a modular project's module: if it no longer resolves (a capability with a `deny` hits the linker's library-wide reject), it's a 409 and the write is rolled back — the newly-created module is deleted, an edited one is restored to its prior content. Test: `test_capability_deny_put_rejected_on_modular_project`.

docs/adr/R-POL-002-compile-agents-from-modules.md
quanghexa94: Fixed in 61965d7. Updated R-POL-002: the 'single bundle reused for all agents / one opa call' claims now describe **per-agent** bundles (each agent compiled from its `(role, agent)` column, agents sharing a column deduped/memoized), and the 'write endpoints do not reject an unresolvable edit' paragraph now documents the write-time 409 rejection (roles PUT validates before writing; module PUT/DELETE reject + roll back).

platform/api/hexgate_api/features/policy_modules/router.py
quanghexa94: Fixed in 61965d7. The roles PUT now validates the **proposed** bindings in memory (`resolves_proposed`) BEFORE calling `set_roles`, so an invalid binding is never written — nothing to be briefly visible to a concurrent GET, and no rollback needed.

platform/api/hexgate_api/core/locks.py
quanghexa94: Fixed in 61965d7. `update_agent` and `register_manifest` (their agents/router endpoints) now run inside `project_lock(project_id)`, so they can't interleave with `recompile_project` and commit bundles out of order.

platform/api/hexgate_api/features/agents/service.py
quanghexa94: Addressed in 61965d7 both ways: `recompile_project`'s classic branch now logs a warning when an agent's `policy_yaml` is the deny-all fallback (so a modular→classic revert that leaves it deny-all is surfaced, not silent), and R-POL-002 documents the behavior. It stays fail-closed (deny-all is the safe direction); the operator re-authors the policy.

platform/api/hexgate_api/models.py
quanghexa94: Fixed in 61965d7 — `agent-policy-dimension-design.md` is now committed to the branch (it documents the whole (role, agent) matrix design), so the reference resolves.

platform/api/hexgate_api/features/policy_modules/service.py
```diff
@@ -249,3 +356,139 @@ async def check(session: AsyncSession, project_id: str):
...
+    return {role: _normalize_cell(cells) for role, cells in roles.items()}
+
+
+async def resolves(session: AsyncSession, project_id: str) -> bool:
+    """Whether the project's STORED modules + bindings compose for every agent.
+
+    A cheap, opa-free precondition for accepting a write that edits a MODULE
+    (the bindings are already stored). ``True`` only if every agent column
+    resolves."""
+    boundaries, capabilities, roles = await _sdk_inputs(session, project_id)
```
victorludvig: _sdk_inputs re-parses every stored module and can raise ValidationError/YAMLError, but it's outside _resolves_with's try — so a stored module that no longer parses turns a PUT into a 500. The old resolves() caught this and returned False (→ 409).

platform/api/hexgate_api/features/policy_modules/router.py
```diff
@@ -65,12 +101,53 @@ async def api_put_policy_module(
...
+        # modular: a byte-identical re-PUT, or any edit to a classic library,
+        # changes no agent, so don't pay a resolve + opa compile for it.
+        if row.content_hash != prev_hash and await service.is_modular(
+            session, project_id
+        ):
+            # An edit that breaks composition (e.g. a capability with a deny, which
+            # the linker rejects) must not be accepted silently: agents would keep
+            # the old bundle while the store holds an uncomposable module. Reject
+            # and restore the previous content (or delete a newly-created module).
+            if not await service.resolves(session, project_id):
```
victorludvig: The module PUT writes the bad module first (upsert_module commits), and only then checks resolves() and rolls back. During that gap, a concurrent GET can see the broken module. The roles PUT was just fixed to validate before writing — the module PUT should do the same instead of commit-then-rollback.

platform/api/hexgate_api/features/policy_modules/service.py
```diff
@@ -85,6 +94,28 @@ async def _get_module(
     ).first()
 
 
+async def get_module_hash(
```
victorludvig: Dead code: get_module_hash has no callers after get_module_peek replaced it.

platform/api/hexgate_api/features/agents/router.py
```diff
@@ -348,10 +353,14 @@ async def api_register_agent(
...
     from hexgate_api.core.keystore import keystore
+    from hexgate_api.core.locks import project_lock
 
-    version, created = await register_manifest(
-        session, project_id, body.manifest, sign=keystore.sign
-    )
+    # Hold the project lock: a first-time register compiles + writes a bundle,
+    # so it must not interleave with recompile_project and commit out of order.
+    async with project_lock(project_id):
+        version, created = await register_manifest(
```
victorludvig, line 370: This lock now serializes every POST /v1/agents — and hexgate serve auto-registers on startup, so a fleet redeploy queues all agents behind it (and behind any in-flight opa recompile). Most of those are no-op re-registers (unchanged content_hash) that compile nothing. Suggestion: check the content_hash no-op before taking the lock.

victorludvig (changes requested):

platform/api/hexgate_api/features/policy_modules/service.py
quanghexa94: Good catch — that was a regression from my refactor. Fixed in 04a0d55: `resolves()` and `resolves_proposed()` now wrap `_sdk_inputs` inside the try, so a stored module that no longer parses folds to `False` (→ 409) instead of escaping as a 500. Regression guard: `test_resolves_false_on_unparseable_stored_module`.

platform/api/hexgate_api/features/policy_modules/router.py
quanghexa94: Fixed in 04a0d55. The module PUT now validates the draft in memory (`resolves_with_module` overlays the unsaved module and resolves) BEFORE `upsert_module`, so nothing is written when it wouldn't compose — no commit-then-rollback window a concurrent GET could observe. Same pattern as the roles PUT.

platform/api/hexgate_api/features/policy_modules/service.py
quanghexa94: Fixed in 04a0d55 — removed `get_module_hash`; the module PUT uses `get_module_peek` for the change check, so it was dead.

platform/api/hexgate_api/features/agents/router.py
quanghexa94: Fixed in 04a0d55. `POST /v1/agents` now takes `project_lock` only when the agent doesn't exist yet (a first registration — the only path that compiles a bundle). A re-register of an existing agent skips the lock, so a fleet redeploy's no-op re-registers don't queue behind it or an in-flight recompile.

victorludvig (approved):
Thanks for the refinements, approved!

PR #122, opened 2026-08-17
feat(sdk): agent-level policy model with admission/agents blocks

quanghexa94:
Add the authoring surface for agent-level authorization: `admission` and `agents` blocks on `AgentPolicy`. This is PR-A1 of the agent-level enforcement series (gating who may run an agent, and who may hand off to another), and it is policy-side only: pure SDK, no platform change, no migration, no runtime wiring yet. It ships the model and the lowering so a policy can be authored and decided offline; the runtime gate, handoff interception, and depth cap follow in later PRs.

## Design

The key idea is that an agent-level rule reuses the tool decision path rather than adding a second one. `admission` and `agents` lower into synthetic tool keys through a computed `AgentPolicy.effective_tools`:

- `admission` becomes `agent.run`.
- each `agents` target becomes one key per `via` mode: `agent.tool:<name>` (agent-as-tool, orchestrator keeps control) and/or `agent.handoff:<name>` (control transfers).

Both engines read `effective_tools` (the pydantic `get_tool_policy` and the Rego compiler), so a lowered `agent.*` key evaluates identically on the pydantic and WASM paths with no engine or grammar change. Tool names are opaque strings in both engines (the Rego compiler emits `input.tool == "<name>"`), so the `.` and `:` in a key round-trip cleanly, the way the existing `net.*` egress tools already do.

Two decisions worth knowing:

- Only the listed rules lower. The fallback for an unlisted target is the runtime gate's job in a later PR, not this loader's. Folding a shared agent default into the single `default_policy` would make an existing agent (no `admission` block, deny-default) deny its own admission, a backward-compat break. Delegation closed-world still comes for free: an unlisted `agent.*` key hits the deny-by-default `default_policy`, so a listed-`agents` policy denies unlisted targets with no extra machinery.
- The `agent.*` key shapes are reserved. An authored tool that collides with `agent.run` / `agent.tool:` / `agent.handoff:` is rejected at load.

## Important files

| File | What to look at |
| --- | --- |
| `hexgate/security/models.py` | the core of the PR: the new fields, `AgentTargetPolicy` with its `via` list, and `lowered_agent_tools` / `effective_tools` |
| `hexgate/security/policy.py`, `hexgate/security/rego.py` | the parity-critical edits: each engine switches its tool lookup from `tools` to `effective_tools` |
| `hexgate/security/policy_set.py` | inheritance carry-through of the new fields, with a fail-open guard so a later mixin cannot null an earlier admission rule; const-ref validation over `effective_tools` |
| `tests/security/test_agent_policy.py` | the 20 cases pinning lowering, via modes, reserved names, enforcement, Rego parity, and inheritance |

## Test plan

Ran the full SDK unit suite: 2001 passed, 6 skipped (opt-in integration), no regressions. The live WASM engine tests are in that run, so the pydantic/WASM parity is exercised end to end, not just predicted. ruff check and ruff format are clean on all touched files.

Checked by hand through the `assert_*` helpers in the new suite:

- a listed handoff target with `approval_required` returns NEEDS_APPROVAL within its `args.depth <= 2` constraint and DENY past it;
- a `via: [tool]` target allows as a tool and denies handoff;
- an unlisted target denies under the deny-by-default policy (closed-world for free);
- inheritance keeps both an inherited `admission` rule and an inherited target after the merge.

A reviewer catches a parity regression if `test_rego_compiler_emits_lowered_agent_keys` or any WASM test fails, and a backward-compat regression if an agent with no agent blocks stops resolving its tools (`effective_tools` must equal `tools` in that case).

## Try it

```
# on this branch, from the repo root
uv run pytest tests/security/test_agent_policy.py -q
```

## Notes

- Blast radius: pure SDK, no platform change, no migration. `effective_tools` equals `tools` when no agent blocks are present, so existing policies are byte-for-byte unaffected.
- The agent default for an unlisted target is deferred to the runtime gate in PR-A2. Until then, closed-world falls out of the deny-by-default `default_policy`.
- A policy *module* (boundary/capability) that carries an agent block is rejected with a `LinkError` rather than silently dropped in the fold, mirroring the existing `file_scope` guard. Full agent-blocks-in-modules support is a later PR.
- Deferred to later PRs: the agent gate seam, admission and handoff interception across the four adapters, and the delegation depth cap.

Follow-up commit `8c19a7b` addressed the code review: the linker guard above, memoizing `effective_tools` off the hot path, and dropping an unused `default_agent_policy` field.
- Design doc: `agent-level-enforcement-design.md` (working note, not committed).
- Tooling: `uv` was not available in my environment, so I ran the equivalents (`ruff` and `pytest` via the project venv) rather than `make lint && make fmt-check && make coverage`. CI runs the make targets.

### DIFF
--- a/hexgate/security/policy_set.py
+++ b/hexgate/security/policy_set.py
@@ -332,6 +332,24 @@ def _resolve_inheritance(
     # user's intent is to override, and silently inheriting an ``allow`` from a
     # parent would be fail-open.
     merged_tools.update(own.tools)
+    # A per-target override replaces the parent's rule wholesale, so a child that
+    # DROPS a via the parent listed would silently un-list that mode (it falls
+    # through to default_policy, fail-open under a permissive default). One
+    # AgentTargetPolicy carries a single mode across its vias, so the parent's
+    # mode for the dropped via cannot be preserved in the merged entry; reject the
+    # narrowing loudly instead, the same stance as the admission merge above.
+    for target, own_rule in own.agents.items():
+        parent_rule = merged_agents.get(target)
+        if parent_rule is None:
+            continue
+        dropped = [via for via in parent_rule.via if via not in own_rule.via]
+        if dropped:
+            raise PolicySetError(
+                f"role {name!r} narrows agent target {target!r} to vias "
+                f"{own_rule.via}, dropping {dropped} that an inherited rule lists. "
+                "A child agent rule cannot silently un-list an inherited via mode. "
+                "Re-declare the full via set for this target."
+            )
     merged_agents.update(own.agents)
     merged_consts.update(own.consts)
     if "default_policy" in own.model_fields_set:
--- a/hexgate/security/policy_set.py
+++ b/hexgate/security/policy_set.py
@@ -332,28 +332,14 @@ def _resolve_inheritance(
     # user's intent is to override, and silently inheriting an ``allow`` from a
     # parent would be fail-open.
     merged_tools.update(own.tools)
-    # A per-target override replaces the parent's rule wholesale, so a child that
-    # DROPS a via the parent listed would silently un-list that mode (it falls
-    # through to default_policy, fail-open under a permissive default). One
-    # AgentTargetPolicy carries a single mode across its vias, so the parent's
-    # mode for the dropped via cannot be preserved in the merged entry; reject the
-    # narrowing loudly instead, the same stance as the admission merge above.
-    for target, own_rule in own.agents.items():
-        parent_rule = merged_agents.get(target)
-        if parent_rule is None:
-            continue
-        dropped = [via for via in parent_rule.via if via not in own_rule.via]
-        if dropped:
-            raise PolicySetError(
-                f"role {name!r} narrows agent target {target!r} to vias "
-                f"{own_rule.via}, dropping {dropped} that an inherited rule lists. "
-                "A child agent rule cannot silently un-list an inherited via mode. "
-                "Re-declare the full via set for this target."
-            )
-    merged_agents.update(own.agents)
-    merged_consts.update(own.consts)
+    # Finalise the effective default before the narrowing guard: a dropped via
+    # resolves to *this* role's default_policy, so the guard must see the child's
+    # override, not the parent's.
     if "default_policy" in own.model_fields_set:
         merged_default = own.default_policy
+    _reject_fail_open_narrowing(name, own, merged_agents, merged_default)
+    merged_agents.update(own.agents)
+    merged_consts.update(own.consts)
     if "admission" in own.model_fields_set:
         merged_admission = own.admission
 
@@ -367,3 +353,40 @@ def _resolve_inheritance(
         admission=merged_admission,
         agents=merged_agents,
     )
+
+
+# How much access each mode grants, low to high.
+_PERMISSIVENESS = {"deny": 0, "approval_required": 1, "allow": 2}
+
+
+def _reject_fail_open_narrowing(
+    role: str,
+    own: AgentPolicy,
+    parent_agents: Mapping[str, AgentTargetPolicy],
+    effective_default: BaseToolPolicy,
+) -> None:
+    """Reject a child that drops an inherited via *and* thereby loosens it.
+
+    A per-target override replaces the parent's rule wholesale, and one
+    :class:`AgentTargetPolicy` carries a single mode across its vias, so the
+    parent's mode for a dropped via cannot be preserved in the merged entry. The
+    dropped via falls through to ``default_policy``. That is only a problem when the
+    default is MORE permissive than the parent's mode (e.g. parent ``deny``, child
+    default ``allow``): it silently loosens a restriction. Under a deny-by-default
+    child, dropping a via only tightens it, a common and safe intent (inherit a tool
+    grant, but never hand off), so it is allowed.
+    """
+    default_rank = _PERMISSIVENESS[effective_default.mode]
+    for target, own_rule in own.agents.items():
+        parent_rule = parent_agents.get(target)
+        if parent_rule is None:
+            continue
+        dropped = [via for via in parent_rule.via if via not in own_rule.via]
+        if dropped and default_rank > _PERMISSIVENESS[parent_rule.mode]:
+            raise PolicySetError(
+                f"role {role!r} drops via {dropped} from inherited agent target "
+                f"{target!r} (parent mode {parent_rule.mode!r}) under a more "
+                f"permissive default_policy ({effective_default.mode!r}), which "
+                "would silently loosen the parent's rule for that mode. Re-declare "
+                "the full via set, or tighten default_policy."
+            )
--- a/hexgate/security/policy_set.py
+++ b/hexgate/security/policy_set.py
@@ -38,7 +38,12 @@ import yaml
 
 from hexgate.security.constraints import iter_const_refs, parse_constraint
 from hexgate.security.decision import Verdict
-from hexgate.security.models import AgentPolicy, BaseToolPolicy, ToolPolicy
+from hexgate.security.models import (
+    AgentPolicy,
+    AgentTargetPolicy,
+    BaseToolPolicy,
+    ToolPolicy,
+)
 
 
 DEFAULT_ROLE_NAME = "default"
@@ -135,7 +140,7 @@ def _validate_const_refs(policies: Mapping[str, AgentPolicy]) -> None:
     """
     for role, policy in policies.items():
         available = set(policy.consts)
-        for tool_policy in (*policy.tools.values(), policy.default_policy):
+        for tool_policy in (*policy.effective_tools.values(), policy.default_policy):
             for raw in tool_policy.constraints:
                 for name in iter_const_refs(parse_constraint(raw)):
                     if name not in available:
@@ -303,15 +308,27 @@ def _resolve_inheritance(
     if not own.inherits:
         return own
     merged_tools: dict[str, ToolPolicy] = {}
+    merged_agents: dict[str, AgentTargetPolicy] = {}
     merged_consts: dict[str, object] = {}
     merged_default: BaseToolPolicy = own.default_policy
+    merged_admission: BaseToolPolicy | None = own.admission
+    merged_default_agent: BaseToolPolicy | None = own.default_agent_policy
 
-    # Merge parents left-to-right (later parents override earlier).
+    # Merge parents left-to-right (later parents override earlier). ``agents``
+    # merges by target name like ``tools`` does. The optional agent-level scalars
+    # (``admission`` / ``default_agent_policy``) only overwrite when a parent
+    # actually sets one, so a later mixin that omits them can't null out an
+    # earlier parent's rule — dropping an agent gate silently would be fail-open.
     for parent_name in own.inherits:
         parent = _resolve_inheritance(parent_name, raw, chain + [name])
         merged_tools.update(parent.tools)
+        merged_agents.update(parent.agents)
         merged_consts.update(parent.consts)
         merged_default = parent.default_policy
+        if parent.admission is not None:
+            merged_admission = parent.admission
+        if parent.default_agent_policy is not None:
+            merged_default_agent = parent.default_agent_policy
 
     # Self overrides everything from parents. Check ``model_fields_set`` rather
     # than comparing against ``BaseToolPolicy()``: a child that explicitly says
@@ -319,9 +336,14 @@ def _resolve_inheritance(
     # user's intent is to override, and silently inheriting an ``allow`` from a
     # parent would be fail-open.
     merged_tools.update(own.tools)
+    merged_agents.update(own.agents)
     merged_consts.update(own.consts)
     if "default_policy" in own.model_fields_set:
         merged_default = own.default_policy
+    if "admission" in own.model_fields_set:
+        merged_admission = own.admission
+    if "default_agent_policy" in own.model_fields_set:
+        merged_default_agent = own.default_agent_policy
 
     return AgentPolicy(
         version=own.version,
@@ -330,4 +352,7 @@ def _resolve_inheritance(
         default_policy=merged_default,
         tools=merged_tools,
         consts=merged_consts,
+        admission=merged_admission,
+        agents=merged_agents,
+        default_agent_policy=merged_default_agent,
     )
--- a/hexgate/security/policy_set.py
+++ b/hexgate/security/policy_set.py
@@ -332,14 +332,10 @@ def _resolve_inheritance(
     # user's intent is to override, and silently inheriting an ``allow`` from a
     # parent would be fail-open.
     merged_tools.update(own.tools)
-    # Finalise the effective default before the narrowing guard: a dropped via
-    # resolves to *this* role's default_policy, so the guard must see the child's
-    # override, not the parent's.
-    if "default_policy" in own.model_fields_set:
-        merged_default = own.default_policy
-    _reject_fail_open_narrowing(name, own, merged_agents, merged_default)
     merged_agents.update(own.agents)
     merged_consts.update(own.consts)
+    if "default_policy" in own.model_fields_set:
+        merged_default = own.default_policy
     if "admission" in own.model_fields_set:
         merged_admission = own.admission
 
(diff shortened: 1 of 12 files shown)


victorludvig (commented):
I approve the logic, here are some Claude's findings to check before merging:


The core design is sound: lowering admission/agents into synthetic agent.* tool keys so both engines share one decision path is the right call, and I verified pydantic/WASM parity actually holds (_default_rules builds its catch-all exclusion list from effective_tools, so lowered keys are correctly excluded). 983 security tests pass.

Three correctness issues, all reproduced.

1. check_default_role_exposure misses every lowered agent grant — hexgate/security/analyzer.py:503,507

Switch these to effective_tools. The PR migrated policy.py, rego.py, and policy_set.py:143 from .tools → .effective_tools but missed the analyzer, which operates on a resolved PolicySet where agent blocks are legal (unlike the linker path, which rejects them).

Reproduced — identical policy shapes, one linted, two silent:

default role grants tool "wire_transfer"     → [('permissive-default', 'wire_transfer')]
default role grants agents{admin-bot: allow} → []
default role grants admission: allow         → []

This is the lint whose entire job is "the default role grants something no named role does, so any caller with an unrecognised role name reaches it." A default role that can hand off to admin-bot is exactly the grant most worth catching, and it lints clean. Still present on current HEAD (4f7ed5d), so PR-A2 didn't fix it.

2. Narrowing via in a child role erases a parent's explicit deny — hexgate/security/policy_set.py:335

merged_agents.update(own.agents) replaces a target's entry wholesale, so dropping a via mode un-lists that mode rather than inheriting the parent's rule for it. Under a permissive default_policy it flips deny → allow:

base:    agents: {admin-bot: {mode: deny,  via: [tool, handoff]}}
support: inherits: [base]
         default_policy: {mode: allow}
         agents: {admin-bot: {mode: allow, via: [tool]}}   # narrowing to tool-only
agent.tool:admin-bot     -> allow   (intended)
agent.handoff:admin-bot  -> allow   (parent said deny)  ← FAIL-OPEN

Under deny-by-default it correctly yields deny, so the blast radius is default_policy: allow policies. Worth fixing because it's the same hazard the PR deliberately guards against one field over: merged_admission uses an is not None check specifically so "a later mixin can't null out an earlier parent's rule — dropping an agent gate silently would be fail-open." The agents map has that exact hole at via-mode granularity. Merging per-(target, via) rather than per-target would close it.

3. cached_property goes stale across model_copy — hexgate/security/models.py:181

q = AgentPolicy(tools={"a": ...});  q.effective_tools   # caches
r = q.model_copy(update={"tools": {"zzz": ...}})
r.tools            -> ['zzz']
r.effective_tools  -> ['a']        ← STALE

Latent, not live: the one model_copy on an AgentPolicy (builder.py:321) updates only inherits/is_mixin, on a freshly built policy whose cache was never populated. But the docstring's justification — "policies are immutable after load" — isn't enforced (AgentPolicy has no frozen=True), and model_copy is already used on this model. Adding frozen=True would make the stated invariant real, or drop the cache in favour of computing once at load.

I checked the other cached_property risks and they're fine: memoization genuinely works, model_dump doesn't leak effective_tools into output, and pydantic v2's __eq__ field-fallback means a populated cache doesn't break equality.

Cleanup

- hexgate/security/linker.py:275 — _reject_agent_blocks is the fourth near-identical guard (alongside _reject_capability_denies, _reject_default_policy_constraints, _reject_file_scope), each hand-enumerating one field the fold doesn't compose. All three cleanup angles converged here independently. The failure mode is structural: _fold_tool reads only .tools, so the next AgentPolicy field ships fail-open unless someone remembers to write guard #5. An allowlist check (model_fields_set - {tools, consts, default_policy, is_mixin, inherits, version} must be empty) fails closed on any new field automatically.
- models.py:185 — lowered_agent_tools rebuilds a bare BaseToolPolicy(mode=..., constraints=...) instead of using target_policy directly, even though AgentTargetPolicy is a BaseToolPolicy. Any field later added to BaseToolPolicy would validate on an agents entry and then silently vanish from effective_tools.
- models.py:95 — _is_reserved_agent_key hardcodes "agent.tool:"/"agent.handoff:" rather than deriving them from the AgentVia literal that agent_target_key already uses. A third via mode added without the parallel edit here lets a colliding authored tool name through.
- Minor: effective_tools returns self.tools by identity when no agent blocks exist (a caller mutating the result mutates the policy); an empty target name in agents mints the nonsense key agent.tool: with no validation.

quanghexa94:
Thanks @victorludvig, all three landed in c8570c1. Rundown, and one call I want your read on (#2).

**1. permissive-default misses agent grants (analyzer.py:503).** Fixed: switched the lint to `effective_tools`, so an `admission` or `agents` grant on the default role now flags. Added a regression test where `agents: {admin-bot: allow}` on default lints both lowered keys.

**2. via-narrowing erases a parent deny (policy_set.py).** I fixed this fail-closed rather than with the per-(target, via) merge, and want your view. The blocker: one `AgentTargetPolicy` carries a single mode across its vias, so a merge that keeps the parent's handoff-deny while the child allows tool-only is not representable in the merged `agents` dict at all. That "tool allow, handoff deny for the same target" state can't even be authored in a single role today; it would only ever arise from inheritance. So a correct per-via merge is a model change, not a merge tweak. For this PR I reject a child that drops an inherited via (`PolicySetError`), which closes the fail-open and matches the admission-merge stance one field over. If the per-via expressiveness is worth having (different modes per via, reachable only through inheritance), I'll spin it as a follow-up PR. Let me know which way you'd prefer.

**3. cached_property goes stale across model_copy (models.py).** Added `frozen=True`. The one `model_copy` on `AgentPolicy` (builder.py) only updates `inherits`/`is_mixin` on a freshly built policy, so the cache can't actually go stale in practice, and frozen makes the "immutable after load" invariant real instead of a docstring promise. Added a frozen test.

**Cleanups.** Generalized the linker guard to an allowlist (`_reject_unsupported_module_fields`), so a new top-level `AgentPolicy` field is rejected automatically instead of needing guard #5. `lowered_agent_tools` now reuses the `AgentTargetPolicy` (a future base field won't silently vanish). `_is_reserved_agent_key` derives its prefixes from `AgentVia`.

Full suite green (2005 passed). The analyzer-coverage gap for the *other* lints (dead-grant, over-grant over agent keys) I'm leaving for the agent-level lints PR, as scoped in the design doc, since those are advisory rather than a fail-open.

victorludvig:
> Thanks @victorludvig, all three landed in [c8570c1](https://github.com/HexamindOrganisation/hexgate/commit/c8570c1839374c33f76c79e7c3697e16d7d903b3). Rundown, and one call I want your read on (#2).
> 
> **1. permissive-default misses agent grants (analyzer.py:503).** Fixed: switched the lint to `effective_tools`, so an `admission` or `agents` grant on the default role now flags. Added a regression test where `agents: {admin-bot: allow}` on default lints both lowered keys.
> 
> **2. via-narrowing erases a parent deny (policy_set.py).** I fixed this fail-closed rather than with the per-(target, via) merge, and want your view. The blocker: one `AgentTargetPolicy` carries a single mode across its vias, so a merge that keeps the parent's handoff-deny while the child allows tool-only is not representable in the merged `agents` dict at all. That "tool allow, handoff deny for the same target" state can't even be authored in a single role today; it would only ever arise from inheritance. So a correct per-via merge is a model change, not a merge tweak. For this PR I reject a child that drops an inherited via (`PolicySetError`), which closes the fail-open and matches the admission-merge stance one field over. If the per-via expressiveness is worth having (different modes per via, reachable only through inheritance), I'll spin it as a follow-up PR. Let me know which way you'd prefer.
> 
> **3. cached_property goes stale across model_copy (models.py).** Added `frozen=True`. The one `model_copy` on `AgentPolicy` (builder.py) only updates `inherits`/`is_mixin` on a freshly built policy, so the cache can't actually go stale in practice, and frozen makes the "immutable after load" invariant real instead of a docstring promise. Added a frozen test.
> 
> **Cleanups.** Generalized the linker guard to an allowlist (`_reject_unsupported_module_fields`), so a new top-level `AgentPolicy` field is rejected automatically instead of needing guard #5. `lowered_agent_tools` now reuses the `AgentTargetPolicy` (a future base field won't silently vanish). `_is_reserved_agent_key` derives its prefixes from `AgentVia`.
> 
> Full suite green (2005 passed). The analyzer-coverage gap for the _other_ lints (dead-grant, over-grant over agent keys) I'm leaving for the agent-level lints PR, as scoped in the design doc, since those are advisory rather than a fail-open.

Hi @quanghexa94 thank you for the revision. Ok for #1 and #3.

What are your thoughts on this example for #2? 
```yaml
base:     is_mixin: true
          tools:  {search_kb: {mode: allow}, read_invoice: {mode: allow}}
          agents: {billing-bot: {mode: allow, via: [tool, handoff]}}
support:  inherits: [base]                                  # deny-by-default
          agents: {billing-bot: {mode: allow, via: [tool]}} # now raises; resolved fine before

support wants the mixin's tools and can ask billing-bot a question, but must never hand the customer
conversation to it — on handoff billing-bot owns the conversation under its own policy, refunds included.
```

quanghexa94:
Good catch, you're right, my first pass over-rejected. Fixed in 0ee8040.

The guard now rejects narrowing only when it would actually loosen the parent's rule, i.e. when the child's effective `default_policy` is more permissive than the parent's mode for the dropped via (`deny < approval_required < allow`). Your example resolves cleanly again:

- parent `billing-bot: allow, via [tool, handoff]`, child `support` is deny-by-default and re-declares `billing-bot: allow, via [tool]`. The dropped `handoff` falls through to `support`'s deny default, so `agent.handoff:billing-bot` resolves to deny, which is exactly "callable as a tool, never hand off." No error.
- The genuine fail-open still rejects: parent `deny`, child drops the via under an `allow` default, which would silently turn the deny into allow.

Added your case as a passing test (asserts `search_kb` allow, `agent.tool:billing-bot` allow, `agent.handoff:billing-bot` deny for `support`), and kept the fail-open case rejecting. Full suite green (2006).

victorludvig:
> Good catch, you're right, my first pass over-rejected. Fixed in [0ee8040](https://github.com/HexamindOrganisation/hexgate/commit/0ee8040538bd1ac77cbea0e48e7fd4809f38bb8a).
> 
> The guard now rejects narrowing only when it would actually loosen the parent's rule, i.e. when the child's effective `default_policy` is more permissive than the parent's mode for the dropped via (`deny < approval_required < allow`). Your example resolves cleanly again:
> 
> * parent `billing-bot: allow, via [tool, handoff]`, child `support` is deny-by-default and re-declares `billing-bot: allow, via [tool]`. The dropped `handoff` falls through to `support`'s deny default, so `agent.handoff:billing-bot` resolves to deny, which is exactly "callable as a tool, never hand off." No error.
> * The genuine fail-open still rejects: parent `deny`, child drops the via under an `allow` default, which would silently turn the deny into allow.
> 
> Added your case as a passing test (asserts `search_kb` allow, `agent.tool:billing-bot` allow, `agent.handoff:billing-bot` deny for `support`), and kept the fail-open case rejecting. Full suite green (2006).

Thanks for the revision. 
Another narrow bug if the parent policy for handoff has contraints and the child agent uses the default policy, then the constrains are dropped:

```
base:     agents: {billing-bot: {mode: allow, via: [tool, handoff],
                                 constraints: ["args.amount <= 100"]}}
support:  inherits: [base]
          default_policy: {mode: allow}
          agents: {billing-bot: {mode: allow, via: [tool]}} -> can use handoff without constraints
```

May be worth fixing with one line: also reject when the ranks tie and the parent rule carries constraints the default doesn't.

quanghexa94:
Right again, thanks. Fixed in 3fc5d9f.

The guard was comparing only mode ranks, so a same-mode tie slipped through even when the parent rule carried a constraint the default drops. Generalized it to "does the default loosen the parent rule for the dropped via":

- more permissive mode (deny → allow), or
- same granting mode but the default lacks a constraint the parent has (`set(parent.constraints) - set(default.constraints)` non-empty).

`deny` ties are never treated as looser, since deny ignores constraints anyway. Your example now rejects (parent handoff `allow` + `args.amount <= 100`, child allow-default → the cap can't silently vanish), and a parent with no constraint under a matching allow-default still passes (nothing is lost). Added both as tests. Full suite green (2008).

hexgate/security/policy_set.py
```diff
@@ -369,24 +369,44 @@ def _reject_fail_open_narrowing(
...
+def _default_loosens(default: BaseToolPolicy, parent_rule: AgentTargetPolicy) -> bool:
+    """True if falling through to ``default`` is looser than ``parent_rule``.
+
+    Looser means a more permissive mode, or the same granting mode with a parent
+    constraint the default drops. ``deny`` ignores constraints, so a same-rank
+    ``deny`` tie is never looser."""
+    default_rank = _PERMISSIVENESS[default.mode]
+    parent_rank = _PERMISSIVENESS[parent_rule.mode]
+    if default_rank != parent_rank:
+        return default_rank > parent_rank
```
victorludvig: The early stopping on mode ranking can lead to constraints loosening.
Consider this example:
```

base:     agents: {billing-bot: {mode: allow, via: [tool, handoff],
                                constraints: ["args.amount <= 100"]}}
support:  inherits: [base]
          default_policy: {mode: approval_required}   # rank 1 < allow's 2 → guard passes
          agents: {billing-bot: {mode: allow, via: [tool]}}
```

The default_loosens returns false because approval_required < allow, however there are no constrains on args.amount for the child billing-bot. So it can be used in handoff mode with the default policy which does not enforce constraints.

We should remove the early stopping and use something like:

```
if default.mode == "deny":
    return False                              # a deny default can never loosen
if _PERMISSIVENESS[default.mode] > _PERMISSIVENESS[parent_rule.mode]:
    return True                               # more permissive mode
if parent_rule.mode == "deny":
    return False                              # deny ignores constraints
return bool(set(parent_rule.constraints) - set(default.constraints))
```

hexgate/security/policy_set.py
```diff
@@ -369,24 +369,44 @@ def _reject_fail_open_narrowing(
...
+
+    * a more permissive mode (e.g. parent ``deny``, child default ``allow``), or
+    * the same granting mode but *fewer constraints* (e.g. parent ``allow`` with
+      ``args.amount <= 100``, child default a bare ``allow`` — the cap vanishes).
+
+    Under a deny-by-default child, dropping a via only tightens it, a common and
+    safe intent (inherit a tool grant, but never hand off), so it is allowed.
     """
-    default_rank = _PERMISSIVENESS[effective_default.mode]
     for target, own_rule in own.agents.items():
```
victorludvig: The guard loops through own.agents, which only contains the agents defined for this role (see https://github.com/HexamindOrganisation/hexgate/pull/122/changes/3fc5d9f3bd414c4ead72a7495e74900d0f9aaba4#diff-9e0b1981644f98ecf4aba53169116a9d4d2eb0acd0e741a6a1f5fae53976294dL307)
Consider a child agent that does not re-define any agent. It inherits the parent agents, however we don't loop through them here. Hence, the guard does not run:

```
base:    agents: {billing-bot: {mode: allow, via: [tool, handoff], constraints: ["args.amount <= 100"]}}
mid:     inherits: [base]        # deny default → narrowing here is legitimate
         agents: {billing-bot: {mode: allow, via: [tool], constraints: ["args.amount <= 100"]}}
support: inherits: [mid]
         default_policy: {mode: allow}   # declares no agents → guard never runs
```

This loads clean and agent.handoff:billing-bot resolves to an uncapped allow (amount 5000 → ALLOW).

victorludvig (changes requested):
Two further cases that slip through the guard were identified, can you have a look?

quanghexa94:
Both real, thanks. And Bug B convinced me the guard was the wrong tool: once a via is dropped up the chain, the merged per-target rule keeps no record of it, so no per-level check can see it. So instead of patching the guard, I fixed the cause in 748caf3.

**Agent reach is now closed-world.** An unlisted `agent.*` key denies regardless of `default_policy`, in both engines:

- pydantic: `get_tool_policy` returns deny for an unlisted agent key; a permissive default only governs ordinary tools.
- rego: the permissive catch-all excludes `agent.*` (`not startswith(input.tool, "agent.")`), so an unlisted agent key matches no rule and denies. Added a pydantic-vs-WASM parity test for it.

That kills both of your cases and the earlier three at once (every dropped or unlisted via just denies), and it let me delete `_reject_fail_open_narrowing` entirely, ranks, constraint-diff, inheritance edge cases, all gone. Your Bug B is now a passing test: `mid` narrows to tool-only, `support` flips default to `allow` and declares no agents, and `agent.handoff:billing-bot` still denies at `amount 5000`.

The one behavior change worth flagging: `default_policy: allow` no longer implicitly grants agent reach, a role must list agents explicitly. For a security layer that felt like the right default, and agent blocks are new so nothing depends on the old behavior. Full suite green (2009).

hexgate/security/rego.py
```diff
@@ -286,11 +294,27 @@ def _rules_for_role(
                 role,
                 role_guard,
                 tool_guard,
-                policy.tools[tool_name],
+                effective[tool_name],
                 tool_name,
                 helpers,
             )
         )
+    if policy.agents and AGENT_RUN_TOOL not in effective:
```
victorludvig: For Rego, a role with no agents block and no admission gets no agent.run rule at all (the opt-in rule is gated on policy.agents), so under a deny default it cannot use the agent. 
On the contrary the pydantic _ADMISSION_OPT_IN_ALLOW fallback applies to every role.

Consider the following example:

admin:  {admission: {mode: allow}}
worker: {tools: {read: {mode: allow}}}   # no agents block, deny default

worker + agent.run: pydantic → ALLOW (opt-in), rego → DENY (no listed rule, no opt-in rule, no permissive catch-all). This is the same non-monotonicity the commit fixes on the pydantic side: admin declaring admission locks worker out of the agent on the WASM engine, without worker's YAML changing. Latent until the PR-A2 gate queries agent.run.

Fix: emit the rule whenever AGENT_RUN_TOOL not in effective (drop the policy.agents gate).

victorludvig (commented):
Thanks for the revision, this simplifies it.
One more edge case to check.

quanghexa94:
Good catch, that gate was my doing and you're right it reintroduced the exact non-monotonicity one engine over. Fixed in 3b9cf95: dropped the `policy.agents` gate, so the `agent.run` opt-in allow rule is emitted for every role that doesn't list admission, matching the unconditional pydantic fallback. Your `admin`/`worker` example now agrees on both engines (worker → allow).

It means every compiled policy carries a per-role `agent.run` rule now, so I updated the rule-count / single-allow structural tests and regenerated the golden fixture. Still latent (bundle admission is deferred, so agent.run isn't queried on WASM yet), but the engines shouldn't diverge on any input. Full suite green (2011).

victorludvig (approved):

PR #123, opened 2026-08-18
feat(sdk): official secret-detection guards (hexgate.plugins)

quanghexa94:
PR3 of the guard line (after #118 runtime and #119 adapters): the first **official plugins**, so `guards=[...]` ships with real, tested guards that double as the usage examples.

## What's here

One shared secret detector, three guards built on it.

**Detector** — `hexgate/plugins/secrets.py`
- **Prefix-first**: high-confidence provider patterns (AWS `AKIA`/`ASIA`, GitHub `ghp_`/`github_pat_`, OpenAI/Anthropic `sk-`/`sk-ant-`, Slack `xox…`, Google `AIza…`, Stripe `sk_live_`, Hexgate `fty_`, PEM private keys).
- **Strict entropy fallback** only when no prefix matched: token-charset + min-length + Shannon threshold, excluding hex digests and UUIDs — so the fail-closed path stays low false-positive (a git SHA, a UUID, a URL, and prose all pass clean).
- `scan_secrets` / `redact_secrets` walk JSON-ish values (dict/list/str), reporting `SecretHit(category, field, fingerprint)` where the fingerprint is a truncated SHA-256 — **never the value**. `safe_reason` (model-facing) and `safe_detail` (operator-facing) name the category and field only.

**Guards** — `hexgate/plugins/__init__.py`
- `secret_guard` — before-guard, **halts** with an actionable, value-free reason.
- `secret_redactor` — before-guard, **strips** the secret from args (`[REDACTED:<category>]`) and `Proceed`s with a `Modification`.
- `secret_watch` — after-guard (observe), **flags** a leaked secret in a result via a value-free warning; never changes the result in v1 (becomes a scrubber when result rewrite lands).

```python
from hexgate import create_agent
from hexgate.plugins import secret_guard, secret_watch

agent, _ = create_agent(model=..., tools=[...], guards=[secret_guard, secret_watch])
```

## Decision record
`docs/adr/R-GUARD-005` — prefix-first, entropy-strict, value-free — with the precision-over-recall rationale for a fail-closed guard.

## Tests
36 plugin tests: per-provider patterns, a false-positive corpus, the entropy bar, redaction (spans + no mutation), the no-leak guarantee on reason/detail, and all three guards end-to-end through `run_guarded`. Full suite green, lint + format clean.

## Follow-ups (not in this PR)
- The docs **site** (`docs/*.mdx`) still doesn't cover guards at all — worth a `docs(sdk)` PR (a guards + plugins page).
- PII / email plugins (observe or redactor) — deferred per the design doc.
- Detector tuning via config/factory, and `secret_scrubber` once result rewrite lands.

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -24,11 +24,11 @@ tracks can be added as their own sections.
 - **PR2: the other three adapters.** Point OpenAI, Google ADK, and Pydantic AI at
   the shared `run_guarded` so all four frameworks share one guard path. Mostly
   deleting duplicated decide-then-invoke.
-- **PR3: official plugins (up as a PR).** `hexgate.plugins`: one prefix-first,
-  entropy-strict secret detector (`scan_secrets` / `redact_secrets`, value-free
-  `SecretHit`s) exposed as `secret_guard` (halt), `secret_redactor` (pre rewrite),
-  and `secret_watch` (post observe), plus a safe-`reason` builder. Detector policy
-  in `docs/adr/R-GUARD-005`.
+- **PR3: official plugins (up as a PR).** `hexgate.plugins`: one prefix-only
+  secret detector (`scan_secrets` / `redact_secrets`, value-free `SecretHit`s)
+  exposed as `secret_guard` (halt), `secret_redactor` (pre rewrite), and
+  `secret_watch` (post observe), plus a safe-`reason` builder. Detector policy in
+  `docs/adr/R-GUARD-005`; gated entropy detection is a later item.
 
 ## Later
 
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -24,9 +24,11 @@ tracks can be added as their own sections.
 - **PR2: the other three adapters.** Point OpenAI, Google ADK, and Pydantic AI at
   the shared `run_guarded` so all four frameworks share one guard path. Mostly
   deleting duplicated decide-then-invoke.
-- **PR3: official plugins.** `hexgate.plugins`: one secret detector exposed as a
-  guard (halt), a redactor (pre rewrite), and a watcher (post observe), plus a
-  safe-`reason` builder.
+- **PR3: official plugins (up as a PR).** `hexgate.plugins`: one prefix-first,
+  entropy-strict secret detector (`scan_secrets` / `redact_secrets`, value-free
+  `SecretHit`s) exposed as `secret_guard` (halt), `secret_redactor` (pre rewrite),
+  and `secret_watch` (post observe), plus a safe-`reason` builder. Detector policy
+  in `docs/adr/R-GUARD-005`.
 
 ## Later
 
--- a/docs/adr/R-GUARD-005-secret-detector-prefix-first.md
+++ b/docs/adr/R-GUARD-005-secret-detector-prefix-first.md
@@ -1,0 +1,58 @@
+# R-GUARD-005: The secret detector is prefix-first, entropy-strict, value-free
+
+**Status:** Accepted · 2026-08-18
+**Applies to:** `hexgate/plugins/**`
+
+## Decision
+
+The official secret detector matches **high-confidence provider prefixes first**
+(AWS `AKIA`/`ASIA`, GitHub `ghp_`/`github_pat_`, OpenAI/Anthropic `sk-`/`sk-ant-`,
+Slack `xox…`, Google `AIza…`, Stripe `sk_live_`, Hexgate `fty_`, PEM private-key
+headers) and falls back to a **strict Shannon-entropy** test only when no prefix
+matched: a leaf must be one contiguous token in the base64 charset, at least
+`_ENTROPY_MIN_LEN` long, over `_ENTROPY_THRESHOLD` bits/char, and **not** a plain
+hex digest or a UUID.
+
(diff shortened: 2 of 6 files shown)


guillaume-hexamind:
3 claude issues: 

1 — hexgate/plugins/secrets.py:43 — hexgate_token never matches a real token (90/100)

\bfty_[A-Za-z0-9]{16,}\b must start at fty_, and [A-Za-z0-9] can't cross the _
after the env segment. Real format is fty_<env>_<project_id>_<biscuit_b64>
(core/biscuits.py:226) with env constrained to test|live (schemas.py:249) — 4 chars,
so {16,} is unsatisfiable for every valid token. Verified 0 hits for both envs with
UUID- and slug-shaped project IDs. A live HEXGATE_API_KEY passes secret_guard and
secret_redactor untouched.
Tests miss it because PROVIDER_SAMPLES["hexgate_token"] (tests/plugins/test_secrets.py:26)
is synthetic filler "fty_" + "A" * 16.
Fix (validated: matches UUID + slug project IDs, no false positives):
("hexgate_token", re.compile(r"\bfty_(?:live|test)[A-Za-z0-9-]{1,64}[A-Za-z0-9_-]{20,}")),
Also replace the test sample with a real-shaped token.

2 — hexgate/plugins/secrets.py:109 — dict keys unscanned but interpolated into Halt.reason (82/100)

The walk recurses on values only, yet builds field via _join(_path, str(key)). Three
consequences:

Leak: {"AKIAIOSFODNN7EXAMPLE": {"token": "sk-ant-…"}} → reason contains
`AKIAIOSFODNN7EXAMPLE.token` — the AWS key goes back to the model, breaking the
module's "never surfaces a matched value" invariant.
Miss: {"AKIA…": "harmless"} → [], and unredacted.
Injection: segments are unescaped, so a key of a`\n\nIGNORE PREVIOUS INSTRUCTIONS.
reaches the model verbatim.
Fix: scan keys as leaves, redact them when rebuilding the mapping, and pass every path
segment through the redactor + a charset allowlist ([A-Za-z0-9_.\[\]-]) before it reaches
safe_reason / safe_detail.

3 — hexgate/plugins/secrets.py:45-51 — header-only PEM match forwards the key body while reporting success (78/100)

The END group is optional, so a truncated block matches the header alone. Confirmed on
all six real header variants (RSA, bare, EC, OPENSSH, DSA, ENCRYPTED):
{"key": "-----BEGIN RSA PRIVATE KEY-----\nMIIEow…KEYMATERIAL\n"}
-> "[REDACTED:private_key]\nMIIEow…KEYMATERIAL\n"
secret_redactor still emits Modification(summary="redacted 1 secret(s): private_key")
and proceeds, so the trail records a redaction that didn't happen. Two related bugs:
fingerprint is a constant for header-only matches (8bcac790… ← truncated in source),
and -----BEGIN PGP PRIVATE KEY BLOCK----- matches nothing.
Fix (validated on complete / truncated / truncated-with-trailer / PGP — no body leak,
surrounding text preserved) — on a header-only match, run to the end of the block: ← rest of sentence truncated in source
re.compile(
r"-----BEGIN (?:[A-Z0-9]+ )?PRIVATE KEY(?: BLOCK)?-----"
r"(?:[\s\S]?-----END (?:[A-Z0-9]+ )?PRIVATE KEY(?: BLOCK)?-----" ← line truncated in source
r"|[\s\S]?(?=\n[ \t]*\n|\Z))"
)

guillaume-hexamind (commented):
A couple of pretty minor issue IMO found by claude above, if you can check them. Otherwise the logic looks good, pretty simple but effective way to demo the hook functionality as of now!

quanghexa94:
All three fixed, thanks — good catches, #2 especially.

**1 — `hexgate_token` never matched a real token.** Right: `[A-Za-z0-9]{16,}` stops at the `_` after the env segment, so no real `fty_<env>_<project>_<biscuit>` ever matched. Now `\bfty_(?:test|live)_[A-Za-z0-9_-]{24,}` — the charset after the env includes the url-safe `_`/`-` the biscuit carries — checked against the format in `hexgate/cloud/biscuit.py`. The synthetic test sample is replaced with a real-shaped token.

**2 — dict keys unscanned but interpolated into the field path (leak / miss / injection).** Keys are now scanned as leaves, so a secret-as-key is detected and redacted on rebuild. And every field-path segment is sanitized before it reaches `safe_reason` / `safe_detail`: any secret in the key is redacted, then a charset allowlist (`[A-Za-z0-9_.:\[\]-]`) drops everything else and the segment is length-bounded. So a crafted key can't inject newlines/instructions into the refusal, and a secret key shows as `[REDACTED:…]`, never verbatim. Tests cover the leak, the miss, and the injection.

**3 — header-only PEM forwarded the key body.** The `END` group was optional, so a truncated block matched the header alone. The pattern now runs to `END` when present, else to the next blank line / end of string, and `(?: BLOCK)?` covers the PGP `PRIVATE KEY BLOCK` variant. Tests cover a truncated block (body redacted, trailing text preserved) and a PGP block.

Full suite green (2040 passed), lint + format clean.

(Also opened #125 — the docs-site guards + plugins page — which documents this module; worth merging this one first.)

guillaume-hexamind (approved):

PR #124, opened 2026-08-19
feat(sdk): agent-level admission enforcement (AgentGate, closed-world agent keys)

quanghexa94:
> **📚 Base of a 2-PR stack.** #142 (runtime reach + depth cap, A3+A4) is stacked on top of this one and targets this branch. Merge order: **this → then #142.**

---

Add agent-level admission enforcement: gate whether a caller, in their role, may run an agent at all, decided at run entry before the model sees anything. This is the original AgentGate PR plus the boundary/capability model reconcile it depends on, folded together into one coherent feature (the reconcile was briefly split out as #137, now folded back here). Pure SDK, no platform or dashboard change, no migration. It is the A2 step of the agent-level enforcement series (A1, the `admission`/`agents` lowering, is already merged); A3 (handoff/delegation interception) and A4 (depth cap) follow.

This PR originally used an opt-in admission model that review here showed had two fail-opens. Rather than patch them in the runtime, the fix is to reconcile agent-level rules with the boundary/capability module model from #115 (R-POL-001) and make agent keys closed-world.

## Design

Agent-level rules (`admission` / `agents`) lower to synthetic `agent.run` / `agent.tool:<t>` / `agent.handoff:<t>` keys that both engines already read through `effective_tools`. This PR makes those keys first-class in the linker fold, closed-world at the engines, and cleanly separates two things #124 had tangled: how rules compose vs. whether the gate is engaged at all.

- Agent keys fold through the linker, exactly like tool keys. A boundary agent-deny is authoritative and role-independent, a ceiling boundary makes an unlisted agent key ineligible, capabilities grant and union, a capability that denies an agent key is a `LinkError`. `admission` / `agents` become allowed module fields.
- Agent keys are closed-world. An unlisted agent key (admission and reach) denies regardless of `default_policy`, so a silent co-role can neither admit an unknown caller nor defeat an explicit deny. This drops `agent.run`'s old opt-in-allow (the `get_tool_policy` fallback and the per-role Rego rule), removing both fail-opens by construction.
- Engagement is opt-in and derived, separate from composition. `PolicyEngine.declares_admission()` / `declares_reach()` decide whether the gate fires: the pydantic engine derives it from whether any resolved role carries the key; the WASM bundle reads an `agent_gating` block recorded in the signed manifest at build time. An agent whose policy never mentions admission is never gated and runs exactly as before, which is what keeps closed-world from locking out agents that do not use the feature.
- `AgentGate` is the runtime, mirroring the egress `Gate`: it reuses `PolicyEnforcer.decide` on the `agent.run` key at run entry, folds approval, and fails closed with `AgentNotAdmittedError`. It is wired in `HexgateAgent` by `enforce_policy` alongside the binding (local, no control plane), threaded through `with_tools` rebuilds, and cleared on the unguarded path. Whether it refuses is decided per run from the current policy, so a hot-reload that adds or drops admission is honored on the next run.

Full rationale and rejected alternatives in `docs/adr/R-AGENT-002`.

## Important files

| File | What to look at |
| --- | --- |
| `hexgate/security/agent_gate.py` | the runtime gate: `check_admission[_async]` reads `declares_admission()` per run, decides `agent.run`, folds approval, fails closed |
| `hexgate/agents/factory.py` | how the gate is wired: attached in `enforce_policy`, threaded through `with_tools`, cleared on the unguarded/guards-only paths, checked at `ainvoke`/`astream_events` entry |
| `hexgate/security/linker.py` | agent keys fold through `effective_tools` in `_fold_tool` / `_tool_names` / `_reject_capability_denies`; `admission`/`agents` in `_MODULE_COMPOSABLE_FIELDS`; resolved policy via `AgentPolicy.resolved()` |
| `hexgate/security/policy.py` | `get_tool_policy` fallback collapsed: every unlisted agent key denies closed-world |
| `hexgate/security/rego.py` | removed the per-role `agent.run` opt-in rule; catch-all excludes exactly the reserved agent keys, not a broad `agent.` prefix |
| `hexgate/security/decision.py`, `policy_set.py`, `bundle.py`, `models.py` | `declares_admission()`/`declares_reach()` on the protocol (pydantic from `effective_tools`, WASM from the manifest); `AgentPolicy.resolved()` builder |
| `docs/adr/R-AGENT-002-...md` | the decision, why, rejected alternatives |
| `tests/security/test_agent_gate.py` | admission verdicts + the multi-role closed-world/permissive-union case |
| `tests/security/test_linker.py`, `test_agent_policy.py`, `test_rego_compile.py` | the behaviour flips: boundary agent-deny composes, capability agent-deny is a LinkError, unlisted `agent.run` denies, no per-role admit rule |

## Test plan

Ran the SDK suite against the hexanlp-demo env: `pytest tests/ -q` gives 2091 passed, 6 skipped (opt-in integration), 20 deselected. `ruff check` clean over `hexgate/`, `tests/`, and the demo; `ruff format --check` clean.

By hand I checked the semantics that changed: (1) a boundary `agents: {x: deny}` folds to a hard deny on `agent.handoff:x` in `effective_tools`, and a capability `agents: {x: deny}` raises `LinkError`; (2) an unlisted `agent.run` under `default_policy: allow` denies on both the pydantic and compiled-WASM paths (parity case in `test_closed_world_agent_parity_pydantic_vs_wasm`); (3) the golden Rego fixture lost exactly the three per-role `agent.run` allow rules and nothing else; (4) the multi-role gate: `admin` (granted) admits, `viewer` (no grant, gate engaged because `admin` declares admission) is refused, `contractor` (`admission: deny`) is refused, `[contractor, admin]` admits via the permissive union, `[viewer, contractor]` is refused. A reviewer catches a regression through the pydantic-vs-WASM parity tests, the golden snapshot, and the multi-role gate test.

## Try it

```
# marimo admission demo against an inline policy (support/billing admitted,
# contractor + unroled refused, closed-world):
uv run marimo edit deploy/agent_gate_demo.py

# the model + linker changes:
uv run pytest tests/security -q
```

## Notes

- No platform or dashboard change; the only manifest touched is the signed WASM bundle manifest (`build_signed_bundle`), which gains an `agent_gating` block. Older bundles without it read `declares_admission()` as False, which is safe (they predate agent-level policy).
- Supersedes #137 (the model reconcile, folded back into this PR). Supersedes the opt-in/closed-world half of R-AGENT-001; its engagement-flag and manifest mechanism survive, now uniform for admission and reach.
- Behaviour change to be aware of when authoring: once a policy declares `admission` on any role, admission is enforced for every caller (closed-world), so grant it on each role that should be admitted (e.g. `admission: allow` on `default` for broad access). An agent that declares no admission is unaffected.
- Follow-ups: A3 (handoff/delegation interception, which consumes `declares_reach()`) and A4 (depth cap). Design doc: `docs/adr/R-AGENT-002-agent-keys-compose-through-the-linker.md`.


tests/agents/test_factory.py
```diff
@@ -384,6 +384,73 @@ async def test_hexgate_agent_astream_raises_before_first_event_when_banned() ->
             async for _ in agent.astream_events({}, {}, version="v2"):
                 pass
```
victorludvig, line 387: 1. Ban-path test lost its assertion — 8/10
tests/agents/test_factory.py:385 — The new admission tests were inserted between the end of test_hexgate_agent_astream_raises_before_first_event_when_banned and its trailing assert graph.astream_event_calls == []. That assert now belongs to the new ..._refused_when_not_admitted test (line ~454), so the banned test no longer verifies the graph never ran. A regression letting a banned caller through astream_events would pass CI. Easy fix: re-add the assert to the ban test.

victorludvig (commented):
I approve the logic, can you check these Claude findings:

2. _apply_approval_handler never updates _agent_gate — 7/10
hexgate/agents/loader.py:347 — For registered/CLI agents, the loader re-wraps every GuardedTool with the CLI approval handler and rebuilds via with_tools, but with_tools threads the old _agent_gate (handler still None) forward. Result: an admission: approval_required policy always raises AgentNotAdmittedError even though the user explicitly wired an interactive handler. This is the most likely real-world bug in the PR.

3. enforce_policy(None) clears the gate but not the binding — 6/10
hexgate/agents/factory.py:567 (also the guards-only branch at ~6nt_gate but with_tools still forwards self._binding, so theand its trailing assert graph.astream_event_calls == []. That assert now belongs to the new ..._refused_when_not_admitted test (line ~454), so the banned test no longer verifies the graph never ran. A regression letting a banned caller through astream_events would pass CI. Easy fix: re-add the assert to the ban test.                                                                                                                                                        
2. _apply_approval_handler never updates _agent_gate — 7/10                                                                                                  hexgate/agents/loader.py:347 — For registered/CLI agents, the lool with the CLI approval handler and rebuilds via with_tools, but with_tools threads the old _agent_gate (handler still None) forward. Result: an admission: approval_required policy always raises AgentNotAdmittedError even though the user explicitly wired an interactive handler. This isug in the PR.

3. enforce_policy(None) clears the gate but not the binding — 6/10
hexgate/agents/factory.py:567 (also the guards-only branch at ~6nt_gate but with_tools still forwards self._binding, so the"unguarded" agent keeps hot-refreshing an enforcer that still-wrapped GuardedTools continue using. Contradicts the documented (None, None) → unguarded rebuild contract. Partly pre-existing, but this PR extends the asymmetry.

4. policy_declares_admission is False for bundles — 5/10
hexgate/security/agent_gate.py:56 — A WASM-bundle policy that dotly unenforced until the manifest-flag follow-up lands.Acknowledged in the PR body, but it's a silent fail-open in a component whose stated design is fail-closed. Consider a warning log or a declares_admission()
method on the PolicyEngine protocol so each engine answers for isinstance chain.

Design / consistency

5. AgentNotAdmittedError is a bare Exception — 5/10
agent_gate.py:33 — PolicyDeniedError, ApprovalRequiredError, ands RuntimeError (hexgate/security/errors.py). Code catchinghexgate refusals via except RuntimeError will miss this one. Tie-in to my earlier answer: raising is right, but it should join the existing hierarchy.

6. Misleading error message — 4/10
agent_gate.py:40 — Reuses Decision.as_error_message(), so refused callers see Tool 'agent.run' is denied ... The tool was not executed. when no tool was ever involved. Worth a dedicated message for the admission case.

7. Duplicated approval-handler normalization — 4/10
agent_gate.py:124 — hexgate/agents/approvals.py exists specifically so adapter copies "don't drift"; AgentGate adds another copy with different behavior (silent close+deny vs. RuntimeError on a coroutine returned to a sync site). If the fail-closed behavior is intentional, it should be one shared helper with a flag.

Efficiency / simplification

8. Unused sync path — 3/10
agent_gate.py:80 — check_admission and _resolve_approval_sync (~caller; HexgateAgent only uses the async variant. Either drop the sync path until the adapters need it, or note that it's forward-looking.

9. policy_declares_admission recomputed every turn — 2/10
agent_gate.py:45 — PolicySet.roles re-sorts the dict on each call; the boolean only changes when refresh_policy swaps the PolicySet object. Cheap to memoize by identity or compute once in PolicySet.__init__. Negligible ints.

10. Ban and admission checks run in series — 2/10
factory.py:404 — Independent checks awaited sequentially on every turn. Could gather, but you'd need to pick which exception wins; probably not worth it unless ban-feed latency is noticeable.

quanghexa94:
Ran a high self-review on this PR. Five findings — three are mechanics we'll fix, two are admission-semantics calls worth a decision.

**1. `enforce_policy(None)` doesn't fully unguard (`factory.py`).** Strongest correctness item. Clearing `_agent_gate` is only half: the tools stay `GuardedTool`-wrapped (`wrap` keeps the old enforcer when `enforcer=None`) and `with_tools` carries `_binding`, so a re-enforce to `None` still denies via the old policy and keeps hot-reloading. Largely pre-existing (the tool-rewrap / binding carry predates this PR; the PR added the `_agent_gate` half). Fix here: clear `_binding` too; the full tool-unwrap is a separate pre-existing `enforce_policy(None)` contract issue.

**2. [design] Opt-in admission admits unknown callers unless `default` denies (`policy.py`).** With admission opt-in-allow (R-AGENT-001), if the `default` role has no `admission` block, any unrecognized role — or a caller with no role — resolves to `default`, so `agent.run` falls through to opt-in ALLOW → admitted. An operator who secures named roles but forgets `admission: deny` on `default` leaves the agent open to unknown callers. This is the admission analog of the existing tool `permissive-default` exposure; the natural mitigation is a `permissive-admission` lint (deferred analyzer work) plus docs — unless we revisit the opt-in default.

**3. [design] `admission: deny` is defeated by any admission-silent co-role (`agent_gate.py`).** A caller carrying `[contractor(deny), viewer(no admission block)]`: the permissive multi-role fold (`ALLOW > NEEDS_APPROVAL > DENY`) plus viewer's opt-in ALLOW short-circuits, so contractor's explicit `deny` never takes effect. It's consistent with how tool multi-role folds (an allow in any role wins), but the sharp edge is that a *silent* role opt-in-grants admission, whereas a silent role does not grant a tool. So `admission: deny` blocks a caller only when none of their other roles are admission-silent.

#2 and #3 are the two faces of the opt-in-allow choice: it fixed the non-monotonic lockout (one role declaring admission locking out silent roles), but traded it for a weaker deny and a fail-open on silent/unknown roles. These are security-semantics decisions, not crashes — flagging for a call on whether opt-in-allow stays the default (mitigated by a lint) or admission should be closed-world (fail-closed, stronger deny, but the lockout returns).

**4. Approval-handler normalization is a third copy (`agent_gate.py`).** The same bool/sync/async fail-closed coercion now lives in `agents/approvals.py`, `egress/gate.py`, and here. Inlined to avoid a `security → agents` import cycle (egress inlines for the same reason), but worth hoisting a shared helper into `hexgate/approvals.py`.

**5. `AgentNotAdmittedError` message mislabels the refusal (`agent_gate.py`).** Reuses `Decision.as_error_message()` → "Tool 'agent.run' is denied … The tool was not executed," but admission is caller-facing and no tool ran. Needs a dedicated admission message.

Confirmed clean: the inheritance resolver threads `admission`/`agents` correctly, the un-decorated async tests aren't false-green (`asyncio_mode = auto`), and the apparent ban-test splice is a display artifact — the `assert` is intact.

quanghexa94:
Superseded by #137. Folded the AgentGate runtime from this PR together with the boundary/capability model reconcile it depends on into a single feature PR, and moved admission from the opt-in model to closed-world agent keys with derived engagement (R-AGENT-002), which fixes the two fail-opens raised in review here. All three commits from this branch are cherry-picked into #137 with the Option B adaptation on top.

winlp4ever:
Second high-effort review pass, scoped to this PR's changeset (`origin/main..d3db1b8`: the Option-B linker reconcile + AgentGate admission runtime + the non-enforcing-adapter warning). The core admission logic verified sound: closed-world engagement via `declares_admission`, pydantic-vs-Rego parity for the removed `agent.run` opt-in rule, hot-reload through the shared enforcer, and the multi-role permissive union. Five findings; after verification, none are must-fix, and two were overstated. Details:

**Not CI failures (the review over-reported these).** Two findings claimed `make fmt-check` / CI would fail on import ordering in `hexgate/security/__init__.py` and `hexgate/adapters/google/wrapper.py`. Checked against the repo config: there is no `[tool.ruff]` section and no `ruff.toml`, so ruff runs its **default** rule set, which does **not** include isort (`I`). `make lint` (`ruff check hexgate tests`) passes on both files, and `ruff format` does not reorder imports. The review reached its conclusion by running `ruff check --select I`, a stricter ruleset than this repo enforces. So the imports are merely not-alphabetical, not a build break. If we *want* import sorting enforced, the right fix is systemic, not these two files: add `I` to a `[tool.ruff.lint] select` and run `ruff check --fix` once across the tree.

**Real but pre-existing / low-priority:**
- `factory.py` `enforce_policy(None)` (empty-pipeline branch) clears `_agent_gate` but returns `with_tools(list(self.tools))`, which keeps the already-guarded tools and carries `_binding` forward — so tool gating and hot-reload persist while only admission is stripped. This is a genuine wart, but it's a pre-existing property of that passthrough branch (this PR only added the `_agent_gate = None` line), and the call pattern (`.enforce_policy(p1).enforce_policy(None)`) is unusual. Worth a follow-up to make the comment ("return unguarded") match reality, not a blocker here.
- `build_signed_bundle` parses the policy twice (`load_policy_set_from_dict(payload)` for the `agent_gating` flags, plus `compile_to_rego(payload)` internally). Minor producer-side efficiency; the resolved set could be reused. Low severity.
- `_admission_unenforced_warned` is an unbounded process-global dedup set keyed by `(framework, agent_name)`; a slow leak only under dynamically-generated agent names. Low severity.

Net: the admission/Option-B core is solid; the remaining items are minor efficiency/consistency and an optional lint-config decision. No changes pushed for this pass.

winlp4ever:
Followed up on the import-ordering finding: even though it isn't a CI blocker under the repo's current ruff config (no isort rule enabled), I sorted the imports so this PR is clean on its own — `refactor(sdk): sort imports in agent-level enforcement files` (ced9845), scoped to only the files this PR touches. Same applied to the stacked #142 (b0857c2). Both branches: `make lint` + `ruff format --check` clean, full suite green (1000 / 2131).

The tree-wide isort gap (~60 pre-existing unsorted spots in files outside these PRs) is untouched here — if we want import sorting enforced going forward, the durable fix is adding `I` to `[tool.ruff.lint] select` and running `ruff check --fix` once across the tree, which I'm happy to do as its own small PR.

quanghexa94:
Second high-effort review pass, scoped to this PR's changeset (`origin/main..d3db1b8`: the Option-B linker reconcile + AgentGate admission runtime + the non-enforcing-adapter warning). The core admission logic verified sound: closed-world engagement via `declares_admission`, pydantic-vs-Rego parity for the removed `agent.run` opt-in rule, hot-reload through the shared enforcer, and the multi-role permissive union. Five findings; after verification, none are must-fix, and two were overstated. Details:

**Not CI failures (the review over-reported these).** Two findings claimed `make fmt-check` / CI would fail on import ordering in `hexgate/security/__init__.py` and `hexgate/adapters/google/wrapper.py`. Checked against the repo config: there is no `[tool.ruff]` section and no `ruff.toml`, so ruff runs its **default** rule set, which does **not** include isort (`I`). `make lint` (`ruff check hexgate tests`) passes on both files, and `ruff format` does not reorder imports. The review reached its conclusion by running `ruff check --select I`, a stricter ruleset than this repo enforces. So the imports are merely not-alphabetical, not a build break. If we *want* import sorting enforced, the right fix is systemic, not these two files: add `I` to a `[tool.ruff.lint] select` and run `ruff check --fix` once across the tree.

**Real but pre-existing / low-priority:**
- `factory.py` `enforce_policy(None)` (empty-pipeline branch) clears `_agent_gate` but returns `with_tools(list(self.tools))`, which keeps the already-guarded tools and carries `_binding` forward — so tool gating and hot-reload persist while only admission is stripped. This is a genuine wart, but it's a pre-existing property of that passthrough branch (this PR only added the `_agent_gate = None` line), and the call pattern (`.enforce_policy(p1).enforce_policy(None)`) is unusual. Worth a follow-up to make the comment ("return unguarded") match reality, not a blocker here.
- `build_signed_bundle` parses the policy twice (`load_policy_set_from_dict(payload)` for the `agent_gating` flags, plus `compile_to_rego(payload)` internally). Minor producer-side efficiency; the resolved set could be reused. Low severity.
- `_admission_unenforced_warned` is an unbounded process-global dedup set keyed by `(framework, agent_name)`; a slow leak only under dynamically-generated agent names. Low severity.

Net: the admission/Option-B core is solid; the remaining items are minor efficiency/consistency and an optional lint-config decision. No changes pushed for this pass.

quanghexa94:
Followed up on the import-ordering finding: even though it isn't a CI blocker under the repo's current ruff config (no isort rule enabled), I sorted the imports so this PR is clean on its own — `refactor(sdk): sort imports in agent-level enforcement files` (ced9845), scoped to only the files this PR touches. Same applied to the stacked #142 (b0857c2). Both branches: `make lint` + `ruff format --check` clean, full suite green (1000 / 2131).

The tree-wide isort gap (~60 pre-existing unsorted spots in files outside these PRs) is untouched here — if we want import sorting enforced going forward, the durable fix is adding `I` to `[tool.ruff.lint] select` and running `ruff check --fix` once across the tree, which I'm happy to do as its own small PR.

quanghexa94:
Applied the review fixes and re-ran a high-effort pass. Summary of what changed (through `f0c698c`):

**Fixed from the first round of findings:**
- **Admission approval handler never reached the gate** — `loader._apply_approval_handler` re-wrapped the tools with the CLI handler but `with_tools` threaded the old `_agent_gate` (handler `None`) forward, so `admission: approval_required` on a registered/CLI agent always failed closed. It now rebuilds the gate with the handler on the same enforcer.
- **`AgentNotAdmittedError` now subclasses `RuntimeError`**, joining `PolicyDeniedError` / `ApprovalRequiredError` / `AgentBannedError`, so `except RuntimeError` catches it.
- **Admission-specific message** instead of reusing `Decision.as_error_message()`, which claimed a "tool" was involved when none was.
- **Ban astream test** regained its `assert graph.astream_event_calls == []` "graph never ran" coverage.

**Two more surfaced by the re-review, both now fixed:**
- **The approval-handler fix was incomplete** — it only rewired the gate in the branch where tools were re-wrapped, so a tool-less agent gated purely by admission still kept its `handler=None` gate. Now it rewires regardless of tools.
- **Ceiling-shadow fail-open (the more important one)** — a capability-declared `admission` grant that a ceiling boundary shadows away dropped `agent.run` from the resolved policy entirely, so `declares_admission()` read False and the gate silently disengaged (admit everyone). For an ordinary tool that drop is the implicit deny; for an agent key it flipped to fail-open. The linker now keeps a shadowed agent key as an explicit deny, so engagement stays on and the verdict is closed-world deny.

Each fix has a regression test. Full SDK suite green (2097 passed), `ruff check` + `ruff format --check` clean.

The re-review verified the closed-world reconcile, the pydantic/Rego parity, the `AgentPolicy.resolved` construct path, and the import re-sort are otherwise sound. Known-deferred items (agent-as-tool reach on OpenAI, agent-aware error rendering) are tracked for the stacked reach PR and R-AGENT-003; nothing else outstanding on this PR's surface.

hexgate/security/linker.py
```diff
@@ -344,12 +361,18 @@ def _fold_tool(
...
+            return BaseToolPolicy(mode="deny") if is_agent_key(tool) else None
 
     # 3+4. Capability grants. No grant → eligible but ungranted → implicit deny.
     grants: list[tuple[ModuleContent, ToolPolicy]] = []
     for cap in capabilities:
-        tp = cap.policy.tools.get(tool)
+        tp = cap.policy.effective_tools.get(tool)
         if tp is not None and tp.mode in GRANT_MODES:
             grants.append((cap, tp))
     if not grants:
```
victorludvig, line 417: Same fail-open as the shadow branch above. For agent keys, returning None here isn't an implicit deny: no capability grants agent.run → None → no "agent.run" entry in the resolved effective_tools → declares_admission() returns False → the gate never engages → every caller is admitted. So a boundary that declares admission with no capability grant ends up fully open instead of denied. Needs the same fix as the shadow branch: return BaseToolPolicy(mode="deny") if is_agent_key(tool) else None.

hexgate/security/models.py
victorludvig, line 1: The docstrings in models.py still describe the pre-R-AGENT-002 semantics and now say the opposite of what the engine does — worth updating in this PR since it changes the behavior they document:
- is_agent_reach_key (L77): "admission … is opt-in, so its absence admits" — no longer true, an unlisted agent.run is now a closed-world deny like reach keys; opt-in survives only as the gate's declares_admission() engagement check.
- is_agent_key (L86): "enforcement splits the two" — it no longer does, both are closed-world at the engine.
- AgentPolicy (L136): "falls to default_policy … the runtime gate refines that fallback in a later PR" — unlisted agent keys never fall to default_policy now (your own test_agent_run_is_closed_world_at_the_engine asserts this), and the "later PR" is this one.

hexgate/agents/loader.py
victorludvig, line 1: Docstring is stale after this PR's own fix: "so the caller knows the handler was silently dropped" only holds for the not touched and gate is None early return now — when there's an admission gate, the handler is applied (re-stamped onto the gate below). Worth rewording so the docstring matches the new behavior; the inline comment and warning text already do.

docs/adr/R-AGENT-002-agent-keys-compose-through-the-linker.md
```diff
@@ -0,0 +1,43 @@
...
+**Applies to:** `hexgate/security/policy.py`, `hexgate/security/rego.py`, `hexgate/security/linker.py`, `hexgate/security/policy_set.py`, `hexgate/security/bundle.py`, `hexgate/security/agent_gate.py`
+
+## Decision
+
+Agent-level rules (`admission` / `agents`, lowered to `agent.run` / `agent.tool:<t>` / `agent.handoff:<t>` keys) MUST compose through the same model as tools (R-POL-001), not a separate runtime fold.
+
+- **Agent keys fold through the linker.** `_fold_tool` and `_tool_names` MUST operate over `effective_tools`, so an `agent.*` key composes exactly like a tool key: a boundary deny wins absolutely and is role-independent; a ceiling boundary makes an unlisted key ineligible; capabilities grant and union; a capability that denies an agent key is a `LinkError`, same as any tool. `admission` / `agents` become allowed module fields.
+- **Agent keys are closed-world.** An unlisted `agent.*` key (admission *and* reach) MUST deny, regardless of `default_policy`. This replaces R-AGENT-001's `agent.run` opt-in-allow: a role that is not granted admission is denied, so a silent co-role can neither admit an unknown caller nor defeat an explicit deny.
```
victorludvig, line 11: R-AGENT-001 is cited here (and on the "replaces"/"supersedes" lines below) but doesn't exist in docs/adr/ or anywhere in git history — every other ADR reference in the repo resolves to a committed file. If it's the Notion/design doc from #122, say so explicitly ("R-AGENT-001 (Notion: …)"), or commit it alongside this one; as written the supersedes-chain can't be verified.

quanghexa94:
Thanks @victorludvig — all four addressed in `bae1844` (and carried into the stacked #142):

- **`linker.py:378` — the twin fail-open.** You're right, the no-grants branch dropped an ungranted agent key exactly like the shadow branch, so a boundary-declared admission with no capability grant left `agent.run` absent and disengaged the gate. Applied the same fix there: `return BaseToolPolicy(mode="deny") if is_agent_key(tool) else None`, with a regression test (`test_admission_declared_with_no_grant_stays_a_deny_not_dropped`) alongside the shadow-branch one. Good catch — my earlier fix only closed one of the two branches.
- **`models.py` docstrings.** Reworded `is_agent_reach_key`, `is_agent_key`, and the `AgentPolicy` block to the shipped R-AGENT-002 model: both admission and reach are closed-world at the engine, and opt-in survives only as the `declares_admission()` / `declares_reach()` engagement check, not an admit-on-absence fallback.
- **`loader.py` docstring.** Reworded — the "handler silently dropped" line is gone; it now says the no-GuardedTool warning is about *tool* approval prompts, and that admission approval still works because the gate is re-stamped.
- **`R-AGENT-001` dangling reference.** Added `docs/adr/R-AGENT-001-agent-gate-defaults.md` (Status: Superseded by R-AGENT-002) recording the original opt-in decision, so the supersedes-chain resolves to a committed file.

Full suite green (2098 on this PR), `ruff check` + `ruff format --check` clean.

On the shape of this round: finding 1 was the only substantive code change — the other three converge to doc/ADR consistency (docstrings and the ADR chain trailing the behavior this PR ships), no logic touched. So I think this has settled. Could you give it one more pass when you get a chance? If it looks good I'll merge.

hexgate/security/models.py
```diff
@@ -173,6 +178,29 @@ def _reject_reserved_tool_names(
                 )
         return value
 
+    @classmethod
+    def resolved(
```
0vl0, line 191: `AgentPolicy.resolved()` bypasses `_reject_reserved_tool_names` via `model_construct` (`models.py:182`), so a resolved policy dumps `tools: {agent.run: …}` that no longer re-validates — breaking the `hexgate policy resolve -o` → `hexgate policy build` round-trip that `cli/policy/main.py:618` promises, for any project declaring `admission:`/`agents:`.

`test_resolved_policy_parity_pydantic_vs_wasm` hits this exact path (`compile_to_rego` → `load_policy_set_from_dict`, `rego.py:123`) but stays green because `floor_bundle` has no agent-level blocks; adding an `admission:` block to a linker fixture reproduces it without OPA.

hexgate/security/linker.py
```diff
@@ -276,21 +281,33 @@ def _reject_file_scope(
 # rejected by _reject_unsupported_module_fields, so a field added to AgentPolicy
 # later fails closed here instead of being silently dropped by _fold_tool.
 _MODULE_COMPOSABLE_FIELDS = frozenset(
-    {"version", "inherits", "is_mixin", "default_policy", "tools", "consts"}
+    {
+        "version",
+        "inherits",
```
0vl0: inherits: does nothing in module files (policies/boundaries/ and policies/capabilities/) — a module that inherits a policy silently gets none of it, not even an error for a base name that doesn't exist, because the linker allowlists the field (linker.py:286) but never resolves it. It works as documented in the policy-set path (policy_set.py:313), so the same YAML behaves differently depending on where the file lives.

0vl0 (changes requested):
Thanks for the revision. Two verified issues to fix, plus 3 optional:

1. tests/security/test_rego_compile.py:379 — assert all("agent.run" in rule for rule in _allow_rules(rego)) is vacuously true now that the rule list is empty; lines 399 and 517 filter for a rule that can no longer be emitted.

2. hexgate/security/agent_gate.py:90 — AgentNotAdmittedError's docstring promises the message never mentions a tool, but it interpolates decision.reason, which reads Policy denied tool "agent.run".

3. deploy/agent_gate_demo.py:188 — _admit is underscore-prefixed, so it's cell-local in marimo and section 4 dies with NameError when the demo opens.

quanghexa94:
@victorludvig — heads-up on what changed since your approval, since `main` moved under this PR and it needed a non-trivial reconcile. The logic you reviewed is intact; this explains the one genuinely new design decision plus the mechanical churn.

## What moved

`#121` (R-POL-002, multi-module policy for *all* agents) landed on `main` while this was in review. It generalizes the modular boundary/capability/role model to the agent dimension, and — importantly for us — it makes a modular agent compile its bundle from the **resolved, role-keyed policy**: `resolve → inline-roles YAML → build_signed_bundle`. I rebased this branch onto it, so the history was rewritten (force-push, now at `23c69c9`) — same commits you saw, replayed on the new base. The linker changes auto-merged cleanly (your closed-world fold and #121's agent-dimension fold coexist); `__init__.py`, `factory.py`, and `openai/runner.py` were resolved by hand to keep both sides.

## The one real new decision: resolved policies couldn't round-trip (finding A)

This is the interesting one, and it only surfaced *because* of #121's compile path. The chain is: original authored policy → linker resolves it → the resolved policy carries the lowered `agent.*` keys (`agent.run`, `agent.tool:x`, `agent.handoff:x`) **in `tools`**, not in `admission`/`agents`. That's deliberate — per-`via` divergence across capabilities (a target allowed via one mode, denied via another, different constraints per mode) can't always be reverse-lowered into a single `AgentTargetPolicy`, so the fold stores them flat.

The problem: to build the signed bundle, #121 serializes that resolved policy to YAML and **re-parses it** through the same loader — which runs the strict `_reject_reserved_tool_names` validator we added to keep authors from hand-writing `agent.*` tool keys. So the machine-resolved artifact tripped a guard meant for humans, and the modular compile path (`hexgate policy build` **and** the platform recompile) failed.

The fix is a `RESOLVED_POLICY_MARKER = "_resolved"` that the resolve serializer stamps onto the dump. The loader validates a marked document under a `{"resolved": True}` pydantic context, and the reserved-name validator skips its check under that context. Hand-written source has no marker, so the authoring guard still fires exactly as before — the guard is an authoring ergonomic, not a security boundary (the engine reads `effective_tools` either way), so scoping it to non-machine input is safe. Regression tests cover both directions (marked round-trips, unmarked still rejected), in both the role-keyed and flat shapes.

## The other re-review findings (a second pass from @0vl0)

- **B — `inherits:` in a module silently did nothing.** `inherits`/`is_mixin` were allow-listed as module-composable, but the module fold composes by *role binding* and never resolves a module's own base — so a boundary/capability declaring `inherits:` got none of it, no error. Dropped them from the allow-list so it fails loud (like `file_scope` already does); `inherits` still works in the single-file policy path.
- **C — vacuous rego-compile asserts.** Post-Option-B, no `agent.run` allow rule is emitted unless a policy declares `admission`, so `all("agent.run" in rule ...)` was `all()` over an empty list — always true. Replaced with the real allow-rule set.
- **D — admission error leaked the synthetic key.** `AgentNotAdmittedError` interpolated `decision.reason`, which is phrased against the lowered `agent.run` key (`tool 'agent.run' is denied…`) — contradicting the class's own promise never to mention a tool. Dropped the interpolation (the `Decision` is still attached for inspection); applied the same to `ReachNotAllowedError` on the stacked PR.
- **E — marimo demo NameError.** `_admit` → `admit` (marimo treats a leading underscore as cell-local, so the consuming cell couldn't see it).

## Stacked #142 + a fresh review pass

I rebased #142 (agent reach + handoff-depth cap) onto this reconciled base and ran another high-effort review over the whole stack. It confirmed the invariants hold (pydantic↔Rego parity on the closed-world keys, identical agent-name canonicalization on both the own-lookup and reach-target sides, name-preserving `model_copy`/`replace` so the governed-root check can't fail open). Three consistency fixes came out of it and are on #142: `HandoffDepthExceededError` now subclasses `RuntimeError` like its siblings; the two reach-unenforced warnings got separate dedup sets; and both handoff seams short-circuit on `declares_reach()` before allocating a `ReachGate`.

Full SDK suite (2202) + platform-api (122) green, ruff lint + format clean. Nothing in your reviewed logic changed semantically — happy to walk through any of the merge resolutions if useful. Good to re-approve / merge from your side?

victorludvig (approved):
Approved.
As a follow-up I suggest adding one platform test that resolves and compiles a modular policy containing an admission: block — this exact path broke in 23c69c9 and nothing in platform/api/tests would catch it breaking again (tracked in #171).

PR #125, opened 2026-08-19
docs(sdk): guards + official-plugins concept page

quanghexa94:
Fills the docs-site gap: guards have shipped (runtime #118, adapters #119) but the published docs never covered them.

## What's here
A new **`concepts/guards.mdx`** page:
- The guard system — the four-step pipeline (before-guards → decide → tool → after-guards), observe / rewrite-args / halt, and the security rule that before-guards run *before* `decide`.
- Authoring — `@before_tool` / `@after_tool`, the flat `guards=[...]` list, `tool_names=` scoping, `observe=`, and what a guard returns/receives.
- Halt-message safety (`reason` vs `detail`, `retryable: false`) and the fail-closed vs observe error tiers.
- Installing guards across every adapter (`guards=` / `HexgateRunner(guards=)`), with a note that the only `hooks=` on the surface is the framework SDK's `RunHooks` on `run`.
- The **official plugins** (`hexgate.plugins`): `secret_guard` / `secret_redactor` / `secret_watch`, and the prefix-only detector rationale (`R-GUARD-005`).

Wired into the **Concepts** nav (after policy-decision) and cross-linked from the **adapters overview**.

## Merge order
The official-plugins section documents `hexgate.plugins`, which lands in **#123** — merge that first (or together) so the published docs don't reference code that isn't in `main` yet. Everything else on the page is already in `main`.

### DIFF
--- a/docs/adapters/overview.mdx
+++ b/docs/adapters/overview.mdx
@@ -19,6 +19,10 @@ logic:
    active [`HexgateContext`](/concepts/user-scope)'s identity (user id, session id, role)
    propagated onto the spans.
 
+Every wrapper also takes [`guards`](/concepts/guards) — before/after-tool functions
+that observe, rewrite arguments, or refuse a call — installed with `guards=[...]`
+(or `HexgateRunner(guards=[...])` on the OpenAI and Google runners).
+
 The four integrations differ in shape because the underlying SDKs do:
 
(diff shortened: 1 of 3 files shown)


guillaume-hexamind (approved):
Good to me, to merge after 123

PR #111, opened 2026-08-06
build(collector): scaffold the OTLP ingestion Collector (Go)

victorludvig:
Design doc: [OpenTelemetry migration design](https://app.notion.com/p/3b2fb45dbae2803ca64adb866704d69b) — row 2 ("Collector scaffold") from the "PR Plan (Design 2)" table.

## What is Changing
- `platform/collector/builder-config.yaml`: an `ocb` (OpenTelemetry Collector Builder) manifest wiring the dual-protocol `otlp` receiver (gRPC + HTTP) through the stock `memory_limiter`/`resource`/`attributes`/`batch` processors to the `kafkaexporter`, with a stubbed `bearertokenauth` extension standing in for the real Biscuit auth (separate PR). All components pinned to the v0.158.0 release train.
- `platform/collector/config.yaml`: the runtime wiring — both listeners, the auth stub, and `kafkaexporter` partitioning traces by `project_id` via `message_key_from_metadata_key` + a `sticky_key` record_partitioner. That key is only ever populated once the real auth extension sets it in client metadata, so today every span falls back to random partitioning.
- `main.go`/`components.go`/`main_others.go`/`main_windows.go`/`go.mod`/`go.sum`: generated by running the builder against the manifest above — see the reviewer note below.
- `Makefile`: `collector-install-builder`, `collector-generate`, `collector-run`.

## Why is this change necessary
Proves the ingestion pipe works end-to-end (receive → auth-gate → process → land on the queue) before the hard part — the real Biscuit auth extension — gets built around it.

## Note for reviewers: first Go code in this repo
Of the 8 files this PR adds under `platform/collector/`, only **2 are actually hand-written**:
- `builder-config.yaml` — which components are compiled into the binary, and at which version.
- `config.yaml` — how those components are wired together at runtime (listeners, auth, the Kafka partition key).

The other 6 are 100% generated by running the `ocb` builder tool against those two files (`make collector-generate`), not hand-written by me or anyone:
- `main.go`, `components.go`, `main_others.go`, `main_windows.go` — Go source; the first two open with `// Code generated by "go.opentelemetry.io/collector/cmd/builder". DO NOT EDIT.`, a standard Go convention that's literally true here.
- `go.mod`, `go.sum` — dependency manifests produced by the builder's own `go mod tidy` step.

[...]

PR #127, opened 2026-08-20
feat(collector): verify Hexgate API keys in the OTLP receiver

victorludvig:
## What is changing

Replaces the collector scaffold's stock `bearertokenauth` placeholder with a real authenticator, `hexgatebiscuitauth` (PR 3 in the OpenTelemetry migration design doc's own numbering, not a PR in this repo). Per request: split the `fty_` envelope → verify the Ed25519 signature chain → run the Datalog authorizer so the token's own checks (TTL caveat) are enforced → read the authority block's facts → resolve `token_id` against a polled snapshot of `devtoken` → stamp `project_id` and `token_id` onto the request's client metadata.

Also adds opt-in integration tests, and Makefile targets `collector-test` / `collector-test-integration`. `collector-check` now covers the two nested Go modules, which `./...` from `platform/collector` cannot see.

The scaffold accepted a single static shared token, so nothing verified real API keys. And nothing produced `project_id`, which `config.yaml`'s batch `metadata_keys` and kafka `message_key_from_metadata_key` were already configured to consume — until now that key was simply absent and every span fell back to random partition assignment.

**Builds on #126 (merged), which made the signed `token_id` fact the api-key row's own id.** That is what lets the revocation cache be keyed on the signed `token_id` rather than on the full secret as the design doc originally specified. The payoff is that the Collector never holds credentials in memory (`secret` is never selected).

### Reviewer notes

- **Where to review: three files carry every trust decision, ~750 lines of the ~3,900-line diff.** In priority order: `extension.go` (320) is the auth boundary — credential extraction, the `Authenticate` flow, and what gets stamped onto or stripped from request metadata; an authentication bypass would live here. `verifier.go` (231) is the cryptography — the Ed25519 signature chain, the Datalog authorizer, and the authority-block-only fact reading that keeps attacker-appendable blocks untrusted. `cache.go` (197) is revocation — the fail-closed lookup and staleness rules. Everything else is support: `envelope.go`/`keyring.go`/`config.go`/`factory.go` are small input-parsing and wiring, ~1,900 lines are tests, and the remainder is go.mod/go.sum, Makefile, and config wiring.
- **A newly minted API key is rejected for up to `poll_interval` (20s).** The cache is a full-table snapshot and a lookup miss is indistinguishable from a revocation, so a just-created key 401s until the next poll. Accepted, not fixed: the human path — copy the key, configure the SDK, start the agent — comfortably exceeds 20s. It does matter for automated mint-then-immediately-use, i.e. the design doc's PR 6 end-to-end test; handle it with a retry there rather than with machinery here.
- The same staleness fails the other way for revocation: a revoked key keeps working for up to one poll interval. The design doc names and accepts that.
- `max_staleness` (2m) is **not** in the design doc. It rejects all traffic once a snapshot goes un-refreshed for that long, so a Postgres outage cannot silently freeze the revocation list instead of failing closed.
- **Two abuse caps, also not in the design doc**, added after a security review pass: a credential over 8 KiB is rejected before any decoding, and a token carrying more than 8 appended blocks is rejected before the signature chain is walked. Attenuation needs no signing key and each block costs an Ed25519 verification plus a Datalog run — work spent even on revoked or expired tokens, which are only rejected afterwards — so without the caps a token's size and depth are chosen by whoever holds one. A minted key is a few hundred bytes and one authority block, so neither cap can touch legitimate traffic.
- **Edge cases sharpened after the same review**: a control-plane outage surfaces to gRPC clients as retryable `UNAVAILABLE` rather than terminal `UNAUTHENTICATED` (HTTP stays 401 — confighttp hardcodes it); a request carrying more than one bearer credential is rejected rather than authenticated by whichever value is found first; and with revocation disabled, a validly-signed token with no `project` fact is rejected rather than stamped with an empty partition key.
- **Fixes from an independent second review**: every key-table read has a 10s timeout, so a black-holed Postgres connection can no longer wedge the poll goroutine permanently (previously: outage outlived the database's recovery, until restart); `metadata_cardinality_limit` is set explicitly with a comment naming it a hard ceiling on lifetime projects per process run — the batch processor's per-`project_id` batcher map never evicts and its default (1000) rejects new projects permanently once hit; the cannot-verify Error logs once per outage instead of per request; the constant Datalog policy/queries are parsed once at init; and `make collector-run` now provides the Postgres/schema/key prerequisites the authenticator made fatal.
- **Config validation refuses the committed dev password on any non-local host** — the collector-side twin of `settings.py`'s `_refuse_dev_password_on_remote_host`, closing the guard PR #111's review deferred to this PR. The DSN is parsed with the same pgconn parser pgxpool connects with, so what is vetted is exactly what would connect, and a malformed DSN now fails at validation instead of at startup.
- The table is `devtoken`, not `apikey` as the doc says — the model was renamed but kept its `__tablename__` to avoid a migration.
- Facts are read from the **authority block only**. `attenuate_token` needs no signing key, so any holder can append blocks to a token that still verifies; honouring an appended `token_id` would hand them the revocation lookup and the choice of which project spans are attributed to.
- The `authorization` header is stripped from the metadata propagated downstream — `include_metadata: true` would otherwise carry the API key into the pipeline.
- Config requires an explicit root public key and enables revocation by default. `revocation.enabled: false` exists for local dev without Postgres and logs a loud warning.

### Found while testing, not fixed here

The envelope is split on its first three underscores, so a project id containing an underscore truncates the project and corrupts the base64 payload, making every token for that project unparseable. This is safe today only by accident: `Project` is absent from `ids.py`'s `_ID_PREFIXES`, so `projects/service.py` uses `uuid4`. Adding `Project: "prj"` there would silently break minting. Worth a guard in `make_envelope`.

## Tests

- **77 unit tests** in the extension module, race-clean. Cover signature failure, a corrupted key, TTL expiry, malformed envelopes and payloads, config validation (including the dev-DSN guard), the size and block-count caps at and past their limits, cache staleness and poll-failure recovery (asserting the gRPC status is retryable), ambiguous and missing credentials, and credential extraction across HTTP and gRPC casing. Two are security-specific to attenuation: an appended block forging `token_id`/`project` must lose to the authority block, and an appended narrowing check must still be enforced. The forgery test asserts the forged fact really is in the token so it cannot pass vacuously.
- **The cross-language guarantee is pinned by a committed fixture.** `testdata/python_minted/` holds a token minted through the real platform code path (`core/biscuits.py`, regenerable via the committed `mint_fixture.py`; throwaway keypair, private half never persisted) plus the raw 32-byte public key in `keystore.py`'s on-disk format. `TestVerifyBiscuit_when_token_was_minted_by_the_python_platform_then_it_verifies` splits the envelope, parses the key through the keyring path, verifies the token, and asserts every authority fact survived — so a wire-format divergence between biscuit-python and biscuit-go breaks `collector-check`, not production.
- **5 integration tests** (`make collector-test-integration`) drive the built binary against real Postgres and Redpanda: valid key → 200 with the Kafka record keyed by `project_id`; revocation taking effect after a poll; unknown `token_id`, wrong signing key, and missing credential all rejected with nothing reaching the exporter. They run the committed `config.yaml` and patch only ports, topic and poll interval through a second `--config` source, so they exercise what actually ships.
- Confirmed the main integration assertion can fail: commenting out `metadata_keys: [project_id]` turns `happy_path` red, so it guards the three-way agreement between extension, batch processor and exporter rather than restating it.
- `make check-all` passes. Integration tests are behind `-tags integration`, so `collector-check` stays hermetic.

Manually verified end to end against a real Python-minted key from `deploy/provision.py`: 200 with the correct partition key, revocation after one poll, and `max_staleness` failing closed ~33s after stopping Postgres. The dev-DSN guard was also exercised through the built binary: `validate` refuses the committed password pointed at a non-local host.


platform/collector/extension/hexgatebiscuitauth/extension.go
```diff
@@ -0,0 +1,333 @@
...
+			a.logger.Error("refusing traffic because the revocation list cannot be trusted; "+
+				"this logs once per outage, not per rejected request", zap.Error(err))
+		}
+		return "", errUnavailable
+	case err != nil:
+		a.logger.Error("unexpected failure resolving an API key", zap.Error(err))
+		return "", errUnavailable
+	}
+
+	if a.unavailable.CompareAndSwap(true, false) {
```
quanghexa94: **Outage flag can stick `true` and suppress later logging**

The `unavailable` flag is only reset on a successful *known-key* lookup. The early `return` in the `errUnknownAPIKey` case (line 182) never reaches this `CompareAndSwap(true, false)` reset.

So after a DB recovers, if post-recovery traffic is dominated by revoked/unknown keys, the flag stays `true`. A subsequent genuine stale/not-loaded outage then hits `CompareAndSwap(false, true)` (line 195), which fails because the flag is already `true` — so that second outage is never logged. The "log once per outage" guarantee is silently lost.

platform/collector/extension/hexgatebiscuitauth/extension.go
```diff
@@ -0,0 +1,333 @@
...
+		if !strings.EqualFold(name, authorizationHeader) {
+			continue
+		}
+		for _, value := range values {
+			if credential := stripBearerScheme(value); credential != "" {
+				credentials = append(credentials, credential)
+			}
+		}
+	}
+	switch len(credentials) {
```
quanghexa94, line 261: **Identical duplicate Authorization values rejected as ambiguous**

A proxy or gRPC client that sets the Authorization header/metadata twice with the *same* envelope produces two identical entries in `credentials`; `len(credentials) == 2` falls to the default branch and returns `errAmbiguousCredential` → 401, rejecting a client whose credential is actually unambiguous.

Deduplicating identical values before the count would avoid the false rejection while still rejecting genuinely conflicting ones.

platform/collector/extension/hexgatebiscuitauth/config.go
```diff
@@ -0,0 +1,167 @@
...
+		if !isLocalHost(host) {
+			return fmt.Errorf(
+				"revocation.dsn points at non-local host %q but still carries the committed "+
+					"dev password — set HEXGATE_COLLECTOR_POSTGRES_DSN to a real credential", host)
+		}
+	}
+	return nil
+}
+
+func isLocalHost(host string) bool {
```
quanghexa94, line 172: **`isLocalHost` misses empty host and the wider 127.0.0.0/8 range**

A DSN carrying the committed dev password but omitting the host (e.g. `postgres://hexgate:hexgate-dev-password@/hexgate`, where pgconn resolves an empty/default host) or using a loopback IP like `127.0.0.2` makes `isLocalHost` return `false`, so `refuseDevPasswordOnRemoteHost` errors with `points at non-local host ""` and refuses a genuinely local configuration.

Fail-safe (it refuses rather than connects), but it blocks a valid local setup.

platform/collector/integration/harness_test.go
```diff
@@ -0,0 +1,617 @@
...
+	return "fty_live_" + fixture.projectID + "_" + base64.URLEncoding.EncodeToString(serialized)
+}
+
+// ---------------------------------------------------------------------------
+// Misc
+// ---------------------------------------------------------------------------
+
+// freeAddress asks the kernel for an unused port so parallel runs and a
+// developer's own collector on 4317/4318 cannot collide.
+func freeAddress(t *testing.T) string {
```
quanghexa94: **TOCTOU window in `freeAddress` can flake the collector startup**

`freeAddress` opens a `:0` listener, records the address, then defers `Close` and returns; `startCollector` later binds that address. Between the close and the collector's bind, another process (or the second `freeAddress` call for the gRPC port) can grab the port, so the collector fails to start and the test flakes intermittently — especially under parallel runs.

platform/collector/extension/hexgatebiscuitauth/keyring.go
```diff
@@ -0,0 +1,79 @@
...
+		base64.StdEncoding,
+	} {
+		decoded, decodeErr = enc.DecodeString(text)
+		if decodeErr == nil {
+			break
+		}
+	}
+	if decodeErr != nil {
+		return nil, fmt.Errorf(
+			"expected either %d raw bytes or a base64-encoded key, got %d bytes that do not decode: %w",
```
quanghexa94, line 70: **Misleading byte count in decode-failure error**

When a base64 public key fails to decode, the error reports `len(raw)` (the untrimmed input, including any surrounding whitespace/newline) rather than `len(text)` (what was actually decoded). An operator debugging a bad `public_key` sees a byte count that doesn't match what was parsed. Cosmetic, but misleading during misconfiguration triage.

platform/collector/extension/hexgatebiscuitauth/extension.go
0vl0: Correct, I added a.markAvailable() in the errUnknownAPIKey switch case to make sure the flag is reset in the case of a db recovery.

Also added TestAuthenticate_when_recovery_is_seen_only_through_revoked_keys_then_the_next_outage_is_still_logged,which walks stale → revoked key → stale and asserts two Errors; it fails with 1 without the fix.

platform/collector/extension/hexgatebiscuitauth/config.go
0vl0: 1) When there is no host, pgx (GO's psql driver) defaults to a filepath, see https://github.com/jackc/pgx/blob/7293fb11125be0373a92f716683f2d494f6fd4b0/pgconn/defaults.go#L49-L63, and the call site https://github.com/jackc/pgx/blob/7293fb11125be0373a92f716683f2d494f6fd4b0/pgconn/defaults.go#L15. Therefore it starts with "/" and the Strings parser succeeds.

2) Correct for the wider loopback 127.0.0.0/8 range, I replaced the switch with a parse:
if host == "localhost" {
    return true
}
if ip, err := netip.ParseAddr(host); err == nil {
    return ip.IsLoopback()
}

platform/collector/extension/hexgatebiscuitauth/extension.go
0vl0: Good catch — deduplicated identical values before the count, so the same envelope twice is one credential while conflicting values are still rejected. Added TestAuthenticate_when_the_same_credential_is_supplied_twice_then_the_request_is_accepted.

platform/collector/extension/hexgatebiscuitauth/keyring.go
0vl0: Good catch. The log now reports len(text) instead of len(raw), so the count matches what was handed to the decoders.

platform/collector/integration/harness_test.go
```diff
@@ -0,0 +1,617 @@
...
+//
+// Driving the committed config rather than a fixture is the point: the
+// properties under test live in that file's processor chain and exporter
+// settings, so a hand-written test config would verify nothing about what
+// actually ships.
+func startCollector(t *testing.T, opts collectorOptions) runningCollector {
+	t.Helper()
+
+	httpEndpoint := freeAddress(t)
+	grpcEndpoint := freeAddress(t)
```
0vl0: Fixed — freeAddress became reserveAddress, which holds the listener until startCollector releases it just before cmd.Start(). Also removes a self-collision: the second call used to ask for a port after the first had already been released, so the kernel could hand back the same one.

Small window remains: the parent has to release before Start(), but the child only binds after it boots and loads its config, so the ports are unowned in between. Can't be closed without handing the child an already-bound socket, which the OTLP receiver can't accept. Losing is loud (address already in use, logged by waitUntilReady), so no retry for now.

platform/collector/integration/harness_test.go
0vl0: Fixed — freeAddress became reserveAddress, which holds the listener until startCollector releases it just before cmd.Start(). Also removes a self-collision: the second call used to ask for a port after the first had already been released, so the kernel could hand back the same one.

Small window remains: the parent has to release before Start(), but the child only binds after it boots and loads its config, so the ports are unowned in between. Can't be closed without handing the child an already-bound socket, which the OTLP receiver can't accept. Losing is loud (address already in use, logged by waitUntilReady), so no retry for now.

quanghexa94 (approved):

PR #128, opened 2026-08-20
feat(platform-api): add ClickHouse batch-insert functions for the enricher job

victorludvig:
## What is changing

Adds three ClickHouse batch-insert functions: `insert_decisions_batch` and `insert_ban_enforcements_batch` in `features/audit/service.py`, and `insert_llm_invocations_batch` in `features/llm_invocations/service.py`. These are the write path the upcoming span-enricher job will call — nothing calls them yet, so this PR is functions plus tests only.

The existing single-row functions (`insert_decision`, `insert_ban_enforcement`, `insert_llm_invocation`) had their row-building extracted into shared helpers so the single and batch paths cannot drift apart. Their behavior is byte-for-byte unchanged: same row shape, same payload-cap boundary and check order (`arguments` → `hint` → `attributes`), same tables, same insert settings.

The enricher job needs to turn each Kafka poll (our broker is Redpanda — Kafka-wire-compatible, so "Kafka" throughout means the protocol) into one multi-row `clickhouse_client.insert()` instead of hundreds of single-row calls. These functions didn't exist anywhere yet — only the single-row inserts did — and they're pure Python with no Go/Kafka dependency, so they land as their own PR, independent of the Collector work (#127).

## Reviewer Notes

- **The live single-row paths are refactored but behavior-identical** — verified line-by-line against `main` by an adversarial review pass, plus an empirical boundary probe: exactly-8192-byte arguments insert, 8193 raises `AuditPayloadTooLarge` naming `arguments` first. A unit test pins single-vs-batch row equivalence for the same event.
- **Per-item `(event, project_id, agent_version_id)` tuples, not a batch-wide project.** Every Kafka record is single-project (the Collector keys them), but a consumer poll aggregates many records across partitions, so one insert batch can span projects and agents.
- **No payload-cap re-checks on the batch path — a deliberate contract, not an omission.** The enricher job truncates and redacts server-side before calling — it becomes the authoritative enforcement point, replacing today's client-side `_truncate_args()`/`_redact()` trust. A dedicated test pins the trust-the-caller contract so changing it is a visible decision. The single-row caps stay: they guard direct external API callers.
- **"Retry-safe", deliberately not "atomic".** ClickHouse commits per block and `clickhouse-connect` splits large batches into ~2MB blocks (and silently re-sends once on a dropped keep-alive), so a failed call can have landed part of the batch. The caller's move on any failure is to retry the whole batch; `ReplacingMergeTree(received_at)` collapses re-inserted `event_id`s. Two documented edges: dedup never crosses the monthly `received_at` partition (a retry straddling month-end midnight can double-count — accepted, seconds-wide window), and a duplicate `event_id` *within* one batch collapses at insert time, last occurrence winning (`optimize_on_insert` applies the collapse to the inserted block itself — observed against a real server).
- **The batch path drops `async_insert`.** It exists to coalesce many small inserts; these calls are already batches, so buffering again would only add a server-side copy and a busy-timeout wait per enricher poll. The single-row paths keep it.

## Tests

- **8 new unit tests** (MagicMock client): one insert call carrying N rows with per-item `project_id`/`agent_version_id`; empty batch never touches ClickHouse; single-vs-batch row equivalence including the legacy `role` → `user_roles` shim and the falsy-attributes rule; the over-cap-is-inserted contract pin; ban and LLM twins.
- **2 new integration tests** (`make platform-api-test-integration`, local ClickHouse): a multi-project batch lands every row; a re-inserted batch collapses to one row per `event_id` under `SELECT … FINAL` — the retry-safety property only a real `ReplacingMergeTree` can prove. Both clean up their randomly-named `test_proj_*` rows.
- **Empirical pass against a real local ClickHouse** (throwaway probes, not committed): a 10,000-row batch inserts in 0.13s with all rows intact; emoji/RTL/control-char/nested-dict payloads round-trip byte-identical; non-UTC `occurred_at` offsets normalize to the stored UTC instants; the cap-boundary regression above.
- `make fmt-check` and `make platform-api-check` pass (506 unit tests); the full integration suite passes (6).

### DIFF
--- a/platform/api/hexgate_api/features/audit/service.py
+++ b/platform/api/hexgate_api/features/audit/service.py
@@ -106,17 +106,14 @@ _DECISION_INSERT_SETTINGS = {
 }
 
 
-def insert_decision(
-    clickhouse_client: Client,
-    *,
-    event: DecisionEvent,
-    project_id: str,
-    agent_version_id: str,
-) -> None:
-    """Write one decision row to policy_decision.
+def _decision_row(
+    event: DecisionEvent, *, project_id: str, agent_version_id: str
+) -> list:
+    """Build one policy_decision row in ``_DECISION_COLUMNS`` order.
 
-    Raises AuditPayloadTooLarge on payload overflow and ClickHouseError on
-    insert failure; both propagate so the caller maps them to transport errors.
+    The single place the row-shaping rules live — payload serialization, the
+    falsy-attributes normalisation, the legacy ``role`` shim — shared by the
+    single-row and batch insert paths so the two cannot drift apart.
     """
     args_json = (
         json.dumps(event.arguments, default=str) if event.arguments is not None else ""
@@ -130,12 +127,6 @@ def insert_decision(
     attributes_json = (
         json.dumps(event.attributes, default=str) if event.attributes else ""
     )
-    if len(args_json.encode("utf-8")) > MAX_ARGS_BYTES:
-        raise AuditPayloadTooLarge("arguments", MAX_ARGS_BYTES)
-    if len(hint_json.encode("utf-8")) > MAX_HINT_BYTES:
-        raise AuditPayloadTooLarge("hint", MAX_HINT_BYTES)
-    if len(attributes_json.encode("utf-8")) > MAX_ATTRIBUTES_BYTES:
-        raise AuditPayloadTooLarge("attributes", MAX_ATTRIBUTES_BYTES)
 
     # ``role`` is an ingest-only compatibility shim, the single place the legacy
     # scalar still exists anywhere in the system. An SDK released before
@@ -146,7 +137,7 @@ def insert_decision(
     # This survives any database recreation, because the SDK is pip-installed.
     user_roles = list(event.user_roles) or ([event.role] if event.role else [])
 
-    row = [
+    return [
         event.event_id,
         event.occurred_at,
         project_id,  # bearer-resolved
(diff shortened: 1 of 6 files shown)


platform/api/hexgate_api/features/audit/service.py
```diff
@@ -173,6 +196,52 @@ def insert_decision(
...
+    Contract with the caller (the span-enricher job): payloads arrive already
+    byte-capped and redacted — that job is the authoritative server-side
+    enforcement point, so unlike ``insert_decision`` there is no cap check
+    here.
+
+    Retry-safe rather than atomic: ClickHouse commits per block (and the
+    driver may re-send once on a dropped keep-alive), so a failed call can
+    have landed part of the batch. The caller's move on any failure is to
+    retry the whole batch — ReplacingMergeTree(received_at) collapses
+    re-inserted event_ids on merges, so duplicates cancel rather than
```
quanghexa94: **Retry contract is eventual-only — worth a docstring caveat**

The "duplicates cancel rather than double-count" guarantee holds only *after* a background merge. No read path applies `FINAL`: `summarize()`, `timeseries()`, and `list_decisions()` (lines ~474/582/645) all query `policy_decision` without it, so after a documented whole-batch retry both copies of each `event_id` are counted in dashboards until `ReplacingMergeTree` merges the parts (minutes-to-hours, or never if merges lag).

Since whole-batch retry is now a first-class expected operation, this duplicate-counting window is routine rather than rare. Suggest the docstring say the cancellation is eventual (post-merge), not immediate. Non-blocking.

platform/api/hexgate_api/features/audit/service.py
```diff
@@ -173,6 +196,52 @@ def insert_decision(
...
+    double-count. Two edges of that guarantee: dedup never crosses the
+    monthly received_at partition, so a retry that straddles a month
+    boundary can double-count (accepted — a seconds-wide window, and only at
+    most once a month); and a duplicate event_id *within* one batch
+    collapses at insert time, last occurrence winning.
```
quanghexa94: **"Collapses at insert time" isn't unconditional — it's per-block**

`optimize_on_insert` dedups only *within a single inserted block*, and clickhouse-connect splits a large insert into ~2MB blocks. Two rows with the same `event_id` that land on opposite sides of a block boundary (a big enricher batch — thousands of rows) are *not* collapsed at insert; they persist as two rows until a background merge.

So the unconditional "a duplicate event_id within one batch collapses at insert time, last occurrence winning" overstates the guarantee (and a test currently pins the unconditional form). Suggest scoping it to "within a block" or dropping the last-occurrence-wins promise. Non-blocking.

platform/api/hexgate_api/features/audit/service.py
```diff
@@ -173,6 +196,52 @@ def insert_decision(
     )
 
 
+def insert_decisions_batch(
```
quanghexa94, line 200: **Three near-identical `insert_*_batch` functions could share one helper**

`insert_decisions_batch`, `insert_ban_enforcements_batch`, and `insert_llm_invocations_batch` differ only in table name, column list, and row-builder — the empty guard, the list-comp, and the `async_insert: 0` pin are identical. A single `_insert_batch(client, table, columns, row_builder, items)` would carry the pin and the guard once; today a change to the batch contract (e.g. the `async_insert` pin) has to be made in three places and can drift. Non-blocking cleanup.

platform/api/hexgate_api/features/audit/service.py
```diff
@@ -173,6 +196,52 @@ def insert_decision(
...
+    clickhouse_client: Client,
+    items: list[tuple[DecisionEvent, str, str]],
```
quanghexa94: **Bare positional tuple invites a silent `project_id`/`agent_version_id` swap**

`items: list[tuple[DecisionEvent, str, str]]` has two adjacent `str` fields. A caller that builds `(event, agent_version_id, project_id)` by mistake unpacks without error, and every row lands with the two ids transposed — silently corrupting scoped dashboard reads, with no type checker to catch it.

The single-row twin `insert_decision` uses keyword-only params (`*, project_id, agent_version_id`) precisely to prevent this. A small `NamedTuple`/dataclass for batch items (`DecisionBatchItem(event, project_id, agent_version_id)`) would restore that protection. Latent (needs a caller mistake), so non-blocking — but the one I'd most consider fixing.

quanghexa94 (commented):
**LGTM overall — non-blocking comments only.**

The refactor is correct and behavior-preserving: I verified the index-derived cap-check mapping and the single-row vs. batch row equivalence, and found no correctness or crash bugs in the changed code.

The four comments below are all non-blocking — two are docstring-accuracy nits about the retry/dedup contract (the "duplicates cancel" and "collapses at insert time" guarantees are weaker than stated), one is a copy-paste cleanup, and one is a latent type-safety footgun. None need to hold the merge; #4 is the one I'd most consider tightening.

winlp4ever (approved):

quanghexa94 (approved):

platform/api/hexgate_api/features/audit/service.py
victorludvig: Fixed in b9a9efb. The docstring now says explicitly that ReplacingMergeTree(received_at) keeps every inserted row on disk and only collapses duplicate sort keys (highest `received_at` wins) when a background merge covers the parts holding them — opportunistic, no upper bound — and that `summarize`/`timeseries`/`list_decisions` read without `FINAL`, so both copies are counted until that merge lands. Mirrored the "eventual" wording in `insert_llm_invocations_batch`.

platform/api/hexgate_api/features/audit/service.py
victorludvig: Fixed in cca897f. The docstring now describes the mechanism: `optimize_on_insert` merges each inserted *block*, the driver splits large batches into ~2MB blocks, so copies in the same block collapse on the way in while copies straddling a boundary land in separate parts and wait for a background merge. Also tells callers to dedup by `event_id` before building the batch rather than rely on it. Mirrored in `insert_llm_invocations_batch`.

One note: I did not find a test pinning the unconditional intra-batch form — `test_when_a_batch_is_reinserted_then_rows_collapse_to_one_per_event` covers two separate calls and reads with `FINAL`, so it is unaffected. If you had a different test in mind, point me at it.

platform/api/hexgate_api/features/audit/service.py
victorludvig: Done in da2d8b9. Added `insert_batch(client, table, columns, row_builder, items)` to `core/clickhouse.py` — it owns the empty guard and the `async_insert: 0` pin (`BATCH_INSERT_SETTINGS`), and the three `insert_*_batch` functions are now one call each. Row builders keep their keyword-only `project_id`/`agent_version_id` through a `RowBuilder` Protocol, so the tuple unpacking happens in exactly one place. Two direct unit tests for the helper; the existing batch tests are unchanged and still pass.

platform/api/hexgate_api/features/audit/service.py
victorludvig: Fixed in 5fcf464. Went with a frozen, slotted, generic `BatchItem[E]` dataclass in `core/clickhouse.py` rather than a NamedTuple: a NamedTuple still accepts `Item(event, "ver_1", "proj_a")` positionally, so it would not actually close the swap hole. `BatchItem` puts both ids behind `KW_ONLY`, so `BatchItem(event, project_id=..., agent_version_id=...)` is the only way to build one — same protection as the single-row `*, project_id, agent_version_id` signature. All three `insert_*_batch` now take `Sequence[BatchItem[...]]`; tests updated, plus one pinning that positional ids raise `TypeError`. Unit + integration suites green locally.

PR #129, opened 2026-08-20
feat(sdk): RunFacts accumulator and run scope

guillaume-hexamind:
Foundation for `run.*` policy constraints (*block after K tool calls / Z seconds*). Adds the per-invocation fact record and opens a scope at every run boundary.

**No behaviour change** — nothing reads the contextvar yet; cost is one `uuid4()` + one `monotonic()` + two `ContextVar.set()` per invocation.

## Commits — review in order

| | |
|---|---|
| `d7e46e8` | `run_facts.py`: the `RunFacts` accumulator, `run_scope` / `use_run_facts`, `DETACHED`, path registry |
| `fa1555b` | Pure refactor — extract `_bind`/`_abind` on the langchain proxy, mirroring the pydantic_ai adapter. **No scope yet**; the 75 existing adapter tests pass untouched |
| `de29cca` | Wire `run_scope` at **11 sites / 16 entry points** (langchain 2, openai 3, pydantic_ai 2, google 2, factory 2) + coverage tests |
| `c7aa11d` | Simplification pass after self-review (−36 lines) |
| `9f9290c` | Rename `run.elapsed_s` → `run.elapsed_seconds` before release |

## Four decisions worth a look

**`DETACHED` drops writes.** A `ContextVar` default is *one process-wide object* — a plain zeroed instance would accumulate for the process lifetime until it exceeded every cap, then deny every tool call. Reads give zeros (fails open, correct for an unwired boundary); `run.id == ""` signals "decided outside a run".

**Scope is not in `HexgateContext.__aenter__`.** Tempting, since every adapter re-enters the context per call — but `HexgateAgent` (`factory.py`) is ambient, so a user's `async with HexgateContext(...)` around a loop of `ainvoke` calls would produce one `RunFacts` for all of them.

**`set()`, not `reset(token)`** in `use_run_facts` — same discipline and reason as `HexgateContext` (`context.py:125`): three entry points are async generators, whose finalizers run in a foreign `Context` where a token reset raises.

**`run_streamed` re-binds rather than mints.** `Runner.run_streamed` hands tools to a background task that snapshots the contextvars, then returns; the consumer-side iterator calls `use_run_facts(facts)` on the *same* object. Minting there would split one invocation across two run ids.

## Registry is `{id, agent, elapsed_seconds}`

`as_namespace()` returns **exactly** `KNOWN_RUN_PATHS`. Counters are accumulated and tested but deliberately **not projected**: a registered path with no writer reads a permanent zero, and `run.tool_calls < 20` against zero never fires — a fail-open cap shipped as a working feature. Each counter joins the registry and the projection together in the follow-up, with the test that reads it.

## Why `elapsed_seconds`, not `elapsed_s`

The unit in the name is load-bearing: the grammar rejects duration literals (`run.elapsed_seconds < 5m` → *"not a valid JSON literal"*), so a policy author writes a bare `300` and the field name is the only thing distinguishing seconds from milliseconds. `_s` was also the SDK's only single-letter unit suffix — `ttl_seconds` sits in the same module, alongside `timeout_seconds` and `call_timeout_seconds`, while `_ms` is reserved for the millisecond symbol. Renamed now because `run.*` paths are a public contract with policy authors.

## What guards what

Two reflective tests, both **verified by deliberately breaking them**:

- `test_run_scope_coverage.py` — derives expected boundaries from method signatures (a public method taking `hexgate_context`), so a *future* unwired entry point fails. Confirmed: removing one scope, and adding an unwired `abatch`, each fail.
- `test_run_facts.py` — discovers `record_*` methods reflectively and asserts each is a no-op on `DETACHED`. Confirmed: a sixth recorder without the guard fails both the behavioural and structural halves.

Plus `test_run_scope_behaviour.py`: one run id per invocation, fresh next time, none on a ban refusal, `DETACHED` on `__getattr__`-bypassed methods (`batch`).

## Not in scope

Engine threading, the `run` key in the constraint context, the load-time linter, and the recorder call sites — all follow-up. Egress is excluded by design: `egress_guard` is one-per-process with a fixed `HexgateContext`, so its decisions aren't inside any run and carry `run_id = ""` truthfully.

`make lint && make fmt-check && make coverage` → 2049 passed, 14 skipped. No existing test modified.

### DIFF
--- a/hexgate/adapters/_common.py
+++ b/hexgate/adapters/_common.py
@@ -20,10 +20,8 @@ from hexgate.tracing._senders import DEFAULT_DRAIN_TIMEOUT, pending_send_tasks
 # values are dropped too). Only bites on an unusually large role list.
 _MAX_METADATA_CHARS = 200
 
-# Fallback when a wrapped agent has no name. Shared so that a nameless agent's
-# audit events (agent_name=...) and its run facts (run.agent) agree on what it
-# is called — two independent literals drifting apart would silently split one
-# agent's trail across two labels.
+# Fallback when a wrapped agent has no name. Shared so a nameless agent's audit
+# events and its run facts agree on the label, rather than drifting apart.
 DEFAULT_AGENT_NAME = "default"
 
 
@@ -81,13 +79,10 @@ def langfuse_propagate_kwargs(context: HexgateContext, tag: str) -> dict[str, An
 async def abind(
     context: HexgateContext, agent_name: str, propagate_kwargs: dict[str, Any]
 ) -> AsyncIterator[None]:
-    """Async run boundary shared by every adapter proxy: identity scope, run
-    facts, then Langfuse propagation — in that order, so the run's facts are
-    live for the whole block a tool call executes in.
+    """Async run boundary shared by every adapter proxy: identity scope, run facts,
+    then Langfuse propagation, so the facts are live wherever a tool call executes.
 
-    ``propagate_kwargs`` is computed by the caller (via
-    :func:`langfuse_propagate_kwargs`), not here, since its ``tag`` embeds the
-    calling method's name.
+    The caller passes ``propagate_kwargs`` because its tag embeds their method name.
     """
     async with context:
         with run_scope(agent_name):
--- a/hexgate/adapters/_common.py
+++ b/hexgate/adapters/_common.py
@@ -20,10 +20,6 @@ from hexgate.tracing._senders import DEFAULT_DRAIN_TIMEOUT, pending_send_tasks
 # values are dropped too). Only bites on an unusually large role list.
 _MAX_METADATA_CHARS = 200
 
-# Fallback when a wrapped agent has no name. Shared so a nameless agent's audit
-# events and its run facts agree on the label, rather than drifting apart.
-DEFAULT_AGENT_NAME = "default"
-
 
 def drain_pending_tasks(
     loop: asyncio.AbstractEventLoop, *, drain_timeout: float = DEFAULT_DRAIN_TIMEOUT
@@ -77,25 +73,24 @@ def langfuse_propagate_kwargs(context: HexgateContext, tag: str) -> dict[str, An
 
 @asynccontextmanager
 async def abind(
-    context: HexgateContext, agent_name: str, propagate_kwargs: dict[str, Any]
+    context: HexgateContext, agent_name: str, tag: str
 ) -> AsyncIterator[None]:
     """Async run boundary shared by every adapter proxy: identity scope, run facts,
     then Langfuse propagation, so the facts are live wherever a tool call executes.
 
-    The caller passes ``propagate_kwargs`` because its tag embeds their method name.
+    Takes the span ``tag`` rather than the built kwargs — it embeds the caller's
+    method name, and resolving it here keeps the attributes read inside the scopes.
     """
     async with context:
         with run_scope(agent_name):
-            with propagate_attributes(**propagate_kwargs):
+            with propagate_attributes(**langfuse_propagate_kwargs(context, tag)):
                 yield
 
 
 @contextmanager
-def bind(
-    context: HexgateContext, agent_name: str, propagate_kwargs: dict[str, Any]
-) -> Iterator[None]:
+def bind(context: HexgateContext, agent_name: str, tag: str) -> Iterator[None]:
     """Sync mirror of :func:`abind`."""
     with context.sync_scope():
         with run_scope(agent_name):
-            with propagate_attributes(**propagate_kwargs):
+            with propagate_attributes(**langfuse_propagate_kwargs(context, tag)):
                 yield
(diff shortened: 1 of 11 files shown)


winlp4ever (commented):
Automated code-review pass (high effort). No high-severity correctness bugs — scope wiring, ban-check ordering, streamed re-bind, and set()-vs-reset() discipline all look correct and consistent with the existing HexgateContext pattern. Four lower-severity notes inline below.

quanghexa94 (commented):
Automated code-review pass (high effort). No high-severity correctness bugs — scope wiring, ban-check ordering, streamed re-bind, and set()-vs-reset() discipline all look correct and consistent with the existing HexgateContext pattern. Four lower-severity notes inline below.

hexgate/adapters/pydantic_ai/agent.py
```diff
@@ -73,19 +73,14 @@ def _setup_observability(self) -> None:
     def _propagate_kwargs(self, context: HexgateContext, method: str) -> dict[str, Any]:
         return langfuse_propagate_kwargs(context, f"pydantic_ai.agent.{method}")
 
-    @asynccontextmanager
-    async def _abind(self, context: HexgateContext, method: str) -> AsyncIterator[None]:
-        """Async HexgateContext scope + Langfuse propagation."""
-        async with context:
-            with propagate_attributes(**self._propagate_kwargs(context, method)):
-                yield
+    def _abind(self, context: HexgateContext, method: str) -> AsyncIterator[None]:
```
quanghexa94: `_abind`/`_bind` dropped their `@asynccontextmanager`/`@contextmanager` decorators and now *return* a context-manager object, but the annotations still say `-> AsyncIterator[None]` / `-> Iterator[None]`. The real return types are now `_AsyncGeneratorContextManager[None]` / `_GeneratorContextManager[None]`, so a strict type checker (mypy/pyright) will flag the mismatch. Harmless at runtime since callers use `async with`/`with`, but misleading. Same twins in `hexgate/adapters/langchain/agent.py` (`_abind`/`_bind`).

hexgate/runtime/run_facts.py
```diff
@@ -0,0 +1,175 @@
...
+    "hexgate_run_facts",
+    default=DETACHED,
+)
+
+
+def get_run_facts() -> RunFacts:
+    """The active run's facts, or :data:`DETACHED` outside a run scope.
+
+    Never ``None``: a missing ``run`` namespace fails every ``run.*`` constraint
+    closed, so an unwired boundary would deny everything for the wrong reason.
```
quanghexa94: This docstring inverts the actual semantics. It says a missing `run` namespace 'fails every `run.*` constraint closed' and 'would deny everything', but `DETACHED` returns `id == ""` and zeroed counters, so `run.elapsed_seconds < 300` / `run.tool_calls < 20` evaluate against 0 and **pass** — it fails *open*, not closed. (The module docstring's 'never needs a fail-soft path' reads the same way.) Whoever builds the constraint engine next may rely on this and assume `run.*` denies outside a scope. Worth rewording to describe the real fail-open behavior.

hexgate/agents/factory.py
```diff
@@ -35,13 +35,15 @@
 from pydantic import BaseModel
 
 # BC re-export — canonical home is hexgate.approvals (framework-agnostic).
+from hexgate.adapters._common import DEFAULT_AGENT_NAME
```
quanghexa94: This makes core `agents/factory.py` take a **module-level** import on the adapter layer (`hexgate.adapters._common`) just to read the `DEFAULT_AGENT_NAME` string — inverting the agents→adapters layering, and pulling `_common`'s eager `langfuse` import into core load. Previously factory imported adapters only lazily (inside functions). A neutral home for the constant (e.g. `hexgate.runtime`) would let both adapters and factory depend downward instead.

hexgate/adapters/_common.py
```diff
@@ -66,3 +73,29 @@ def langfuse_propagate_kwargs(context: HexgateContext, tag: str) -> dict[str, An
         "session_id": context.session_id,
         "metadata": {"user_roles": roles},
     }
+
+
+@asynccontextmanager
+async def abind(
+    context: HexgateContext, agent_name: str, propagate_kwargs: dict[str, Any]
```
quanghexa94: `abind`/`bind` receive `propagate_kwargs` already computed at the call site (before the `context` / `run_scope` is entered), whereas the original per-proxy `_abind`/`_bind` computed `langfuse_propagate_kwargs` lazily *inside* the scope. Harmless today because it only reads static fields off `HexgateContext`, but a latent trap: if propagation ever needs a value only established once the context/run scope is active (e.g. a contextvar-derived attribute), the eager computation would silently capture the pre-scope value.

quanghexa94 (commented):
Automated code-review pass (high effort). No high-severity correctness bugs — scope wiring, ban-check ordering, streamed re-bind, and set()-vs-reset() discipline all look correct and consistent with the existing HexgateContext pattern. Four lower-severity notes inline below.

quanghexa94 (approved):
LGTM but try to see if u can patch minor feedbacks from claude

guillaume-hexamind:
> LGTM but try to see if u can patch minor feedbacks from claude

Thanks for the review, doing that before merge

PR #130, opened 2026-08-20
feat(sdk): define otel span semconv, expose redaction pipeline

victorludvig:
## What is Changing

- New `hexgate/tracing/semconv.py`: the OTLP wire contract for audit events carried as OTel spans — instrumentation scope names (`hexgate.audit` / `hexgate.usage` / `hexgate.bans`) and the `sec_ai.*` / `gen_ai.*` span attribute keys, with the contract rules (occurred_at = span start time, event_id as a string-UUID attribute, lists as native string arrays, dicts as JSON strings, project_id never on the span) documented in the module docstring.
- `hexgate/audit.py`: the redaction/truncation pipeline gains public names (`redact`, `truncate_json`, `bounded_violations`, the two sensitive-key regexes). Pure aliases — no behavior change.

The upcoming span-enricher job (platform side) and the future SDK span emitter must agree byte-for-byte on scope names and attribute keys; this module is the single source both import, so neither hardcodes a string. The public aliases let server-side ingestion import the exact redaction/truncation rules instead of keeping its own copy that would drift. The `gen_ai.*` keys are inlined as literals rather than imported from `opentelemetry-semantic-conventions` because they only exist in its private `_incubating` namespace, which may rename them at any time — and these strings are a wire contract that must only change deliberately.

## Tests

- `tests/tracing/test_semconv.py` (new): no two constants share a wire string, prefixes match their namespace (scopes `hexgate.*`, vocabulary `sec_ai.*`, official keys `gen_ai.*`), and the inlined `gen_ai.*` literals match the official convention verbatim.
- Full SDK suite green: 1997 passed. `make lint && make fmt-check && make coverage` clean.

### DIFF
--- a/hexgate/audit.py
+++ b/hexgate/audit.py
@@ -141,6 +141,15 @@ def _truncate_json(payload: dict[str, Any], *, cap: int) -> dict[str, Any]:
         preview_bytes //= 2
 
 
+# Public aliases so server-side ingestion can import this pipeline instead of
+# keeping its own copy.
+redact = _redact
+truncate_json = _truncate_json
+bounded_violations = _bounded_violations
+SENSITIVE_ARG_KEY_RE = _SENSITIVE_ARG_KEY_RE
+SENSITIVE_ATTR_KEY_RE = _SENSITIVE_ATTR_KEY_RE
+
+
 @dataclass(frozen=True, slots=True)
 class AuditEvent:
     """Decision plus caller identity from the active HexgateContext scope.
--- a/tests/tracing/test_semconv.py
+++ b/tests/tracing/test_semconv.py
@@ -1,0 +1,45 @@
+"""Invariants of the OTel span wire contract (hexgate.tracing.semconv)."""
+
+from __future__ import annotations
+
+from hexgate.tracing import semconv
+
+_SCOPES = {"SCOPE_AUDIT", "SCOPE_USAGE", "SCOPE_BANS"}
+_GEN_AI = {
+    "GEN_AI_REQUEST_MODEL",
+    "GEN_AI_USAGE_INPUT_TOKENS",
+    "GEN_AI_USAGE_OUTPUT_TOKENS",
+}
+
+
+def _constants() -> dict[str, str]:
+    return {
+        name: value
+        for name, value in vars(semconv).items()
+        if not name.startswith("_") and isinstance(value, str)
+    }
+
+
+def test_when_all_values_are_collected_then_none_collide() -> None:
+    """Two constants sharing a wire string would silently merge two fields."""
+    values = list(_constants().values())
+    assert len(values) == len(set(values))
+
+
+def test_when_names_are_grouped_then_prefixes_match_their_namespace() -> None:
+    """Scopes name the emitting library; attributes name the shared vocabulary."""
+    for name, value in _constants().items():
+        if name in _SCOPES:
+            assert value.startswith("hexgate."), name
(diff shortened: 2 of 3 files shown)


hexgate/audit.py
```diff
@@ -141,6 +141,15 @@ def _truncate_json(payload: dict[str, Any], *, cap: int) -> dict[str, Any]:
         preview_bytes //= 2
 
 
+# Public aliases so server-side ingestion can import this pipeline instead of
+# keeping its own copy.
+redact = _redact
```
quanghexa94, line 146: Minor / non-blocking: these redaction aliases are exposed from `hexgate.audit`, which transitively imports `httpx` via `_senders`. A server-side consumer that wants only the redaction pipeline drags in the full HTTP stack, which slightly undercuts the PR's own "zero-dependency single source of truth" rationale. If the constants/redaction helpers are meant to be imported broadly, a leaf module with no heavy imports would be a cleaner home. Just a thought — not a blocker.

tests/tracing/test_semconv.py
```diff
@@ -0,0 +1,45 @@
...
+    for name, value in _constants().items():
+        if name in _SCOPES:
+            assert value.startswith("hexgate."), name
+        elif name in _GEN_AI:
+            assert value.startswith("gen_ai."), name
+        else:
+            assert value.startswith("sec_ai."), name
+
+
+def test_when_gen_ai_names_are_read_then_they_match_the_official_semconv() -> None:
```
quanghexa94, line 40: Minor / non-blocking: this test asserts the `gen_ai.*` constants against literals defined in the same repo, so it's a change-detector (guards against accidental edits) rather than real conformance to the upstream OpenTelemetry semconv spec. That's a fine regression guard — just worth not reading it as "verified against the official convention," since nothing here pins it to an external source of truth.

quanghexa94 (commented):
**LGTM — non-blocking comments only.**

Clean, additive change: a new semconv constants module, pure re-export aliases, and tests. No correctness bugs — I verified the aliases are exact and the constants are internally consistent. The two comments below are low-confidence observations surfaced by a high-effort pass; neither needs to hold the merge.

quanghexa94 (approved):

hexgate/audit.py
victorludvig: Correct, the upcoming enricher python job will import SENSITIVE_ARG_KEY_RE, redact from this file, without using the rest of hexgate/audit.py, which costs 0.1MB memory and a few ms.

But the real cost is actually hexgate/__init__.py. Importing hexgate.tracing.semconv which has zero imports of its own costs about 80MB and 630ms, because Python executes first the parent package's __init__.py, and hexgate/__init__.py eagerly imports agents, cloud, guard, manifest, runtime, security, tools. That fans out to langchain, langfuse, opentelemetry, google, pydantic, wasmtime, cryptography, and httpx. The whole SDK, plus every framework adapter, loads to read a handful of string constants.

The fix would be to make hexgate/__init__.py lazy, but that's an SDK-wide public-API change, which is out of scope for this PR.

And the slight overhead does not matter much for the enricher which is a long-running process. 80MB and 0.6s once at startup is not much.

tests/tracing/test_semconv.py
victorludvig: Agreed: it's a change-detector, not upstream conformance. Pinning to the real source would mean importing the private _incubating namespace the module deliberately avoids.

PR #131, opened 2026-08-24
fix(platform-api): cover llm_invocation in the startup schema check

victorludvig:
## What is Changing

- The written-schema startup guard (exception + check loop, formerly private to `features/audit/service.py`) moves to `core/clickhouse.py` as `verify_written_columns` + `AuditSchemaOutOfDate`, next to the `table_columns` helper it uses.
- Each feature now verifies the tables it writes with its own thin `verify_schema` wrapper: `features/audit` checks `policy_decision` + `ban_enforcement` (unchanged scope), and `features/llm_invocations` gains one for `llm_invocation`.
- The API lifespan calls both checks at startup.

The startup check was born in #120 scoped to the two tables that PR touched; `llm_invocation` was never added, so a ClickHouse volume missing an llm column passed startup and then failed every insert at runtime with a raw driver error instead of the actionable "recreate the volume" boot error. Owning the check per feature (with the generic machinery in core) closes the gap without features importing each other's table definitions. The upcoming span-enricher job reuses the same wrappers for its own startup check.

## Tests

- New in `test_llm_invocation.py`: `verify_schema` passes on a current schema; a missing `input_tokens` column raises with the table named.
- Existing verify_schema suite in `test_audit.py` unchanged in behavior (now exercising the core machinery through the audit wrapper).
- `make fmt-check && make platform-api-check`: 500 passed.

### DIFF
--- a/platform/api/hexgate_api/main.py
+++ b/platform/api/hexgate_api/main.py
@@ -16,14 +16,14 @@ from hexgate_api.features.agents.router import router as agents_router
 from hexgate_api.features.agents.service import backfill_bundles
 from hexgate_api.features.audit.router import router as audit_router
 from hexgate_api.features.audit.service import verify_schema as verify_audit_schema
-from hexgate_api.features.llm_invocations.service import (
-    verify_schema as verify_llm_schema,
-)
 from hexgate_api.features.auth.router import include_auth_routers, mount_oauth_routers
 from hexgate_api.features.bans.router import router as bans_router
 from hexgate_api.features.chat.router import router as chat_router
 from hexgate_api.features.invitations.router import router as invitations_router
 from hexgate_api.features.llm_invocations.router import router as llm_invocations_router
+from hexgate_api.features.llm_invocations.service import (
+    verify_schema as verify_llm_schema,
+)
 from hexgate_api.features.members.router import router as members_router
 from hexgate_api.features.orgs.router import router as orgs_router
 from hexgate_api.features.policy_modules.router import router as policy_modules_router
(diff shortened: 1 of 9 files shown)


platform/api/hexgate_api/main.py
```diff
@@ -159,8 +162,9 @@ async def lifespan(app_: FastAPI):
     else:
         # Behind this build, every insert is rejected and dropped by the SDK —
         # silently, from this side. Refuse to boot so the previous deployment
-        # keeps serving.
+        # keeps serving. Each feature checks the tables it writes.
         verify_audit_schema(get_clickhouse())
```
quanghexa94, line 166: Minor / non-blocking: these two checks run sequentially, so if `verify_audit_schema` raises `SchemaOutOfDate` first, `verify_llm_schema` never runs — a boot against a volume where *both* audit and llm tables are stale reports only the audit gaps. Low impact, since the prescribed remedy (recreate the volume) restores all tables at once; it only bites an operator applying a narrow per-table `ALTER`. If you want one boot to surface every gap, combining both features' `(table, columns)` into a single `verify_written_columns` call — or aggregating the missing-column dicts before raising — would do it.

quanghexa94 (commented):
**LGTM — safe to merge.**

Clean refactor-plus-gap-fill: the written-schema check moves to `core/clickhouse.py` as the feature-neutral `SchemaOutOfDate` + `verify_written_columns`, and `llm_invocation` — previously unguarded, so a stale volume booted then failed every insert at runtime — is now covered. I verified `_LLM_INVOCATION_COLUMNS` matches `schema.sql` (server-defaulted `received_at` correctly excluded), the code-60/error-degradation semantics moved intact, no stale `AuditSchemaOutOfDate` references remain, and all three insert tables are now guarded. One low-confidence, non-blocking observation below.

winlp4ever (approved):

quanghexa94 (approved):

platform/api/hexgate_api/main.py
victorludvig: Agreed — I'll fix this in a follow-up PR by catching SchemaOutOfDate per feature and aggregating the gaps before raising, rather than merging the column lists (keeps features from knowing each other's tables); the enricher (#133) runs the same two checks at startup, so a shared helper is worth doing once both call sites exist.

PR #133, opened 2026-08-24
feat(platform-api): add the span-enricher job (redpanda to clickhouse)

victorludvig:
## What is Changing

New standalone job `hexgate_api/jobs/enricher/` (`python -m hexgate_api.jobs.enricher`, `make enricher-run`): an aiokafka consumer that reads the Collector's OTLP records from `hexgate.otlp.raw`, decodes and validates spans back into the platform's event models, and batch-inserts into ClickHouse.

- `decode.py` — record bytes → per-span units (`ExportTraceServiceRequest` protobuf, AnyValue → Python).
- `coerce.py` — generic attribute-type coercion (string→int, JSON-string→list/dict), lossy coercions logged, failures reject the span.
- `mapping.py` — the semconv wire contract (`hexgate.tracing.semconv`, PR #130) → `DecisionEvent` / `LlmInvocationEvent` / `BanEnforcementEvent`; `occurred_at` from span start time; Pydantic + the ingest window are the validation layer.
- `enforcement.py` — authoritative server-side redaction/byte caps, importing the SDK pipeline (same rules as `AuditEvent.as_payload`).
- `resolver.py` — one DB session per poll, `agent_version_id` per unique (project, agent); unresolved → `""` like the HTTP ingest.
- `dlq.py` + `consumer.py` — per-span dead-lettering to `hexgate.otlp.dlq` (JSON envelopes; attributes redacted with the substring `SENSITIVE_ARG_KEY_RE`, the JSON-string dict fields `arguments`/`hint`/`attributes` parsed first so the key match reaches inside them; anything that doesn't parse to a dict is replaced by `[UNPARSEABLE]` — the `source` pointer still locates the raw record). Keyless records are decoded and parked span by span through the same envelope; only undecodable bytes go out raw. Events are deduped by `(project_id, event_id)` within a poll (first copy wins; a tenant reusing another's id is a different event) before the batch inserts, per the batch functions' contract. ClickHouse inserts and DLQ sends share one retry loop (`_retry_until_acked`: capped backoff, stop-aware, never commits on the way out), offset commit only after both, SIGTERM-clean shutdown with the client `start()` calls inside the `try/finally`.
- Wiring: `redpanda_*`/`enricher_*` settings, `aiokafka` + explicit `opentelemetry-proto` deps, `.env.sample`, stale DLQ comment fixed in `create-topics.sh`, `jobs/` layout line in `platform/api/CLAUDE.md`.
- Makefile: new `postgres-init` target (`postgres-up` + `init_db()`), used by `collector-run`, `collector-test-integration` and the new `enricher-run`, which runs the job with `DATABASE_URL=$(POSTGRES_DSN)` — the same Postgres as `platform-api-pg`, so `agent_version_id` resolves instead of silently falling back to the SQLite file.

Built on #128 (batch inserts), #130 (semconv), #131 (schema guard) and #136 (`verify_all`) — all merged; rebased onto #136 so the startup check is a single `verify_all(...)` call run via `asyncio.to_thread`, reporting every stale table in one boot.

This is the consumer the OTLP pipeline has been building toward: Collector (#111, #127) authenticates and buffers spans into Redpanda; the batch inserts (#128) gave ClickHouse a bulk write path; this job connects the two. Offsets commit only after ClickHouse acks, so a crash replays the poll and `event_id` dedup absorbs it — at-least-once end to end with no ack-then-drop. Poisoned input (undecodable records, keyless records, invalid spans) is parked per-span on the DLQ so one bad span never wedges a partition, while infra failures — ClickHouse or the DLQ topic — halt consumption with backoff instead of dropping data or crashing the job.

The DLQ topic has 30-day retention and no ACLs, and there is no SDK span emitter yet, so this job is the authoritative redaction point: a rejected span's tool arguments must never reach the DLQ unredacted, whether the span was rejected for validation, clock skew, or a missing record key.

## Tests

- 69 unit tests (`tests/jobs/enricher/`): decode round-trips, coercion rules, per-scope mapping incl. ms-precision timestamps and rejection paths, SDK-parity redaction/caps, DLQ envelope shapes incl. redaction inside JSON-string dict fields and the unparseable case, keyless records parked span by span (redacted, no raw bytes), resolver caching, and `_process_poll` ordering (inserts → DLQ → commit), insert retry-then-single-commit, DLQ-send retry-then-single-commit, stop mid-outage (ClickHouse and DLQ) → no commit, intra-poll `event_id` dedup, sibling survival, producer-start failure still stops the consumer, two stale tables reported in one `SchemaOutOfDate`.
- 1 integration test (real ClickHouse): full poll round-trip into all three tables; reprocessing the same records leaves `FINAL` counts unchanged.
- Manual smoke (2026-08-27, after the review fixes) against live Redpanda + ClickHouse + Postgres via `make enricher-run`'s command line: valid record → correct `policy_decision` row; one record carrying the same span twice → 1 row without `FINAL` (intra-poll dedup); garbage record → `decode` envelope with raw bytes; span missing `tool_name` with `{"password": …}` in `sec_ai.arguments` → `validation` envelope with `"password": "[REDACTED]"` inside the parsed dict; keyless valid record → `missing_key` span envelope, redacted, no `record_value_base64`; SIGTERM → `LeaveGroup request succeeded`, exit within 1 s, no tracebacks. `make postgres-init` verified live.
- `make fmt-check && make platform-api-check` (586 passed) and `make platform-api-test-integration` (7 passed) green after the rebase onto #136.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -253,6 +253,10 @@ collector-generate: ## Regenerate + compile the collector from builder-config.ya
 # The authenticator makes boot fatal without Postgres, the devtoken schema,
 # and the root public key, so this target now provides the first two and
 # checks for the third — the pre-auth collector booted on Redpanda alone.
+.PHONY: enricher-run
+enricher-run: redpanda-topics clickhouse-up ## Run the span-enricher job (Kafka → ClickHouse)
+	cd platform/api && uv run python -m hexgate_api.jobs.enricher
+
 collector-run: postgres-up redpanda-topics ## Run the collector binary against config.yaml
 	cd platform/api && DATABASE_URL=$(POSTGRES_DSN) uv run python -c \
 		"import asyncio; from hexgate_api.core.db import init_db; asyncio.run(init_db())"
--- a/Makefile
+++ b/Makefile
@@ -178,6 +178,14 @@ POSTGRES_DSN ?= postgresql+asyncpg://hexgate:hexgate-dev-password@localhost:5433
 postgres-up: ## Start local Postgres and wait until healthy
 	$(COMPOSE) up -d --wait postgres
 
+# The schema belongs to platform-api; anything else that reads the relational
+# store (collector, enricher, integration tests) creates it through here so a
+# fresh volume never surfaces as "no such table" on the first request.
+.PHONY: postgres-init
+postgres-init: postgres-up ## Create the platform-api tables on local Postgres (idempotent)
+	cd platform/api && DATABASE_URL=$(POSTGRES_DSN) uv run python -c \
+		"import asyncio; from hexgate_api.core.db import init_db; asyncio.run(init_db())"
+
 .PHONY: postgres-stop
 postgres-stop: ## Stop Postgres (keeps the data volume)
 	$(COMPOSE) stop postgres
@@ -253,16 +261,17 @@ collector-generate: ## Regenerate + compile the collector from builder-config.ya
 # The authenticator makes boot fatal without Postgres, the devtoken schema,
 # and the root public key, so this target now provides the first two and
 # checks for the third — the pre-auth collector booted on Redpanda alone.
-.PHONY: enricher-run
-enricher-run: redpanda-topics clickhouse-up ## Run the span-enricher job (Redpanda → ClickHouse)
-	cd platform/api && uv run python -m hexgate_api.jobs.enricher
-
-collector-run: postgres-up redpanda-topics ## Run the collector binary against config.yaml
-	cd platform/api && DATABASE_URL=$(POSTGRES_DSN) uv run python -c \
-		"import asyncio; from hexgate_api.core.db import init_db; asyncio.run(init_db())"
+collector-run: postgres-init redpanda-topics ## Run the collector binary against config.yaml
 	@test -f platform/api/data/hexgate.pub \
 		|| { echo "platform/api/data/hexgate.pub is missing — run 'make platform-api' once to generate the root keypair, or set HEXGATE_COLLECTOR_PUBLIC_KEY_FILE"; exit 1; }
 	cd platform/collector && ./hexgate-collector --config=config.yaml
+
+# Same Postgres as platform-api-pg: the resolver reads agent/agent_version
+# there, and pointing the job at the SQLite fallback instead would resolve
+# every agent_version_id to "" without an error.
+.PHONY: enricher-run
+enricher-run: postgres-init redpanda-topics clickhouse-up ## Run the span-enricher job (Redpanda → ClickHouse)
+	cd platform/api && DATABASE_URL=$(POSTGRES_DSN) uv run python -m hexgate_api.jobs.enricher
 
 .PHONY: collector-test
 collector-test: ## Unit tests for the collector's own Go modules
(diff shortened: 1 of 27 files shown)


guillaume-hexamind:
1. Unbounded token/latency ints wedge a partition forever — 72/100

schemas.py:541-543 bounds input_tokens/output_tokens/latency_ms with ge=0 only, but the columns are UInt32 (schema.sql:94-96). An over-range value (a foreign emitter sending latency_ms in ns, say) passes as_int and Pydantic, then clickhouse_connect raises DataError: value out of range — a permanent, client-side error. _retry_until_acked (consumer.py:175) catches bare Exception, so it retries forever and never commits; _insert_all also re-inserts the valid decision/ban rows on every cycle. This breaks "one bad span never wedges a partition".

Fix: add le=UINT32_MAX to the three fields in schemas.py → the span goes to the DLQ instead (and the HTTP route returns 422, not 500). Optionally also treat DataError/ProgrammingError as permanent in _retry_until_acked.

2. Uncapped DLQ envelopes wedge the same way — 65/100

dlq.py:115 base64-encodes the whole raw record (+33%) and span_envelope applies no byte cap, unlike the accepted path (enforcement.py caps arguments at 8 KiB). The producer uses aiokafka's default max_request_size=1048576, and _serialize raises MessageSizeTooLargeError client-side on every attempt → same infinite retry at consumer.py:323. Triggered by an undecodable record >~750 KB, or a rejected span with a few hundred KB of sec_ai.arguments.

Fix: truncate raw_value before b64 (add a record_value_truncated flag) and run span attributes through truncate_json with a DLQ cap; treat MessageSizeTooLargeError as permanent (drop with a log, or send a stub envelope — _source still locates the original).

3. None record value / non-UTF-8 key crash-loops the process — 50/100

decode_record(None) raises TypeError (verified: ParseFromString(None) → TypeError, not DecodeError), and record.key.decode("utf-8") at consumer.py:197 is unguarded. Neither is caught anywhere — the exception escapes run() and __main__.main(), the offset never commits, and a restart replays the same record. Only reachable from a foreign producer (the Collector always keys with the auth-derived project_id), but that's the documented threat model for this unauthenticated topic.

Fix: park a None value through the existing record_envelope path (error_class="decode", raw_value=b""), and wrap the key decode in a helper returning None on UnicodeDecodeError so it falls through the existing missing_key branch.

4. Postgres is the one infra dependency with no retry — 48/100

resolve_versions (consumer.py:265) sits outside _retry_until_acked, so a connection failure or pool exhaustion propagates out and kills the process — a backoff-free restart loop, unlike ClickHouse and the DLQ. No data loss (no commit), but it contradicts "infra failures … halt consumption with backoff instead of crashing the job".

Fix: wrap it in the same closure pattern as _insert_all:
versions: dict[tuple[str, str], str] = {}

async def _resolve() -> None:
    nonlocal versions
    versions = await resolve_versions(pairs)

if not await self._retry_until_acked(_resolve, "agent version resolve"):
    return
Minor, same file: resolver.py takes the session factory from module scope rather than by injection, and issues one query per unique pair where get_latest_agent_versions_map would do it in one.

guillaume-hexamind (commented):
Logic looks solid overall.

Above are 4 fact checked claude issues to investigate, the first 2 seeming blocking; PR can be merged afterward

victorludvig (commented):
1. UInt32 overflow (3bc5612): added a UINT32_MAX ceiling on the three fields, so over-range spans now fail validation and go to the DLQ (HTTP path returns 422).
2. Uncapped DLQ envelopes (2811226): raw record capped at 64 KiB before base64 (with a record_value_truncated flag), span attributes run through the same truncate_json as the accepted path with a 32 KiB cap. As a backstop, MessageSizeTooLargeError on a DLQ send is now treated as permanent — logged and dropped instead of retried (went with drop-with-log rather than a stub envelope; with the caps in place it should be unreachable, and _source still locates the original).
3. decode_record(None) now raises RecordDecodeError, so a null value parks through the existing decode path with raw_value=b""; the key decode moved into a helper that returns None on UnicodeDecodeError, so a garbled key falls through the existing missing_key branch.
4. Postgres retry (7606965): resolve_versions now runs through the same _retry_until_acked closure as the inserts. Also took the minor: the resolver takes the session factory by injection and resolves in two batched queries instead of one per pair.

guillaume-hexamind (approved):
Thanks, looks good to me

PR #135, opened 2026-08-25
fix(sdk): reject exponentially-backtracking regexes at policy load

### LINKED ISSUE #134: Regexes accepted by `matches` can hang the pydantic engine (linear under RE2)
## Summary

`matches` constraints accept regexes that are linear under RE2 and exponential
under Python's `re`, so the pydantic engine can be stalled by a tool argument
while the WASM bundle would decide instantly. A branch with a fix, tests and an
ADR is ready — I can't open a PR (the repo limits that to collaborators), so
I'm filing this instead.

**Branch:** https://github.com/Dev-next-gen/hexgate/tree/lc/fix/reject_exponential_regex
**Diff:** https://github.com/HexamindOrganisation/hexgate/compare/main...Dev-next-gen:hexgate:lc/fix/reject_exponential_regex

## The problem

`_validate_re2` refuses regex *syntax* RE2 cannot run, so a policy cannot mean
one thing on the pydantic engine and another in a WASM bundle. That covers what
the engines **express**; it says nothing about what they **cost**.

Python's `re` backtracks, RE2 does not. `^(a+)+$` is accepted by both. Timed
through `check_constraints`, with the pattern written as an email allowlist
`^([a-zA-Z0-9_-]+)*@corp[.]com$`:

| argument length | 20 | 24 | 26 | 28 | 30 |
|---|---|---|---|---|---|
| time to decide one call | 0.04s | 0.54s | 2.21s | 8.74s | 37.8s |

(20–28 through `check_constraints`; 30 with the same `re.search` call
`_eval_call` makes. About four times longer per two characters added.)

Tool arguments are model-written and caller-influenced, and `hexgate/security/`
has no timeout and no argument-size cap — so the component that decides whether
a call is allowed can be stalled by the argument it exists to filter. Reachable
wherever the pydantic engine runs, which per `source.py` includes the documented
local mode and the "platform served no compiled bundle" fallback.

## What the branch does

Refuses such patterns at **policy load**, where the RE2 syntax check already
lives, so nothing changes on the decision path (the check sits inside the
already-cached `parse_constraint`; worst case measured at 3ms, once).

Detection is exponential ambiguity — two distinct paths leaving a state and
returning to it over one word — which is exactly what RE2's construction rules
out. A syntactic "nested quantifier" rule would be wrong both ways: it rejects
`^([a-z0-9-]+[.])+corp[.]com$` (nested and linear, because the mandatory `[.]`
forces one split per iteration) and misses `^(a|aa)+$` (not nested, exponential).

It also refuses **atomic groups** `(?>a+)+` and **possessive quantifiers** `a*+`,
which pass `_validate_re2` on `main` and are undefined under RE2 —
`opa eval 'regex.match("(?>a+)+", "aaa")'` returns undefined while Python
matches. Same divergence class as the existing list; they were missing because a
source-text rule cannot tell `a*+` from `\++`.

## Validation

- For 50 randomly generated patterns the check refused, an input was hunted that
  measurably explodes: 50/50 confirmed, no false positive left standing.
- Reverse direction: of 118 generated patterns, 27 were passed as linear and
  each attacked with anchored pumped inputs; none exploded, 0 false negatives.
- Every `matches` pattern this repo ships still loads.
- Full SDK suite unchanged, new module at 94.2% branch coverage, ruff clean,
  `test_engine_parity.py` run against a real `opa` binary rather than skipped.

## Known limits, stated rather than implied

- **Polynomial** cost is out of scope: `^a*a*a*$` takes 67.6s on a
  4 000-character argument and still loads. A detector was written and dropped
  because its precision could not be demonstrated — `^.*foo.*$` has the same
  structure and is linear (measured flat to 8 000 characters), since the pattern
  matches as soon as `foo` appears, so no input both exercises the ambiguity and
  fails. Closing it needs the automaton to model acceptance.
- Anchors (`^`, `\b`) are treated as consuming nothing: an over-approximation I
  found no failing case for but have not proved safe.
- The 4-character bound threshold is a judgement call, not a derived constant.

## One question for you

The refusal applies even to a deployment that only ever runs the WASM engine,
where such a pattern would be safe. That is deliberate — one policy language,
one behaviour, whichever engine is underneath — but it is a strictness decision
rather than a technical necessity. If you would rather have it as an
`analyzer.py` lint than a hard refusal, that is a small change.

Happy to open this as a PR if you'd like to grant access, or you're welcome to
pull the branch directly.

Dev-next-gen:
Closes #134.

## The gap

`_validate_re2` keeps a policy's `matches` patterns to syntax RE2 can run, so a
policy tested on the pydantic engine cannot mean something else once it ships as
a WASM bundle. That covers what the two engines can **express**. It says nothing
about what they **cost**.

Python's `re` backtracks; RE2 does not. `^(a+)+$` is accepted by both, linear
under RE2, exponential under `re`. Measured through the evaluator's own entry
point (`check_constraints`), with the pattern written as an email allowlist:

```
matches(args.email, "^([a-zA-Z0-9_-]+)*@corp[.]com$")
```

| argument length | 20 | 24 | 26 | 28 | 30 | 32 | 34 |
|---|---|---|---|---|---|---|---|
| time to decide one call | 0.04s | 0.54s | 2.21s | 8.74s | 37.8s | 149s | 596s |

Lengths 20-28 were timed through `check_constraints`; 30 and beyond with the
same `re.search` call `_eval_call` makes, since with this change in place the
pattern no longer loads. The rate holds across the range at roughly four times
longer per two characters added (x4.1, x3.9, x4.0), so a 34-character argument
already costs ten minutes. Continuing that slope puts the low forties in the
hours — that last step is an extrapolation, not a measurement. `hexgate/security/` has no
timeout and no argument-size cap, so the component deciding whether a tool call
is allowed can be stalled by the argument it exists to filter — and those
arguments are model-written and caller-influenced.

This is reachable wherever the pydantic engine runs, which per `source.py`
includes the documented local mode and the "platform served no compiled bundle"
fallback, not only development.

## The change

Patterns are refused at **policy load**, where the RE2 syntax check already
lives, so nothing changes on the decision path: the check sits inside the
already-cached `parse_constraint`, and its worst case over the patterns this
repo ships measured 3ms, once.

Detection is exponential ambiguity — whether some state of the automaton can be
reached from itself, over one word, by two distinct paths. That is precisely
the property RE2's construction rules out, which is why RE2 can promise linear
time, so the check tests the divergence itself rather than a syntax that
correlates with it.

A syntactic "nested quantifier" rule would be wrong in both directions, and both
cases are in the test corpus:

- it would **reject** `^([a-z0-9-]+[.])+corp[.]com$` — nested, and linear,
  because the mandatory `[.]` forces one split per iteration;
- it would **miss** `^(a|aa)+$` — not nested, and exponential.

Three verdicts, each with a defined outcome: a witness refuses the pattern, a
proof of linearity accepts it, and an exhausted search budget refuses it as
unverifiable rather than guessing. The error names the pattern and an example:

```
regex '^(a+)+$' in 'matches(args.v, "^(a+)+$")' can backtrack exponentially:
the segment 'aa' can be matched more than one way, so on an input like
'aaaaaaaaaa!' the number of combinations Python's engine must try doubles as
the input grows, while the WASM engine stays linear. Rewrite the repetition so
each iteration has only one way to match (a required separator between
iterations, or a single quantifier instead of a nested one)
```

### Wide repetition bounds

A bound wider than 4 is read as unbounded. Unrolling a bound exactly leaves an
automaton with no cycle, and "a state reachable from itself two ways" is the
whole basis of the analysis — so exact modelling would certify every bounded
pattern as linear however slow it really is. `^(a{1,10}){1,10}$` spans 100
characters and takes 14.7s on a 40-character argument.

Narrow bounds stay exact, so patterns that can only ever span a few characters
are not swept up: `^(a{1,4}){1,4}$`, and the ordinary `^(\d{1,3}\.){3}\d{1,3}$`
and `^(\w{1,63}\.){1,10}[a-z]{2,6}$` shapes.

## A second, smaller finding

Two constructs pass `_validate_re2` on `main` and are undefined under RE2:
**atomic groups** `(?>a+)+` and **possessive quantifiers** `a*+`. Verified with
the method the existing comment cites:

```
$ opa eval 'regex.match("(?>a+)+", "aaa")'
{}                     # undefined under RE2
>>> re.search(r"(?>a+)+", "aaa")
<re.Match object; span=(0, 3), match='aaa'>
```

They are the same divergence class the existing list covers; they were missing
because a source-text rule cannot tell `a*+` from `\++`, a repeated literal
plus. Detected from the parse tree instead.

Fixing this is required by the change either way: without it, those two patterns
escaped as an unexpected exception type instead of `ConstraintParseError`, which
the grammar's no-crash guarantee forbids.

## Validation

- **Every flagged pattern was confirmed**: for 50 randomly generated patterns
  the check refused, an input was hunted that measurably explodes. 50/50
  confirmed, no false positive left standing. One needed a hand-built string
  (`'0-' + 'a'*12 + '!'`: 22k steps at 8 characters, 114k at 10, budget
  exhausted at 12) that the generic generator never produced.
- **And the reverse direction**: of 118 generated patterns, 27 were passed as
  linear and each was then attacked with anchored, pumped inputs drawn from its
  own alphabet. None exploded — 0 false negatives.
- Both directions were re-run after every change to the model, since tightening
  one moves the other. Measurement uses a step-bounded backtracking simulator
  rather than wall-clock, so the comparison is deterministic and cannot hang.
- The corpus holds every `matches` pattern this repo ships (docs, tests,
  examples); all still load.
- New module at 97.0% branch coverage; the six uncovered lines are defensive
  guards for constructs the parser does not produce in practice.
- The corpus covers wildcards and negated classes too: `^(.+)+$` and
  `^([^,]+)+$` are refused, while `^.+$`, `^[^,]+$` and
  `^([^,]+,)+[^,]+$` load as before.
- Full SDK suite: identical pass/fail sets before and after. This machine has 24
  pre-existing failures on `main` — bundle-hash tests that mismatch under
  Windows line endings — and the same 24 after, with no additions.
- `test_engine_parity.py` and `test_rego_compile.py` ran against a real `opa`
  binary rather than skipping.

## What this does not cover

Cost that grows **polynomially** rather than by doubling is out of scope, and
that is a decision rather than an oversight.

| pattern | 1 000 chars | 2 000 chars | 4 000 chars |
|---|---|---|---|
| `^a*a*a*$` | 1.06s | 8.55s | 67.6s |

Two different repetitions absorbing the same text; no crafted string needed,
just a long argument, and RE2 stays linear on it. A detector for it was written
and then dropped from this change, because its precision could not be
demonstrated: it reports the *structure*, and the structure is not always
reachable as a slow input. `^.*foo.*$` has it and is linear (measured flat to
8 000 characters), because the pattern matches as soon as `foo` appears, so no
input both exercises the ambiguity and fails.

Closing that needs the automaton to model acceptance, so the analysis can
require a failing continuation after the pumped section. That is a larger change
and belongs in its own review. It is written up in the ADR with the numbers.

Two further limits, stated rather than implied: anchors (`^`, `\b`) are treated
as consuming nothing, which is an over-approximation I did not find a failing
case for but have not proved safe; and the 4-character bound threshold is a
judgement call, not a derived constant.

## One judgement call worth your view

The refusal applies even to a deployment that only ever runs the WASM engine,
where such a pattern would be safe. That is deliberate — one policy language,
one behaviour, whichever engine is underneath — and the rewrite is usually
trivial (`(a+)+` → `a+`). But it is a strictness decision rather than a
technical necessity, and if you would rather have it as an `analyzer.py` lint
than a hard refusal, that is a small change to make.

### DIFF
--- a/docs/adr/R-CONSTRAINT-001-reject-exponential-regex.md
+++ b/docs/adr/R-CONSTRAINT-001-reject-exponential-regex.md
@@ -1,0 +1,137 @@
+# R-CONSTRAINT-001: `matches` regexes must be linear-time, not just RE2-shaped
+
+**Status:** Accepted · 2026-08-24
+**Applies to:** `hexgate/security/constraints.py`, `hexgate/security/regex_safety.py`
+
+## Decision
+
+A constraint's `matches` pattern is refused at **policy load** when it is
+**exponentially ambiguous** — when some state of its automaton can be reached
+from itself, over one word, by two distinct paths. Detection is exact (a search
+over the product automaton), not a syntactic "nested quantifier" rule, and the
+error carries a concrete example string plus the rewrite to apply.
+
+Three verdicts, and each has a defined outcome:
+
+| verdict | meaning | outcome |
+|---|---|---|
+| witness | two distinct paths found | refuse, showing the example |
+| `False` | provably linear | accept |
+| `None` | search budget exhausted | refuse as unverifiable |
+
+## Why
+
+`_validate_re2` already refuses regex *syntax* RE2 cannot run, so a policy
+cannot mean one thing on the pydantic engine and another in a WASM bundle. That
+covers what the engines can **express**; it says nothing about what they
+**cost**.
+
+Python's `re` backtracks. RE2 does not. `^(a+)+$` is accepted by both, is linear
+under RE2, and is exponential under `re`. Measured on the evaluator's own entry
+point (`check_constraints`), with the pattern as an email allowlist
+`^([a-zA-Z0-9_-]+)*@corp[.]com$`:
+
+| argument length | 20 | 24 | 26 | 28 | 30 | 32 | 34 |
+|---|---|---|---|---|---|---|---|
+| decision time | 0.04 s | 0.54 s | 2.21 s | 8.74 s | 37.8 s | 149 s | 596 s |
+
+Lengths 20-28 were timed through `check_constraints`; 30 and beyond with the
+same `re.search` call `_eval_call` makes, since with this change in place the
+pattern no longer loads. The rate holds across the whole range at roughly four
+times longer per two characters added (x4.1, x3.9, x4.0), so a 34-character
+argument already costs ten minutes. Continuing that slope puts the low forties
+in the hours — that last step is an extrapolation, not a measurement.
+
+The arguments a constraint tests are written by a model and influenced by the
+caller, and `hexgate/security/` carries no timeout or argument-size cap — so
+the component deciding whether a tool call is allowed can be stalled by the
+input it exists to filter. The local and pydantic-fallback modes are supported
+deployment shapes (`source.py`), not just dev conveniences, so this is reachable
+in production.
+
+Exponential ambiguity is the right property to test because it is *exactly* what
+RE2's construction rules out — which is why RE2 can promise linear time. Testing
+for it tests the divergence itself rather than a syntax that correlates with it.
+
+## Not covered
+
+Cost that grows *polynomially* rather than by doubling is **out of scope here**,
+and deliberately so rather than by oversight.
+
+Measured, on the same engine entry point:
+
+| pattern | 1 000 chars | 2 000 chars | 4 000 chars |
+|---|---|---|---|
+| `^a*a*a*$` | 1.06s | 8.55s | 67.6s |
+
+Two different repetitions can absorb the same text, so the engine tries every
+way to divide the input between them. It needs no crafted string, only a long
+argument, and RE2 stays linear on it, so it is the same parity gap one tier
+down.
+
+A detector for it was written and then dropped from this change, because it
+could not be shown to be precise: it reports the *structure*, and the structure
+is not always reachable as a slow input. `^.*foo.*$` has it and runs in linear
+time (measured flat to 8 000 characters) because the pattern matches as soon as
+`foo` appears, so no input both exercises the ambiguity and fails. Reporting it
+would mean warning about patterns that are fine.
+
+Closing that gap needs the automaton to model *acceptance*, so the analysis can
+require a failing continuation after the pumped section. That is a larger change
+and belongs in its own review, not bundled here.
+
+## Consequences
+
+- A pattern that is safe under RE2 but exponential under `re` no longer loads,
+  even for a deployment that only ever runs the WASM engine. Deliberate: one
+  policy language, one behaviour, whichever engine is underneath. The rewrite is
+  usually trivial (`(a+)+` → `a+`).
+- The check runs inside the already-cached `parse_constraint`, so it costs
+  nothing per decision.
+- `None` (budget exhausted) fails closed. No policy pattern in this repo comes
+  near the budget — every shipped pattern resolves under 20 000 steps against a
+  200 000 default.
+- An epsilon cycle alone does not make a pattern exponential and is not treated
+  as ambiguity. `(a?)+` has one (an empty iteration) but every iteration matches
+  exactly one character, so there is a single way to split the input and it runs
+  linearly; `(a*)*` has the same cycle *and* iterations that can absorb any
+  number of characters, so it is exponential. Routes are therefore counted
+  without reusing an edge, which is what separates the two.
+- A repetition bound wider than 4 is read as unbounded. Unrolling a bound
+  exactly leaves an automaton with no cycle, and the analysis looks for a state
+  reachable from itself — so exact modelling would certify every bounded pattern
+  as linear however slow it is (`^(a{1,10}){1,10}$` spans 100 characters and
+  takes 14.7s on a 40-character argument). Narrow bounds stay exact, so
+  `^(a{1,4}){1,4}$`, `^(\d{1,3}\.){3}\d{1,3}$` and `^(\w{1,63}\.){1,10}[a-z]{2,6}$`
+  are not swept up.
+- The analysis models the RE2 subset the grammar allows. A construct it cannot
+  model raises `UnsupportedRegex` rather than answering "linear".
+
+## Rejected alternatives
+
+- **A syntactic "nested quantifier" rule.** Wrong in both directions. It rejects
+  `^([a-z0-9-]+[.])+corp[.]com$`, which is nested *and* linear because the
+  mandatory `[.]` forces one split per iteration; and it misses `^(a|aa)+$`,
+  which has no nesting at all and is exponential. Both are in the test corpus.
+- **A runtime timeout.** Python's `re` cannot be interrupted from another
+  thread, `signal.alarm` is Unix-only, and a subprocess per constraint would
+  dominate decision latency. It would also convert a policy bug into a runtime
+  failure on the enforcement path, which is the worst place to discover it.
+- **Capping argument length before matching.** Changes semantics for legitimate
+  long arguments, and the cap would have to be brutal to help: 30 characters
+  already costs 38 seconds.
+- **Swapping `re` for a linear-time engine** (`google-re2`, or a Thompson
+  simulation of the supported subset). The soundest fix and the largest: a new
+  dependency or a second regex engine to keep in parity with RE2 forever. Worth
+  revisiting if `matches` ever needs to accept patterns this ADR refuses; until
+  then, refusing them is the smaller and more reviewable change.
+
+## Verify
+
+```
+pytest tests/security/test_regex_safety.py
+```
+
+passes: every blow-up shape is refused, every `matches` pattern this repo ships
+still loads, and each reported example is a string the engine genuinely fails to
+match (which is what forces it to try every split).
--- a/docs/policy/writing-conditions.mdx
+++ b/docs/policy/writing-conditions.mdx
@@ -95,6 +95,25 @@ pattern anywhere in the string. Add `^` and `$` when you mean the *whole* value,
 like the ticket-ID example above.
 </Tip>
 
+<Warning>
+A pattern must be one a regex engine can match in **linear time**. Policies are
+refused at load when a repetition can match the same text more than one way —
+`^(a+)+$`, `^([a-z]+)*$`, `^(a|aa)+$` — because the engine then has to try every
+split. A 34-character argument already takes ten minutes to decide, and the cost
+keeps quadrupling every two characters. The error names the pattern and an
+example string.
+
+The fix is to leave only one way to match each repetition: `^(a+)+$` is just
+`^a+$`, and a repeated group needs a separator that cannot be absorbed by the
+group itself — `^([a-z0-9-]+[.])+corp[.]com$` is fine, because the `[.]` ends
+every iteration.
+
+A milder version of the same problem is not refused today: two repetitions that
+can both absorb the same text, like `^a*a*a*$`, still load, and matching cost
+then grows with the square or cube of the argument length rather than doubling.
+Worth avoiding, but it will not stop a policy from loading.
+</Warning>
+
 ## Rules over a list
 
 When an argument is a **list**, `every` and `any` let you constrain its
--- a/hexgate/security/constraints.py
+++ b/hexgate/security/constraints.py
@@ -84,6 +84,10 @@ from functools import lru_cache
 from typing import Any
 
 from hexgate.security.errors import PolicyDeniedError
+from hexgate.security.regex_safety import (
+    UnsupportedRegex,
+    find_exponential_ambiguity,
+)
 
 
 class ConstraintParseError(ValueError):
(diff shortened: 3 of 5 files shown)


victorludvig:
Nice finding and excellent write-up, but I think we should revisit the approach before merging.

Out of curiosity, did you hit this in practice while testing Hexgate, or did it come from analyzing the code? If there's a real pattern that triggered it, that's worth knowing to evaluate the priority.

The bug is narrower than the fix. The pydantic engine only runs when WASM compilation failed ([service.py#L132-L139](https://github.com/HexamindOrganisation/hexgate/blob/fb195cd3ce2f159b0695c36ebc3482f7e1c8f9ff/platform/api/hexgate_api/features/agents/service.py#L132-L139) → [source.py#L207](https://github.com/HexamindOrganisation/hexgate/blob/fb195cd3ce2f159b0695c36ebc3482f7e1c8f9ff/hexgate/security/source.py#L207)) or in local mode, and it needs a developer to have written a pattern like `(a+)+` where a repetition can end anywhere. Real allowlists almost always have a delimiter between repetitions (`\.` in domains, `/` in paths, `@` in emails) that pins each boundary, which prevents backtracking. The analyzer, on the other hand, runs on every path, WASM included.

An automated review reproduced these on the PR head:

- Any `{m,n}` with n > 4 is treated as unlimited, so `^(a{1,4}){1,5}$` is refused while `^(a{1,5}){1,4}$` is accepted — both are ≤ 20 chars and match instantly. Ordinary shapes like `^(?:[a-z0-9-]{1,63}\.){1,5}corp\.com$` fail to load for the same reason.
- `.` is assumed to match `\n`, which it doesn't without `DOTALL`. In `^(.+\n)+$` the `\n` is really a delimiter that `.+` can't cross, so the pattern is linear — but the analyzer thinks `.+` can swallow it and refuses the pattern, while the equivalent `^(?:[^\n]+\n)+$` passes.
- The check itself is super-linear and runs synchronously in `api_validate_policy`: `^` + `a?`×40 + `$` takes 3.5 s, `a?`×60 33 s — one validate call stalls the event loop for every tenant.

Fixing those properly is a lot more work, and the analyzer would still miss `^a*a*a*$`.

Proposal: evaluate `matches` with `google-re2` in the pydantic engine. Same engine Rego runs, so `_RE2_INCOMPATIBLE`, `_validate_re2` and this module go away. Installation isn't a concern: google-re2 publishes [pre-built wheels](https://pypi.org/project/google-re2/1.1.20251105/#files) for Python 3.13 and 3.14 on macOS, Linux and Windows (arm64 and x86_64), so developers never compile anything. We already ship native packages (`wasmtime`, `cryptography`, `biscuit-python`).

If we'd rather not add the dependency, the blunt alternative is to refuse any repetition nested inside a repeated group (`(a+)+`, `(\w*)*`…) and cap the argument length. It wrongly refuses some valid patterns like `([a-z]+\.)+` and misses `(a|aa)+`, but it's ~20 lines, can't hang or crash, and the restriction is easy to document. Still simpler than the analyzer.

Waiting on @quanghexa94's opinion.

quanghexa94:
Thanks @victorludvig — agree we should revisit before merging, and I lean toward the `google-re2` route too. Dropping `_validate_re2` + this whole module for the same engine Rego already runs is a real simplification, and the prebuilt 3.13 wheels make it painless.

A few things not yet raised, mostly on the docs side:

**Docs**
- The user-facing page `docs/policy/writing-conditions.mdx` copies the ADR's benchmark rationale (the "34-char → 10 min, quadruples every 2 chars" line) and documents the current `^a*a*a*$` gap. Both are internal/rationale, and the second is expected to change — so it'll go stale on the public site. I'd keep just the actionable rule + fix there and link to the ADR for the why, so those facts live in one place.
- The ADR is a great write-up but a bit long for an ADR — `Consequences` / `Not covered` mostly explain the detector's internals (epsilon cycles, bound-unrolling, the product-automaton search). Those read more like code comments; the ADR could stay at decision + why.

**Couple of code notes (only matter if we keep the analyzer):**
- A fully-negated class gives an empty charset, and `_pick_char` does `charset[0][0]` → `IndexError` instead of `ConstraintParseError`, so a crash can escape policy load.
- Besides the super-linear case you measured, the NFA build itself is unbudgeted — it unrolls bounded repeats per nesting level, so nested `{100}` (e.g. `((a{100}){100}){100}`) explodes before the product-search budget even engages.
- Minor: some rejection messages print an example that actually matches fine (`^(.+)+$` → `aaaaaaaaaa`), so it doesn't reproduce the blow-up it's warning about.

If we go with re2 most of this (and the module) disappears anyway. 👍

Dev-next-gen:
Thanks both — this is a better read of the problem than mine, and I agree with the re2 direction.

**To your question:** code analysis, not a case I hit while running Hexgate. I was reading `constraints.py`, saw that `_validate_re2` rejects RE2-incompatible *syntax* with the explicit rationale that the engines must not diverge, and noticed nothing covered *cost*. So treat the priority accordingly: I have no evidence of a real policy triggering it. Your point that real allowlists carry a delimiter that pins each boundary matches what I found — every `matches` pattern in the repo is linear, and the corpus in the PR is built from them.

**On the defects.** I reproduced four of the six on the PR head:

- The bound inconsistency is real: `^(a{1,4}){1,5}$` refused, `^(a{1,5}){1,4}$` accepted, both instant. `_UNROLL_CAP` is applied to `max` alone, which is arbitrary.
- `.` is real too. I mapped `ANY` to the full codepoint range, so `^(.+\n)+$` loses the delimiter that makes it linear.
- The build is unbudgeted, as you say: `^((a{100}){100}){100}$` takes 7.5s before the product search ever starts.
- **The super-linear check is the one that decides it.** I measure 5.8s for `a?`×40 (0.18s at 20, 1.3s at 30), so worse than your 3.5s. A validation endpoint that blocks for seconds on an adversarial policy is the same failure I was trying to remove, moved one layer up. I only ever benchmarked the check against *benign* patterns, which is how I missed it.

Two I could not reproduce on `e24a173`, in case they came from an earlier commit: `^(?:[a-z0-9-]{1,63}\.){1,5}corp\.com$` loads fine here, and a fully-negated class (`[^\s\S]`) returns "linear" rather than raising `IndexError`. Happy to look at the repro if you have it.

You're also right that the cost lands everywhere while the bug only reaches the pydantic path. I flagged the strictness as a judgement call in the PR, but I under-weighted it: refusing patterns on WASM-only deployments to fix something that cannot happen there is a bad trade.

**On re2:** I listed it in the ADR's rejected alternatives as "the soundest fix and the largest", and argued the dependency wasn't worth it. With prebuilt 3.13/3.14 wheels next to `wasmtime` and `biscuit-python`, that argument is weak. It also removes the syntax divergence, not just the cost one — `_RE2_INCOMPATIBLE`, `_validate_re2` and this module all go.

Happy to do that work: a PR that swaps `matches` to `google-re2` in the pydantic engine. I'd keep the adversarial corpus from this branch as parity tests, since it's a ready-made set of patterns where the two engines used to disagree. I'll leave this PR open until you've settled the direction rather than pre-empt it — say the word and I'll close it in favour of the re2 one.

@quanghexa94 agreed on the docs — the benchmark rationale belongs in the ADR only, and the ADR should stop at decision + why. And you're right that a rejection example that matches is a bug in its own right, not just a caveat.

Dev-next-gen:
Closing this in favour of #139, which takes the `google-re2` route you both
preferred.

Your read was right, and reproducing it made that clear rather than just
plausible. Four of the six points landed on my head:

- the bound handling was inconsistent — `^(a{1,4}){1,5}$` refused while
  `^(a{1,5}){1,4}$` was accepted, both instant;
- `.` was mapped to the full codepoint range, so `^(.+\n)+$` lost the delimiter
  that makes it linear;
- the automaton build was unbudgeted: `^((a{100}){100}){100}$` took 7.5s before
  the product search even started;
- and the check was super-linear. I measure 5.8s for `a?`×40, worse than the
  3.5s you saw. A validation endpoint that blocks for seconds on an adversarial
  policy is the same failure I was trying to remove, moved one layer up. I only
  ever benchmarked the check against benign patterns, which is how I missed it.

Two I could not reproduce on the PR head, in case they came from an earlier
commit: `^(?:[a-z0-9-]{1,63}\.){1,5}corp\.com$` loaded fine, and a fully-negated
class returned "linear" rather than raising.

The deeper point is the one I had in the ADR's rejected alternatives and argued
my way past: the root cause was never the patterns, it was two engines being
kept in agreement by hand. #139 deletes that difference instead of policing it —
`_RE2_INCOMPATIBLE` and the whole detector go, 1 002 lines of new code become 50,
and `^a*a*a*$`, which my analysis missed entirely, decides a 4 000-character
argument in 0.03ms.

Thanks for taking the time on this one.

PR #136, opened 2026-08-26
refactor(platform-api): aggregate schema gaps across features at startup

victorludvig:
## What is Changing

- New `verify_all(client, checks)` in `core/clickhouse.py`: runs every feature's `verify_schema`, collects each `SchemaOutOfDate.missing`, and raises a single `SchemaOutOfDate` naming every gap. Only `SchemaOutOfDate` is collected — anything else escaping a wrapper is a bug and still surfaces as one.
- `main.py` lifespan calls `verify_all(get_clickhouse(), (verify_audit_schema, verify_llm_schema))` instead of the two checks in sequence.

Follow-up to the review on #131. The two per-feature checks ran one after another, so a volume behind on both audit and `llm_invocation` reported only the audit gaps; the operator would migrate, restart, and fail again on the llm gap. Since #131's error now points at per-table migrations before a volume reset, that two-boot cycle is the expected path, not an edge case.

Aggregating in a shared helper (rather than merging the column lists) keeps the feature boundary #131 set up: `main.py` lists which features it writes, and no feature learns another's tables. The enricher job (#133) runs the same two checks at its own startup and will switch to `verify_all` when it rebases.

## Tests

- `tests/core/test_clickhouse.py`: happy path (all checks run, no raise); two stale features → one exception whose `.missing` is the union; a stale first check does not short-circuit the later ones; a non-`SchemaOutOfDate` failure is not swallowed.
- Full platform-api suite: 515 passed, 7 skipped. Integration (`-m integration`, live ClickHouse): 6 passed. `ruff check` / `ruff format --check` clean.

### DIFF
--- a/platform/api/hexgate_api/core/clickhouse.py
+++ b/platform/api/hexgate_api/core/clickhouse.py
@@ -7,7 +7,7 @@ from __future__ import annotations
 
 import logging
 import re
-from collections.abc import Sequence
+from collections.abc import Callable, Sequence
 from dataclasses import KW_ONLY, dataclass
 from functools import lru_cache
 from typing import Generic, Protocol, TypeVar
@@ -194,5 +194,27 @@ def verify_written_columns(
         gaps = sorted(set(columns) - present)
         if gaps:
             missing[table] = gaps
+    if missing:
+        raise SchemaOutOfDate(missing)
+
+
+def verify_all(client: Client, checks: Sequence[Callable[[Client], None]]) -> None:
+    """Run every feature's ``verify_schema`` and raise one
+    :class:`SchemaOutOfDate` naming every gap.
+
+    Calling the checks one after another would stop at the first stale
+    feature, so a volume behind on two tables reports one, gets migrated,
+    and fails again on the next boot. Aggregating here keeps the per-feature
+    wrappers ignorant of each other's tables while giving the operator the
+    whole list in a single boot. Only :class:`SchemaOutOfDate` is collected:
+    the wrappers already degrade unreachable/denied schemas to warnings, so
+    anything else escaping is a bug and should surface as one.
+    """
+    missing: dict[str, list[str]] = {}
+    for check in checks:
+        try:
+            check(client)
+        except SchemaOutOfDate as exc:
+            missing.update(exc.missing)
     if missing:
         raise SchemaOutOfDate(missing)
--- a/platform/api/hexgate_api/core/clickhouse.py
+++ b/platform/api/hexgate_api/core/clickhouse.py
@@ -175,6 +175,10 @@ def verify_written_columns(
     the schema is not evidence that it is stale, and the error this would
     otherwise raise tells the operator to destroy the volume. Degrading is
     safe — a genuinely broken table makes inserts 503, which the SDK retries.
+
+    Degrading stops the scan but never discards a gap already found: a stale
+    earlier table is evidence on its own, and returning quietly would let the
+    process boot and drop every insert into it.
     """
     missing: dict[str, list[str]] = {}
     for table, columns in tables:
@@ -184,12 +188,12 @@ def verify_written_columns(
             _log.warning(
                 "ClickHouse unreachable during %s schema check: %s", table, exc
             )
-            return
+            break
         except DatabaseError as exc:
             if _server_error_code(exc) != _UNKNOWN_TABLE_CODE:
                 # ACCESS_DENIED on a scoped grant, quota, metadata blip…
                 _log.warning("cannot verify the %s schema: %s", table, exc)
-                return
+                break
             present = set()
         gaps = sorted(set(columns) - present)
         if gaps:
@@ -215,6 +219,9 @@ def verify_all(client: Client, checks: Sequence[Callable[[Client], None]]) -> No
         try:
             check(client)
         except SchemaOutOfDate as exc:
-            missing.update(exc.missing)
+            # Merge, don't clobber: two checks naming the same table must
+            # both surface, or the operator is back to one gap per reboot.
+            for table, columns in exc.missing.items():
+                missing[table] = sorted(set(missing.get(table, ())) | set(columns))
     if missing:
         raise SchemaOutOfDate(missing)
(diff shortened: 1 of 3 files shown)


platform/api/hexgate_api/core/clickhouse.py
```diff
@@ -196,3 +196,25 @@ def verify_written_columns(
...
+def verify_all(client: Client, checks: Sequence[Callable[[Client], None]]) -> None:
+    """Run every feature's ``verify_schema`` and raise one
+    :class:`SchemaOutOfDate` naming every gap.
+
+    Calling the checks one after another would stop at the first stale
+    feature, so a volume behind on two tables reports one, gets migrated,
+    and fails again on the next boot. Aggregating here keeps the per-feature
+    wrappers ignorant of each other's tables while giving the operator the
+    whole list in a single boot. Only :class:`SchemaOutOfDate` is collected:
+    the wrappers already degrade unreachable/denied schemas to warnings, so
```
quanghexa94, line 214: **Worth fixing before merge — the aggregation guarantee has a hole.**

This docstring states the invariant `verify_all` relies on: a wrapper either raises `SchemaOutOfDate` or degrades unreachable/denied schemas to warnings, and "anything else escaping is a bug." But `verify_written_columns` can **return without raising even after it has already recorded a real gap** — which silently defeats the guard.

Scenario: audit's `verify_schema` checks `policy_decision` then `ban_enforcement` in one call. `policy_decision` has a genuine missing column (recorded in `missing`), then `DESCRIBE ban_enforcement` raises a non-`UNKNOWN_TABLE` `DatabaseError` — e.g. `ACCESS_DENIED` on a scoped grant, or a transient quota blip. `verify_written_columns` hits its early `return` on that error, so `SchemaOutOfDate` is never raised, the `policy_decision` gap is discarded, the API boots, and every `policy_decision` insert is then silently dropped by the SDK — the exact failure this guard exists to prevent.

The root cause is the early `return` in `verify_written_columns` (unchanged in this PR, but this PR is what makes the "must raise to aggregate" contract load-bearing). Fix: don't discard `missing` when a later table in the same call degrades — raise `SchemaOutOfDate(missing)` if anything was already accumulated before returning/degrading. Worth a test that records a gap on table A, then throws a non-UNKNOWN_TABLE error on table B, and asserts A's gap still surfaces.

quanghexa94 (commented):
Nice — this is the aggregation I hoped for on the earlier PR: one boot names every gap instead of one-per-restart, and the per-feature wrappers stay ignorant of each other's tables. The refactor is clean and the new tests (happy path, cross-feature aggregation, continue-after-stale, non-swallowing of unexpected exceptions) cover the design well.

One genuine bug I'd fix before merge (#1 below), plus two latent/robustness notes that are safe to defer. Details inline.

platform/api/hexgate_api/core/clickhouse.py
```diff
@@ -196,3 +196,25 @@ def verify_written_columns(
...
+    anything else escaping is a bug and should surface as one.
+    """
+    missing: dict[str, list[str]] = {}
+    for check in checks:
+        try:
+            check(client)
+        except SchemaOutOfDate as exc:
+            missing.update(exc.missing)
```
quanghexa94: Non-blocking (latent): `dict.update` clobbers rather than merges. If two feature checks ever report gaps for the *same* table name, `missing.update(exc.missing)` keeps only the last check's column list and drops the first's. That would reproduce the very multi-boot loop this PR eliminates: the operator migrates the surviving columns, reboots, and the other check's columns are still missing.

Harmless today — the two features write disjoint tables — but cheap to make robust now: `missing.setdefault(table, [])` then merge + dedupe the column lists, rather than `update`.

platform/api/hexgate_api/main.py
```diff
@@ -162,9 +162,9 @@ async def lifespan(app_: FastAPI):
     else:
         # Behind this build, every insert is rejected and dropped by the SDK —
         # silently, from this side. Refuse to boot so the previous deployment
-        # keeps serving. Each feature checks the tables it writes.
-        verify_audit_schema(get_clickhouse())
-        verify_llm_schema(get_clickhouse())
+        # keeps serving. Each feature checks the tables it writes; combined so
+        # one boot names every gap rather than one per restart.
+        verify_all(get_clickhouse(), (verify_audit_schema, verify_llm_schema))
```
quanghexa94, line 167: Non-blocking (robustness): the check set is a hardcoded tuple here, so a new event-table feature is protected only if someone remembers to add its `verify_schema` to this line. Forget it and that table's schema is silently unguarded — stale schema never caught, inserts dropped by the SDK. A small registry / auto-discovery of feature `verify_schema` functions would make the guard complete by construction rather than by remembering to edit this call site. Fine to leave for now with two features.

victorludvig: Agreed, leaving as-is for now. Plan for a follow-up refactor once #133 rebases (second call site): declare each table once as a WrittenTable(name, columns) in core/clickhouse.py that registers itself, and make both the single-row and batch insert helpers take it instead of a bare table name, so a table can't be written without being registered. verify_all(client) then checks every registered table, no hardcoded list.

One deliberate shift from #131: the guard becomes "every table this build declares" rather than "the tables this process writes". A job that only writes llm_invocation would still refuse to boot on a stale policy_decision. Stricter, but one build ships one schema contract (the error already says "behind this build"), and a per-process registry would be fragile anyway: any shared import silently widens the check set.

platform/api/hexgate_api/core/clickhouse.py
victorludvig: Fixed by updating the dict with the sorted set of missing table columns: 
`merged = set(missing.get(table, ())) | set(columns)`

Also added a unit test.

platform/api/hexgate_api/core/clickhouse.py
victorludvig: Fixed. The two early returns in verify_written_columns are now break, so gaps found before a degrade still raise. Test added for the A-stale / B-denied ordering.

victorludvig (commented):
1-2 fixed, 3 will be a follow-up PR once the enricher redpanda->clickhouse job PR is merged (#133

PR #137, opened 2026-08-26
feat(sdk): agent-level admission enforcement (AgentGate, closed-world agent keys)

quanghexa94:
Add agent-level admission enforcement: gate whether a caller, in their role, may run an agent at all, decided at run entry before the model sees anything. This folds in the former #124 (AgentGate runtime) and the model reconcile it depends on, so it lands as one coherent feature instead of a stacked pair. Pure SDK, no platform or dashboard change, no migration. It is the A2 step of the agent-level enforcement series (A1, the `admission`/`agents` lowering, is already merged); A3 (handoff/delegation interception) and A4 (depth cap) follow.

This supersedes #124, which was built on an opt-in admission model that review showed had two fail-opens. Rather than patch them in the runtime, the fix is to reconcile agent-level rules with the boundary/capability module model from #115 (R-POL-001) and make agent keys closed-world.

## Design

Agent-level rules (`admission` / `agents`) lower to synthetic `agent.run` / `agent.tool:<t>` / `agent.handoff:<t>` keys that both engines already read through `effective_tools`. This PR makes those keys first-class in the linker fold, closed-world at the engines, and cleanly separates two things #124 had tangled: how rules compose vs. whether the gate is engaged at all.

- Agent keys fold through the linker, exactly like tool keys. A boundary agent-deny is authoritative and role-independent, a ceiling boundary makes an unlisted agent key ineligible, capabilities grant and union, a capability that denies an agent key is a `LinkError`. `admission` / `agents` become allowed module fields.
- Agent keys are closed-world. An unlisted agent key (admission and reach) denies regardless of `default_policy`, so a silent co-role can neither admit an unknown caller nor defeat an explicit deny. This drops `agent.run`'s old opt-in-allow (the `get_tool_policy` fallback and the per-role Rego rule), removing both fail-opens by construction.
- Engagement is opt-in and derived, separate from composition. `PolicyEngine.declares_admission()` / `declares_reach()` decide whether the gate fires: the pydantic engine derives it from whether any resolved role carries the key; the WASM bundle reads an `agent_gating` block recorded in the signed manifest at build time. An agent whose policy never mentions admission is never gated and runs exactly as before, which is what keeps closed-world from locking out agents that do not use the feature.
- `AgentGate` is the runtime, mirroring the egress `Gate`: it reuses `PolicyEnforcer.decide` on the `agent.run` key at run entry, folds approval, and fails closed with `AgentNotAdmittedError`. It is wired in `HexgateAgent` by `enforce_policy` alongside the binding (local, no control plane), threaded through `with_tools` rebuilds, and cleared on the unguarded path. Whether it refuses is decided per run from the current policy, so a hot-reload that adds or drops admission is honored on the next run.

Full rationale and rejected alternatives in `docs/adr/R-AGENT-002`.

## Important files

| File | What to look at |
| --- | --- |
| `hexgate/security/agent_gate.py` | the runtime gate: `check_admission[_async]` reads `declares_admission()` per run, decides `agent.run`, folds approval, fails closed |
| `hexgate/agents/factory.py` | how the gate is wired: attached in `enforce_policy`, threaded through `with_tools`, cleared on the unguarded/guards-only paths, checked at `ainvoke`/`astream_events` entry |
| `hexgate/security/linker.py` | agent keys fold through `effective_tools` in `_fold_tool` / `_tool_names` / `_reject_capability_denies`; `admission`/`agents` in `_MODULE_COMPOSABLE_FIELDS`; resolved policy via `AgentPolicy.resolved()` |
| `hexgate/security/policy.py` | `get_tool_policy` fallback collapsed: every unlisted agent key denies closed-world |
| `hexgate/security/rego.py` | removed the per-role `agent.run` opt-in rule; catch-all excludes exactly the reserved agent keys, not a broad `agent.` prefix |
| `hexgate/security/decision.py`, `policy_set.py`, `bundle.py`, `models.py` | `declares_admission()`/`declares_reach()` on the protocol (pydantic from `effective_tools`, WASM from the manifest); `AgentPolicy.resolved()` builder |
| `docs/adr/R-AGENT-002-...md` | the decision, why, rejected alternatives |
| `tests/security/test_agent_gate.py` | admission verdicts + the multi-role closed-world/permissive-union case |
| `tests/security/test_linker.py`, `test_agent_policy.py`, `test_rego_compile.py` | the behaviour flips: boundary agent-deny composes, capability agent-deny is a LinkError, unlisted `agent.run` denies, no per-role admit rule |

## Test plan

Ran the SDK suite against the hexanlp-demo env: `pytest tests/ -q` gives 2091 passed, 6 skipped (opt-in integration), 20 deselected. `ruff check` clean over `hexgate/`, `tests/`, and the demo; `ruff format --check` clean.

By hand I checked the semantics that changed: (1) a boundary `agents: {x: deny}` folds to a hard deny on `agent.handoff:x` in `effective_tools`, and a capability `agents: {x: deny}` raises `LinkError`; (2) an unlisted `agent.run` under `default_policy: allow` denies on both the pydantic and compiled-WASM paths (parity case in `test_closed_world_agent_parity_pydantic_vs_wasm`); (3) the golden Rego fixture lost exactly the three per-role `agent.run` allow rules and nothing else; (4) the multi-role gate: `admin` (granted) admits, `viewer` (no grant, gate engaged because `admin` declares admission) is refused, `contractor` (`admission: deny`) is refused, `[contractor, admin]` admits via the permissive union, `[viewer, contractor]` is refused. A reviewer catches a regression through the pydantic-vs-WASM parity tests, the golden snapshot, and the multi-role gate test.

## Try it

```
# marimo admission demo against an inline policy (support/billing admitted,
# contractor + unroled refused, closed-world):
uv run marimo edit deploy/agent_gate_demo.py

# the model + linker changes:
uv run pytest tests/security -q
```

## Notes

- No platform or dashboard change; the only manifest touched is the signed WASM bundle manifest (`build_signed_bundle`), which gains an `agent_gating` block. Older bundles without it read `declares_admission()` as False, which is safe (they predate agent-level policy).
- Closes/supersedes #124. Supersedes the opt-in/closed-world half of R-AGENT-001; its engagement-flag and manifest mechanism survive, now uniform for admission and reach.
- Behaviour change to be aware of when authoring: once a policy declares `admission` on any role, admission is enforced for every caller (closed-world), so grant it on each role that should be admitted (e.g. `admission: allow` on `default` for broad access). An agent that declares no admission is unaffected.
- Follow-ups: A3 (handoff/delegation interception, which consumes `declares_reach()`) and A4 (depth cap). Design doc: `docs/adr/R-AGENT-002-agent-keys-compose-through-the-linker.md`.

### DIFF
--- a/deploy/agent_gate_demo.py
+++ b/deploy/agent_gate_demo.py
@@ -1,0 +1,287 @@
+"""Hexgate agent-gate demo (marimo) — gate whole agents by policy, in code.
+
+A code-only tour of agent-level enforcement, the sibling to the egress gate. No
+platform, no API key, no model call. Hexgate normally gates the *tool call* the
+model proposes; this gate sits one level up and asks two questions from the same
+policy engine:
+
+  * admission — may this caller, in this role, *run this agent at all*?
+  * reach     — which *other* agents may it call as a tool or hand off to?
+
+Both lower to synthetic tool keys (`agent.run`, `agent.tool:<name>`,
+`agent.handoff:<name>`), so the same engine decides them through the same path as
+any tool. Admission is enforced at run entry today; the handoff interception in
+the framework adapters is the next slice, but the policy already decides it.
+
+Run with `uv run --with marimo marimo edit deploy/agent_gate_demo.py`.
+"""
+
+import marimo
+
+__generated_with = "0.23.10"
+app = marimo.App(width="medium")
+
+
+@app.cell
+def _():
+    import yaml
+
+    import marimo as mo
+
+    from hexgate import HexgateContext
+    from hexgate.security import (
+        AGENT_RUN_TOOL,
+        AgentNotAdmittedError,
+        agent_target_key,
+        resolve_agent_gate,
+    )
+    from hexgate.security.enforcer import build_enforcer
+    from hexgate.security.policy_set import load_policy_set_from_dict
+
+    return (
+        AGENT_RUN_TOOL,
+        AgentNotAdmittedError,
+        HexgateContext,
+        agent_target_key,
+        build_enforcer,
+        load_policy_set_from_dict,
+        mo,
+        resolve_agent_gate,
+        yaml,
+    )
+
+
+@app.cell
+def _(mo):
+    mo.md(
+        """
+        # 🚪 Hexgate — gating whole agents
+
+        The tool gate answers "may the model run *this tool* with *these args*?".
+        The **agent gate** sits one level up and answers two more, from the same
+        policy engine:
+
+        - **admission** — may this caller, in this role, *start this agent at all*?
+          Checked at run entry, before the model sees anything.
+        - **reach** — which *other* agents may it call as a tool, or hand the whole
+          conversation off to?
+
+        Both lower to synthetic tool keys (`agent.run`, `agent.tool:<name>`,
+        `agent.handoff:<name>`), so there is no second decision path: the same
+        engine, constraints, and audit apply. This demo is **code-only** — no
+        platform, no API key, no model call.
+        """
+    )
+    return
+
+
+@app.cell
+def _(mo):
+    mo.md("""## 1 · The policy""")
+    return
+
+
+@app.cell
+def _(load_policy_set_from_dict, yaml):
+    # A refund agent that can consult a billing-bot. support may run it and call
+    # billing-bot as a tool, but must never hand the customer conversation off.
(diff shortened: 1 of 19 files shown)


quanghexa94:
Folded back into #124, which is the PR already under review. #124 now carries the full feature: the AgentGate admission runtime plus this boundary/capability model reconcile (R-AGENT-002, closed-world agent keys). Continuing there.

PR #139, opened 2026-08-26
feat(sdk): evaluate `matches` with RE2 on the pydantic engine

Dev-next-gen:
Supersedes #135, following the review there.

## Why this instead of #135

#135 detected exponentially-backtracking patterns and refused them at policy
load. @victorludvig and @quanghexa94 showed that approach was the wrong shape,
and they were right on every count I could reproduce:

- the analysis was itself super-linear on adversarial input — `^` + `a?`×40 + `$`
  took 3.5s in `api_validate_policy`, which is the same failure one layer up;
- it still missed polynomial blow-ups like `^a*a*a*$`;
- it refused patterns on WASM-only deployments, where they were never a problem;
- and it had real defects of its own (bound handling, `.` versus `\n`, an
  unbudgeted automaton build).

The root cause was never the patterns. It was that the pydantic engine evaluated
`matches` with a *different engine* from the one Rego runs, so parity had to be
maintained by hand. This PR removes that difference instead of policing it.

Concretely, on the same measurements:

| | #135 | this PR |
|---|---|---|
| `^a*a*a*$` on 4 000 chars | 67.6s (not covered) | **0.03 ms** |
| `^([a-zA-Z0-9_-]+)*@corp[.]com$`, 28-char argument | refused at load | **0.015 ms**, decided |
| validating `^` + `a?`×40 + `$` | 3.5s | **0.33 ms** |
| patterns wrongly refused | some | none |
| new code to maintain | 1 002 lines | 50 |

## What this does

`matches` is evaluated with **RE2** (`google-re2`), the engine Rego already
runs. Patterns compile when the policy loads, so one RE2 cannot run is refused
there with RE2's own reason, and the compiled object is cached for evaluation.

`_RE2_INCOMPATIBLE` — the hand-maintained list of Python constructs to reject —
is removed. RE2 is the judge now.

**Syntax.** The list was incomplete: atomic groups `(?>a+)+` and possessive
quantifiers `a*+` passed it and are undefined under RE2
(`opa eval 'regex.match("(?>a+)+", "aaa")'` returns undefined while Python
matches). A source-text rule cannot reliably catch those, since `a*+` and `\++`
differ only by parse. All ten constructs the list rejected are still rejected —
now by the engine itself, which cannot drift from itself.

**Cost.** `re` backtracks, RE2 does not. Timed through `check_constraints` with
`^([a-zA-Z0-9_-]+)*@corp[.]com$`, deciding one call took 0.04s on a 20-character
argument, 8.74s at 28 and 596s at 34 — about four times longer per two
characters. The verdict was always `DENY`; only reaching it was the problem.

**Dependency.** `google-re2` ships prebuilt wheels for 3.13 and 3.14 on macOS,
Linux and Windows (x86_64 and arm64), verified against the index, so nothing is
compiled on install.

## One behaviour change

`$` now means end-of-value only. Python's `$` also matches just before a
trailing newline, so `^abc$` against `"abc\n"` was true on the pydantic engine
and false in a compiled bundle — `opa eval` confirms false. **The change removes
that divergence rather than introducing one**, but it is a change for anyone
relying on the Python reading, so it is called out in the ADR.

A differential run over 35 patterns × 31 inputs found 29 patterns identical
across both engines and 6 divergent — every divergence being that same
trailing-newline case, and in every one of them `main` disagreed with the
compiled bundle.

## Validation

- Full SDK suite: identical pass/fail sets before and after. This machine has 24
  pre-existing failures on `main` (bundle-hash tests under Windows line endings)
  and the same 24 here, with no additions.
- `test_engine_parity.py` and `test_rego_compile.py` run against a real `opa`
  binary rather than skipping: 262 passing.
- `platform/api`: 482 passing. (`tests/core/test_keystore.py` is flaky on
  Windows — 1 to 3 failures per run — on `main` as well as here; it looks
  unrelated to this change.)
- 40 tests added. Every one of the 50 added lines is covered.
- The test corpus is adversarial on purpose: each pattern in it is one Python's
  engine takes exponential time on, so a regression to `re` would hang the suite
  rather than fail it quietly.

## Also worth a separate look

While auditing the rest of the SDK's regex use, two things came up that are out
of scope here and that I'd rather raise on their own than bundle in:

1. `tools/files/grep.py` compiles a **model-supplied** pattern with `re` and runs
   it over every line of every file in the workspace. `(a+)+$` against one
   31-character line takes 75s; a 41-character one does not finish. A malformed
   pattern also escapes as an uncaught `re.error`.
2. A parity divergence unrelated to regex: Rego's `in` iterates an object's
   *values*, so `every(args.x, …)` over a dict is satisfied in a compiled bundle
   while the pydantic engine fails it closed — production allowing what the local
   engine denies.

I have both fixed and tested locally, with a differential fuzzer that found the
second one. Happy to open them as follow-ups if useful; say the word and I'll
file them.

Thanks for the review on #135 — it changed the approach for the better.


hexgate/security/constraints.py
```diff
@@ -795,9 +819,10 @@ def _eval_call(node: Call, context: dict[str, Any]) -> bool:
     if node.fn == "contains":
         return v in x
     if node.fn == "matches":
-        # re.ASCII pins \d/\w/\s/\b to ASCII, matching Rego's RE2 (Go) engine —
-        # without it Python treats them as Unicode and diverges on non-ASCII args.
-        return re.search(v, x, re.ASCII) is not None
+        # RE2 is the engine Rego runs, so the two engines agree by
+        # construction instead of by keeping a list of Python features to
+        # avoid. It is linear-time, so no pattern can stall the decision.
+        return _compile_re2(v).search(x) is not None
```
victorludvig: Malformed tool-argument edge case (rare, but worth a guard):

A policy constrains a tool argument with matches, e.g. matches(args.ticket, "^INC-[0-9]+$"). If the agent calls the tool with an argument containing a truncated emoji escape, e.g. {"ticket": "INC-42 \ud83d"} (😀 is \ud83d\ude00 in JSON; only the first half got emitted), json.loads accepts this and produces the string 'INC-42 \ud83d', where \ud83d is half a character (a "lone surrogate") that UTF-8 cannot encode.

On main, re.search evaluates it and the constraint fails → DENY. On this branch, re2 must UTF-8-encode the value before matching, so this line raises UnicodeEncodeError instead; evaluate_tool_call only catches PolicyDeniedError, so the enforcer's decide() raises rather than returning a verdict.

Reproduce on this branch:
import json
from hexgate.security.constraints import check_constraints
args = json.loads('{"ticket": "INC-42 \\ud83d"}')
check_constraints(['matches(args.ticket, "^INC-[0-9]+$")'], args, "update_ticket", role="support")
// UnicodeEncodeError: 'utf-8' codec can't encode character '\ud83d' ... surrogates not allowed

I suggest catching UnicodeEncodeError here and returning False, so the constraint fails closed like the non-string case above.

pyproject.toml
```diff
@@ -51,6 +51,12 @@ dependencies = [
     # enforcement as @agent_tool functions. Was a transitive dep
     # already; pinned explicitly now since we import from it directly.
     "mcp>=1.0",
+    # RE2 (Go regexp / Rego's `regex.match`) evaluates `matches` on the
+    # pydantic engine, so a constraint behaves the same there as in a
+    # compiled WASM bundle -- same syntax accepted, and linear time either
+    # way. Prebuilt wheels for 3.13/3.14 on macOS, Linux and Windows, so
+    # nothing is compiled on install. See docs/adr/R-CONSTRAINT-001.
+    "google-re2>=1.1.20251105",
```
victorludvig, line 61: platform/api depends on the SDK as an editable path dep, so platform/api/uv.lock also records hexgate's dependencies, and it still lacks google-re2 (cd platform/api && uv lock --check fails on this branch). CI's plain uv sync re-resolves silently so nothing breaks there, but a --locked/--frozen install would miss re2 and the API imports hexgate.security.constraints. 

Could you run uv lock in platform/api and commit the result?

docs/adr/R-CONSTRAINT-001-matches-runs-on-re2.md
```diff
@@ -0,0 +1,74 @@
...
+timeout on the decision path, that difference is the one that matters.
+
+**Dependency.** `google-re2` publishes prebuilt wheels for the Python versions
+this project supports, alongside the native packages already shipped
+(`wasmtime`, `cryptography`, `biscuit-python`), so contributors compile nothing.
+
+## Consequences
+
+- `$` now means end-of-value only. Python's `$` also matches just before a
+  trailing newline, so `^abc$` against `"abc\n"` was true on the pydantic engine
```
victorludvig, line 44: There's a second small change worth listing here: RE2's \s is [\t\n\f\r ], while Python's ASCII \s also includes \v, so a pattern relying on \s no longer matches a vertical tab (verified: re2.search(r"\s", "\v") → None, re.search(r"\s", "\v", re.ASCII) → match). Same direction as the $ change — it aligns with the bundle — but the test docstring calls $ "the one behaviour change", so a one-line mention (and maybe a \v case in the tests) would keep that accurate.

victorludvig (commented):
Hi @Dev-next-gen, thank you for this new revision with the proposed google-re2 approach, I approve the logic. Three comments to check from Claude's findings, mostly nits.

Dev-next-gen:
Thanks — all three reproduce, and the first one is not a nit. Pushed in 94b2744.

**Unencodable argument.** Confirmed exactly as you describe: `json.loads` accepts
a lone surrogate, RE2 has to encode the value as UTF-8 before matching, and the
`UnicodeEncodeError` escapes `decide()` because `evaluate_tool_call` only catches
`PolicyDeniedError`. On `main` the same input was evaluated and denied. It now
fails the constraint closed, like every other unmatchable value in that function.

While reproducing it I hit the same limit one level up: a *pattern* carrying an
unpaired surrogate leaked the same exception, from `_validate_re2` at load. That
one is a malformed policy rather than a denial, so it is now refused there with
the other regex errors:

```
regex '^INC-\ud83d' in '…' is not valid UTF-8 (it carries an unpaired
surrogate), so RE2 cannot compile it
```

**`platform/api/uv.lock`.** Confirmed — the committed lockfile had no
`google-re2`, and `uv lock --check` failed against it. Relocked and committed; it
resolves clean now.

**`\s` and the vertical tab.** Confirmed: `re2.search(r"\s", "\v")` is `None`
while `re.search(r"\s", "\v", re.ASCII)` matches. Same direction as `$` — it
agrees with the bundle, which never matched it either — so it is now listed
alongside `$` in the ADR's consequences, and the test docstring no longer calls
`$` "the one behaviour change". Added a case pinning the vertical tab as denied
and the other five whitespace characters as still matching.

Three tests added (43 in that file). Full SDK suite: 2083 passing, with the same
pre-existing failures as `main` on this machine and none added.

hexgate/security/constraints.py
Dev-next-gen: Reproduced, and fixed in 94b2744: the `matches` branch now catches `UnicodeEncodeError` and returns `False`, so the constraint fails closed like the non-string case above it rather than escaping `decide()`.

The same limit applies to the pattern side — a regex carrying an unpaired surrogate leaked the same exception out of `_validate_re2` at load. That one is a malformed policy rather than a denial, so it is refused there with the other regex errors. Both have tests.

pyproject.toml
Dev-next-gen: Confirmed: the committed `platform/api/uv.lock` had no `google-re2`, and `uv lock --check` failed against it (I restored the old file to reproduce it before relocking). Ran `uv lock` in `platform/api` and committed the result in 94b2744 — it resolves clean now.

docs/adr/R-CONSTRAINT-001-matches-runs-on-re2.md
Dev-next-gen: Confirmed: `re2.search(r"\s", "\v")` is `None` while `re.search(r"\s", "\v", re.ASCII)` matches. Listed in the ADR's consequences next to `$`, since it points the same way — the bundle never matched it either.

You were right that the wording had gone stale: the test docstring no longer calls `$` "the one behaviour change". Added a case pinning the vertical tab as denied and the other five whitespace characters as still matching.

PR #142, opened 2026-08-26
feat(sdk): enforce agent reach + admission at runtime, cap handoff depth (A3+A4)

quanghexa94:
> **📚 Stacked on #124 — review/merge #124 first.** This PR's base branch is `lhq/feat/agent-admission-gate` (#124's head), so the diff below shows **only** the reach/handoff commits on top of #124, not #124's own changes. GitHub retargets this base to `main` automatically once #124 merges. Merge order: **#124 → then this.**
>
> **Update (as-tool reach on OpenAI):** OpenAI agent-as-tool reach is now enforced at the **top level** too — the SDK tags an `Agent.as_tool()` with its target, so `wrap_tool` gates it under `agent.tool:<target>`. Nested/deeper as-tool edges remain name-gated (a follow-up); see `nested-agent-as-tool-reach-design.md`.

---

Enforce agent-to-agent reach (handoff and agent-as-tool) and admission at runtime across the framework adapters, and cap handoff-chain depth. This is A3 + A4 of the agent-level enforcement series, building on the policy model and gates from #124 (which it is stacked on). Pure SDK, no platform or dashboard change, no migration.

Before this PR the reach policy layer was complete but nothing enforced it at runtime (`declares_reach()` only fed the bundle manifest), and admission was enforced only on the native agent. This wires both into the OpenAI and Google adapters at their real handoff/delegation seams, adds a handoff-depth runaway cap, and warns loudly where a framework exposes no target handle.

Scope was chosen deliberately: only OpenAI (handoff) and Google (handoff + agent-as-tool) expose a first-class target-agent handle at the seam, so enforcement lands there; pydantic_ai and the native single-graph agent get a loud warn-and-defer. Full rationale and the explicit follow-ups are in `docs/adr/R-AGENT-003`.

## Design

- **`ReachGate`** (`security/agent_gate.py`): the reach counterpart to `AgentGate`, sharing a new `_PolicyGate` base (enforcer + approval fold + fail-closed). `check_reach(target, via)` decides `agent.{via}:<target>` against the source agent's policy, engages per-seam via `declares_reach()`, and raises `ReachNotAllowedError` on a deny.
- **`canonical_agent_name`** (`security/naming.py`): one derivation for an agent's own name and its name as a reach target, so a target authored under the name it registers with matches at the seam. Replaces two divergent per-adapter spellings.
- **OpenAI**: a per-run `_HexgateReachHooks` intercepts the SDK `on_handoff` (awaited before the transfer completes, so raising vetoes it); admission is checked at run entry for the top-level agent. Only reach from the governed source is gated.
- **Google ADK**: a `_HexgateReachPlugin` (`BasePlugin.before_tool_callback`) intercepts `transfer_to_agent` (handoff) and `AgentTool` (agent-as-tool); admission is checked at run entry in `run`/`run_async`.
- **A4 depth cap**: `HexgateRunner(max_handoff_depth=N)` on both adapters. A handoff transfers control forward, so the per-run handoff count is the chain depth; past the cap the seam raises `HandoffDepthExceededError`. OpenAI counts on the per-run hook; Google counts per `invocation_id` and clears in `after_run_callback`.
- **Warn-and-defer**: `warn_if_reach_unenforced` on pydantic_ai, LangChain, and the native agent (native keeps enforcing admission, so reach-only warning).

## Important files

| File | What to look at |
| --- | --- |
| `hexgate/security/agent_gate.py` | `_PolicyGate` base; `ReachGate.check_reach`; `ReachNotAllowedError`, `HandoffDepthExceededError`; `warn_if_reach_unenforced` |
| `hexgate/security/naming.py` | `canonical_agent_name` / `canonical_name` — the one name derivation |
| `hexgate/adapters/openai/runner.py` | `_HexgateReachHooks.on_handoff` (reach + depth), run-entry admission, `_merge_hooks` always composing |
| `hexgate/adapters/google/runner.py` | `_HexgateReachPlugin.before_tool_callback` (transfer_to_agent + AgentTool, reach + per-invocation depth), run-entry admission |
| `docs/adr/R-AGENT-003-...md` | the runtime-seam decisions, the source-governed rule, and the deferred follow-ups |
| `tests/security/test_reach_gate.py`, `tests/adapters/{openai,google}/...` | the reach/admission/depth behaviour, driven at the seams |

## Test plan

Ran the SDK suite against the hexanlp-demo env: `pytest tests/ -q` gives 2128 passed, 6 skipped (opt-in integration), 20 deselected. `ruff check` and `ruff format --check` clean over `hexgate/` and `tests/`.

Verified against the installed SDKs (not assumed): the OpenAI `on_handoff` is awaited via `asyncio.gather` before the SDK returns `NextStepHandoff`, so raising there vetoes the transfer; ADK's `before_tool_callback` sees `transfer_to_agent(agent_name=...)` and `AgentTool` (with `tool_context.agent_name` identifying the source), and `after_run_callback` carries the `invocation_id` used to clear the depth counter. By hand I checked: an unlisted/deny reach target raises and stops the transfer; a non-admitted caller is refused at run entry before the underlying run; a handoff past `max_handoff_depth` raises; a handoff from an un-governed source is not gated; ordinary tools fall through to the wrapped enforcer. A reviewer catches a regression through `test_reach_gate.py` (gate semantics) and the per-adapter seam tests (governed-source, via distinction, depth cap, admission refuse/allow).

## Try it

```
uv run pytest tests/security/test_reach_gate.py tests/security/test_naming.py -q
uv run pytest tests/adapters/openai/test_runner_reach_admission.py -q
uv run pytest tests/adapters/google/test_runner.py -q
```

## Notes

- Stacked on #124 (`lhq/feat/agent-admission-gate`); review/merge that first, or read this PR's diff against it. No platform/dashboard change.
- Fail-closed throughout: a denied reach/admission or an over-depth handoff raises and aborts the run rather than continuing ungoverned.
- Deferred (warned in the interim, tracked in R-AGENT-003): per-sub-agent/post-handoff admission (needs sub-agent policy resolution); agent-as-tool reach on OpenAI (no target handle at the seam); pydantic_ai/native reach; the manifest drift lint (reach key vs declared sub-agents), which needs a cross-package `AgentManifest` change.

### DIFF
--- a/docs/adapters/google-adk.mdx
+++ b/docs/adapters/google-adk.mdx
@@ -112,6 +112,15 @@ if __name__ == "__main__":
   executes inside `propagate_attributes(user_id=..., session_id=...,
   metadata={"user_roles": ...}, tags=["google.runner.run.<agent_name>"])`.
 
+## Agent-level enforcement
+
+The Google runner has the fullest [agent-level](/concepts/agent-level-enforcement)
+coverage, because ADK expresses both delegation modes as tool calls. It enforces
+**admission** at run entry, and **reach** at `before_tool_callback` for both
+`transfer_to_agent` (handoff, `agent.handoff:<name>`) and `AgentTool`
+(agent-as-tool, `agent.tool:<name>`) — a denied delegation is refused before it
+runs. Set `HexgateRunner(max_handoff_depth=N)` to cap a runaway handoff chain.
+
 ## Runnable example
 
 `examples/devops_google.py` — `HexgateRunner` (Google ADK) end-to-end with
--- a/docs/adapters/langchain.mdx
+++ b/docs/adapters/langchain.mdx
@@ -119,6 +119,17 @@ so the wrapper resolves the policy for an agent called `LangGraph` and fails
 with a 404 unless one is registered under that name.
 </Note>
 
+## Agent-level enforcement
+
+<Note>
+[Agent-level policy](/concepts/agent-level-enforcement) (`admission` / `agents`)
+is not yet enforced on the LangChain adapter — a declared block logs a warning at
+setup rather than gating. A sub-agent exposed as a tool is still gated normally by
+its **tool name**. (The native `create_agent` orchestrator does enforce admission;
+this adapter, which wraps a pre-built LangChain/LangGraph agent, is a planned
+follow-up.)
+</Note>
+
 ## Runnable example
 
 `examples/devops_langchain.py` — `wrap_langchain_agent` (LangChain / LangGraph)
--- a/docs/adapters/openai.mdx
+++ b/docs/adapters/openai.mdx
@@ -97,18 +97,23 @@ if __name__ == "__main__":
 ## Agent-level enforcement
 
 Beyond tool calls, `HexgateRunner` enforces [agent-level policy](/concepts/agent-level-enforcement):
-**admission** (may this caller run the agent) at run entry, and **handoff reach**
+**admission** (may this caller run the agent) at run entry, **handoff reach**
 (`agent.handoff:<name>`) at the SDK's `on_handoff` seam — a denied handoff is
-vetoed before the target runs. Set `HexgateRunner(max_handoff_depth=N)` to cap a
-runaway handoff chain.
+vetoed before the target runs — and **agent-as-tool reach** (`agent.tool:<name>`)
+at the tool seam. Set `HexgateRunner(max_handoff_depth=N)` to cap a runaway
+handoff chain.
 
-<Warning>
-**Agent-as-tool reach is not enforced on OpenAI.** `Agent.as_tool()` produces a
-plain function tool with no link back to the target agent, so the
-`agent.tool:<name>` reach key can't be resolved here — a policy that declares one
-logs a warning at setup. Gate a sub-agent-as-tool by its **tool name** instead
-(an ordinary `tools:` rule), which works normally.
-</Warning>
+<Note>
+**Agent-as-tool reach is enforced at the top level.** The SDK tags an
+`Agent.as_tool()` with its target agent, so when the top-level agent's policy
+declares an `agents` block, that as-tool call is gated under `agent.tool:<name>`
+(closed-world, `via`/constraints honored) rather than by tool name. A reach edge
+one level deeper — a sub-agent that itself calls another agent as a tool — is not
+gated yet, because the sub-agent runs unwrapped through the SDK's own runner;
+those deeper edges fall back to tool-name gating. On an `agents` SDK too old to
+expose the tool origin, even the top level falls back to name-gating and a
+declared `via: tool` rule warns at setup.
+</Note>
 
 ## Runnable example
 
--- a/docs/adapters/openai.mdx
+++ b/docs/adapters/openai.mdx
@@ -94,6 +94,22 @@ if __name__ == "__main__":
 
 `run_sync` and `run_streamed` work the same way.
 
+## Agent-level enforcement
+
+Beyond tool calls, `HexgateRunner` enforces [agent-level policy](/concepts/agent-level-enforcement):
+**admission** (may this caller run the agent) at run entry, and **handoff reach**
+(`agent.handoff:<name>`) at the SDK's `on_handoff` seam — a denied handoff is
+vetoed before the target runs. Set `HexgateRunner(max_handoff_depth=N)` to cap a
+runaway handoff chain.
+
+<Warning>
+**Agent-as-tool reach is not enforced on OpenAI.** `Agent.as_tool()` produces a
+plain function tool with no link back to the target agent, so the
+`agent.tool:<name>` reach key can't be resolved here — a policy that declares one
+logs a warning at setup. Gate a sub-agent-as-tool by its **tool name** instead
+(an ordinary `tools:` rule), which works normally.
+</Warning>
+
 ## Runnable example
 
 `examples/devops_openai.py` — `HexgateRunner` (OpenAI Agents SDK) end-to-end.
--- a/docs/adapters/pydantic-ai.mdx
+++ b/docs/adapters/pydantic-ai.mdx
@@ -96,6 +96,18 @@ if __name__ == "__main__":
   the caller identity. Global tracing is enabled via `Agent.instrument_all()` on
   construction.
 
+## Agent-level enforcement
+
+<Note>
+pydantic_ai has **no native handoff primitive** — multi-agent here is
+agent-as-tool, where one agent is called from inside another's tool. So there is
+no transfer seam for Hexgate to gate: a declared [`agents` or `admission`
+block](/concepts/agent-level-enforcement) logs a warning at setup rather than
+enforcing. Gate a sub-agent-as-tool by its **tool name** with an ordinary
+`tools:` rule, which works normally. Admission and reach enforcement on this
+adapter are planned follow-ups.
+</Note>
+
 ## Runnable example
 
 `examples/devops_pydantic_ai.py` — `wrap_pydantic_agent` (Pydantic AI) end-to-end.
--- a/docs/adr/R-AGENT-003-reach-and-admission-runtime-seams.md
+++ b/docs/adr/R-AGENT-003-reach-and-admission-runtime-seams.md
@@ -33,7 +33,7 @@ Warn-and-defer keeps closed-world honest across frameworks: enforcement lands on
 ## Deferred (follow-ups, warned in the interim)
 
 - **Per-sub-agent / post-handoff admission** (B re-checking admission at its run entry): needs resolving sub-agent policies and handling unregistered sub-agents, so it is out of this slice. Reach from the governed source is the boundary control today.
-- **Agent-as-tool reach on OpenAI** (`Agent.as_tool`): no metadata links the produced function tool back to a target agent name at the seam.
+- **Agent-as-tool reach on OpenAI** (`Agent.as_tool`): no metadata links the produced function tool back to a target agent name at the seam, so it cannot be *enforced* — but a policy that declares a `via: tool` target now *warns* (`warn_if_tool_reach_unenforced`, gated on `declares_tool_reach()` so handoff-only policies are not spammed), so the gap is no longer silent.
 - **pydantic_ai and native reach**: no runtime target handle.
 - **Manifest drift lint** (reach key vs declared sub-agents/handoffs): needs `AgentManifest` to carry target declarations, a cross-package change.
 
--- a/docs/adr/R-AGENT-003-reach-and-admission-runtime-seams.md
+++ b/docs/adr/R-AGENT-003-reach-and-admission-runtime-seams.md
@@ -1,0 +1,48 @@
+# R-AGENT-003: Reach and admission are enforced at per-framework runtime seams
+
+**Status:** Accepted · 2026-08-26 · builds on R-AGENT-002
+**Applies to:** `hexgate/security/agent_gate.py`, `hexgate/security/naming.py`, `hexgate/adapters/openai/runner.py`, `hexgate/adapters/google/runner.py`, `hexgate/adapters/pydantic_ai/wrapper.py`, `hexgate/adapters/langchain/wrapper.py`, `hexgate/agents/factory.py`
+
+## Decision
+
+Agent-level admission (`agent.run`) and reach (`agent.handoff:<t>` / `agent.tool:<t>`) are enforced at framework-specific runtime seams by two gates that reuse `PolicyEnforcer.decide`, mirroring the egress `Gate`. R-AGENT-002 defined the policy model; this defines where and how it runs.
+
+- **One canonical name.** A target's name at a seam derives identically to an agent's own name via `canonical_agent_name` (trim, blank/None → `"default"`, no case folding). Own-name and target-name must agree or a reach key would not match the name its target registers with. This is the one derivation; the two prior per-adapter spellings are gone.
+- **Reach is governed by the source agent's policy, at the seam, before control transfers.** `ReachGate.check_reach(target, via)` decides the lowered key and raises `ReachNotAllowedError` on a deny.
+  - **OpenAI:** the SDK `on_handoff` hook (awaited before the transfer completes, so raising vetoes it) → `via="handoff"`.
+  - **Google ADK:** a `BasePlugin.before_tool_callback` intercepting the two shapes ADK expresses as tool calls: `transfer_to_agent` (`via="handoff"`) and `AgentTool` (`via="tool"`).
+  - Only reach from a **Hexgate-governed source** (the resolved top-level agent) is gated; a transfer from an un-governed sub-agent is left alone (sub-agent governance is a later slice).
+- **Admission is enforced at run entry for the top-level agent**, inside the context scope so the caller's role is visible, on the OpenAI and Google runners. A non-admitted caller is refused before the run starts.
+- **Handoff depth is a per-run runaway cap, independent of reach policy.** A handoff transfers control forward, so the count of handoffs within one run is the chain depth; past `max_handoff_depth` the seam raises `HandoffDepthExceededError`. OpenAI counts on the per-run hook; Google counts per `invocation_id` on the shared plugin and clears it in `after_run_callback`.
+- **Where the framework hides the target, warn — never silently pass.** pydantic_ai (delegation inside a tool body), the native single-graph agent (no handoff seam), and agent-as-tool on OpenAI expose no target handle, so a declared block is surfaced by `warn_if_admission_unenforced` / `warn_if_reach_unenforced` at wrap time.
+
+## Why
+
+The two gates share `_PolicyGate` (enforcer + approval fold + fail-closed), so admission and reach behave identically and are defined once. Enforcing at the framework seam (not a wrapper around the whole run) is what lets reach fire at the exact moment control would transfer, so a denied handoff never starts the target. Governing reach by the **source** policy matches the model: the `agents` block on A says who A may reach; B's own admission is a separate direction (its run entry).
+
+Counting handoffs as depth is exact for these SDKs because a handoff is a forward transfer with no return, so a run is a chain and the handoff count is its length. The cap is deliberately policy-independent: it stops a runaway even when every hop is allowed.
+
+Warn-and-defer keeps closed-world honest across frameworks: enforcement lands only where a target handle exists, and everywhere else the gap is loud (one log per framework/agent) rather than a silent no-op, so an operator is never misled into thinking a declared block is enforced.
+
+## Consequences
+
+- Reach and admission are enforced end-to-end on OpenAI (handoff) and Google (handoff + agent-as-tool); the interim `warn_if_admission_unenforced` is gone from those adapters.
+- A denied reach or admission fails closed by raising, aborting the run rather than degrading to an ungoverned continuation.
+- `HexgateRunner(max_handoff_depth=N)` is opt-in on both adapters (None = no cap).
+
+## Deferred (follow-ups, warned in the interim)
+
+- **Per-sub-agent / post-handoff admission** (B re-checking admission at its run entry): needs resolving sub-agent policies and handling unregistered sub-agents, so it is out of this slice. Reach from the governed source is the boundary control today.
+- **Agent-as-tool reach on OpenAI** (`Agent.as_tool`): no metadata links the produced function tool back to a target agent name at the seam.
+- **pydantic_ai and native reach**: no runtime target handle.
+- **Manifest drift lint** (reach key vs declared sub-agents/handoffs): needs `AgentManifest` to carry target declarations, a cross-package change.
+
+## Verify
+
+```
+pytest tests/security/test_reach_gate.py tests/security/test_naming.py \
+       tests/adapters/openai/test_runner_reach_admission.py \
+       tests/adapters/google/test_runner.py -q
+# reach denies an unlisted/deny target and vetoes the transfer; admission refuses a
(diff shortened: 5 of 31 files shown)


winlp4ever:
Second high-effort review pass. It confirmed the two fixes from the first pass are resolved (not re-flagged) and surfaced 5 more. Three are fixed in b61c973; two are documented limitations.

**Fixed**

1. **OpenAI agent-as-tool reach was silently neither enforced nor warned** (unlike the other adapters, which at least warn). Added `declares_tool_reach()` to the `PolicyEngine` protocol (PolicySet scans `effective_tools` for `agent.tool:` keys; PolicyBundle reads a new `reach_tool` manifest flag) and `warn_if_tool_reach_unenforced`, called from the OpenAI runner. It fires only when a `via: tool` target is declared, so a handoff-only policy (which OpenAI *does* enforce) is not spammed.
2. **Google `_HexgateReachPlugin._depth` leaked one entry per aborted run** — a raise from `before_tool_callback` (depth cap or reach deny) skips `after_run_callback`, so the invocation's counter was never popped on the shared plugin. Now popped on the abort path.
3. **The four copy-pasted NEEDS_APPROVAL folds** in `AgentGate`/`ReachGate` are collapsed behind `_PolicyGate._cleared_sync/_cleared_async`, so the pass/refuse condition lives once and can't drift.

**Documented (won't-fix by design)**

4. **`run_streamed` + async `approval_handler`**: a sync entrypoint can't await an async handler, so a NEEDS_APPROVAL admission verdict fails closed there. This matches the gate's own documented sync/async behavior (and the ban gate's); admission is checked before the stream task spawns so a refused run yields nothing. Noted in the `run_streamed` docstring; use the async `run` for async approval.
5. **Hot-reload doesn't re-warn**: the unenforced-adapter warnings fire at wrap time and dedup permanently, so a policy that newly declares admission/reach after a `refresh_policy` won't emit a fresh warning. Low severity (declaring a new block on a non-enforcing adapter mid-process is an edge case); leaving as a known limitation rather than moving the warning to the per-run hot path.

Full suite green (2131 passed), ruff clean.

quanghexa94:
Ran a high-effort review over the reach/admission/depth cluster. It verified the cross-engine parity of the closed-world change, that `declares_admission`/`declares_reach` exist on both engines, and that the OpenAI `on_handoff` and ADK `before_tool_callback`/`after_run_callback` signatures and context attributes match what the seams use. Five findings; the two on this PR's surface are fixed in d1462e7.

**1. Google emitted a false "admission not enforced" warning (fixed).** `wrap_google_agent` still called `warn_if_admission_unenforced` after the Google runner started enforcing admission at run entry, so a policy with an `admission` block logged that admission was bypassed when it was actually enforced, a contradictory, security-relevant message. Removed it (Google enforces both admission and reach now).

**2. Reach-key parity asymmetry (fixed).** `lowered_agent_tools` built the reach key from the verbatim authored `agents:` target while `ReachGate` canonicalizes the runtime target, so a padded authored name (e.g. `agents: {" Bot ": ...}`) would never match and a policy-allowed handoff would fall to the closed-world deny. Canonicalized the target at lowering too, so both ends normalize identically; added a parity test.

The other three are minor or out of scope for this PR, noted for the record:

- **OpenAI reach hook allocates a `ReachGate` per handoff even when no reach/cap is set.** Negligible in practice (a handoff is an LLM-round-trip-gated event, not a hot loop), and the depth counter must run on every handoff regardless. Leaving as is.
- **`build_signed_bundle` parses the policy twice** (once in `compile_to_rego`, once for the `agent_gating` flags). Real but pre-existing on the base branch (#124), not this PR's diff.
- **Warn-dedup sets / Google per-invocation depth map grow across distinct keys.** Bounded by distinct agent names in practice; the depth entry is cleared in `after_run_callback` and only leaks on an abnormally terminated invocation. Low severity; can revisit if it matters for a long-lived server.

quanghexa94:
Second high-effort review pass. It confirmed the two fixes from the first pass are resolved (not re-flagged) and surfaced 5 more. Three are fixed in b61c973; two are documented limitations.

**Fixed**

1. **OpenAI agent-as-tool reach was silently neither enforced nor warned** (unlike the other adapters, which at least warn). Added `declares_tool_reach()` to the `PolicyEngine` protocol (PolicySet scans `effective_tools` for `agent.tool:` keys; PolicyBundle reads a new `reach_tool` manifest flag) and `warn_if_tool_reach_unenforced`, called from the OpenAI runner. It fires only when a `via: tool` target is declared, so a handoff-only policy (which OpenAI *does* enforce) is not spammed.
2. **Google `_HexgateReachPlugin._depth` leaked one entry per aborted run** — a raise from `before_tool_callback` (depth cap or reach deny) skips `after_run_callback`, so the invocation's counter was never popped on the shared plugin. Now popped on the abort path.
3. **The four copy-pasted NEEDS_APPROVAL folds** in `AgentGate`/`ReachGate` are collapsed behind `_PolicyGate._cleared_sync/_cleared_async`, so the pass/refuse condition lives once and can't drift.

**Documented (won't-fix by design)**

4. **`run_streamed` + async `approval_handler`**: a sync entrypoint can't await an async handler, so a NEEDS_APPROVAL admission verdict fails closed there. This matches the gate's own documented sync/async behavior (and the ban gate's); admission is checked before the stream task spawns so a refused run yields nothing. Noted in the `run_streamed` docstring; use the async `run` for async approval.
5. **Hot-reload doesn't re-warn**: the unenforced-adapter warnings fire at wrap time and dedup permanently, so a policy that newly declares admission/reach after a `refresh_policy` won't emit a fresh warning. Low severity (declaring a new block on a non-enforcing adapter mid-process is an edge case); leaving as a known limitation rather than moving the warning to the per-run hot path.

Full suite green (2131 passed), ruff clean.

guillaume-hexamind:
## Review: 6 findings on the reach/admission seams

Reviewed at `25050d0`. Every claim below was checked against the installed `google-adk` 1.32.0 and `openai-agents` 0.15.1 sources, not against the PR description. Scores are severity × confidence out of 100.

Two blockers (#1, #2), four to fix before merge (#3–#6).

---

### 1. Reach plugin is appended last, so a caller plugin can silently preempt it — 94/100

**Issue.** `google/runner.py:171` appends `_HexgateReachPlugin` *after* the caller's plugins. ADK's `PluginManager._run_callbacks` (`plugins/plugin_manager.py:283-306`) iterates plugins in registration order and **returns on the first non-`None` result** — its own class docstring says so: *"if any plugin callback returns a non-`None` value, the execution of subsequent plugins for that specific event is halted."* Order is preserved end to end (`runners.py:268` `plugins = app.plugins` → `runners.py:209` `PluginManager(plugins=plugins)` → `register_plugin` appends at `plugin_manager.py:93`), and in `flows/llm_flows/functions.py:512-531` a non-`None` return becomes the tool response and the tool is never invoked.

So any preceding plugin that returns a dict from `before_tool_callback` short-circuits the loop and **neither the reach check nor the depth cap runs** — a silent fail-open on the security seam this PR adds. The inline comment at `:169-170` ("Appended after the caller's plugins so it always runs") has the ADK semantics exactly backwards.

This is not hypothetical: ADK *ships* plugins that return non-`None` from `before_tool_callback` — `cli/plugins/replay_plugin.py:90` (returns a recorded tool response) and `tools/environment_simulation/environment_simulation_plugin.py:40` (returns `simulate(...)`). Anyone using ADK replay or environment simulation loses reach enforcement and the depth cap with no diagnostic.

**Suggested fix.** Register first instead of last, and correct the comment:

```python
# Reach enforcement at the ADK transfer/AgentTool seam. Inserted *before* the
# caller's plugins: ADK's PluginManager early-exits on the first plugin whose
# before_tool_callback returns non-None, so a caller plugin that overrides a tool
# result (replay, environment simulation, a cache) would otherwise skip this gate.
plugins.insert(0, _HexgateReachPlugin(self))
```

Safe as-is: the plugin only ever returns `None` or raises, so it cannot itself short-circuit anything downstream. Position relative to `HexgateUsagePlugin` is irrelevant (that one only implements `after_model_callback`, `google/usage.py:28`).

---

### 2. `arun_streamed` checks admission through the *sync* path, so an async approval handler always fails closed — 90/100

**Issue.** `arun_streamed` (`openai/runner.py:405`) delegates to `_launch_streamed`, which calls `_check_admission_sync(binding)` at `:447`. On the async path with an async `approval_handler`, `_PolicyGate._resolve_approval_sync` (`security/agent_gate.py:240-250`) detects the awaitable, closes it, logs `"admission approval_handler is async on a sync run; denying (fail-closed)"`, and returns `False` → `AgentNotAdmittedError`.

`hexgate serve` wires exactly this combination:

- `cli/_common.py:379` — `HexgateRunner(approval_handler=approval_handler)`
- `cli/serve.py:126` — `RelayApprovalHandler.__call__` is `async def`
- `cli/_common.py:385` → `astream_openai` → `openai/streaming.py:142` `await runner.arun_streamed(...)`

**Concrete failure:** a policy with `admission: {mode: approval_required}` served over `hexgate serve` is *always* refused. The approval request never reaches the dashboard, so there is no way to approve it on that path.

This isn't the documented sync limitation. `run_streamed`'s docstring (`:356-359`) states the constraint and points callers at the async entrypoint — and `arun_streamed`'s own docstring (`:388-397`) says it exists precisely to await the async variants, then inherits the sync admission check anyway.

**Suggested fix.** Resolve admission in each entrypoint, at the right sync/async-ness, and let `_launch_streamed` skip it. `HexgateContext.sync_scope` saves/restores through a stack (`runtime/context.py:139-147`), so re-entering the scope inside `_launch_streamed` is harmless:

```python
async def arun_streamed(self, ...) -> RunResultStreaming:
    ...
    with hexgate_context.sync_scope():
        await self._check_admission_async(binding)   # in-scope: reads the caller's role
        return self._launch_streamed(..., admitted=True)
```

and in `_launch_streamed`, guard the existing call:

```python
with hexgate_context.sync_scope():
    if not admitted:
        # Before run_streamed spawns its task, so a non-admitted run yields nothing.
        self._check_admission_sync(binding)
```

`run_streamed` keeps passing `admitted=False` and its documented sync-handler caveat. A test that drives `arun_streamed` with an `async def` handler and an `approval_required` admission policy would lock this down — the current suite only covers the sync entrypoints for admission.

---

### 3. `_render_reach_error` is also used for guard halts, producing a false "was not invoked" — 65/100

**Issue.** `openai/tools.py:146` hands `_render_reach_error(reach_target)` to `run_guarded_async` as `render_error`, but that renderer reaches four sites in `guards/runner.py`, not just the policy branch: pre-guard halt (`:403`), the enforcer decision (`:430`), and both post-guard paths (`:438`, `:454` → `:347`).

So an `@after_tool` guard that halts on an `Agent.as_tool()` result renders:

> `reach to agent 'billing-bot' is not permitted by this agent's policy. The sub-agent was not invoked.`

— after the sub-agent has already run and the guard has read its output. Both halves are false: a guard rejected the output, not a reach policy, and the sub-agent *was* invoked. Pre-guard halts get the same wrong attribution (less harmful, since nothing ran).

Partially self-limiting: `_halt_to_decision` tags a DENY halt `guard_denied` (`guards/runner.py:162-182`) and `_render_reach_error` reads `decision.error_type` first, so the marker is right even when the prose isn't. It also needs guards configured to trigger.

**Suggested fix.** Scope the reach wording to the policy decision only. Add a `render_policy_error: RenderError | None = None` parameter to `run_guarded_async` / `run_guarded` alongside the existing `policy_key` / `policy_args` overrides, use it at the enforcer branch:

```python
return render_error(decision) if render_policy_error is None else render_policy_error(decision)
```

and have the OpenAI reach path pass `render_error=_render_error, render_policy_error=_render_reach_error(reach_target)`. Every other caller keeps a single renderer, as with `policy_key`/`policy_args` today.

---

### 4. As-tool reach engages on `declares_tool_reach()` while Google engages both seams on `declares_reach()` — 62/100

**Issue.** `openai/tools.py:134` engages as-tool reach on `enforcer.policy.declares_tool_reach()` (`policy_set.py:175` — any `agent.tool:` key). `google/runner.py:112` engages **both** seams on `declares_reach()` (`policy_set.py:168` — any reach key). The two adapters therefore disagree on the same policy.

With `agents: {billing_bot: {via: [handoff], mode: allow}}`:

- **Google** builds the gate, decides `agent.tool:billing_bot`, hits `_AGENT_CLOSED_WORLD_DENY` (`policy.py:65-66`), and denies the `AgentTool`.
- **OpenAI** skips the reach key entirely and falls back to plain name-gating, so a `tools:` allow or a permissive `default_policy` lets the delegation through.

The docs commit to Google's reading: `agent-level-enforcement.mdx:68` — *"Once you declare agent rules, any agent you did not list is denied"* — and `:53` (`via: [tool]  # callable as a tool; handoff is denied`) establishes that `via` lists are exclusive in both directions. The Note at `:108-118` carves out only *"when the policy declares no `agents` block"*; the handoff-only case isn't covered either way. `test_tools_reach.py:156-173` locks in the OpenAI behavior as intended.

To be clear on blast radius: this is an inconsistency, not a fail-open. The as-tool still falls to name-gating on OpenAI, so it needs an explicit `tools:` allow or a permissive default to get through — which is exactly what that test encodes.

**Suggested fix.** Pick one semantics and make both adapters and the docs agree. Recommended (smallest change, preserves the deliberate no-spam-on-handoff-only behavior): narrow Google's `AgentTool` branch to match OpenAI, leaving `transfer_to_agent` on `declares_reach()`:

```python
if is_transfer:
    if not self._runner._binding.enforcer.policy.declares_reach():
        return None
    target, via = tool_args.get("agent_name"), "handoff"
elif isinstance(tool, AgentTool):
    if not self._runner._binding.enforcer.policy.declares_tool_reach():
        return None  # no via:tool target declared — as-tools stay name-gated
    target, via = getattr(tool.agent, "name", None), "tool"
```

Then amend the Note in `agent-level-enforcement.mdx` to say that a policy declaring **no `via: tool` target** (not merely no `agents` block) leaves agent-as-tool gated by tool name — otherwise `:68` and `:53` still read as promising a closed-world deny.

---

### 5. `ReachNotAllowedError` / `HandoffDepthExceededError` are destroyed into a bare `RuntimeError` on Google — 62/100

**Issue.** ADK wraps every plugin exception into a **new** exception (`plugin_manager.py:298-303`):

```python
raise RuntimeError(f"Error in plugin '{plugin.name}' during '{callback_name}' callback: {e}") from e
```

The original survives only as `__cause__`. So on the Google adapter `except ReachNotAllowedError as e: e.decision` and `except HandoffDepthExceededError as e: e.depth` never match, and `.decision` / `.depth` / `.cap` are unreachable. The ADR promises the typed error twice — *"raises `ReachNotAllowedError` on a deny"* and *"A denied reach or admission fails closed by raising"*. The message also leaks the internal plugin name `hexgate_reach` to whoever surfaces it.

`except RuntimeError` still works, since all three errors subclass it by design (`agent_gate.py:137`, `:163`, `:189`) — so refusal behavior is intact and this is an API-contract break rather than a broken gate.

Invisible to the suite because every Google reach test calls `plugin.before_tool_callback(...)` directly (`tests/adapters/google/test_runner.py:617-690`) instead of driving a real ADK `Runner`.

**Suggested fix.** Unwrap in the runner, where the ADK boundary is crossed — restores the ADR contract without fighting the plugin manager:

```python
_TYPED_SEAM_ERRORS = (ReachNotAllowedError, HandoffDepthExceededError)

def _unwrap_plugin_error(exc: RuntimeError) -> BaseException:
    """ADK's PluginManager re-raises every plugin exception as a fresh RuntimeError
    (plugin_manager._run_callbacks), which would erase our typed seam errors and
    their .decision/.depth payload. Recover the original when it is one of ours."""
    cause = exc.__cause__
    return cause if isinstance(cause, _TYPED_SEAM_ERRORS) else exc
```

Apply it around the `run_async` iteration in both `run` and `run_async`, re-raising the unwrapped error. Worth pairing with one integration test that drives a real `Runner` with a preceding plugin — the same test catches #1 and #5 together, which is why both slipped through.

---

### 6. Google double-gates an `AgentTool` (reach key *and* tool name) where OpenAI substitutes — 65/100

**Issue.** `AgentTool.__init__` sets `name=agent.name` (`adk/tools/agent_tool.py:123`, reinforced by the `populate_name` validator at `:127`). `wrap_google_agent` takes `agent.tools` (`google/wrapper.py:52`) — where an `AgentTool` lives — and gates each by `base.name` (`google/tools.py:62`, `:73`). So one delegation is decided **twice**: `agent.tool:billing-bot` by the plugin (`google/runner.py:106-118`), then `billing-bot` by the wrapped enforcer — and two audit decision events are emitted per delegation.

OpenAI substitutes instead: `openai/tools.py:140` sets `policy_key = agent_target_key("tool", reach_target)` and `guards/runner.py:413` uses `policy_key or call.tool_name`, so the tool name is never decided. (The SDK forces an explicit `tool_name` on `as_tool()` and keeps origin metadata separately, so name and target are independent there.)

**This breaks the documented example.** `agent-level-enforcement.mdx:43-54` shows `default_policy: { mode: deny }` with `refund-bot: {via: [tool], mode: allow}`. On Google that grant is overridden by the name gate: `refund-bot` isn't in `tools:`, so `policy.py:67` returns `default_policy` deny and the delegation is refused despite an explicit `via: tool` allow. The same page marks Google "enforced" for agent-as-tool.

**Suggested fix.** Give Google the same substitution OpenAI has, and drop the now-redundant plugin branch. In `google/tools.py:wrap_tool`, detect an `AgentTool` and pass the reach key through the parameters `run_guarded_async` already accepts:

```python
reach_target = canonical_name(base.agent.name) if isinstance(base, AgentTool) else None
...
if reach_target is not None and enforcer.policy.declares_tool_reach():
    policy_key = agent_target_key("tool", reach_target)
    policy_args = {"agent": enforcer.agent_name, "target": reach_target, "via": "tool"}
    render_error = ...  # reach wording, via render_policy_error from #3
```

Then remove the `isinstance(tool, AgentTool)` branch from `_HexgateReachPlugin._gate_tool`, leaving the plugin responsible for `transfer_to_agent` and the depth cap only. That collapses the two decisions into one, emits one audit event, makes the doc example behave as written, and lands the same substitution semantics on both adapters.

If you'd rather keep the plugin branch, the doc needs to say that a Google `via: tool` grant *additionally* requires a matching `tools:` allow — but that contradicts the OpenAI behavior and the "grants union" framing in `:60-64`.

---

<sub>Cross-cutting note: #1 and #5 both exist because the Google reach tests exercise `before_tool_callback` directly rather than through an ADK `Runner`. One integration test driving a real `Runner` with a preceding plugin would have caught both.</sub>

guillaume-hexamind (commented):
Overall a good design, agent reach problems are starting to become pretty complex.

A couple of blocking issues above, will approve once addressed

quanghexa94:
Thanks @guillaume-hexamind — this was an exceptional review (the ADK `PluginManager` order/wrap analysis especially). All six are addressed in `bbe5dc7`; here's the mapping.

**Blockers**

1. **Plugin ordering / fail-open** — the reach plugin is now `plugins.insert(0, …)`, so it runs before any caller plugin that could short-circuit `before_tool_callback`. Comment corrected to say exactly that. You were right that the old "appended so it always runs" had the ADK semantics backwards.
2. **`arun_streamed` sync admission** — `arun_streamed` now runs the **async** admission check itself (inside a `sync_scope`) and passes `admitted=True` into `_launch_streamed`, which skips its sync check. So an async `approval_handler` is awaited instead of fail-closed; `hexgate serve`'s `approval_required` admission now works over the streamed path. The sync `run_streamed` keeps its documented sync-handler caveat (and I added a test pinning that it *is* fail-closed there, so the contrast is intentional).

**The rest**

3. **`_render_reach_error` bleeding into guard halts** — added `render_policy_error` to `run_guarded_async`/`_sync` (alongside `policy_key`/`policy_args`). Reach wording now renders **only** a policy denial; a guard `Halt` renders through `render_error`, so an after-guard rejecting the sub-agent output is no longer mislabeled "sub-agent was not invoked."
4. **OpenAI vs Google engagement** — resolved by #6: Google's agent-as-tool reach now engages on `declares_tool_reach()` in `wrap_tool`, matching OpenAI, so a handoff-only policy leaves as-tools name-gated on both adapters.
5. **Typed seam errors destroyed on Google** — the runner now unwraps ADK's `RuntimeError` re-wrap (`_unwrap_plugin_error`, checking `__cause__`), so `ReachNotAllowedError` / `HandoffDepthExceededError` and their `.decision`/`.depth` survive and the internal plugin name stops leaking.
6. **Google double-gate + broken doc example** — agent-as-tool reach moved from the plugin into `wrap_tool` with the same reach-key substitution OpenAI uses. One delegation is now decided **once** (one audit event), and the documented `default deny + refund-bot: {via: [tool], allow}` example is honored instead of overridden by the name gate. The plugin now handles `transfer_to_agent` + the depth cap only. The `Note`/matrix in `agent-level-enforcement.mdx` are updated to say engagement is on a declared `via: tool` target.

**On your cross-cutting note** — you nailed it: #1 and #5 slipped through because the Google reach tests drove `before_tool_callback` directly. Added the integration test you suggested — it drives a real `PluginManager` with a preceding short-circuiting plugin and asserts (a) the gate still fires (order) and (b) the typed error survives ADK's wrap — one test catching both. Plus tests for async-admission on `arun_streamed`, guard-halt wording, and Google as-tool reach (doc example / closed-world / handoff-only fallback).

Full SDK suite 2440 passing, ruff clean, rebased on current `main`. Two boundaries I kept deliberate (both pre-existing, not new here): the sync `run_streamed` async-handler limitation stays documented, and *nested* as-tool reach (B calls C as a tool) remains a follow-up — the design for it is in `nested-agent-as-tool-reach-design.md`. Ready for another look when you have a moment.

guillaume-hexamind (approved):

PR #146, opened 2026-08-28
feat(sdk): emit audit events as otlp spans

victorludvig:
## What is Changing

The SDK stops POSTing JSON to the three `/v1/audit/*` FastAPI endpoints and instead emits every audit event as one OpenTelemetry span over OTLP/HTTP to the Hexgate Collector — the SDK half of the OTel migration ("OpenTelemetry migration design", section V). The wire contract is `hexgate/tracing/semconv.py` (#130), already decoded by the span-enricher job (#133).

- **`hexgate/tracing/_senders.py`** — `AuditSender` is now one `TracerProvider` → `BatchSpanProcessor` → `OTLPSpanExporter` per `api_key`, with three tracers (`hexgate.audit` / `hexgate.usage` / `hexgate.bans`); the instrumentation scope tells the platform which event type a span is. The httpx clients, both semaphores, the asyncio task set, the sync-thread fallback and the loop-rebinding logic are deleted — `emit()` just enqueues onto the processor's thread-safe queue, so it behaves the same on a loop thread, in an executor, or with no loop anywhere. Registry keyed by `api_key` alone (all three event types share one sender per key).
- Two safeguards beyond section V: `sampler=ALWAYS_ON` and an empty parent `Context` per span, so a customer's own OTel tracing can neither parent our spans nor apply its sampling rate to them. Export timeout is 5s (`DEFAULT_EXPORT_TIMEOUT`) instead of OTel's 30s — enforced by a `shutdown()` override (`_BoundedShutdownProcessor`), because `BatchSpanProcessor`'s `export_timeout_millis` constructor param is dead upstream (open-telemetry/opentelemetry-python#4555) and the default shutdown path would otherwise block exit for up to 30s against an unreachable Collector.
- Queue saturation is detected in `emit()`: OTel's bounded deque silently evicts the *oldest* queued span when full (no signal, no log — upstream forbids logging on that path), so the sender counts evictions itself and warns on the first drop, then every 10th. The warning stays on the stdlib logger, never OTLP: it must reach stderr precisely when the OTLP pipeline is the thing failing.
- **Events** — `AuditEvent` / `LlmUsageEvent` / `BanEnforcementEvent` gain a `SCOPE` classvar and `span_attributes()` replaces `as_payload()`: keys are `semconv.*` constants, dict fields (`hint`/`arguments`/`attributes`) travel as JSON strings, unset optionals are omitted. The redaction/truncation pipeline is unchanged.
- **`hexgate/config/env.py`** — new `HEXGATE_OTLP_ENDPOINT`; falls back to `HEXGATE_API_URL + /v1/traces`.
- **Deleted** `pending_send_tasks` / `drain_pending_tasks` and their call sites in the google and openai runners — they patched an asyncio-task lifecycle that no longer exists.
- **Deps** — `opentelemetry-sdk` and `opentelemetry-exporter-otlp-proto-http` become explicit (already installed transitively via langfuse at 1.38; nothing new resolves).
- **Docs** — `audit-pipeline.md` and `ban-enforcement.md` describe the span format, endpoint resolution, and the shutdown contract (`await hexgate.audit.shutdown()` before exit; the provider's `atexit` hook is a best-effort safety net, not the contract).

Coexistence: the legacy `/v1/audit/*` endpoints are **not** touched. Hexkit consumes `hexgate` via PyPI (`>=0.2.8`), so it keeps POSTing JSON until it upgrades — the version pin is the coexistence mechanism, and migration step 3 (removing the endpoints) stays gated on that.

PRs 0–5 of the migration stand up Collector → Redpanda → enricher → ClickHouse, but nothing emits real OTLP spans into it yet; PR 6's end-to-end staging test can't run until the SDK does. Moving the SDK to OTel also removes the most fragile code in the package (loop affinity, non-daemon fallback threads, per-adapter drains) in favour of the standard batch exporter, and lets customers point the same spans at their own OTel backend.

Deployment: the Collector will be exposed behind the API host (proxy route `/v1/traces` → collector), so the SDK's default endpoint works with zero config. The ordered rollout checklist (deploy compose + proxy route + staging verification before PyPI publish) is in the "OpenTelemetry migration design" Notion doc, "Deployment checklist" section.

## Tests

- `tests/tracing/test_sender.py` rewritten around an injected `InMemorySpanExporter`: scope selection, `start == end == occurred_at` (ns precision), root span even inside a caller's active span, always sampled, emit from a plain thread, JSON-string dict fields / native list fields, `close()` flush + idempotency, real exporter construction (endpoint, bearer, timeout). Plus: saturation drop counting with the rate-limited warning, a pin on the OTel private queue attributes the detection reads (an `opentelemetry-sdk` upgrade that moves them fails loudly), and `close()` forwarding the 5s bound into the processor's `shutdown()`.
- `tests/tracing/test_senders.py`: registry keyed by `api_key`, `HEXGATE_OTLP_ENDPOINT` precedence, local-mode gate, `shutdown()`.
- `tests/audit/test_event.py`, `tests/tracing/test_usage_event.py`, `tests/security/test_bans.py`: every wire-mapping test ported to `span_attributes()` with `semconv` keys; new assertions that unset dict fields are absent and token counts stay `int`.
- `tests/config/test_env.py`: `resolve_otlp_endpoint` precedence.
- `tests/conftest.py`: autouse fixture swaps the exporter for an in-memory one in every unit test (the first run of the suite actually POSTed to `app.hexgate.ai` at teardown); `real_span_exporter` marker opts out.
- `tests/audit/test_integration.py` (opt-in `-m integration`) now drives configure → emit → flush against a live Collector.
- `make check-all` green: SDK 2055 passed, platform-api 586 passed (enricher tests import the SDK's redaction pipeline), dashboard 205 passed, collector check OK.
- `tests/adapters/*/test_integration.py` (opt-in `-m integration`) — **these were broken by this PR and are now fixed.** The shared fixture never set `HEXGATE_OTLP_ENDPOINT`, so `resolve_otlp_endpoint()` fell back to `<api url>/v1/traces`; the control plane serves no such route, so every span export died with a 405 the SDK only logs and all four suites failed on a bare "row never landed in ClickHouse". The fixture now pins the endpoint to the Collector's OTLP/HTTP receiver, `poll_until`'s budget goes 10s → 30s (sized for the Collector's 5s batch timeout + the enricher's Kafka poll, not the old direct POST), and `OTEL_BSP_SCHEDULE_DELAY` drops the exporter's idle wait to 500ms. Docstrings and the `integration-tests` skill were refreshed too: they still described ClickHouse + platform-api alone, and the Collector authenticates keys against the `devtoken` table in **Postgres**, so platform-api must run on Postgres (`make platform-api-pg`), not SQLite.
- Verified against the full local stack (Postgres, ClickHouse, Redpanda, platform-api-pg, Collector, enricher): **5 passed** where it was 4 failed / 1 passed, rows in both `policy_decision` and `llm_invocation`, DLQ empty on all partitions, no Collector or enricher errors. Note CI runs no `-m integration` job, so the PR checks do not cover this.


guillaume-hexamind:
## Review findings

Four issues, verified against this branch and `opentelemetry-sdk` 1.38.0. Score = severity × confidence, out of 100.

---

### 1. Audit serialization moved onto the caller's thread, unguarded — **82/100**

`hexgate/tracing/_senders.py:192-197`

`emit()` now calls `event.span_attributes()` synchronously, which runs `json.dumps(..., default=str)` three times plus `_truncate_json` (`hexgate/audit.py:129`, `:205-222`). Previously `as_payload()` ran on the background task/thread (`_send()` / `_send_sync()`), so a serialization failure only lost the event.

The call chain is now `guards/runner.py:351` → `enforcer.decide()` → `enforcer.record()` → `emit()`, and `record()` (`hexgate/security/enforcer.py:169-171`) guards only the decision observer below it — not the emit. A self-referential argument dict survives `_snapshot`'s `copy.deepcopy` and raises `ValueError: Circular reference detected` inside `json.dumps`; a value whose `__str__` raises does the same via `default=str`. Either one propagates out of `decide()` and kills the tool call.

**Fix:** wrap the body of `emit()` in `try/except Exception: _log.exception(...)`, the way `emit_llm_usage` already wraps itself. Losing an audit event is acceptable; breaking enforcement is not.

---

### 2. Registry `shutdown()` lost `return_exceptions=True` and lost concurrency — **72/100**

`hexgate/tracing/_senders.py:315-318`

```python
senders = list(_senders.values())
_senders.clear()
for sender in senders:
    await sender.close()
```

vs. the previous `await asyncio.gather(*(s.close() for s in senders), return_exceptions=True)`. Two regressions:

- **Aborts on first failure.** With two api_keys, if sender A's `provider.shutdown()` raises, sender B is never flushed — its queued events are silently lost — and the exception escapes `await hexgate.audit.shutdown()` in the host's teardown.
- **Serial 5s budget.** `close()` blocks up to `DEFAULT_EXPORT_TIMEOUT` each, so N keys is now N×5s instead of 5s total — which defeats the point of `_BoundedShutdownProcessor` ("must not hold a host application's exit").

**Fix:** restore the `gather(..., return_exceptions=True)`, and log the exceptions it returns rather than dropping them.

---

### 3. Zero-config default endpoint points at a route that doesn't exist — **68/100**

`hexgate/config/env.py:59-62`

`resolve_otlp_endpoint()` falls back to `<HEXGATE_API_URL>/v1/traces`, i.e. the FastAPI control plane. `git grep 'v1/traces' -- platform/api/hexgate_api/` returns zero hits — the route does not exist, as this PR's own fixture comment states (`tests/adapters/conftest.py:90-94`: *"the platform API serves no such route — every export then dies with a 405 the SDK only logs"*).

The legacy `/v1/audit/*` endpoints are untouched but no longer written to, so anyone upgrading before the collector proxy route is deployed loses 100% of decision, usage and ban telemetry, with an `opentelemetry` ERROR log line as the only signal. This is what broke the four adapter integration suites.

**Fix:** this is a deploy-ordering dependency, not something a code change alone makes safe — gate the release on the proxy route being live. In the meantime, have `configure()` warn when the resolved endpoint is the API URL rather than a dedicated collector.

---

### 4. The provider is isolated from the customer's sampler and parent, but not from their OTel environment — **60/100**

`hexgate/tracing/_senders.py:147`

`TracerProvider(sampler=ALWAYS_ON, resource=_RESOURCE)` passes no `span_limits`, so `SpanLimits()` still reads process env (`sdk/trace/__init__.py:1212`, `:636-684`):

- **`OTEL_SDK_DISABLED=true`** → `get_tracer()` returns `NoOpTracer()` (`:1231-1232`), whose `start_span` accepts every kwarg `emit()` passes and returns `INVALID_SPAN`. `configure()` still returns a live-looking `AuditSender`, `emit()` still succeeds, and every audit event is silently discarded — a compliance-relevant silent failure a customer triggers by disabling *their own* tracing.
- **`OTEL_SPAN_ATTRIBUTE_VALUE_LENGTH_LIMIT`** / **`OTEL_ATTRIBUTE_VALUE_LENGTH_LIMIT`** (a common payload-trimming setting) → applied as `BoundedAttributes(..., max_value_len=...)` at `:802-805`, truncating the JSON-string `sec_ai.arguments` / `hint` / `attributes` mid-string. The enricher's `as_json_dict` then raises `SpanRejected` (`jobs/enricher/coerce.py:100-105`) for the **whole span** — so the decision record is lost, not just the one field.
- **`OTEL_SPAN_ATTRIBUTE_COUNT_LIMIT`** below the attribute count drops attributes, and `required(attrs, TOOL_NAME)` then rejects the span.
- A malformed or negative value for any of those raises `ValueError` from `SpanLimits._from_env_if_absent` (`:689-714`), which propagates through `TracerProvider.__init__` → `AuditSender.__init__` → `configure()`, failing agent bootstrap.

**Fix:** pass an explicit `SpanLimits(...)` with the limits this pipeline needs, and read `OTEL_SDK_DISABLED` yourself (or construct the tracer without going through the env-configured path) so a customer's setting can't silently mute the audit trail.

guillaume-hexamind (commented):
Logic seem good overall. Only conflicts to solve, and to check the claude issues above.

victorludvig:
Review findings addressed:

1. **Unguarded emit on the caller's thread** — fixed in c75c099: `emit()` body wrapped in `try/except`, logged and costs one event; an unserializable event can no longer kill the tool call. Follow-up for the redaction cycle guard: #164.
2. **shutdown() lost gather/return_exceptions** — fixed in e87289f: senders close concurrently again (N keys ≈ one 5s budget), one failing close no longer skips the rest, exceptions logged per sender.
3. **Default endpoint has no route until the proxy is deployed** — deploy-ordering, gated by the rollout checklist in the design doc (PyPI publish only after the staging end-to-end check). No `configure()` warning: post-rollout, `<api url>/v1/traces` *is* the recommended zero-config endpoint, so the warning would fire on every correctly configured install forever.
4. **Provider inherits the host's OTel env** — fixed in 6ef091e: explicit `SpanLimits` pinned to `UNSET` (no env reads, no truncation, no ValueError at configure), and `OTEL_SDK_DISABLED` forced off for the audit provider — it's the customer's telemetry kill switch, not the audit trail's (`HEXGATE_LOCAL_MODE` is that). Each vector pinned by a test.

guillaume-hexamind (approved):

PR #149, opened 2026-08-31
feat(sdk): serve OpenAI, Google ADK, and Pydantic AI agents (framework-agnostic streaming)

quanghexa94:
Serve every supported agent framework, not just the native one. `hexgate serve` relays a local agent to the dashboard Playground over a WebSocket, streaming normalized `StreamEvent`s; until now it only worked for the native (LangChain) `HexgateAgent` and `AttributeError`'d on an OpenAI, Google, or Pydantic agent. This makes all four frameworks serve, with policy enforcement, audit, and approvals intact. Blast radius: pure SDK, no platform change, no migration.

## Design

serve was welded to LangChain in two places: `build_runtime_from_local_agent` called `HexgateAgent.enforce_policy`, and `_run_chat_turn` streamed through `stream_agent` + `normalize_langchain_events`. The fix introduces one seam, `AgentRuntime.astream_normalized(input, ctx, query)`, bound per framework in `build_runtime_from_local_agent` (dispatching on `manifest.framework`, which `create_manifest` already computes). serve itself is now framework-blind.

Each framework maps its native run stream onto the existing `StreamEvent` contract (run-start, text/reasoning block deltas, tool start/end, run-end, error). Since that bookkeeping is identical across adapters, it lives once in `StreamAccumulator` (sequence, flat run id, blocks, tool-call correlation, `_tool_end_state` inference, JSON-safe `raw_output`, finish/error bracketing) driven by `run_normalizer`; each adapter is a thin `consume()` plus a small driver.

| Framework | How it streams | Notes handled |
| --- | --- | --- |
| OpenAI | `HexgateRunner.run_streamed().stream_events()` | args are a JSON string (parsed); errors are raised out of the stream |
| Google ADK | `HexgateRunner.run_async(..., RunConfig(SSE))` | session-based history (driver owns an in-memory session, fresh per conversation); SSE emits a duplicate final aggregate, deduped to partial deltas |
| Pydantic AI | `agent.iter()` node graph, flattened | tool failures are in-band (`ToolReturnPart.outcome` / `RetryPromptPart`), mapped to an explicit FAILED state; history carried across turns |

Enforcement ownership differs by design: the native path calls `enforce_policy`; OpenAI/Google/Pydantic each let their own runner/proxy resolve and hot-reload the policy binding, so the driver builders skip `enforce_policy`.

## Important files

| File | What to look at |
| --- | --- |
| `hexgate/streaming/_accumulator.py` | the shared base + `run_normalizer`; read this first, the adapters lean on it |
| `hexgate/adapters/{openai,google,pydantic_ai}/streaming.py` | each `consume()` mapping + the serve driver (Google session mgmt, Pydantic node flattening) |
| `hexgate/cli/_common.py` | `build_runtime_from_local_agent` dispatch + the shared `_serve_runtime_envelope` |
| `hexgate/cli/serve.py` | `_run_chat_turn` now drives the seam (dropped the direct `stream_agent` coupling) |
| `tests/streaming/test_accumulator.py`, `tests/adapters/*/test_streaming.py` | the base + per-framework normalizer coverage; `tests/cli/test_serve.py` for dispatch |

## Test plan

Full SDK suite green (exit 0) on `pytest -m "not integration and not framework_compat"`, ruff check + format clean. New coverage: the shared accumulator (13 cases: blocks, tool correlation, callless FIFO, state inference, run bracketing, `run_normalizer` clean + raise paths), the OpenAI/Google/Pydantic normalizers (~10 cases each, incl. the SSE dedup, the pydantic outcome→FAILED mapping, fatal-error termination), and serve dispatch for all four frameworks with the runners/wrappers mocked so no network is touched. I also verified against the actually-installed `agents`, `google.adk`, and `pydantic_ai` that `create_manifest` routes each demo agent to the right `framework` with the expected tools. A reviewer catches a regression by running the adapter test suites; the mapping tests use hand-built fakes shaped like each SDK's real events. Not yet done: the live browser BYOK smoke (needs provider + platform keys), which is the "Try it" below.

## Try it

```
make demo-notebook-build   # one-time
HEXGATE_NOTEBOOK=deploy/openai_demo_notebook.py   make demo-notebook   # or google_/pydantic_
# open http://localhost:2718, paste an OpenAI key, Start, open the Playground
# get_order_status streams + is allowed; refund <= $100 prompts for approval; > $100 is denied
```

## Notes

- Re-lands #147, which was accidentally merged and then reverted in #148. Same change, replayed onto current `main`; the full suite is green on the new base.
- Unchanged on purpose: the LangChain normalizer (`hexgate/streaming/normalize.py`) and `hexgate chat` (a separate REPL loop; this PR is serve only).
- Rollback: pure SDK, revert the three commits; no state or schema touched.
- Known limitations, documented in code: a pydantic `denied` outcome renders as FAILED (the `StreamEvent` contract has no DENIED state yet); builtin/hosted pydantic tools are not rendered (provider-side, not enforced, event shape in flux); an OpenAI tool error the SDK folds into plain output text reads as COMPLETED (no structured marker). Enforcement and audit are unaffected in every case.

### DIFF
--- a/deploy/google_demo_notebook.py
+++ b/deploy/google_demo_notebook.py
@@ -1,0 +1,217 @@
+"""Hexgate live-demo notebook (marimo) — Google ADK edition.
+
+Twin of ``openai_demo_notebook.py`` / ``demo_notebook.py``, but the agent is a
+raw Google ADK ``Agent``. It exercises the Google serve path end to end:
+``serve_manager`` runs the ``hexgate serve`` loop bound to the live ``Agent``,
+``hexgate serve`` dispatches on the manifest framework to the Google runtime
+(``GoogleServeDriver`` owns the ADK session + SSE streaming), and every turn
+streams through the Google → hexgate ``StreamEvent`` normalizer to the dashboard
+Playground.
+
+BYOK is an OpenAI key: the ADK agent runs on OpenAI via ``LiteLlm``. Your key
+lives only in this throwaway container and is never written to disk. Run by
+boot.py via ``marimo edit``.
+"""
+
+import marimo
+
+__generated_with = "0.23.10"
+app = marimo.App(width="medium")
+
+
+@app.cell
+def _():
+    import sys
+    from pathlib import Path
+
+    import marimo as mo
+
+    sys.path.insert(0, str(Path(__file__).resolve().parent))
+    import serve_manager
+
+    return Path, mo, serve_manager
+
+
+@app.cell
+def _(mo):
+    mo.md("""
+    # 🛡️ Hexgate — run a live **Google ADK** agent
+
+    A **throwaway sandbox** — everything vanishes when it scales down. The
+    Google-framework twin of the native demo: same gate, same dashboard, a
+    `google.adk.agents.Agent` instead of a hexgate agent.
+
+    1. **Define your tools** (plain Python functions).
+    2. **Define your agent** (a factory — built on Start, after your key).
+    3. **Enter your OpenAI key and start it** (the ADK agent runs on OpenAI via LiteLlm).
+    4. **Open the playground** to chat and watch policy decisions stream.
+    """)
+    return
+
+
+@app.cell
+def _(mo):
+    mo.md("""
+    ## 1 · Define your tools
+    """)
+    return
+
+
+@app.cell
+def _():
+    # Plain Python functions — ADK wraps them as tools. The function name is the
+    # tool name the policy references below.
+    def get_order_status(order_id: str) -> str:
+        """Look up the delivery status of an order by its id."""
+        return f"Order {order_id}: shipped, arriving Tuesday."
+
+    def refund_order(order_id: str, amount: float) -> str:
+        """Issue a refund of `amount` USD for `order_id`. A side-effecting tool."""
+        return f"Refunded ${amount:.2f} for order {order_id}."
+
+    TOOLS = [get_order_status, refund_order]
+    return (TOOLS,)
+
+
+@app.cell
+def _(mo):
+    mo.md("""
+    ## 2 · Define your agent
+    """)
+    return
+
+
+@app.cell
+def _(TOOLS):
+    # A factory — built on Start (step 3), after your key. The `name` is the
+    # manifest name and the policy lookup key on the platform.
+    from google.adk.agents import Agent
+    from google.adk.models.lite_llm import LiteLlm
+
+    def build_agent():
+        return Agent(
+            name="demo_google_agent",
+            model=LiteLlm(model="openai/gpt-4o-mini"),
+            instruction=(
+                "You are a customer support agent. Help with orders and refunds. "
+                "Confirm details before issuing a refund."
+            ),
+            tools=TOOLS,
+        )
+
+    return (build_agent,)
+
+
+@app.cell
+def _(mo):
+    mo.md("""
+    ## 3 · Add your OpenAI key & start
+    """)
+    return
+
+
+@app.cell
+def _(mo):
+    api_key = mo.ui.text(kind="password", placeholder="sk-...", full_width=True)
+    start = mo.ui.run_button(label="▶ Start agent")
+    mo.vstack([mo.md("**OpenAI API key**"), api_key, start])
+    return api_key, start
+
+
+@app.cell
+def _(api_key, build_agent, mo, serve_manager, start):
+    # serve_manager is framework-agnostic — build_runtime_from_local_agent
+    # dispatches on the manifest framework and builds the Google runtime.
+    import os
+    import time
+
+    if start.value:
+        if not api_key.value:
+            out = mo.md("⚠️ **Enter your OpenAI key above**, then click Start.")
+        else:
+            os.environ["OPENAI_API_KEY"] = api_key.value  # BYOK
+            agent = build_agent()
+            serve_manager.apply(agent)
+            time.sleep(3)  # let it build the runtime, auto-register + dial /v1/serve
+            st = serve_manager.status()
+            if st == "running":
+                out = mo.md(
+                    f"✅ **Agent running** (key `…{api_key.value[-4:]}`). "
+                    "Open the playground below."
+                )
+            elif st.startswith("error"):
+                out = mo.md(f"❌ **Failed to start:** `{st}`")
+            else:
+                out = mo.md(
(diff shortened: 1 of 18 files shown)


guillaume-hexamind:
### Review findings (verified against installed google-adk 1.32.0 / openai-agents 0.15.1)

**1. Tool calls extracted from partial SSE events — 78/100**
`hexgate/adapters/google/streaming.py:48-65` runs `get_function_calls()` / `get_function_responses()` on every ADK event; only text/reasoning is gated on `partial` (line 68). But `PROGRESSIVE_SSE_STREAMING` is `default_on=True` in ADK 1.32 (`features/_feature_registry.py:143`), and in that mode `StreamingResponseAggregator.process_response` marks every intermediate chunk `partial=True` — including chunks carrying a `function_call` — then `close()` re-emits it in the `partial=False` aggregate. ADK's own flow guards against this (`base_llm_flow.py:982`, *"Skip partial function call events"*); we don't.
- Non-streaming args: duplicate `ToolStartEvent` with the same `adk-` id, and `finalize_blocks()` fires twice (`_accumulator.py:205`), splitting the text block.
- Streaming args (`fc.partial_args` / `will_continue`): continuation chunks arrive with an empty `fc.id`, so `populate_client_function_call_id` (`base_llm_flow.py:103`) mints a fresh `adk-<uuid>` per chunk → N phantom tool rows, empty args, name `"tool"`, stuck at `STARTED` forever (only the aggregate's id ever gets a `ToolEndEvent`).

Not caught because the demo notebook uses `LiteLlm`, whose streaming path only emits function calls in the final aggregate, and because `tests/adapters/google/test_streaming.py` builds every `_call(...)` with the default `partial=False`.

**Fix:** gate lines 48–65 on `not getattr(event, "partial", False)`, mirroring `base_llm_flow.py:982`. Add a test with `partial=True` + a function call.

---

**2. Blocking sync HTTP on the serve event loop, every turn — 62/100**
`astream_openai` (`hexgate/adapters/openai/streaming.py:140`) calls `runner.run_streamed(...)` straight from the asyncio loop, and `run_streamed` uses the **sync** entry points: `_binding_for()` (first turn: blocking `resolve_policy` → `get_agent`), `binding.refresh()` and `ban_gate.check()` (`runner.py:302-307`). Both sources are ETag-cached but have no time-based gate, so each is a real HTTP round-trip per turn. A blocking call inside a coroutine freezes the whole loop thread — which defeats the deliberate task-per-frame dispatch at `serve.py:461-463` (*"critical for approval.reply arriving mid-chat"*): no approval reply and no ping/pong is serviced meanwhile. Every other path in the PR correctly uses `refresh_async` / `check_async` (both `asyncio.to_thread`), including this runner's own `async def run` (`runner.py:200-204`).

**Fix:** add `arun_streamed` that awaits `refresh_async()` / `check_async()` first, then calls `Runner.run_streamed` on-loop inside the `hexgate_context` scope. A naive `to_thread(run_streamed, ...)` won't work — it ends in `asyncio.create_task`, which needs a running loop in the calling thread, and the spawned task must inherit the contextvar scope.

---

**3. `ttl_seconds` dropped and `session_id` overwritten for Google agents — 45/100**
`serve.py:292-297` parses both off the Playground's `user_attenuation`, but `GoogleServeDriver.astream` (`google/streaming.py:171-176`) rebuilds a fresh `HexgateContext` with only `user_id` / `user_roles` / `attributes`, dropping `ttl_seconds` and replacing `session_id` with the ADK session UUID. Both siblings pass `ctx` through unchanged (`pydantic_ai/streaming.py:164`, `cli/_common.py:389`) — Google is the only outlier.
- `ttl_seconds` feeds the Biscuit lifetime caveat (`agents/factory.py:266` → `cloud/attenuate.py:100-103`), so an operator-set short TTL is silently replaced by the default.
- `session_id` lands in every audit row (`security/enforcer.py:172` → `audit.py:215`), so Google turns can't be correlated with the same conversation under another framework.

**Fix:** carry `ttl_seconds=ctx.ttl_seconds if ctx else None` through, and keep the ADK session id out of `HexgateContext.session_id` — it's already held in `self._session_id` and passed to `run_async` separately.

guillaume-hexamind (commented):
3 claude review high issues above, then fact checked, to investigate

Otherwise the logic seems good, I approve; we could even think later to ship the event translation as a separate package independant from hexgate

quanghexa94:
Thanks Guillaume, great catches. All three fact-checked findings are fixed and pushed (`e7e57f4`, `d56dcde`, `e63dfb6`).

**1. Partial SSE tool calls (google).** Gated the `get_function_calls()` / `get_function_responses()` extraction on the non-partial aggregate, matching `base_llm_flow`'s "skip partial function call events." Text comes from the partials, tools from the aggregate. Added a test feeding a partial function call then the aggregate, asserting a single `ToolStart`.

**2. Blocking sync HTTP on the serve loop (openai).** Added `HexgateRunner.arun_streamed` (awaits `refresh_async()` / `check_async()`, then `Runner.run_streamed` on-loop, sharing a `_launch_streamed` helper with the sync path), and `astream_openai` now awaits it. Added ordering tests: async refresh before launch, and the ban gate awaited before the background task spawns.

**3. ttl_seconds / session_id (google).** `ttl_seconds` now carries through. On `session_id` I went a slightly different route than passing the ADK id straight through: the google `run_async` reads the ADK session from `hexgate_context.session_id`, so instead I added an optional `session_id` param to `run_async` that overrides it. `GoogleServeDriver` now mints/evicts the ADK session id internally and passes it via that param, while the caller's own `session_id` + `ttl_seconds` flow through `HexgateContext` to audit unchanged. Net effect is what you asked for (audit stays correlatable, the ADK uuid stays out of the audit session), and it also lets ADK memory reset per conversation independently.

For good measure I ran `/code-review high` a couple more passes on top and fixed the fallout: unbounded ADK session growth, aggregate-only / multi-segment text handling (now via suffix emission so the SSE aggregate is deduped without dropping a non-streamed segment), id-less calls colliding on the tool name, and a tool left open when a run ends mid-call (now closed as FAILED so it doesn't hang at STARTED). Full SDK suite green, ruff clean.

On shipping the event translation as a standalone package: agree, that's a clean seam. The shared accumulator + per-framework `consume()` are already framework-independent, so splitting `hexgate.streaming` out is mostly mechanical. Happy to do it as a follow-up rather than grow this PR.

guillaume-hexamind (approved):
Good to me!

PR #150, opened 2026-08-31
feat(sdk): run namespace and writer

guillaume-hexamind:
> 📘 Reviewer's mental model (concepts, examples, no file-level detail):
> [How `run.*` caps work — PR #150](https://app.notion.com/p/3cdfb45dbae281749595c17273c89590)

## What

Makes the `run.*` namespace a real policy input. The previous PR in the stack
(`gp/feat/run_facts_and_scope`) added the `RunFacts` accumulator and opened a run
scope at every adapter boundary, but nothing wrote to it and no engine read it.
This PR wires **the writers, the readers, and the load-time validation**, so a
policy can now stop a runaway invocation:

```yaml
default_policy:
  mode: allow
  constraints:
    - run.tool_calls < 20         # runaway-agent circuit breaker
    - run.elapsed_seconds < 300   # wall-budget breaker
    - run.calls_of_this_tool < 5  # per-tool cap, without naming any tool
    - run.errors < 5
    - run.total_tokens < 200000
```

## Changes

**Writers — facts now accumulate**
- `guards/runner.py`: both the sync and async guarded paths record tool
  executions, errors, denials and approval gates. Executions are counted around
  `invoke`, before the post-guards, so a post-guard halt can't un-count a call
  whose side effect already happened. A denied call consumes no tool budget; it
  counts as a denial instead.
- `tracing/usage.py`: `emit_llm_usage` records tokens into the run **before** the
  sender check, so a token cap still works in local mode / with no platform key.

**Readers — both engines see `run`**
- `PolicyEnforcer.decide` builds the `run` snapshot once per decision (not per
  role, so roles can't disagree about the same run) and threads it through.
- Pydantic path: `PolicySet.evaluate` → `evaluate_tool_call` → `check_constraints`
  exposes it as the `run.<name>` namespace.
- WASM path: `PolicyBundle.evaluate` → `WasmPolicy.decide` adds `input.run` to the
  composed input document. `PolicyEngine` protocol updated accordingly.

**Projection — 13 paths**
`RunFacts.as_namespace(tool_name)` now projects the full set. `tool_name` is
required because `calls_of_this_tool` is the one value that isn't monotone across
a run. `total_tokens` and `elapsed_seconds` are derived (the grammar has no
arithmetic or time functions), and `tools_used` is first-use-ordered and
deduplicated. Paths are split into `SCALAR_PATHS` / `LIST_PATHS`.

**Load-time validation — no silent failure modes**
`_validate_run_refs` runs at `PolicySet` construction, alongside the existing
const-ref check, and rejects two otherwise-silent mistakes:
- an unknown or too-deep path (`run.tool_call`, `run.id.value`) — would resolve to
  missing and deny *every* call;
- a list-valued path in scalar position (`run.tools_used not in [...]`) — would
  silently *pass* every call. The error suggests `not any(run.tools_used, . == "x")`.

Supported by a new `iter_cmp_operands` in `constraints.py`, which keeps the
operator and side (unlike `iter_arg_refs`) so `count(x) <= 6` can be told apart
from `x <= 6`.

**Testing surface**
- `security.testing.run_namespace(tool, **facts)` — build a run namespace over a
  zeroed run; raises on an unregistered keyword so a typo can't silently leave the
  counter at 0. `assert_allows/denies/needs_approval` take `run=`, defaulting to a
  freshly-started run.
- `hexgate policy test --run-facts '{"tool_calls": 20}'` — dry-run a circuit
  breaker at its threshold, on both `--engine pydantic` and `--engine wasm`.

**Docs**
- `docs/policy/constraints.mdx`: a `run.*` section — the 13-path table, what each
  counter counts (denials vs. executions, approval fired vs. granted, pre- vs.
  post-guard halts), the three places counts lag reality (parallel calls, tokens
  trailing a turn, sub-agents not rolling up), and where facts aren't collected
  (egress, `agent.batch()`, raw threads).
- `docs/policy/writing-conditions.mdx`: a short "stopping a run that's gone wrong"
  section pointing at the reference.

## Notes for review

- **Circuit breakers, not quotas.** Under `N` concurrent tool calls a cap of `K`
  permits up to `K - 1 + N` executions — all concurrent calls read the counter
  before any of them records. Documented rather than locked, deliberately.
- **Fail-open outside a run scope.** Where no run is open (egress, `batch()`),
  every counter reads 0 and a cap passes. An unwired boundary shouldn't brick the
  agent.
- Stacked on `gp/feat/run_facts_and_scope` — review that one first.

## Tests

+1600 lines across `tests/guards/test_runner_run_facts.py`, `tests/security/`
(`test_run_refs_validation.py`, `test_constraints.py`, `test_enforcer.py`,
`test_engine_parity.py`, `test_rego_compile.py`, `test_builder.py`,
`test_analyzer.py`), `tests/tracing/test_usage_emit.py`, `tests/cli/test_policy.py`
and `tests/egress/test_gate.py`. Engine parity is asserted for `run.*` the same
way it is for `ctx.*`.


hexgate/guards/runner.py
```diff
@@ -287,6 +326,8 @@ async def _run_post_async(
         res = await _call_guard_async(guard, call, outcome)
         if isinstance(res, Halt):
             halt_decision = _halt_to_decision(res, call)
+            # No execution recorded: the tool already counted around ``invoke``.
+            _record_run_decision(halt_decision.outcome)
```
quanghexa94, line 338: **Post-guard halt double-counts as both an execution and a denial.** A tool runs (`tool_calls++` around `invoke`), then a post-guard returns `Halt` → this `_record_run_decision(halt_decision.outcome)` bumps `run.denials`. So one call counts as *both* an execution and a denial. But `docs/policy/constraints.mdx` documents an after-the-tool halt only as "still counts the call" (not a denial), and frames `run.denials` as refused calls that consume no tool budget. Effect: a `run.denials < N` breaker trips earlier than documented, and output-filtering post-guards get conflated with an agent asking for forbidden things. Mirrored in the sync path (`_run_post_sync`, ~line 509).

hexgate/guards/runner.py
```diff
@@ -363,9 +408,12 @@ async def run_guarded_async(
                     _notify(pipeline, call, mods, blocked=True)
                 return render_error(decision)
 
+    # Before dispatch, not after: see _record_run_execution.
+    _record_run_execution(call.tool_name)
```
quanghexa94, line 420: **The "no `await` between `decide()` and `record_execution`" atomicity guarantee doesn't hold for approval-gated calls.** `_record_run_execution`'s docstring (lines 231–247) relies on there being no `await` between the counter read in `enforcer.decide()` and this increment, making read-then-increment atomic on the event loop. But when the outcome is `NEEDS_APPROVAL` and granted, `await resolve_approval_async(...)` (~line 401) sits between the read and this line. With an auto-approving handler and N concurrent approval-gated calls, all N read the same pre-increment `tool_calls` before any records — so a `run.tool_calls < K` cap admits more than K executions, the exact parallel-overshoot the PR fixes for the non-approval path. Worth either documenting this as a known gap or recording the execution before awaiting approval.

hexgate/security/testing.py
```diff
@@ -25,19 +32,71 @@
...
+    """
+    unknown = sorted(set(facts) - KNOWN_RUN_PATHS)
+    if unknown:
+        raise ValueError(
+            f"unknown run.* path(s) {unknown} "
+            f"(this build knows: {', '.join(sorted(KNOWN_RUN_PATHS))})"
+        )
+    for name, value in facts.items():
+        _check_run_value(name, value)
+    return {**_zeroed_run(tool), **facts}
```
quanghexa94: **`run_namespace` never recomputes derived fields, so `total_tokens` can diverge from production.** This flat merge `{**_zeroed_run(tool), **facts}` overlays raw kwargs onto the already-projected zeroed dict. `run_namespace(input_tokens=100, output_tokens=50)` yields `total_tokens=0` (from the zeroed base), while production `RunFacts.as_namespace` derives `total_tokens=150`. A test asserting a `run.total_tokens < 120` breaker with those inputs reads 0, passes, and silently green-lights a cap that fires for real. Safer to build a `RunFacts` and project it via `as_namespace`, so derived invariants stay consistent, rather than merging into the projected dict.

hexgate/security/testing.py
```diff
@@ -11,12 +11,19 @@
 
 ``policy`` may be an :class:`AgentPolicy` (single role) or a :class:`PolicySet`
 (role-aware — pass ``role=``).
+
+Assert a ``run.*`` cap by supplying the run's facts via :func:`run_namespace`:
+
+    assert_denies(policy, "refund", run=run_namespace(tool_calls=20))
```
quanghexa94: **This documented example makes `run.calls_of_this_tool` silently read 0.** `run_namespace(tool_calls=20)` leaves `tool` defaulting to `""`, so `calls_of_this_tool = _calls_by_tool.get("", 0) = 0` regardless of the `"refund"` tool being asserted — the `tool` that keys `calls_of_this_tool` is decoupled from the tool passed to `assert_denies`. A policy on `refund` using `run.calls_of_this_tool >= 5` would evaluate against 0 in such a test and never exercise the per-tool cap it appears to test. Consider defaulting `run_namespace`'s tool from the asserted tool, or at least showing `run_namespace("refund", calls_of_this_tool=5)` in the example.

hexgate/cli/policy/main.py
```diff
@@ -802,12 +820,39 @@ def _main_test(args: argparse.Namespace) -> int:
...
+    """Parse ``--run-facts`` over a zeroed run, so an unset path reads zero
+    rather than failing the dry-run closed. Raises :class:`ValueError` with a
+    printable message; the caller renders it."""
+    try:
+        parsed: Any = json.loads(raw)
+    except json.JSONDecodeError as exc:
+        raise ValueError(f"--run-facts is not valid JSON: {exc}") from exc
+    if not isinstance(parsed, dict):
+        raise ValueError("--run-facts must be a JSON object (dict).")
+    unknown = sorted(set(parsed) - KNOWN_RUN_PATHS)
```
quanghexa94, line 847: **Dead duplication of the unknown-path check.** `_resolve_run_facts` computes `unknown = sorted(set(parsed) - KNOWN_RUN_PATHS)` and raises here, then calls `run_namespace(tool, **parsed)` (~line 848), which recomputes the identical unknown-key check (`testing.py:44–49`) before its type check. The CLI-side check can never fire once `run_namespace` is reached — it's a second copy with a slightly different message to keep in sync. Dropping it and letting `run_namespace` be the single validator (wrapping its `ValueError` for both unknown and wrong-typed cases) removes the drift risk.

quanghexa94 (commented):
Automated code-review pass (high effort). The PR is well-considered and heavily commented; genuine defects are few. Five notes inline — the two in `guards/runner.py` are the substantive ones (behavior contradicts the documented `run.*` accounting), the two in `testing.py` are test-helper traps, and the CLI one is a cleanup.

guillaume-hexamind:
Thanks — reviewed all five. Four addressed in 75a69dc; the fifth left as-is with reasoning below.

**1. Post-guard halt counts as execution + denial — no code change.** Behaviour is deliberate and already pinned (`test_post_guard_halt_still_counts_the_execution`), and it matches the reference table's definition of `run.denials` ("calls the policy or a guard refused"). The conflation premise doesn't hold either: post-guards can't filter output — `_reject_post_proceed` rejects `Proceed(result=...)` as "a later phase" — so a post-guard `Halt` is only ever a refusal. What *was* wrong is one word of prose: `constraints.mdx` said a denied call counts toward `run.denials` "instead", which reads as exclusive. The after-halt bullet now says it counts both.

**2. Approval gate breaks the parallel-cap guarantee — confirmed, documented.** Real, but not for the stated reason. `bool`, sync callables and non-suspending coroutines never yield, so the cap holds; it needs a handler that *actually* suspends — which is every realistic one. Also async-only, and only the policy's own gate (a pre-guard's approval awaits *before* `decide()`, leaving the window closed). Documented as a fifth circuit-breaker caveat plus a note on `_record_run_execution`, and pinned by four tests: three asserting the cap holds when nothing suspends, one characterisation test asserting the overshoot with a note to delete both if it ever closes. Not fixed in code: recording before the await would count executions for approvals that get refused, breaking the documented `run.approvals` contract. A real fix re-checks the cap after the grant — separate PR.

**3. `run_namespace` doesn't derive `total_tokens` — fixed.** `run_namespace(input_tokens=100, output_tokens=50)` now yields `150`. Not via the suggested "build a `RunFacts` and project it": all 13 paths incl. derived ones are in `KNOWN_RUN_PATHS`, so that would make `total_tokens` and `elapsed_seconds` unsettable. Derived only when the split is supplied and the total isn't; an explicit value still wins.

**4. `calls_of_this_tool` reads 0 — fixed, and it was worse than reported.** The `tool` parameter was *entirely inert*: `_zeroed_run` builds an empty `_calls_by_tool`, so `calls_of_this_tool` was 0 and `tools_used` `[]` for any tool string. Defaulting `tool` from the asserted tool therefore wouldn't have helped. It now credits the named tool, so `run_namespace("refund", tool_calls=20)` projects byte-identically to a real 20-call run — which also fixes the `tool_calls=20` / `tools_used=[]` state no run can reach.

**5. Duplicated unknown-path check in the CLI — left as-is.** The duplication earns its keep: `_resolve_run_facts` wraps every `ValueError` from `run_namespace` as "wrong-typed value", so dropping the pre-check makes a typo print `--run-facts has a wrong-typed value: unknown run.* path(s) ['tool_call']`. Worth doing only with distinct exception types; low value against one shared `KNOWN_RUN_PATHS` constant.

3 and 4 both flow through to `hexgate policy test --run-facts`. These two dry-runs silently reported ALLOW before and correctly DENY now:

```
--run-facts '{"input_tokens": 100, "output_tokens": 50}'  →  ✗ run.total_tokens < 120
--run-facts '{"tool_calls": 5}'                           →  ✗ run.calls_of_this_tool < 5
```

guillaume-hexamind:
@quanghexa94 all 5 addressed, ready for another review!

hexgate/guards/runner.py
quanghexa94: Agreed — since `_reject_post_proceed` means a post-guard halt can only ever be a refusal (no output rewrite in v1), counting it toward `run.denials` as well as the call is the right call. The reworded bullet matches the behavior now, and the two tests pin both halves. Good.

hexgate/guards/runner.py
quanghexa94: Thanks for digging into this. Keeping the behavior and documenting it is the right move — the caveat + characterization test capture it well, and I agree recording before the await would break the `run.approvals` contract. Could we open a tracking issue for the re-check-after-grant fix so the follow-up doesn't get lost? Happy either way.

hexgate/security/testing.py
quanghexa94: Nice — deriving `total_tokens` only when the split is supplied (and letting an explicit total win) is cleaner than what I suggested, and avoids making the derived paths unsettable. 👍

hexgate/security/testing.py
quanghexa94: Good catch that the `tool` param was completely inert before — crediting the named tool so `run_namespace("refund", tool_calls=20)` projects like a real run is exactly right.

hexgate/cli/policy/main.py
quanghexa94: Fair enough — since the wrapper reports every `ValueError` as "wrong-typed value", dropping the pre-check would misreport an unknown path. Fine to leave as-is.

quanghexa94 (approved):
Re-checked against 75a69dc — this all looks good.

- #3 / #4 fixed cleanly; the token-total derivation is actually nicer than my first suggestion, and crediting the named tool closes a state a real run could never reach.
- #1's reworded bullet matches the real behavior now that post-guards can't rewrite output, and it's pinned by tests.
- Happy to ship the approval-gate overshoot as a documented caveat with the fix deferred — left a note about tracking that follow-up.

Approving. Thanks for the thorough writeup.

quanghexa94 (approved):

quanghexa94:
LGTM, rebase first though!

PR #153, opened 2026-08-31
 feat(clickhouse, platform-api): run_id and run counters on policy_decision

guillaume-hexamind:
> 📘 Reviewer's mental model (concepts, examples, no file-level detail):
> [How run_id ties a decision to its run — PR #153](https://app.notion.com/p/3cdfb45dbae28137a5c5de0a5d75e05e)

## What

Gives the audit log a place to put the `run.*` tally #129 built and #150 made
enforceable. Both live entirely inside the SDK process — the tally is gone the
moment it exits. This PR adds `run_id` plus a snapshot of the run counters to
every `policy_decision` row (and `run_id` alone to `llm_invocation`, so a model
call joins back to the run it belongs to):

```sql
run_id            UUID     DEFAULT toUUID('00000000-0000-0000-0000-000000000000')
run_tool_calls    UInt32   DEFAULT 0
run_llm_calls     UInt32   DEFAULT 0
run_denials       UInt32   DEFAULT 0
run_total_tokens  UInt32   DEFAULT 0
run_elapsed_ms    UInt32   DEFAULT 0
```

Today `policy_decision` is a stream of isolated rows — nothing says "these 20
decisions were one invocation." That blocks a per-run timeline in the
dashboard, blocks giving a customer a real distribution of tool-calls-per-run
to calibrate their own `run.tool_calls < K` cap against, and makes "why was
this call denied" a single row with no context.

**Platform-only, no SDK change.** Every row still reads the zero UUID / zero
counters until the SDK is taught to stamp the `run_ns` snapshot the enforcer
already builds once per decision (`security/enforcer.py`, since #150) onto the
outgoing payload — that's the next PR in the stack. Shipping the schema first
means the columns exist everywhere before any SDK build starts sending real
values, so API and SDK never need a coordinated deploy.

## Changes

**Schema**
- `platform/clickhouse/init/schema.sql`: the six columns on `policy_decision`,
  `run_id` on `llm_invocation`. Additive, `DEFAULT`-ed, appended (no `AFTER`) to
  match `_DECISION_COLUMNS`' "order matches schema.sql" convention.
- `platform/clickhouse/migrations/0002_add_run_columns.sql`: hand-applied ALTER
  for both tables, same header discipline as `0001` (must be applied **before**
  deploying this API build).

**Wire model + ingest**
- `schemas.py`: `DecisionEvent` gains `run_id: UUID | None` + the five counters
  (`Field(default=0, ge=0)`); `LlmInvocationEvent` gains `run_id`. `None` is the
  wire default — the zero-UUID substitution happens at insert time, not on the
  model, so the model reflects what the caller actually sent.
- `core/clickhouse.py`: `ZERO_RUN_ID` sentinel — the columns are `UUID`, not
  `Nullable(UUID)`, matching every other advisory column in this schema.
- `features/audit/service.py`, `features/llm_invocations/service.py`: both
  `_*_row` builders substitute `event.run_id or ZERO_RUN_ID`.

**Fixed while wiring `llm_invocations`**
- `features/llm_invocations/router.py`: added the missing `ProgrammingError` →
  503 branch (was falling through to a permanent-looking 422). `run_id` is the
  first column ever added to `llm_invocation` since its original `CREATE
  TABLE`, so "schema behind this build" became a reachable deploy-ordering
  state for the first time — mirrors the decision endpoint's existing handling.

## Notes for review

- **This branch was rebased after opening.** The base of the stack
  (`gp/feat/run_facts_policy_run`) picked up a large upstream refactor that
  generalized the audit schema-guard into shared `SchemaOutOfDate` /
  `verify_written_columns` / `verify_all` machinery in `core/clickhouse.py`,
  plus batch-insert support (`BatchItem`, `insert_batch`) for the
  span-enricher job. That collided with this PR's changes to the same three
  files. Resolution: adopted the new shared machinery as-is (it's strictly
  better than this PR's original per-file `verify_schema` loop) and merged
  `ZERO_RUN_ID` + the `run_id`/counter columns on top of it — no functional
  change to what this PR set out to do.
- No new columns in `_LIST_COLUMNS` or any dashboard response schema — a
  per-run timeline view is unblocked by this PR, not shipped by it.
- `Decision`/audit-payload wiring on the SDK side (`hexgate/security/enforcer.py`,
  `hexgate/audit.py`) is deliberately out of scope — next PR in the stack.

## Tests

New/extended coverage in `tests/features/audit/test_audit.py` and
`tests/features/llm_invocations/test_llm_invocation.py`: wire-model defaults +
validation, insert round-trip (zero-UUID default and explicit-value paths),
`verify_schema` schema-drift coverage, the `ProgrammingError` regression test,
and both real-ClickHouse integration round-trips extended to assert the new
columns.

`make fmt-check` clean, `ruff check .` clean, 527 unit + 6 integration tests
passing (`make clickhouse-up` for the integration run).

### DIFF
--- a/platform/api/hexgate_api/features/llm_invocations/service.py
+++ b/platform/api/hexgate_api/features/llm_invocations/service.py
@@ -4,9 +4,16 @@ from clickhouse_connect.driver.client import Client
 
 from datetime import datetime
 
-from hexgate_api.core.clickhouse import BatchItem, insert_batch, verify_written_columns
+from hexgate_api.core.clickhouse import (
+    ZERO_RUN_ID,
+    BatchItem,
+    insert_batch,
+    verify_written_columns,
+)
 from hexgate_api.query_scope import scope_filters
 from hexgate_api.schemas import LlmInvocationEvent
+
+LLM_INVOCATION_TABLE = "llm_invocation"
 
 # Order matches schema.sql; received_at absent (server-stamped via column default).
 _LLM_INVOCATION_COLUMNS = [
@@ -23,9 +30,8 @@ _LLM_INVOCATION_COLUMNS = [
     "latency_ms",
     "status",
     "error_code",
+    "run_id",
 ]
-
-LLM_INVOCATION_TABLE = "llm_invocation"
 
 
 def verify_schema(client: Client) -> None:
@@ -66,6 +72,7 @@ def _llm_invocation_row(
         event.latency_ms,
         event.status,
         event.error_code,
+        event.run_id or ZERO_RUN_ID,
     ]
 
 
@@ -190,7 +197,7 @@ def summarize_llm_invocations(
     where_sql = " AND ".join(where)
     summary_sql = (
         f"SELECT {', '.join(_SELECT_COLS)} "
-        f"FROM llm_invocation WHERE {where_sql} GROUP BY {_GROUPING_SETS}"
+        f"FROM {LLM_INVOCATION_TABLE} WHERE {where_sql} GROUP BY {_GROUPING_SETS}"
     )
     result = client.query(summary_sql, parameters=params)
 
--- a/platform/api/hexgate_api/features/llm_invocations/service.py
+++ b/platform/api/hexgate_api/features/llm_invocations/service.py
@@ -122,7 +122,7 @@ def insert_llm_invocations_batch(
     """
     insert_batch(
         clickhouse_client,
-        "llm_invocation",
+        LLM_INVOCATION_TABLE,
         _LLM_INVOCATION_COLUMNS,
         _llm_invocation_row,
         items,
--- a/platform/api/hexgate_api/schemas.py
+++ b/platform/api/hexgate_api/schemas.py
@@ -543,6 +543,17 @@ class DecisionEvent(AuditEnvelope):
     # against. Advisory: contextvar-sourced and client-assertable, exactly like
     # ``role`` and ``user_id``.
     attributes: Optional[dict] = None
+    # Run attribution — same advisory + client-assertable tier as user_id /
+    # user_roles. Populated from hexgate/security/enforcer.py's run_ns dict;
+    # an SDK that doesn't yet send it, and a decision made outside a run
+    # scope, omit run_id here and the platform substitutes the zero UUID at
+    # insert time (core.clickhouse.ZERO_RUN_ID) — never NULL.
+    run_id: Optional[UUID] = None
+    run_tool_calls: int = Field(default=0, ge=0)
+    run_llm_calls: int = Field(default=0, ge=0)
+    run_denials: int = Field(default=0, ge=0)
+    run_total_tokens: int = Field(default=0, ge=0)
+    run_elapsed_ms: int = Field(default=0, ge=0)
 
 
 class LlmInvocationEvent(AuditEnvelope):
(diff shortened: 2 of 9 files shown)


platform/api/hexgate_api/schemas.py
```diff
@@ -543,6 +543,17 @@ class DecisionEvent(AuditEnvelope):
     # against. Advisory: contextvar-sourced and client-assertable, exactly like
     # ``role`` and ``user_id``.
     attributes: Optional[dict] = None
+    # Run attribution — same advisory + client-assertable tier as user_id /
+    # user_roles. Populated from hexgate/security/enforcer.py's run_ns dict;
+    # an SDK that doesn't yet send it, and a decision made outside a run
+    # scope, omit run_id here and the platform substitutes the zero UUID at
+    # insert time (core.clickhouse.ZERO_RUN_ID) — never NULL.
+    run_id: Optional[UUID] = None
+    run_tool_calls: int = Field(default=0, ge=0)
```
quanghexa94: These five `run_*` counters have only `ge=0` — **no `le=UINT32_MAX` upper bound**, unlike the sibling `UInt32` fields just below (`input_tokens`/`output_tokens`/`latency_ms`, lines 563–569) whose comment documents exactly why the cap is mandatory.

A caller — or the future SDK / span-enricher — sending `run_elapsed_ms` for a run older than ~49.7 days (`elapsed_seconds*1000 > 2**32-1`), or a runaway `run_total_tokens > 4.29e9`, passes Pydantic (only `ge=0` is checked) and then fails the ClickHouse `UInt32` write at serialization (`struct 'I' requires 0<=n<=4294967295`). On the single-row path that's a permanent insert error (422, or an unhandled 500 if the `struct.error`/`OverflowError` isn't wrapped as a `ClickHouseError` the router catches); on the batch/enricher path it's a permanent error the job retries forever — precisely what the `le=UINT32_MAX` caps on the sibling fields were added to prevent.

Suggest `Field(default=0, ge=0, le=UINT32_MAX)` on all five, matching those fields.

platform/api/hexgate_api/features/llm_invocations/service.py
```diff
@@ -66,6 +72,7 @@ def _llm_invocation_row(
         event.latency_ms,
         event.status,
         event.error_code,
+        event.run_id or ZERO_RUN_ID,
```
quanghexa94, line 75: Minor: this `event.run_id or ZERO_RUN_ID` None→zero-UUID substitution is duplicated in the two row builders (here and `audit/service.py:152`). The "wire None means unattributed" contract now lives in two places, so a future change to the sentinel or the rule (e.g. distinguishing "outside a run" from "SDK too old") has to be made in both, and a miss silently diverges the two event streams. A one-line helper next to `ZERO_RUN_ID` in `core.clickhouse` (e.g. `run_id_or_zero(event.run_id)`) would keep the rule in one place. Not blocking.

quanghexa94 (commented):
Two notes — the schema one is worth fixing before merge, the other is an optional cleanup. Nice clean additive change otherwise; the deploy-ordering framing and the ProgrammingError→503 fix both look right.

guillaume-hexamind:
@quanghexa94 thanks for the feedback, ready for another review

winlp4ever (approved):

quanghexa94 (approved):

PR #155, opened 2026-08-31
feat(sdk): run facts on the audit payload, and policy-level constraints

guillaume-hexamind:
PR A–C made `run.*` a live, enforced policy namespace and gave ClickHouse the columns to store it. This PR makes the SDK actually *send* those columns, and makes a run-wide cap writable once instead of once per tool.

## Commits

**1. `feat(sdk, platform-api): attach run facts to the decision audit payload`**

`run_id` + five counters now reach the audit wire. New `RunAttribution` value object on `Decision` owns the whole namespace→wire mapping: the `run.*` key names, the seconds→milliseconds conversion, and the `"" → None` rule. Stamped at both `Decision` construction sites — `PolicyEnforcer.decide` (reusing the snapshot the verdict was evaluated against, so the record and the decision cannot disagree) and `_halt_to_decision` for guard halts. `LlmUsageEvent` carries `run_id` too, so `llm_invocation` rows join to the same run.

Deliberately **not** added to `as_error_payload` / `as_error_message`: the model may learn *that* a constraint tripped, never how close it is to its budget. Pinned by a negative test.

**2. `feat(sdk): policy-level constraints applied to every tool`**

New top-level `constraints:` on `AgentPolicy`, applied to every tool the role can reach, on both engines. `default_policy.constraints` only covers tools that fall *through* to the default, so a run cap on a role listing ten tools had to be repeated ten times.

- Unions across `inherits` instead of overriding — the only field here that does. Override would let `inherits: [read_only]` silently drop the mixin's cap, which is fail-open on a security restriction. Deduplicated, first-seen order.
- Can only narrow: `mode: deny` short-circuits before constraints on both engines, so nothing here resurrects a denied tool or an unlisted reach key.
- Both load-time validators (`consts.*` and `run.*`) reach the new list, so a typo is still rejected at `PolicySet` construction rather than denying silently at runtime.

**3. `fix(cli): localize a grammar error in a policy-level constraint`**

`hexgate policy check` reports it under `<policy>` instead of a raw pydantic `ValidationError`.

**4–5. Docs + comment trim.** `constraints.mdx` gains a "where a constraint can live" section (the three locations, the inheritance union, the module-composition restriction) and a note on what a denied model is told.

## Three things the plan missed, found while grounding

1. **`run_id: ""` on the wire would have dropped whole audit rows.** The platform types it `UUID | None`; pydantic won't coerce `""`, FastAPI returns 422, and the sender discards any 4xx without retrying. Every decision made outside a run scope would have lost its entire record — not just its attribution. The `"" → None` conversion lives in one place and has a cross-package test on both sides of the wire.
2. **Guard halts are a second `Decision` construction site.** Stamping only in `decide()` would leave a halt inside a live run writing the zero UUID — losing exactly the row that explains why the run stopped.
3. **`_gated_rules` has three call sites, not two.** The third is the synthetic `agent.run` admission rule, whose policy comes from a fallback constant rather than the document. Missing it would allow on WASM what pydantic denies. The new parameter is required and keyword-only so a fourth call site can't be added silently; a parity case covers `agent.run` specifically.

## Testing

`make check-all` green — 2307 SDK + 530 platform tests. `opa` was on PATH, so the cross-engine parity suite ran for real rather than skipping.

Commit 1 was verified releasable on its own (commits 2–5 stashed, both suites green against that tree alone). The golden Rego snapshot is unchanged, so no existing project's `source_hash` moves on upgrade.

## Deploy ordering

Needs PR C **deployed** (migration + API), not just merged. Against an un-migrated ClickHouse ingest returns 503 and the sender retries; against a migrated DB with an old API the new fields are silently ignored — lossy, not fatal.

### DIFF
--- a/hexgate/security/decision.py
+++ b/hexgate/security/decision.py
@@ -200,6 +200,23 @@ _ERROR_TYPE_BY_OUTCOME: dict[DecisionOutcome, str] = {
 # ``run.elapsed_seconds`` is a float; the platform column is UInt32 ms.
 _MILLISECONDS_PER_SECOND = 1000
 
+# The range the platform's ``DecisionEvent`` validates every run counter against
+# (``ge=0, le=UINT32_MAX``, backing a UInt32 ClickHouse column). Duplicated
+# rather than imported: the SDK does not depend on the platform package.
+_COUNTER_MIN = 0
+_COUNTER_MAX = 2**32 - 1
+
+
+def _as_counter(value: Any) -> int:
+    """Coerce one ``run.*`` value into the platform's UInt32 counter range.
+
+    Clamped, not passed through: counters are monotone within a run, so an
+    out-of-range value does not cost one row — the validator rejects every
+    later decision for that run too, and the DLQ takes the rest of the trail
+    with it. A clamped counter misreports; a rejected span loses the record.
+    """
+    return min(max(int(value), _COUNTER_MIN), _COUNTER_MAX)
+
 
 @dataclass(frozen=True, slots=True)
 class RunAttribution:
@@ -220,17 +237,18 @@ class RunAttribution:
     def from_namespace(cls, run: Mapping[str, Any] | None) -> "RunAttribution":
         """Project the ``run.*`` namespace onto the five persisted counters.
 
-        A subset: the other paths have no column yet.
+        A subset: the other paths have no column yet. Every counter is clamped
+        to the wire's range (see :func:`_as_counter`).
         """
         if not run:
             return DETACHED_RUN
         return cls(
             run_id=str(run.get("id", "")),
-            tool_calls=int(run.get("tool_calls", 0)),
-            llm_calls=int(run.get("llm_calls", 0)),
-            denials=int(run.get("denials", 0)),
-            total_tokens=int(run.get("total_tokens", 0)),
-            elapsed_ms=int(
+            tool_calls=_as_counter(run.get("tool_calls", 0)),
+            llm_calls=_as_counter(run.get("llm_calls", 0)),
+            denials=_as_counter(run.get("denials", 0)),
+            total_tokens=_as_counter(run.get("total_tokens", 0)),
+            elapsed_ms=_as_counter(
                 float(run.get("elapsed_seconds", 0.0)) * _MILLISECONDS_PER_SECOND
             ),
         )
--- a/hexgate/security/decision.py
+++ b/hexgate/security/decision.py
@@ -10,7 +10,9 @@ from __future__ import annotations
 from collections.abc import Callable, Mapping, Sequence
 from dataclasses import dataclass
 from enum import Enum
-from typing import Any, Protocol, runtime_checkable
+from typing import Any, Final, Protocol, runtime_checkable
+
+from hexgate.tracing import semconv
 
 
 class DecisionOutcome(str, Enum):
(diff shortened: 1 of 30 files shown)


hexgate/security/policy.py
```diff
@@ -99,7 +99,9 @@ def evaluate_tool_call(
 
     try:
         check_constraints(
-            tool_policy.constraints,
+            # Policy-level first, so a run-budget denial is the one reported.
+            # Order is all it affects — every constraint must pass either way.
+            [*policy.constraints, *tool_policy.constraints],
```
quanghexa94, line 104: **Policy-level constraints are applied to the synthetic `agent.run` admission key too, so an `args.*` constraint here can lock the agent out at ingress.** If a role has an `admission:` block (putting `agent.run` in `effective_tools`) plus a top-level `constraints: [args.amount <= 500]` meant for tools, the admission check evaluates `args.amount` against `agent.run`, which carries no arguments → missing ref → fails closed → `agent.run` denied → the agent is refused before it starts. Same on the rego side (`rego.py:294` applies `policy_constraints` to every effective tool). The "Where a constraint can live" docs say "every tool the role can reach" but don't mention `agent.run`/admission. Worth either excluding `agent.run` from policy-level constraints, or calling this interaction out loudly in the docs.

hexgate/security/models.py
```diff
@@ -155,11 +163,21 @@ class AgentPolicy(BaseModel):
     inherits: list[str] = Field(default_factory=list)
     is_mixin: bool = False
     default_policy: BaseToolPolicy = Field(default_factory=BaseToolPolicy)
+    # Applied to *every* tool this role can reach, before the tool's own.
+    # Unlike ``default_policy.constraints``, which only reaches tools that fall
+    # through to the default. Can only narrow: ``mode: deny`` short-circuits
+    # before constraints on both engines.
+    constraints: list[str] = Field(default_factory=list)
```
quanghexa94, line 179: **This new field silently activates a previously-ignored top-level `constraints:` key on upgrade.** `AgentPolicy` uses the default `extra="ignore"`, so before this PR a stray/leftover top-level `constraints:` in a policy YAML was silently dropped. After upgrade the same key is parsed and enforced against every tool the role can reach — newly denying calls (and moving the compiled `source_hash` for that policy) with no migration guard or warning. Probably fine given how new the surface is, but worth a conscious call — at minimum a changelog note, or a one-time warning when a top-level `constraints:` first takes effect.

hexgate/security/decision.py
```diff
@@ -195,6 +197,66 @@ def _bounded_reason(header: str, clauses: Sequence[str]) -> str:
...
+        """
+        if not run:
+            return DETACHED_RUN
+        return cls(
+            run_id=str(run.get("id", "")),
+            tool_calls=int(run.get("tool_calls", 0)),
+            llm_calls=int(run.get("llm_calls", 0)),
+            denials=int(run.get("denials", 0)),
+            total_tokens=int(run.get("total_tokens", 0)),
+            elapsed_ms=int(
```
quanghexa94: **Counters are sent unclamped, so a very long-lived run can DLQ its own later rows instead of clamping.** PR C added `le=UINT32_MAX` validation to the platform's `DecisionEvent`/`LlmInvocationEvent`. Here the SDK sends the counters verbatim: a monotone run whose `total_tokens` exceeds ~4.29e9, or `elapsed_ms` exceeds ~49.7 days, produces `RunAttribution` values above `UINT32_MAX` → platform validator rejects → `SpanRejected` → DLQ, dropping *every subsequent* decision/usage row for that run rather than persisting a clamped value. Since the two halves meet exactly here, clamping to `UINT32_MAX` at construction (with the platform cap as the backstop) keeps the run observable.

hexgate/security/decision.py
```diff
@@ -195,6 +197,66 @@ def _bounded_reason(header: str, clauses: Sequence[str]) -> str:
...
+                float(run.get("elapsed_seconds", 0.0)) * _MILLISECONDS_PER_SECOND
+            ),
+        )
+
+    def as_span_attributes(self) -> dict[str, Any]:
```
quanghexa94, line 264: Minor: `as_span_attributes` always emits the five run counters, even for detached (non-run) decisions, so every no-run span carries `RUN_TOOL_CALLS=0 … RUN_ELAPSED_MS=0`. The platform's `_run_fields` already defaults absent counters to 0, so these carry no signal — pure per-span serialization overhead on the common no-run path. Omitting them when detached (the way `RUN_ID` is already correctly omitted when empty, a few lines down) would keep the non-run span lean.

hexgate/security/policy.py
quanghexa94, line 104: Nit: `[*policy.constraints, *tool_policy.constraints]` allocates a fresh combined list on every `evaluate_tool_call` (the per-invocation hot path), though `policy.constraints` is fixed for the role's lifetime. A role-level precomputed union, or just iterating the two lists in sequence in `check_constraints`, avoids the repeated allocation.

quanghexa94 (commented):
Really solid PR — the RunAttribution single-mapping, the inherits-union-not-override call, and the three grounding notes (""→None, the halt Decision site, the third _gated_rules site) are all well judged. Five notes below; #1 and #2 are the two I'd want a decision on before merge (both fail closed on upgrade), #3 completes the UInt32 clamp from PR C on the SDK side, #4/#5 are cleanups.

guillaume-hexamind:
Thanks for the review — three of the five are addressed on the branch, two I'd like to leave as they are.

## Fixed

**`9e5db01` `fix(sdk, platform-api): clamp run counters to the audit wire's uint32 range`** (comment 3)

Right that the two halves meet exactly here. `RunAttribution.from_namespace` now clamps each counter to `[0, UINT32_MAX]` at projection; the platform validator stays the backstop for a foreign emitter. Four unit tests plus a cross-package test in `test_mapping.py` that drives the real SDK emitter through `map_span` and asserts the span validates rather than raising `SpanRejected`.

One correction to the blast radius: `llm_invocation` rows are unaffected — they carry only `run_id` from the run namespace, and their token/latency fields are per-call, not cumulative. The loss was decision rows only. The `elapsed_ms` arm is also the only one a real run can reach (~49.7 days of wall clock, versus 4.29e9 tokens in a single invocation).

**`62e8458` `docs(sdk): policy-level args.* constraints reach agent.run admission`** (comment 1)

Reproduced exactly as described, on both engines. Documented rather than carved out, because the behaviour is not `agent.run`-specific: a policy-level `args.amount <= 500` equally denies any ordinary tool that takes no `amount`, and fail-closed on a missing ref is the rule everywhere else. Excluding `agent.run` would break "every tool the role can reach" without fixing the general case.

`constraints.mdx` now names `agent.run` in the table, shows the policy that denies its own agent at ingress, and documents the escape hatch — `tool == "agent.run" or args.amount <= 500` (verified on both engines before writing it down). The existing parity case only covered an *unlisted* `agent.run`, i.e. closed-world deny; the declared-admission combination is now pinned too, escape hatch included.

**`17544aa` `docs(sdk): note that top-level constraints activate on upgrade`** (comment 2)

Confirmed — `AgentPolicy` is `extra="ignore"`, so a leftover top-level block was dropped before and is enforced now, moving that policy's `source_hash` with it. Composed module policies are unaffected: `extra="forbid"` at every scope, so they rejected the key then and now. No changelog in the repo, so it landed as an `### Upgrading` section under "Where a constraint can live".

## Left as is

**Counters always emitted on detached spans** (comment 4) — accurate, and the omission would be safe (`run_id == ""` implies detached implies all-zero on the production path). But it is a deliberate, pinned choice: `test_event.py` asserts `RUN_TOOL_CALLS == 0` on a detached decision, and `test_span_attribute_key_set_is_the_wire_contract` treats the key set as fixed. A uniform key set seems worth more than five integer attributes on a span that already carries the argument and attribute blobs and up to 4096 chars of reason. Easy to flip if you'd rather have the lean non-run path.

**Per-call list allocation** (comment 5) — real, but not where the cost sits: `check_constraints` already builds a fresh context dict per call with `dict(args)` / `dict(ctx)` / `dict(run)` copies beside it, and `parse_constraint` is `@lru_cache(maxsize=2048)`, so parsing is a dict lookup rather than a re-parse. A precomputed union also can't be a plain field on the frozen `AgentPolicy` — it would need another `cached_property` duplicating state `_resolve_inheritance` already owns. Happy to take it with a profile behind it.

---

`make check-all` green on the branch — 2459 SDK + 638 platform, with `opa` on PATH so the parity suite ran for real.

Unrelated nit spotted while in there: the PR description credits commit 3 to `hexgate policy check`, but the `<policy>` localization is in `hexgate policy validate` — `check` lints module bundles under `policies/`.

hexgate/security/policy.py
quanghexa94: Agreed the docs approach is the right one — it isn't `agent.run`-specific, it's the general fail-closed-on-missing-ref rule, and a carve-out would be inconsistent without fixing the general case. Naming it in the table with the worked example and the prefix escape hatch makes it discoverable, which is what it needed. Good.

hexgate/security/models.py
quanghexa94: The `### Upgrading` section covers it, and nice that composed module policies are safe via `extra="forbid"` at every scope. Works for me.

hexgate/security/decision.py
quanghexa94: The clamp plus that docstring are exactly right — clamp-not-reject because one out-of-range value would take the whole run's trail down with it. And thanks for the blast-radius correction: agreed it's decision rows only, since `llm_invocation` carries just `run_id`. 👍

hexgate/security/decision.py
quanghexa94: Fair — a uniform key set as the pinned wire contract is worth more than saving five ints on a span that already carries the arg/attribute blobs and the reason string. Happy to leave it as-is.

hexgate/security/policy.py
quanghexa94: Agreed, this is dwarfed by the per-call `dict(args)/dict(ctx)/dict(run)` copies already in `check_constraints`, and parsing is cached — not worth complicating the frozen model for. Leave it; easy to revisit with a profile if it ever shows up.

quanghexa94 (approved):
Re-checked against 3bf8103 — all good.

- Counter clamp (#3) fixed cleanly, with the platform validator kept as the backstop.
- The `agent.run` admission interaction (#1) and the on-upgrade activation of a top-level `constraints:` (#2) are both documented now, with a worked example and escape hatch — the right call for behavior that's consistent with the general fail-closed rule.
- The two I raised as cleanups (#4 detached counters, #5 per-call allocation) are fine left as they are; both are pinned/justified.

One tiny thing for the description whenever you touch it: commit 3's localization is in `hexgate policy validate`, not `policy check`. Not blocking.

Approving — nice close to the run-facts series.

PR #156, opened 2026-09-01
docs(sdk): marimo notebook demoing guards + plugins across frameworks

quanghexa94:
A runnable **marimo** notebook that showcases the guard system end-to-end: `examples/marimo_guards_demo.py`.

## What it shows
1. **Writing a guard** — the heart of it: a **global** before-guard vs a **tool-specific** one (`tool_names=`), a before-guard that **rewrites args**, an **after**-guard that withholds a leaky result, and an **observe** watcher — each called directly on a `ToolCall` so you see exactly what a guard receives and returns.
2. **The official plugins** — `secret_guard` / `secret_redactor` / `secret_watch` from `hexgate.plugins`.
3. **The same guards across every framework** — one pipeline (`[secret_redactor, block_secrets]`) attached at each adapter's tool layer — **native Hexgate / LangChain**, **OpenAI Agents**, **Google ADK**, **Pydantic AI** — redacting a secret before the tool runs. Plus a "wiring into your real agent" section with the `guards=[...]` calls per surface.

## Runs offline
No API keys, no LLM calls, no platform — an allow-all policy, a fake tool, and a fake AWS key. The tool layer exercised here is exactly what runs on a live agent call.

```
marimo edit examples/marimo_guards_demo.py
```

Verified end-to-end via `marimo export` (all cells execute, redaction fires through all four framework layers). Pairs with the guards docs page.

### DIFF
--- a/examples/marimo_guards_demo.py
+++ b/examples/marimo_guards_demo.py
@@ -1,0 +1,412 @@
+# /// script
+# requires-python = ">=3.13"
+# dependencies = ["marimo", "hexgate"]
+# ///
+"""Hexgate guards, live — write a guard, then watch it fire across every framework.
+
+Run it:  `marimo edit examples/marimo_guards_demo.py`
+Everything runs offline: no API keys, no LLM calls — guards are exercised directly
+against the tool layer, which is exactly what runs on a real agent call.
+"""
+
+import marimo
+
+__generated_with = "0.9.0"
+app = marimo.App(width="medium")
+
+
+@app.cell
+def _():
+    import marimo as mo
+
+    return (mo,)
+
+
+@app.cell
+def _(mo):
+    mo.md(
+        r"""
+        # Hexgate guards, live
+
+        A **guard** is a small function you attach *before* and *after* a tool call to
+        **observe** it, **rewrite its arguments**, or **refuse** it — the open extension
+        point next to the policy check (`decide`).
+
+        ```
+          the agent wants to call a tool
+                    │
+          1. before-guards   look at the arguments; tweak them, or refuse
+                    │
+          2. decide          allow / deny / needs-approval  (the policy)
+                    │
+          3. run the tool
+                    │
+          4. after-guards    look at the result; refuse to pass it back
+        ```
+
+        This notebook: **write a guard**, use the **official secret plugins**, then
+        watch the same guards fire through **every adapter** — native Hexgate,
+        LangChain, OpenAI Agents, Google ADK, and Pydantic AI.
+        """
+    )
+    return
+
+
+@app.cell
+def _():
+    # Shared setup — the whole guard toolkit, plus a fake secret and an allow-all
+    # policy so the demo runs with no keys and no platform.
+    from hexgate.guards import (
+        Halt,
+        Proceed,
+        ToolCall,
+        ToolOutcome,
+        after_tool,
+        before_tool,
+        build_pipeline,
+    )
+    from hexgate.plugins import (
+        scan_secrets,
+        secret_guard,
+        secret_redactor,
+        secret_watch,
+    )
+    from hexgate.runtime import HexgateContext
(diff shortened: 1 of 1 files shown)


guillaume-hexamind:
Two blocking issues in `examples/marimo_guards_demo.py`. Both verified against the published wheel and by executing the pipeline.

---

### 1 · `:3` — the declared dependencies can't resolve `hexgate.plugins`

```python
# dependencies = ["marimo", "hexgate"]
```

`hexgate` resolves to the PyPI latest, **0.3.0**, which does not ship the `plugins` package:

```
hexgate-0.3.0-py3-none-any.whl
hexgate subpkgs: [adapters, agents, cli, cloud, config, egress, guards,
                  manifest, mcp, runtime, security, streaming, tools, tracing, utils]
has plugins: False    has guards: True
```

`hexgate.plugins` landed in 94a1ba2 (2026-08-24), six days after the `v0.3.0` tag (db3daab, 2026-08-18), and no newer release has been cut — `pyproject.toml:12` is still `version = "0.3.0"`. So `uv run examples/marimo_guards_demo.py` and `marimo edit --sandbox …` both die in the setup cell with `ModuleNotFoundError: No module named 'hexgate.plugins'`, and every downstream cell errors with it.

The documented command at `:7` (`marimo edit …`, no `--sandbox`) is unaffected — it uses the ambient env — but the sandbox is the usual way a shared notebook gets run, and the header is wrong as written.

**Suggested fix** — the header can't be made correct until a release ships, so either drop it and state the requirement:

```python
"""...
Run it:  `uv sync && uv run marimo edit examples/marimo_guards_demo.py`
Requires a local checkout — `hexgate.plugins` is unreleased.
"""
```

or keep it and pin `dependencies = ["marimo", "hexgate>=0.3.1"]`, gating this PR on cutting 0.3.1.

---

### 2 · `:285` — `block_secrets` is dead in the shared pipeline, and the pairing is the documented anti-pattern

```python
pipe = build_pipeline([secret_redactor, block_secrets])
```

`secret_redactor` rewrites the args first, so `block_secrets` always scans clean args and can never halt. Running it:

```
pre order: ['secret_redactor', 'block_secrets']   post: []
result: tool ran with: {'text': 'my key is [REDACTED:aws_access_key]'}
block_secrets halted? False
# same guards, reversed:
{'ok': False, 'error': {'type': 'guard_denied', 'message': 'Refused', ...}}
```

This is structural, not incidental: `build_pipeline` preserves order (`guards/types.py:282`), the pre-loop applies each `Proceed(args=…)` to the working call (`runner.py:399` / `:576`), and `scan_secrets` and `redact_secrets` are backed by the same `_spans` detector over the same JSON walk (`plugins/secrets.py:121`, `:161`) — so the second guard is *guaranteed* to see nothing.

`hexgate/plugins/__init__.py:16-17` warns against exactly this:

> Do not register both on the same tool — the redactor would clean the args before the guard ever sees them.

Since this is the pipeline section 3 presents as the cross-framework showcase, it's the snippet readers will copy — with half of it inert.

Related: `pipe.post` is empty, so no framework cell exercises an after-guard at all. `audit` (`:181`) and `secret_watch` (imported `:72`) are declared and never run, though section 2 advertises `secret_watch` as one of the three plugins.

**Suggested fix** — one guard per tier, which also gives the after-guard tier something to do:

```python
@app.cell
def _(build_pipeline, secret_redactor, secret_watch):
    # Redactor (before) + watcher (after) — one guard per tier, no shadowing.
    pipe = build_pipeline([secret_redactor, secret_watch])
    return (pipe,)
```

and update the prose at `:273` to match. If you want a halt in the showcase, give it its own pipeline (`build_pipeline([secret_guard])`) in a separate cell rather than stacking it behind the redactor.

guillaume-hexamind (commented):
2 claude issues posted above + merge conflicts to solve

Otherwise demo looks good, will approve once updated!

guillaume-hexamind (approved):

PR #157, opened 2026-09-01
build(collector): ship the full otlp pipeline in the deploy stack

victorludvig:
## What is Changing

`docker-compose.deploy.yml` gains the OTLP ingest pipeline — the deploy stack goes from postgres + clickhouse + api to the full set of services, so a self-hosted platform can actually receive the SDK's audit spans:

- **`collector`** — new `platform/collector/Dockerfile` (plain `go build` of the committed ocb output; `bookworm-slim` so the TCP healthcheck has bash — the ocb manifest builds no health extension). The build stage also runs `hexgate-collector validate --config` — the same gate `make collector-check` ends with, and otherwise the only automated path that touches `config.yaml`, since CI has no collector job. Runs as uid 10002: on the shared `/keys` volume it can read the 0644 `hexgate.pub` but never the 0600 private half. Publishes `127.0.0.1:${HEXGATE_OTLP_PORT}:4318`, same loopback contract as the api. The receiver bind addresses in `config.yaml` become `${env:…}` lookups (loopback default keeps `make collector-run` identical). Revocation DSN is keyword/value form with the password via `PGPASSWORD` — pgx reads it itself, so the secret has no URL-safety constraint, matching the api's `DATABASE_URL` assembly.
- **`redpanda`** — pinned image, single listener advertised as the service name (dev's second listener exists only for host-native clients), **no host port**: the broker is PLAINTEXT with no auth, safe only while private. Capped at `--smp 1 --memory 1G`: `--mode dev-container` sets `--reserve-memory 0M`, and an uncapped seastar broker budgets (near) all host RAM — per broker, with two stacks on one box. Healthcheck greps `Healthy: true` rather than trusting `rpk cluster health`'s exit code, which is 0 as soon as the admin API answers. Topics come from a one-shot **`redpanda-init`** service running the same `create-topics.sh` as dev; the script gains `HEXGATE_REDPANDA_ADMIN_API` (because `rpk cluster config set` talks to the admin API on :9644, which is only `localhost` when the script runs inside the broker container — the dev `docker exec` path, unchanged) and a bounded retry around every rpk call, so a transient failure in the one-shot cannot block the services gated on it for good.
- **`api-init`** — one-shot from the api image running the two lifespan side effects other services depend on: `init_db()` (the Postgres tables the enricher's resolver reads) and `keystore.ensure_keypair()` (the `/keys/hexgate.pub` the collector verifies against). `collector`, `enricher` and `api` gate on it with `service_completed_successfully` instead of on api health, so an api-only redeploy (or an api that refuses to boot on a stale ClickHouse schema, deliberately) no longer takes the ingest path down with it. The api lifespan still runs both steps; they are idempotent.
- **`enricher`** — the api image with `command:` overridden to `python -m hexgate_api.jobs.enricher` and the inherited `/health` HEALTHCHECK disabled (it probes uvicorn, which this process doesn't run). Same reason `make enricher-run` runs `postgres-init` first: the tables must exist before the first span is consumed, which `api-init` guarantees.
- **`DEPLOY.md` + `.env.sample`** — one new reverse-proxy route per stage: path `/v1/traces` on the api hostname → `HEXGATE_OTLP_PORT`, matched before the catch-all. That is what will let the OTLP-emitting SDK's default endpoint (`<api url>/v1/traces`) work with zero client config once that SDK ships; the SDK in this tree still posts to `/v1/audit/*`, and the doc says so. Diagram, ports, bridged-proxy guidance and the backup table cover the new services. §4 gains a verify step for the route (a POST to `/v1/traces` gets 401 from the collector; a JSON 404 means the api's SPA catch-all answered and the route is missing or misordered). §6 documents the one-time step an existing box needs for the new required key: `HEXGATE_OTLP_PORT` must be added to the `/hexgate/<stage>` secret and re-pulled with `make platform-env-pull`, because `platform-up` never refreshes an existing `.env.<stage>` and compose aborts every subcommand on a missing `:?` variable. The compose error text points there too.

The collector, Redpanda and the enricher existed only as host-native dev processes (`make collector-run` / `redpanda-topics` / `enricher-run`), so a deployed or self-hosted platform had no ingest path at all — spans had nowhere to land, and standing the pipeline up meant an ocb build plus three hand-run processes. This was the "deploy compose + proxy route" item on the OTel migration rollout checklist; after it, self-hosting is `make platform-up` plus one proxy route.

## Follow-ups (out of scope here, tracked)

Two review findings need code changes outside the deploy files and are recorded as "known gap" comments in the compose:

- #165 — platform-api: the enricher has no health signal and retries forever (needs a heartbeat file + retry ceiling in the job), and `hexgate.pub` is written only on the keystore's first boot (a volume restored from `hexgate.priv` alone crash-loops the collector). DEPLOY.md carries an interim restore warning until then.
- #166 — collector: the TCP healthcheck stays green while the revocation snapshot is stale or Redpanda is unreachable; needs the healthcheckv2 extension plus component status reporting in the biscuit-auth extension.

## Tests

Verified end-to-end on a scratch stack (`-p hexgate-smoke`, fresh volumes, a Postgres password deliberately containing `@ : / #` to prove the `PGPASSWORD` path), before the review-hardening commit:

- First boot converges all-healthy in one shot; `redpanda-init` exits 0 with both topics created and auto-creation disabled.
- A hand-built OTLP/JSON `hexgate.audit` span POSTed to the published port with a freshly minted key returns 200 and lands in `hexgate_audit.policy_decision` with the **auth-derived** `project_id` — a value never present in the span, so the biscuit → client-metadata → Kafka-key → enricher tenancy chain is exercised end to end.
- A garbage bearer gets 401; the DLQ stays empty; no collector/enricher errors.

After the review-hardening commit (`api-init`, redpanda caps/healthcheck, rpk retry, `validate` in the Dockerfile):

- `docker compose config` renders the deploy file with a sample env; `api-init` and the new `depends_on` graph come out as intended; with `HEXGATE_OTLP_PORT` removed, compose aborts with the migration hint.
- `hexgate-collector validate --config=config.yaml` exits 0 on a local `CGO_ENABLED=0` build with none of the `HEXGATE_*` env set — the exact condition of the Dockerfile's build stage.
- `bash -n` on the topics script.

Not covered: the reverse-proxy leg itself (Caddy/nginx path routing exists only as DEPLOY.md prose) — the smoke test hits the published loopback port the proxy would forward to. The hardening commit has not been re-run against a live stack (a fresh-volume `platform-up` on staging is the intended check). No CI job builds these images or runs this stack.


guillaume-hexamind (approved):
Looks good to me.

We'll really need to test well by redeploying staging and make sure the full end to end logic works.

PR #159, opened 2026-09-02
fix(platform-scripts): realign seed_audit with the audit column contract

guillaume-hexamind:
## Why

`platform/scripts/seed_audit.py` was last updated for the `attributes` column (#113) and never followed the multi-role change (#120), which dropped the scalar `role` in favour of `user_roles` + `deciding_role`.

Its hand-written positional rows were **17 values wide against an 18-column contract**, with the old `"default"` role literal sitting in the `outcome` slot — so every field after `tool_name` was shifted by one. `_validate_row_width` aborted before the insert, so the script could not run at all; nothing was ever silently corrupted.

## What changed

- **Rows are dicts keyed by column name**, ordered onto `_SEED_COLUMNS` by `_project_row()`. A column added upstream now surfaces as a named missing key rather than a positional shift or the driver's count-mismatch error, which names neither side. This replaces `_validate_row_width` and removes the bug class that caused the drift.
- **Roles**: stable per-user role sets (`ROLE_SETS`), so the dashboard's `has(user_roles, …)` breakdown tells a consistent story across seeds; `deciding_role` is empty on a deny, matching the schema's "empty when every role denied".
- **Runs and sessions**: decisions are grouped into short runs (`RunProgress`), each carrying its own `session_id` — previously always `''`, so the dashboard's session filter had nothing to bite on. Counters are snapshot-then-record, matching how the enforcer reads the run namespace before the decision lands, and a deny increments `denials` without consuming `tool_calls` (same rule as `RunFacts.record_denial`).
- **Forward-compatible run attribution**: `RunSnapshot.as_fields()` narrows to the run columns the audit contract actually declares. The six `run_*` columns are not on `main` yet (they arrive with the run-attribution work), so this one script serves an API image from either side of that change — writing zero-run attribution today and real run counters once those columns land, with no follow-up edit.
- `seed()` returns `(normal, anomalous)` counts instead of the CLI re-deriving them from `ROWS_PER_USER × users`, which is now only approximate since runs are quantized.

## Verification

Inserted and cleared against a live ClickHouse in both directions.

**Against this branch's contract (18 columns, `run_*` absent):**

```
columns 18 / validate_columns ok / inserted normal=900 anomalous=33
outcome          rows  uniq(run_id)  uniq(session_id)  max(run_denials)
allow            714   1             123               0
deny             130   1              71               0
needs_approval    89   1              66               0
```

`uniq(run_id) = 1` and `run_denials = 0` are the column defaults — correct for a build with no run attribution.

**Against the post-merge contract (24 columns, simulated by extending `_DECISION_COLUMNS`):**

```
columns 24 / validate_columns ok / inserted normal=900 anomalous=33
outcome          rows  uniq(run_id)  max(run_denials)  max(run_tool_calls)  max(run_elapsed_ms)
allow            714   123           3                 11                   302000
deny             130    71           32                 9                   300000
needs_approval    89    66           2                 10                   286000
```

The anomaly spike shows as `run_denials` climbing to 32 inside one run, which is the shape the anomaly detector is meant to catch.

`make lint` and `make fmt-check` clean. Those targets now include `platform/scripts`, which nothing covered before — that gap is why #120 broke this script without CI noticing. `platform-scripts` is also added to CLAUDE.md's commit-scope list, where repo history (#84) already used it.

## Deploy note

No compose file mounts this script — `platform/scripts/` is not in the api image either (the Dockerfile copies only `platform/api`). It is run as a one-off against a deployed stack, with the mount supplied on the command line:

```bash
docker compose -p hexgate-prod --env-file platform/.env.prod \
  -f platform/docker-compose.deploy.yml run --rm --no-deps \
  -v "$PWD/platform/scripts:/app/platform/scripts:ro" \
  api uv run --no-sync python /app/platform/scripts/seed_audit.py --project_id … --agent_name …
```

Because the script imports `_decision_row` and `DecisionEvent` from the image's `hexgate_api`, **the mounted script and the running image must be the same commit.** For an existing ClickHouse volume, `clickhouse/init/` only runs once, so the `attributes` migration (`0001`) must already be applied — `_validate_columns()` refuses the insert and names the gap otherwise.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -22,5 +22,5 @@ tests/                # hexgate package tests (agents, cli, security, tracing, s
 
 ## Rules & Constants
 - **Branches:** `{initials}/{type}/{short_description}` (e.g., `vl/feat/web_search`)
-- **Commits:** `type(scope): description` (lowercase, imperative, no period). Scopes: `platform-api`, `dashboard`, `sdk`, `cli`, `clickhouse`, `redpanda`, `collector`. Types: `feat`, `fix`, `docs`, `build`, `refactor`, `test`.
+- **Commits:** `type(scope): description` (lowercase, imperative, no period). Scopes: `platform-api`, `platform-scripts`, `dashboard`, `sdk`, `cli`, `clickhouse`, `redpanda`, `collector`. Types: `feat`, `fix`, `docs`, `build`, `refactor`, `test`.
 - **Envs:** All prefixed with `HEXGATE_`. Never commit private keys.
--- a/Makefile
+++ b/Makefile
@@ -74,19 +74,19 @@ coverage-html: ## Coverage with a browsable HTML report under htmlcov/
 
 .PHONY: lint
 lint: ## Static check via ruff
-	$(UV) ruff check hexgate tests
+	$(UV) ruff check hexgate tests platform/scripts
 
 .PHONY: lint-fix
 lint-fix: ## Apply ruff autofixes
-	$(UV) ruff check --fix hexgate tests
+	$(UV) ruff check --fix hexgate tests platform/scripts
 
 .PHONY: fmt
-fmt: ## Format with ruff (SDK + platform/api)
-	$(UV) ruff format hexgate tests platform/api
+fmt: ## Format with ruff (SDK + platform/api + platform/scripts)
+	$(UV) ruff format hexgate tests platform/api platform/scripts
 
 .PHONY: fmt-check
 fmt-check: ## Check formatting without writing changes
-	$(UV) ruff format --check hexgate tests platform/api
+	$(UV) ruff format --check hexgate tests platform/api platform/scripts
 
 .PHONY: check
 check: lint fmt-check test ## Python CI parity: lint + fmt-check + test (no coverage overhead)
--- a/platform/scripts/seed_audit.py
+++ b/platform/scripts/seed_audit.py
@@ -7,18 +7,16 @@
 from __future__ import annotations
 
 import json
+import logging
 import random
+import sys
+from dataclasses import dataclass
 from datetime import datetime, timedelta, timezone
-from uuid import uuid4
+from pathlib import Path
+from typing import Any, Iterator, Tuple
+from uuid import UUID, uuid4
 
 from clickhouse_connect.driver.client import Client
-
-import sys
-from pathlib import Path
-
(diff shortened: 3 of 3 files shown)


victorludvig:
One thing to double check, the number of medium anomalies has risen a lot, which may be related to the new sampling algo:
Screenshot 2026-09-03 at 11 40 32

To resolve it, I suggest increasing the values of these two variables:
https://github.com/HexamindOrganisation/hexgate/blob/318a907c2a3db14eb6f507b7ab15669443d0af32/platform/scripts/seed_audit.py#L105-L106

Other Claude issues to check
- The run attribution machinery writes nothing. The six run columns do not exist in _DECISION_COLUMNS, the schema, or migrations on this branch or main. They live only on the unmerged origin/gp/feat/run_id_audit_columns branch. The filter in as_fields silently returns an empty dict for every row, while the docstrings say the columns "landed". Either rebase onto that branch or drop the run code until it merges.
- needs_approval counts as a tool call. The RunProgress counter increments tool calls for every non-deny outcome, contradicting the RunFacts semantics it claims to mirror. Masked today because the columns are dropped, wrong the moment they land.
- Timestamps can land in the future. A run starting minutes before now walks forward up to about 8 minutes past it. With 6 or more users the fixed seed produces rows with a future occurred_at. Clamp the start time or walk backward.
- Tail runs violate the minimum. The remainder clamp produces 1- or 2-decision runs for about a quarter of users. Fold small remainders into the previous run. The header comment's "~300 rows per user" is also wrong, the loop yields exactly 300.
- Rows are hand-built instead of reusing the service. The seed could construct a DecisionEvent and call the API's own row builder, then splice in received_at. That removes the drift class that caused #120 by construction.
- Script is neither linted nor tested. No ruff target or test covers platform/scripts, which is why #120, #128 and #131 broke it unnoticed. The PR body's "bind-mounted into the api container" note describes a setup that does not exist in any compose file.
- Duplicate row builders. The normal and anomaly builders differ in four values and already show minor drift in how roles are copied.
- Commit scope platform-scripts is not in CLAUDE.md's list. Nearest sanctioned scope is platform-api.
- Batch insert uses single-row settings. The repo's own convention for multi-row inserts is BATCH_INSERT_SETTINGS. Minor, pre-existing line.

victorludvig (approved):
LGTM
Image

PR #168, opened 2026-09-04
build(platform-scripts): add the otlp pipeline smoke test as a post-deploy check

victorludvig:
## What is Changing

- `platform/scripts/otlp_smoke.py`: a smoke test of a live stage's OTLP pipeline, from the SDK's sender onward. It sends one of every event type the SDK can emit — `allow` / `deny` / `needs_approval` decisions, one LLM usage event, one ban enforcement — through the SDK's sender (`audit.configure()` → `emit()` → `shutdown()`), then logs into the dashboard API and polls the read endpoints (`/audit/decisions`, `/audit/ban-enforcements`, `/llm/summary`) until every row is visible. Prints a per-event `ok` / `MISSING` report and exits non-zero on failure. Without dashboard credentials it prints the ClickHouse queries to run on the box instead.
- `make platform-smoke STAGE=prod|staging`: maps the stage to its origin (override with `HEXGATE_API_URL`) and runs the script. Needs `HEXGATE_API_KEY` minted on that stage, plus `HEXGATE_SMOKE_EMAIL` / `HEXGATE_SMOKE_PASSWORD` for the read-back.
- `platform/DEPLOY.md`: §4 Verify now ends with the smoke test as the standard post-`platform-up` step; §3's "the SDK in this tree still posts to `/v1/audit/*`" paragraph is replaced — that stopped being true when #146 merged.

**Scope.** The script starts at the SDK's sender, so a PASS proves the deployment: proxy route, collector auth and project keying, Redpanda, enricher, ClickHouse, and the dashboard read API, for this SDK version. It runs no agent. The `PolicyEnforcer`, the adapter hooks that produce usage and ban events, identity from the `HexgateContext` scope, and policy evaluation are upstream of where it starts and remain covered by `pytest -m integration` against a local stack. It is a pipeline smoke test, not an agent-level end-to-end test.

#157 shipped the collector / Redpanda / enricher stack and #146 switched the SDK to emit OTLP spans. Deploying both to staging and prod today (2026-09-04) showed the only thing that proves the deployment is a real span going all the way to a dashboard-readable row: `docker ps` is green and `curl /v1/traces` returns 401 even when the enricher is wedged or the nginx route is missing. This was the script used to sign off both stages; it belongs in the repo and in the runbook rather than in a scratch directory.

Two of the five paths (`needs_approval`, ban enforcement) have no integration test coverage; this script is currently the only place they are exercised against the real pipeline. Adding them to `pytest -m integration` is a follow-up. The event fixtures are deliberately not shared between the script and the tests: both build the same SDK dataclasses, so a shape change breaks both at import, and only arbitrary sample values could drift.

## Tests

- `uvx ruff check` / `ruff format --check` clean on the new script.
- `make -n platform-smoke STAGE=staging` / `STAGE=prod` expand to the expected command; `HEXGATE_API_URL` override honoured.
- The identical script (before lint reformatting) ran against **staging and prod** today and returned `PASS` on both, with all five rows visible in the dashboards.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -399,13 +399,21 @@ platform-logs: _require-stage-env ## Tail a deploy stack's logs: make platform-l
 # clickhouse → read API); it runs no agent, so enforcer/adapter behaviour is
 # out of scope — that's `pytest -m integration` against a local stack. Needs
 # HEXGATE_API_KEY minted on that stage, plus HEXGATE_SMOKE_EMAIL/_PASSWORD (a
-# dashboard login there) for the read-back; without those it prints the
-# ClickHouse queries to run on the box instead. Stage → origin below; set
-# HEXGATE_API_URL to target anything else (a local stack, a new hostname).
+# dashboard login there, project admin — the ban read is admin-gated) for the
+# read-back; without those it prints the ClickHouse queries to run on the box
+# instead. Stage → origin below; set HEXGATE_API_URL to target anything else (a
+# local stack, a new hostname). Unlike its siblings this target has no
+# _require-stage-env dep — it needs no env file, just network access — so it
+# guards the stage itself: an unknown STAGE would expand SMOKE_URL_ to empty,
+# and the SDK reads an empty HEXGATE_API_URL as unset and falls back to the
+# prod default, i.e. a typo would smoke-test production.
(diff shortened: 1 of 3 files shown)


guillaume-hexamind (approved):
Good to me, nice to add these kind of check

PR #170, opened 2026-09-04
docs(platform-api): describe the otlp pipeline as the audit transport

victorludvig:
## What is Changing

Docs only. Every page that still described `POST /v1/audit/decisions` as the audit transport now describes the OTLP pipeline that has been live on prod and staging since 2026-09-04.

- `concepts/audit-trail`: rewritten around "one OTel span per event → `/v1/traces`"; covers the three event types and their tables; drops the legacy curl.
- `platform/audit-clickhouse`: "Ingest endpoint" → "How events arrive": a linear Mermaid diagram (SDK → proxy → collector → Redpanda → enricher → ClickHouse), one paragraph per service, local `make` targets, pointer to `platform/DEPLOY.md` and `make platform-smoke`. The legacy HTTP endpoints move into a note marked slated for removal.
- `internals/audit-pipeline`: §1 overview diagram replaced with the full Mermaid graph from the Notion "Hexgate data pipeline" page; §4 rewritten as Collector → Redpanda → span-enricher (auth extension, revocation snapshot, project keying, topics, per-poll contract, DLQ, trust boundary; legacy HTTP as §4.5); §5.2 notes the batch inserts; §8 platform failure table rewritten for the pipeline; §9 item 1 replaced by the pipeline-health gap (the old item described a debug endpoint §7 already says was removed); status line and PR references updated.
- `reference/environment`: adds `HEXGATE_OTLP_ENDPOINT` and `HEXGATE_REDPANDA_BOOTSTRAP_SERVER`.
- `platform/control-plane`: `POST /v1/traces` in the endpoint list, flagged as served by the collector via the proxy rule; new "The full stack" section listing the three extra processes and pointing at the deploy compose and runbook.
- `internals/development.md`: new "Run the platform locally" section — the five processes (`platform-api-pg`, `collector-run`, `enricher-run`, `dashboard`, `serve`), why the API must be on Postgres, the `HEXGATE_OTLP_ENDPOINT` override needed without a proxy, and how to verify. The targets table lists the collector and enricher.
- `make demo-platform`: the printed recipe now matches (it said `make platform-api` + two terminals; an agent run that way 404s its audit spans). `.claude/skills/integration-tests` points at the docs section as the source of truth.
- `two-paths`: audit-destination cell updated. `docs/README.md`: index line updated.

#146 switched the SDK to OTLP and #157 shipped the collector/Redpanda/enricher stack; both are deployed. Three public pages still told readers audit events are POSTed to `/v1/audit/decisions`, `HEXGATE_OTLP_ENDPOINT` was undocumented, and nothing on the docs site said a self-hosted platform now needs three more services or a `/v1/traces` proxy rule. The internals spec had its SDK half updated in #146 but its platform half still described the HTTP ingest. The only written recipe for running the full local stack was a Claude skill file; the developer-facing one (`make demo-platform`, `development.md`) predated the pipeline entirely.

## Tests

- Rendered locally with `pnpm exec mintlify dev` (Node 22; mintlify refuses Node 25+). The audit-log page diagram, dark theme:
Screenshot 2026-09-04 at 16 27 23

- These are the first Mermaid blocks in the docs. There is no per-PR Mintlify preview for this repo (the check reports "no eligible deployments"); the `.mdx` diagram won't render on GitHub, the `.md` one will.
- `grep -rn "v1/audit/decisions" docs/` now hits only the two legacy notes and nothing else.


guillaume-hexamind:
Two accuracy issues in `docs/internals/audit-pipeline.md`, both about the same thing — dedup semantics. Neither is a regression from this PR, but §4.3's new wording is what brings them into conflict, and this PR is already rewriting the file.

### 1. §5.1's DDL contradicts the §4.3 added here, and the real schema

`docs/internals/audit-pipeline.md:457-460` still reads:

```sql
ENGINE = MergeTree
PARTITION BY toYYYYMM(occurred_at)
ORDER BY (project_id, agent_name, outcome, occurred_at)
TTL toDateTime(occurred_at) + INTERVAL 90 DAY
```

`platform/clickhouse/init/schema.sql` has been this for a while:

```sql
ENGINE = ReplacingMergeTree(received_at)
PARTITION BY toYYYYMM(received_at)
ORDER BY (project_id, agent_name, outcome, occurred_at, event_id)
TTL toDateTime(received_at) + INTERVAL 90 DAY
```

Four differences: the engine, the partition key, the TTL anchor, and `event_id` missing from the sort key (which is what makes dedup possible at all).

The bullet at :465 — "retention/partitioning key off `occurred_at`" — asserts exactly what the schema's own comment rules out:

> Partition + TTL anchor on server-stamped received_at, not the client-supplied occurred_at (clock skew would break retention).

Since :391 now states the tables *are* `ReplacingMergeTree`, the document contradicts itself 66 lines apart. A reader who trusts §5.1 concludes dedup is impossible, and that a client with a skewed clock can land rows in the wrong partition or expire them early.

This block is stale on `main` too, so merging won't fix it.

### 2. §4.3 overstates replay safety

:389-391:

> Committing only after the ack means a crash anywhere in the cycle replays the poll. That is safe for the tables — `event_id` is the idempotency key and the tables are `ReplacingMergeTree` — and merely duplicates DLQ envelopes […]

"That is safe for the tables" reads as unconditional. The `insert_decisions_batch` docstring (`platform/api/hexgate_api/features/audit/service.py`) documents three ways it isn't:

- **The read paths don't use `FINAL`.** `summarize`, `timeseries` and `list_decisions` all query without it (`grep -c FINAL` in that file returns 2, both inside the docstring), so after a replay both copies of each `event_id` are counted until a background merge lands — and merges are "opportunistic, with no upper bound on when they run."
- **Dedup never crosses a partition.** A replay straddling a month boundary double-counts **permanently** — the docstring's word. Accepted upstream as a seconds-wide window at most once a month, but the spec shouldn't imply it can't happen.
- **Duplicates within one batch aren't reliably collapsed** — `optimize_on_insert` works per ~2MB block, so copies straddling a block boundary wait for a background merge like any other.

The practical cost of the current wording: an engineer treats an enricher crash-replay as invisible to the dashboard, then mis-diagnoses inflated decision counts. The code's own qualifier — dedup is *eventual*, not immediate — is the sentence that belongs here.

### Suggested fix

Refresh the §5.1 DDL block and its two bullets from `schema.sql`, and add the eventual-dedup qualifier to §4.3 with a pointer at the docstring for the edges.

guillaume-hexamind (commented):
Branch not up to date with main -> the doc references things that are on main but not on  this branch.

Otherwise no blocking issues, once updated + the claude comments checked will approve

victorludvig:
Both addressed, plus the rebase — `f524ab4`.

**Rebased on `main`.** Was 2 commits behind (#124, #168). #168 was the one that mattered: the branch referenced `make platform-smoke` in five places (`audit-clickhouse.mdx`, `control-plane.mdx`, `development.md` ×2, `audit-pipeline.md` §9) and neither the target nor `platform/scripts/otlp_smoke.py` existed here, so every one of those was a dead reference. Rebase was clean — no docs conflicts.

**1. §5.1 DDL.** Refreshed from `schema.sql`: `ReplacingMergeTree(received_at)`, `PARTITION BY toYYYYMM(received_at)`, `event_id` appended to the sort key, TTL anchored on `received_at`. Kept both schema comments (the dedup one and the clock-skew one) inline in the block so the *why* travels with it, and rewrote the two bullets — the `occurred_at` one now says what anchoring retention on a client clock would actually cost.

Two extras while I was in there: §7 repeated the same pre-`event_id` sort key, so it's corrected too; `grep` for the old engine/partition/sort key across `docs/` is now empty.

**TTL left at 90 days** — #174 owns that number and rewrites this exact block. It'll conflict with this PR either way: #174 also edits the §1 ASCII diagram that this PR replaced with Mermaid. I'll rebase #174 once this lands.

**2. §4.3 replay safety.** Rewritten to say the dedup is *eventual*, with all three edges from the `insert_decisions_batch` docstring — no `FINAL` on the read paths, no dedup across the monthly partition (permanent double-count on a month-boundary replay), within-batch copies collapsing only per insert block — and a pointer at the docstring as the reference.

Unrelated, spotted while checking: that docstring names the decision sort key as `(project_id, occurred_at, event_id)`, which is `ban_enforcement`'s. Left alone — separate fix.

guillaume-hexamind (approved):
Thanks, good to me!

PR #132, opened 2026-08-24
feat(platform-api, dashboard): policy-module editor

quanghexa94:
The in-dashboard **policy-module editor**: browse and edit boundary + capability modules as a file tree, edit `roles.yaml`, watch the composed policy per role resolve live, and test a tool call against it. One combined PR (backend endpoints + frontend), stacked on #121 (retargets to `main` when #121 merges).

## What it does

A three-pane editor at `/policy-modules` ("Modules" in the sidebar). The tree is a rendering of the store: the platform has no files, so the `policy_module` rows are shown as the SDK's on-disk layout: `boundaries/`, `capabilities/` (nestable subfolders), and a synthetic `roles.yaml` at the root.

- **Tree (left):** new / rename / delete modules; **drag-to-move** between folders (same-tier only, since the move endpoint is same-tier); every destructive action gets a sonner Undo toast. A capability rename cascades to the role bindings that imported it, server-side in one transaction.
- **Editor (center):** the existing CodeMirror `PolicyEditor`, reused. Explicit Save (+ Discard). Client YAML parse errors and semantic `/check` lints render in the gutter, anchored to the offending tool's line.
- **Inspector (right):** **Resolved** (per-role tool→mode table, or the composed policy as one **YAML document** — the "compiled" `hexgate policy resolve` output), **Lints**, and a **decision tester** (role + JSON call → tri-state ALLOW / DENY / APPROVAL_REQUIRED verdict with the failing constraint).
- **Live preview (debounced):** typing fires `POST /policy/preview` only after a ~600ms pause, and only when the draft parses client-side, so the Resolved + Lints panes reflect the unsaved edit without a Save. Latest-wins; Save stays explicit.

Three backend endpoints do the work, all composing over the existing SDK (`resolve_for_project` / `analyze_project` / `PolicySet.evaluate`), no policy logic reimplemented:

| Endpoint | What it does |
| --- | --- |
| `POST /policy/preview` | resolve + lint with an optional unsaved-edit overlay, no write. Always 200 (diagnostics-as-data). |
| `POST /policy/test` | evaluate one call against the whole resolved policy for a role (pydantic engine, no opa). 422 if it won't compose, 404 for an unknown role. |
| `PATCH /policy-modules/{tier}/{path}` | rename/move a module; a capability rename cascades to its role bindings. 409 on clash, 404 if absent. |

## Try it

[...]

PR #172, opened 2026-09-07
refactor(dashboard): move workspace switcher into a collapsible sidebar

quanghexa94:
Extracts the app-shell / navigation plumbing that was bundled into the policy-module editor PR (#132) so it can merge on its own, independent of the policy-authoring rewrite that supersedes the rest of that branch.

## What it does
- **AppShell** — restructures the nav into a collapsible sidebar, moves the org/project switcher into it, and adds a user dropdown (settings / orgs / sign-out).
- **OrgProjectSwitcher** — redesigns the trigger into a two-line block (project name primary, org beneath), OpenAI-style, while one dropdown still switches both.
- **lib/ui** — a small persisted zustand store so the sidebar's collapsed state survives a reload.

## Try it
- Collapse/expand the sidebar — the state persists across a reload.
- Switch org/project from the sidebar trigger.

## Important files
- `platform/dashboard/src/components/AppShell.tsx`
- `platform/dashboard/src/components/OrgProjectSwitcher.tsx`
- `platform/dashboard/src/lib/ui.ts` (new)

## Notes
- Lifted from `lhq/feat/policy-editor` (#132), **minus** the `/policy-modules` nav item + route and the `PolicyFolder` id change — those belong to the policy-module work, not this refactor. No policy/tier coupling remains.
- `tsc`, `eslint`, and `prettier --check` all pass.

### DIFF
--- a/platform/dashboard/src/components/AppShell.test.tsx
+++ b/platform/dashboard/src/components/AppShell.test.tsx
@@ -1,0 +1,194 @@
+/**
+ * Tests for the sidebar app-shell.
+ *
+ * Four surfaces worth pinning:
+ *   1. The workspace nav renders its links, and collapsing the sidebar
+ *      hides the labels (icon-only mode).
+ *   2. The collapse toggle works from both the button and Cmd/Ctrl-B.
+ *   3. The account chip shows the signed-in user and signs out.
+ *   4. The bootstrap effect auto-picks a first org + project, and heals
+ *      a stale (removed-org / deleted-project) selection.
+ *
+ * The dropdown/label plumbing inside OrgProjectSwitcher has its own
+ * test — here we only assert the shell's own behaviour.
+ */
+
+import { act, fireEvent, screen, waitFor } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+import { AppShell } from "@/components/AppShell";
+import { useActive } from "@/lib/active";
+import { useUi } from "@/lib/ui";
+import { renderWithProviders } from "@/test/render";
+
+const USER = { id: "user-1", email: "dev@example.com", is_verified: true };
+const ORG_A = {
+  id: "org-a",
+  slug: "org-a",
+  name: "Org Alpha",
+  created_at: "2026-01-01T00:00:00Z",
+  role: "owner" as const,
+};
+const ORG_B = {
+  id: "org-b",
+  slug: "org-b",
+  name: "Org Beta",
+  created_at: "2026-01-02T00:00:00Z",
+  role: "member" as const,
+};
+const PROJ_A1 = { id: "proj-a1", name: "alpha-one", org_id: "org-a" };
+const PROJ_A2 = { id: "proj-a2", name: "alpha-two", org_id: "org-a" };
+
+/** Stub fetch with canned JSON keyed by path; 404 for anything else so
+ * a missed route surfaces as an obvious failure rather than a hang. */
+function stubFetch(routes: Record<string, unknown>): void {
+  vi.spyOn(window, "fetch").mockImplementation(
+    async (input: RequestInfo | URL) => {
+      const url = typeof input === "string" ? input : input.toString();
+      const path = url.split("?")[0];
+      if (path in routes) {
+        return new Response(JSON.stringify(routes[path]), {
+          status: 200,
+          headers: { "Content-Type": "application/json" },
+        });
+      }
+      return new Response("not found", { status: 404 });
+    },
+  );
+}
+
+/** A fully-wired workspace: a signed-in user, two orgs, two projects
+ * under org-a. Individual tests seed the active selection they want. */
+function stubWorkspace(): void {
+  stubFetch({
+    "/v1/users/me": USER,
+    "/v1/orgs": [ORG_A, ORG_B],
+    "/v1/orgs/org-a/projects": [PROJ_A1, PROJ_A2],
+    "/v1/orgs/org-b/projects": [],
+    "/v1/auth/cookie/logout": {},
+    "/v1/auth/google/authorize": {},
+  });
+}
+
+describe("AppShell", () => {
+  beforeEach(() => {
+    useUi.setState({ sidebarCollapsed: false });
+    act(() => {
+      useActive.setState({
+        activeOrgId: ORG_A.id,
+        activeProjectId: PROJ_A1.id,
+      });
+    });
+    stubWorkspace();
+  });
+
+  afterEach(() => {
+    vi.restoreAllMocks();
+    window.localStorage.clear();
+  });
+
+  it("renders the workspace nav links", async () => {
+    renderWithProviders(<AppShell />);
+    await waitFor(() => {
+      expect(screen.getByText("Agents")).toBeInTheDocument();
+    });
+    for (const label of [
+      "Policies",
+      "Graph",
+      "Playground",
+      "Audit",
+      "API keys",
+    ]) {
+      expect(screen.getByText(label)).toBeInTheDocument();
+    }
+  });
+
+  it("collapsing the sidebar hides the nav labels", async () => {
+    renderWithProviders(<AppShell />);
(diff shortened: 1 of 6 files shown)


guillaume-hexamind:
## Code review — 3 findings

Reviewed at high effort and fact-checked each claim against source (CodeMirror/Radix dist included). Test suite passes and `tsc --noEmit` is clean, so all three are behavioral, not build breaks. Scored out of 100.

---

### 1. Global Cmd/Ctrl-B handler has no modifier or target guards — 70/100

`platform/dashboard/src/components/AppShell.tsx:167`

**What's wrong.** The handler matches on `(e.metaKey || e.ctrlKey) && e.key.toLowerCase() === "b"` and calls `preventDefault()` unconditionally, with no check for other modifiers, key repeat, or whether the event came from an editable target.

The concrete collision, verified end to end:

- `@codemirror/commands@6.10.3` dist:1665 binds `{ key: "Ctrl-b", run: cursorCharLeft, ... }` in `emacsStyleKeymap`, and dist:1739 concats it into `standardKeymap` as `{ mac: b.key, run: b.run, shift: b.shift }`.
- `BASIC_SETUP` in `PolicyEditor.tsx:39` disables `foldGutter`, `autocompletion`, `bracketMatching` and `closeBrackets` — but not `defaultKeymap`, so `standardKeymap` is live.
- `@codemirror/view@6.43.1` dist:9129-9145: `runHandlers` stops propagation *only* when the binding sets `stopPropagation`, which the mac-mapped binding does not. CodeMirror calls `preventDefault` (the command returns true), but that never stops propagation.
- `App.tsx:55,61` — `PoliciesPage` is a child route of `AppShell`, so both are mounted together.

**Result:** on macOS, pressing Ctrl-B in the policy YAML editor to move the cursor one char left *also* toggles the sidebar, on every press.

Three smaller issues in the same handler: no `altKey`/`shiftKey` exclusion means Ctrl+Shift+B and Ctrl+Alt+B both match and get `preventDefault()`ed (CodeMirror guards this case itself at view dist:9153 — `!(browser.windows && event.ctrlKey && event.altKey)`, *"Ctrl-Alt may be used for AltGr on Windows"*); no `e.repeat` guard, so holding the combo strobes the sidebar and writes `localStorage` on each repeat; and no editable-target check (see finding 2).

**Suggested fix.** Narrow the match before `preventDefault()`:

```ts
const isEditable = (t: EventTarget | null): boolean =>
  t instanceof HTMLElement &&
  (t.isContentEditable ||
    ["INPUT", "TEXTAREA", "SELECT"].includes(t.tagName) ||
    t.closest(".cm-editor") !== null);

const onKey = (e: KeyboardEvent) => {
  if (e.repeat || e.altKey || e.shiftKey) return;
  if (!(e.metaKey || e.ctrlKey) || e.key.toLowerCase() !== "b") return;
  if (isEditable(e.target)) return;
  e.preventDefault();
  toggleSidebar();
};
```

`AppShell.test.tsx:128` only fires `{ key: "b", metaKey: true }` — worth adding cases for a repeat, an extra modifier, and a keypress originating inside `.cm-editor`.

---

### 2. Collapsing the sidebar unmounts `OrgProjectSwitcher`, destroying its open dialogs — 50/100

`platform/dashboard/src/components/AppShell.tsx:198` · `OrgProjectSwitcher.tsx:42-43,151-155`

**What's wrong.** `OrgProjectSwitcher` owns `createOrgOpen`/`createProjectOpen` as local state and renders `<CreateOrgDialog>` / `<CreateProjectDialog>` in its own fragment, but it is rendered inside `{!sidebarCollapsed && (…)}`. Collapsing the sidebar therefore unmounts the dialogs along with the trigger.

Reachable path: open "New project", type a name, press Cmd/Ctrl-B. There is no `stopPropagation` anywhere in `@radix-ui/react-dialog@1.1.15`, `react-focus-scope@1.1.10` or `react-dismissable-layer@1.1.11`, so the keydown escapes the focus trap and reaches `window`; the switcher unmounts and the dialog vanishes silently. The typed name is react-hook-form state inside `CreateProjectDialog`, so it's gone with no confirmation.

Worth noting this path is *entirely* gated on finding 1 — the collapse button sits behind the modal overlay with `pointer-events: none` on body, so the keyboard shortcut is the only way in. The `isEditable` guard above closes the data-loss case on its own. The structural coupling is still worth fixing, since it will resurface with any other collapse trigger.

Separately, and more minor than it first looked: this PR removes the old header, and while `/orgs` can still change the active org (`Orgs.tsx:38,110`), the switcher is now the only place that *displays* the active org/project and the only place to change the *project*. Since `sidebarCollapsed` persists, a collapsed reload shows no workspace context — recoverable in one click, but worth a thought.

**Suggested fix.** Hoist the dialogs out of the collapsed branch so their lifecycle is independent of the trigger. Lift the two `useState` flags into `AppShell` (or a small `WorkspaceDialogs` component) and render the dialogs unconditionally, passing `open`/`onOpenChange` down to the switcher. A nice bonus would be rendering an icon-only switcher trigger when collapsed instead of dropping it entirely.

---

### 3. `switcherLabel` renders "No project" as the primary line while projects are still loading — 60/100

`platform/dashboard/src/components/OrgProjectSwitcher.tsx:49,166-171`

**What's wrong.** Only `orgsQuery.isLoading` is threaded into `switcherLabel`, but `useProjects` is a separate query. Once orgs have resolved and projects have not, `activeProject` is `null` and the helper returns `"No project"` as the **primary** line.

The deterministic case is an org switch: `setActiveOrg` clears `activeProjectId` by design (`lib/active.ts:43-49`), and `orgsQuery` is cached (`staleTime: 60_000`, so `isLoading` is false), so the primary line reads "No project" for a full network round trip on *every* org switch. `OrgProjectSwitcher.test.tsx:79` already asserts that clearing behavior. Same on a first visit, where the projects query is `enabled: false` until bootstrap picks an org. On a reload with a persisted org it's a race — both queries fire in parallel — so the flash is likely but not guaranteed there. If the projects query errors (403/404 on a stale org), the label reads "No project" permanently while the store still holds a valid project id.

This is a regression from the old `SwitcherLabel`, which kept the org name as the primary text and demoted the missing project to a muted `· no project` suffix — so the dominant label was never wrong.

**Suggested fix.** Thread the projects query's loading state in, mirroring the guard the repo already uses in `useProjectScoped` (`lib/active.ts:99-102`):

```ts
const label = switcherLabel(
  activeOrg,
  activeProject,
  orgsQuery.isLoading || (activeOrgId !== null && projectsQuery.isLoading),
);
```

`switcherLabel` needs no change — `loading` already short-circuits to `{ project: "Loading…", org: "" }`. Optionally keep the org name visible during that window (`{ project: "Loading…", org: activeOrg?.name ?? "" }`) so the line doesn't lose context mid-switch.

---

*Reviewed with [Claude Code](https://claude.com/claude-code)*

guillaume-hexamind (commented):
A couple of minor comments above, logic and code look good overall

quanghexa94:
Thanks @guillaume-hexamind — all three addressed in the latest commit:

1. **Cmd/Ctrl-B guards.** The handler now bails on `e.repeat`, `altKey`/`shiftKey`, and any editable target — an `input`/`textarea`/`select`, `contentEditable`, or anything inside `.cm-editor` — before `preventDefault()`. So the CodeMirror Ctrl-B (move-left) no longer double-fires the sidebar, and Ctrl+Alt-B / Ctrl+Shift-B pass through. Added tests for the editable-target, repeat, and extra-modifier cases.
2. **Dialog lifecycle.** `CreateOrgDialog` / `CreateProjectDialog` and their open-state moved up to `AppShell`, rendered unconditionally; the switcher only fires `onNewOrg` / `onNewProject`. Collapsing the sidebar can no longer unmount an open dialog and drop a half-typed form.
3. **Loading label.** Threaded the projects query's loading state into `switcherLabel` (`orgsQuery.isLoading || (activeOrgId !== null && projectsQuery.isLoading)`), so the primary line reads "Loading…" (keeping the org name for context) instead of flashing "No project" on every org switch. Added a test for that window.

Typecheck + lint clean; AppShell coverage 75% → 93%.

guillaume-hexamind (approved):

PR #174, opened 2026-09-07
fix(platform-api): raise audit log retention to six months

### LINKED ISSUE #173: Raise audit log retention above the AI Act six-month floor
[AI Act Art. 19(1) (providers)](https://ai-act-service-desk.ec.europa.eu/en/ai-act/article-19#019.001) and [Art. 26(6) (deployers)](https://ai-act-service-desk.ec.europa.eu/en/ai-act/article-26#026.006) both require high-risk AI system logs to be kept "for a period appropriate to the intended purpose […] of at least six months"; the three hexgate_audit tables currently TTL at 90 days. Change the TTL in init/schema.sql and recreate the volumes — no migration, since no environment holds data worth preserving. Open question for the PR: fixed 180-day constant, or per-plan configurable retention.

victorludvig:
Closes #173.

## What is Changing

Audit-log retention goes from 90 days to 180 days.

- `platform/clickhouse/init/schema.sql` — the `TTL toDateTime(received_at) + INTERVAL 90 DAY` clause on all three event tables (`policy_decision`, `ban_enforcement`, `llm_invocation`) becomes `INTERVAL 180 DAY`.
- `platform/api/hexgate_api/query_scope.py` — `RETENTION_WINDOW` becomes `timedelta(days=180)`. This is the load-bearing half of the change: the same constant gates the ingest guard (`validate_event_window`, which 400s an `occurred_at` older than retention) and the dashboard query clamp (`prepare_date_range`, which pins `start_date` to `end_date - RETENTION_WINDOW`). Left at 90, the extra ClickHouse retention would exist on disk but be unreachable through the API. The stale comment above `WINDOW_HOURS` calling 90d "the TTL ceiling" is reworded — 90d is now just where the presets stop.
- Tests that hardcoded 90/91 days are re-expressed against `RETENTION_WINDOW` so they track the constant instead of restating it; names and section comments now read "retention" rather than "90d".
- `docs/concepts/bans.mdx` — blocked-attempt retention stated as 180 days.

No ClickHouse migration ships. `init/schema.sql` runs once on an empty volume; there are no customers and no live data worth preserving, so recreating the volumes is the sanctioned path per that file's own header comment.

The EU AI Act requires logs of high-risk AI systems to be kept "for a period appropriate to the intended purpose […] of at least six months":

- [Art. 19(1)](https://ai-act-service-desk.ec.europa.eu/en/ai-act/article-19#019.001) — providers
- [Art. 26(6)](https://ai-act-service-desk.ec.europa.eu/en/ai-act/article-26#026.006) — deployers

At 90 days Hexgate sat below that floor, so a customer using it as their Art. 19/26 logging mechanism would have been non-compliant through no fault of their own. Six months is the Act's **floor, not a target** — this change clears the minimum, it does not aim at a considered retention policy.

**Deliberately deferred:**

- **No `180d` dashboard preset.** `AuditWindow`, `dashboard-range-picker.tsx`, `bans/constants.ts`, `date-range.ts` and `WINDOW_HOURS` all stay at 24h/7d/30d/90d. The picker's existing "Custom" date-range mode already reaches past 90 days once the clamp is raised, so the new retention is reachable today without a new button.
- **Retention stays a flat constant, not a per-plan setting.** Whether retention should become configurable (and whether the floor should then be enforced as a minimum rather than the value) is a real question, but a separate one.
- **`Settings.tsx`'s "Retention" card is untouched** — it is an unimplemented roadmap placeholder, not live behaviour.
- ~~**`docs/internals/audit-pipeline.md` and `docs/internals/ban-enforcement.md` still say 90 days**, as does `docs/concepts/user-scope.mdx`.~~ Done after all, in `docs(clickhouse): align retention references with the 180-day TTL`. The only remaining `90d` mentions in those files describe the dashboard window presets, which genuinely still stop at 90d.

## Tests

`make check-all` passes green from the repo root — SDK lint/fmt/coverage, `platform-api-check` (618 passed, 8 skipped), dashboard lint/typecheck/fmt/coverage, and `collector-check`.

The two ingest-guard tests that asserted a rejection at 91 days (`test_too_old_occurred_at_rejected`, `test_when_occurred_at_is_too_old_then_400_is_returned`) failed on the first run — 91 days is now inside the window — which is exactly the coupling this PR is about. Both, plus the clamp tests, now derive their out-of-window instant from `RETENTION_WINDOW` itself, so they stay honest if the constant moves again.


guillaume-hexamind (approved):
Good to me

PR #175, opened 2026-09-07
feat(sdk): compose policy.yaml grammar front-end

quanghexa94:
Adds the `policy.yaml` authoring front-end — the first SDK PR of the policy-authoring revamp. It parses the position-wildcard grammar and lowers it onto the **existing** linker fold (which already composes agent keys via #124), so the rego → wasm → signing path is reused unchanged, not reimplemented.

## What it does
- **`hexgate/security/compose/`** — a single-file `policy.yaml` grammar: `agents → roles → (tools | reach | mcp | boundary)`, with **scope-by-depth** wildcards — a block at the top level applies to all agents/roles, in an agent body to that agent's roles, in a role body to that one cell. Reach lowers to `agents:` blocks, `mcp` is readability sugar for tools, `boundary` is a closed-world ceiling.
- Resolves **per `(agent, role)`** through `link_policy_set`, assembling the same `ProjectLinkResult` that `resolve_for_project` returns — which is what lets a per-agent / per-role `boundary` apply only to its own cell.
- **CLI:** `hexgate policy resolve --file policy.yaml [--agent X] [--role Y]` (mutually exclusive with `--dir`; `--agent` now selects an agent column in either layout).
- **Analyzer fix:** the dead / redundant / constraint-erased lint passes now read `effective_tools`, so composed reach/admission grants are linted like ordinary tools — a gap #124 left when it made agent keys compose.

## Try it
```sh
cat > policy.yaml <<'YAML'
boundary:
  tools: { refund_order: { constraint: "args.amount <= 1000" } }
tools:
  refund_order: { mode: allow, constraint: 'args.currency == "USD"' }
YAML
hexgate policy resolve --file policy.yaml
```
The effective `refund_order` shows the boundary ceiling **and** the grant constraint AND-ed together; a tool the ceiling omits is shadowed away.

## Important files
- `hexgate/security/compose/{grammar,parse,lower,resolve}.py` — parse → lower → resolve
- `hexgate/cli/policy/main.py` — `resolve --file` / `--agent`
- `hexgate/security/analyzer.py` — lint passes read `effective_tools`
- `tests/security/compose/test_compose.py` — golden-parity + reach + error paths

## Notes
- **Golden parity:** a compose policy and the equivalent `policies/` tree resolve to byte-identical effective policy — proof the fold is reused, not forked.
- **Closed-world reach:** a reach the boundary doesn't list is denied even if a capability grants it (the confused-deputy guard).
- **Errors are source-named:** bad constraints, reserved `agent.*` tool keys, empty `as:`, `tools`/`mcp` key collisions, and non-mapping documents all surface as a `LinkError` naming the file.
- **Deferred to the next SDK PR** (parsed, but rejected with a clear error for now): the import graph — `import:` / `export:`, position-as-scope, cycle detection, globs. `boundary` is a ceiling only (no floor) and roles are per-named-agent, both per the ratified grammar.
- Builds on #124 (agent keys compose through the linker); no engine / rego / signing change. 1179 SDK security + CLI tests pass.

### DIFF
--- a/hexgate/cli/policy/main.py
+++ b/hexgate/cli/policy/main.py
@@ -33,6 +33,7 @@ from hexgate.security.testing import run_namespace
 from hexgate.security import (
     AgentPolicy,
     DecisionOutcome,
+    DEFAULT_AGENT,
     DEFAULT_ROLE_NAME,
     OpaNotFoundError,
     PolicySetError,
@@ -261,10 +262,28 @@ def add_parser(subparsers: argparse._SubParsersAction) -> None:
             "exactly what the engines will enforce."
         ),
     )
-    p_resolve.add_argument(
+    resolve_src = p_resolve.add_mutually_exclusive_group()
+    resolve_src.add_argument(
         "--dir",
         default=".",
         help="Repo root containing a policies/ tree (default: current dir).",
+    )
+    resolve_src.add_argument(
+        "--file",
+        default=None,
+        metavar="PATH",
+        help=(
+            "Resolve a single-file policy.yaml (the compose grammar) instead of a "
+            "policies/ tree. Mutually exclusive with --dir."
+        ),
+    )
+    p_resolve.add_argument(
+        "--agent",
+        default=DEFAULT_AGENT,
+        help=(
+            f'The executing agent to resolve for (default: the generic "{DEFAULT_AGENT}" '
+            "agent). Selects an agent's column in either layout."
+        ),
     )
     p_resolve.add_argument(
         "--role",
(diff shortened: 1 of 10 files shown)


victorludvig (approved):
Hi @quanghexa94 can you add a link to the design in the description for next PRs? This gives more contexts for reviewing agents. 

A couple of suggestion:
- I think using 'front-end' for this PR is confusing, at least to me. I would rename it to 'feat(sdk): policy.yaml parser and composition layer.'
- This predates this PR, but when there are conflicting boundary and grants the tool is still 'allow', although the tool is not usable. We should fix this in a follow-up with a lint:

| Boundary              | Grant                 | Effective                        |
| --------------------- | --------------------- | -------------------------------- |
| `args.amount <= 1000` | `args.amount >= 10`   | both, AND-ed                     |
| `args.amount >= 10`   | `args.amount <= 1000` | both, AND-ed                     |
| `args.amount <= 5`    | `args.amount >= 10`   | both, AND-ed, mode still `allow` |

Some non-blocking Claude findings:

- **No lint path for `--file`.** `resolve --file` never calls the analyzer, and `check` has no `--file`. The dead-grant, constraint-erased and unknown-tool lints work when invoked directly but are unreachable for the compose grammar — including the `effective_tools` fix this PR adds. Not listed in the Deferred section.

- **Provenance is hardcoded.** `_cap`/`_boundary` emit `source=f"policy.yaml#{name}"` and `lower()` takes no `source`, so `resolve --file deploy/billing.yaml` reports a file that doesn't exist. This is what the signed bundle's `layers` records; once `import:` lands every fragment from every file claims the same origin. Thread `source` through `lower()` → `_cell()` → `_cap()`/`_boundary()`.

- **An agent named `"*"` is parsed then silently discarded.** `RESERVED_NAMES` omits it, so it validates; `lower()` then skips it for every agent, and its grants/roles/boundary are dead with no diagnostic. The CLI compounds this with `note: ... also defines agents ['*'] — pass --agent NAME`, which resolves to the same empty baseline. `"*"` is a valid column key in the tier layout, so migrating authors will write it. Add it to `RESERVED_NAMES`.

- **Non-UTF-8 policy file crashes.** The `--file` load path catches only `OSError`; `UnicodeDecodeError` is a `ValueError`, so it escapes to a traceback. The `--dir` branch catches `(ValueError, OSError)`. Same gap in `_read_and_parse`.

- **Stale `--help`.** The `resolve` subparser description still says it loads modules from `<dir>/policies/`, printed above a usage line offering `--file PATH`. `--role`'s help is
likewise dir-only. `docs/internals/policy-modules.md` and `docs/adr/R-POL-lso document only `--dir`.

### Nits

- `CeilingSpec`'s docstring says a `deny` "subtracts"; it deletes the tool
- Title reads as frontend work. Suggest `feat(sdk): policy.yaml parser and composition layer`.
- `tests/security/compose/` is the only test package without `__init__.py`
- `version:` is accepted unvalidated, so `version: 99` resolves under v1 semantics.

PR #176, opened 2026-09-07
feat(sdk): compose import graph (position=scope imports)

quanghexa94:
Adds `import:` to the compose front-end — PR-2 of the policy-authoring stack, on top of #175. A policy can import a leaf capability fragment at any scope; the resolver splices it onto that scope and the linker unions it like any other capability. Imports **compose, never replace**, and the fold is commutative so order never matters.

**Stacked on #175** — this targets the `lhq/feat/policy-authoring` branch, so the diff here is just the import graph. It retargets to `main` automatically when #175 merges.

## What it does
- **`import:` at any scope** (top / agent body / role body) — *position is scope*: a fragment imported in a role applies to that cell, in an agent body to that agent, at the top level to everything.
- **Forms:** `caps.yaml` (whole-file leaf blocks) and `caps.yaml#refunds` (a named `export:`), resolved **transitively** and **relative to the importing file's directory**, sandboxed to the project tree (no `..` / absolute / symlink escape).
- **Safety:** cycles (including importing the entry file back) → `LinkError`; imports are **grant-only** (a `boundary` in an import is rejected — an imported ceiling would intersect and could silently deny every grant); imported files must be **leaf-only** (no `agents:`/`roles:` yet). Per-agent resolution is isolated — an unrelated agent's broken import can't abort a valid one. Provenance names the real source file.

## Try it
```sh
mkdir demo && cd demo
printf 'export:\n  refunds:\n    tools: { refund_order: { mode: allow } }\n' > caps.yaml
printf 'import: [ caps.yaml#refunds ]\ntools: { read_ticket: { mode: allow } }\n' > policy.yaml
hexgate policy resolve --file policy.yaml   # refund_order + read_ticket resolve
```

## Important files
- `hexgate/security/compose/imports.py` — the resolver (new): loader, path sandbox, cycle/`seen` de-dup, leaf/grant-only checks
- `hexgate/security/compose/{grammar,lower,resolve}.py` — `import:` on every scope; splice into the fold; shared `file_loader`
- `hexgate/cli/policy/main.py` — `resolve --file` resolves imports relative to the file
- `tests/security/compose/test_compose.py` — position=scope, transitive, cycles, sandbox, diamond de-dup, provenance, …

## Notes
- **Deferred to a follow-up:** globs (`dir/*`, needs a listing loader) and structural imports (an import carrying `agents:`/`roles:`).
- Imported ≡ inlined (golden test). 1211 SDK security + CLI tests pass; ruff clean. Three review rounds (7 → 7 → 2 findings), all addressed.

### DIFF
--- a/hexgate/cli/policy/main.py
+++ b/hexgate/cli/policy/main.py
@@ -606,18 +606,26 @@ def _main_resolve(args: argparse.Namespace) -> int:
     # --file (compose grammar) and --dir (policies/ tree) are the two front-ends;
     # both produce the same ProjectLinkResult, so the print path below is shared.
     if args.file is not None:
-        from hexgate.security.compose import parse_entry, resolve_entry
+        from hexgate.security.compose import file_loader, parse_entry, resolve_entry
 
+        file_path = Path(args.file)
         try:
-            text = Path(args.file).read_text(encoding="utf-8")
+            text = file_path.read_text(encoding="utf-8")
         except OSError as exc:
             print(f"load error: {exc}", file=sys.stderr)
             return 1
         # Parse once: resolve_entry reuses the Entry, and the agent hint below reads
         # its declared agents. resolve_entry surfaces every failure as LinkError.
+        # `import:` refs resolve relative to the entry file's directory (sandboxed).
         try:
             entry = parse_entry(text, source=args.file)
-            result = resolve_entry(entry, agent=args.agent, source=args.file)
+            result = resolve_entry(
+                entry,
+                agent=args.agent,
+                source=args.file,
+                loader=file_loader(file_path.parent),
+                entry_path=file_path.name,
+            )
         except LinkError as exc:
             print(f"link error: {exc}", file=sys.stderr)
             return 1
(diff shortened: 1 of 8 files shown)


victorludvig (approved):
Hi @quanghexa94, nice work.

One edge case to check:

1. A policy fragment saved in any non-UTF-8 encoding crashes hexgate policy resolve --file with a traceback

Realistic example — the PR's own "Try it" flow, run by a Windows author. PowerShell 5.1's > redirection writes UTF-16LE by default, so echo 'tools: { refund_order: { mode: allow } }' > caps.yaml produces a UTF-16 fragment. Equally realistic for this codebase: a French-authored caps.yaml with # accès remboursement saved as latin-1. Both give:

$ hexgate policy resolve --file policy.yaml
  File ".../pathlib/_abc.py", line 633, in read_text
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xff in position 0: invalid start byte

instead of link error: caps.yaml: …. UnicodeDecodeError is a ValueError, so file_loader's read_text escapes the except OSError in _load — the same gap Victor flagged on #175 ("UnicodeDecodeError is a ValueError, so it escapes to a traceback"), now reproduced on the new import path.

https://github.com/HexamindOrganisation/hexgate/blob/3f5842e545b20dc01cc1f100f13865a547fc63fe/hexgate/security/compose/imports.py#L145-L151
https://github.com/HexamindOrganisation/hexgate/blob/3f5842e545b20dc01cc1f100f13865a547fc63fe/hexgate/security/compose/resolve.py#L42-L49

PR #179, opened 2026-09-07
feat(platform-api): compose policy-file store + resolve/check/preview API

quanghexa94:
Adds the platform store the compose front-end points at — PR-3 of the policy-authoring stack, on top of #176. A project's compose policy lives as `PolicyFile` rows (a small in-DB filesystem); the API lists/upserts/deletes files, resolves them through the SDK compose loader, and compiles the signed bundle. The existing resolve/check/compile paths are reused unchanged — a project auto-dispatches to compose (when an entry file exists) or the legacy tier/matrix model.

**Stacked on #176** — this targets the `lhq/feat/policy-authoring-imports` branch, so the diff here is just the platform store. It retargets down the stack automatically as #175 and #176 merge.

## What it does
- **File store** — `PolicyFile(project_id, name, content, content_hash)`, one row per `(project, name)`. The entry file is `policy.yaml`; other files are pulled in via `import:` from it. A DB-backed loader feeds the SDK's compose resolver, so a project is effectively a small in-DB filesystem — no tier column, since roles/imports live inside the file content.
- **Endpoints** (all under `/v1/projects/{id}`):
  - `GET /policy-files` — list a project's files.
  - `PUT /policy-files/{name:path}` — upsert one file. Structural validation first (**422**), then a whole-project resolve for **every** declared agent (**409** if the save would break resolution), then — if this is the flip from classic to modular — the bundle must rebuild or the write **rolls back fail-closed** (never leaves `is_modular` true on a stale classic bundle).
  - `DELETE /policy-files/{name:path}` — **404** if absent, **409** if another file still imports it.
  - `POST /policy/preview` — resolve a **draft** (name + content, unsaved) and return the resolved YAML + lints inline, so the editor can preview before saving.
- **Auto-dispatch** — `resolve_auto` / `check_auto` / `resolved_*_auto` pick compose vs. the tier model by whether an entry file is present, so `GET .../policy`, the check endpoint, and the agent bundle compile all work unchanged for both models.

## Try it
```sh
cd platform/api
uv run pytest tests/features/policy_modules/ -q
# 46 pass — file CRUD, cross-file import resolve via the DB loader,
# 422 invalid content, 409 break-on-save + still-imported delete,
# preview (draft resolve + inline error), and the fail-closed modular flip.
```

## Important files
- `platform/api/hexgate_api/features/policy_modules/service.py` — the compose section: file store, `_db_loader`, `compose_resolve`/`_check`/`_preview`, `resolve_auto`/`check_auto`, the fail-closed flip, and `resolved_*_auto` compile dispatch.
- `platform/api/hexgate_api/features/policy_modules/router.py` — the four endpoints + `_rebuild_or_none` (the flip-rollback helper).
- `platform/api/hexgate_api/models.py` — the `PolicyFile` table; `schemas.py` — the file + preview DTOs; `core/ids.py` — the `pfl_` prefix.
- `platform/api/tests/features/policy_modules/test_policy_modules.py` — the compose file-store tests.

## Notes
- The SDK seam is untouched — this reuses `compose.resolve_text(...)` and `build_signed_bundle` exactly as PR-1/PR-2 shipped them; the loader is the only new injection point.
- **Next (PR-4):** the dashboard editor that consumes these endpoints (file tree, live preview, resolved YAML, decision tester, policy graph).
- 629 platform-api tests pass (46 in policy_modules), ruff clean. Three review rounds; the 3rd round's 7 findings are all fixed and regression-tested (fail-closed flip rollback and preview-validates-every-agent both have new tests).

---
**Update — graph + decision-test endpoints.** Added the two read/compute endpoints in the same API family so the dashboard (PR-4) can wire the policy graph and the decision tester:
- `GET /policy/graph` — the resolved project as agent/tool/mcp **nodes** with call / reach / admission **edges** (`?role=` filters to one role). 422 if it can't compose.
- `POST /policy/test` — evaluate one tool call against the resolved policy for a role, with an optional unsaved-draft overlay. 422 if it doesn't compose, 404 for an unknown role.

Both auto-dispatch compose vs tier through the same resolve seam and share one model-agnostic graph builder. +9 tests (638 platform pass). One more `/code-review high` round (6 findings: 4 fixed, 2 documented as by-design).

### DIFF
--- a/platform/api/hexgate_api/core/ids.py
+++ b/platform/api/hexgate_api/core/ids.py
@@ -7,6 +7,7 @@ from hexgate_api.models import (
     AgentVersion,
     ApiKey,
     Ban,
+    PolicyFile,
     PolicyModule,
     RoleBinding,
     Tool,
@@ -21,6 +22,7 @@ _ID_PREFIXES: dict[type, str] = {
     ApiKey: "tok",
     Ban: "ban",
     PolicyModule: "pmd",
+    PolicyFile: "pfl",
     RoleBinding: "rbd",
 }
 
--- a/platform/api/hexgate_api/features/agents/service.py
+++ b/platform/api/hexgate_api/features/agents/service.py
@@ -127,7 +127,7 @@ async def _modular_bundle(
     from hexgate_api.features.policy_modules import service as modules
 
     try:
-        policy_yaml = await modules.resolved_policy_yaml(
+        policy_yaml = await modules.resolved_policy_yaml_auto(
             session, project_id, agent=agent
         )
     except modules.compose_error_types() as exc:
@@ -157,7 +157,9 @@ async def _resolved_yaml_or_none(
     from hexgate_api.features.policy_modules import service as modules
 
     try:
-        return await modules.resolved_yaml_by_agent(session, project_id, agent_names)
+        return await modules.resolved_yaml_by_agent_auto(
+            session, project_id, agent_names
+        )
     except modules.compose_error_types() as exc:
         logger.warning(
             "modular project %s does not resolve; no bundles: %s", project_id, exc
--- a/platform/api/hexgate_api/features/policy_modules/router.py
+++ b/platform/api/hexgate_api/features/policy_modules/router.py
@@ -10,6 +10,7 @@ docs/adr/R-POL-002).
 
 from __future__ import annotations
 
+import json
 import logging
 
 from fastapi import APIRouter, Depends, HTTPException, Response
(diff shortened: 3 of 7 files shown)


victorludvig:
Starting the review, 2 comments, can we drop the 'front-end' wording that I think is confusing, and can you add the design doc link to the PR description please.

victorludvig (approved):
Approved.

nit (no need to fix, nice-to-have): Test naming — the 20 new tests at platform/api/tests/features/policy_modules/test_policy_modules.py:986-1305 don't use test_<fn>_happy_path / test_when_X_then_Y.

PR #180, opened 2026-09-08
feat(dashboard): compose policy editor (file store, live preview, graph, tester)

quanghexa94:
The dashboard editor for the compose policy model — the final PR of the policy-authoring stack, on top of #179. It replaces the classic single-file `/policies` editor with a three-pane compose editor that reads/writes #179's policy-file store and renders the composed policy, lints, a decision tester, and the policy graph.

**Stacked on #179** — targets `lhq/feat/policy-authoring-store`, so the diff here is just the dashboard. It retargets down the stack (#175 → #176 → #179) automatically as they merge. Design context: this is PR-4 of the compose stack — #175 (grammar), #176 (import graph), #179 (platform store + resolve/check/preview + graph/test endpoints), and this dashboard on top.

## What it does
- **Three-pane editor** at `/policies`: a **file tree** (the project's `policy_file` rows as a flat, name-spaced filesystem, entry `policy.yaml` pinned), a **CodeMirror editor** with open tabs + per-file unsaved buffers, and an **inspector** (composed policy per role, lints, decision tester, policy graph).
- **Live preview** — the Resolved + Lints panes reflect the unsaved edit via a debounced `POST /policy/preview`; Save is explicit (`PUT /policy-files/{name}`), and the server's **422 (invalid) / 409 (breaks resolution)** surface **inline** in the editor's problems bar, not just a toast.
- **Decision tester + policy graph** — wired to #179's `POST /policy/test` and `GET /policy/graph`: "would this call be allowed?" probes (with an unsaved-draft overlay) and a force-directed graph of agents, tools, and reach/admission edges with per-edge verdict/constraint/role cards and constraint-bypass flagging.
- **Auto-scoped** — the compose editor shows for every project; classic (non-compose) projects skip the resolve/check round-trips and read as "classic" in the inspector.

## Try it
```sh
cd platform/dashboard
pnpm install && pnpm test        # 275 pass (33 files)
pnpm dev                         # then open a project's Policies tab
```
Add a `policy.yaml`, import a `caps.yaml`, watch the Resolved pane compose live; open the graph and the tester from the inspector.

## Important files
- `src/routes/Policies.tsx` — the three-pane page (replaces the classic editor; keeps the `/policies` route + nav).
- `src/components/policy_files/` — `FileTree`, `EditorPane` (tabs/dirty buffers, inline 422/409), `InspectorTabs` (Resolved/Lints + agent picker), `TestPanel`, `PolicyGraphDialog`.
- `src/lib/policy_files.ts` — React Query hooks over the file store + resolve/check/graph/test + debounced preview.
- `src/lib/file_tree.ts` — the flat-name → nested-tree builder; `src/lib/api.ts` — the compose client methods + wire types.

## Notes
- **Salvaged** the graph dialog + editor theme from the demo branch and rewired them from the old tier/matrix model onto the flat file store; the graph/tester matched #179's endpoint shapes (the endpoints were ported from the same demo).
- Coverage: **89% on the new files** (`vitest run --coverage`); typecheck + lint clean. `@xyflow` (the graph) can't render under jsdom, so its tests mock it — **worth a visual pass in-browser** for the CodeMirror + graph render.
- Three review rounds across the stack; PR-4's own `/code-review high` surfaced 8 findings (incl. two data-integrity bugs — a cross-project write and a delete-resurrection), all fixed with regression tests.
- **Deferred:** folder CRUD / drag-move, and server-side agent enumeration (the agent field is a free-text `*`-default input). The classic editor's now-unused helpers (`lib/policy_graph.ts`, `components/graph/`, `lib/policy.ts`) are left in place for a follow-up cleanup.


victorludvig (commented):
Hi @quanghexa94, nice work. Could you add screenshots / videos in PR description for next frontend PRs?
1 issue to solve then we can merge:

### Cross-project write: a project switch carries the previous project's unsaved buffer, and Save writes it into the new project

`EditorPane` is not keyed by `projectId` ([`Policies.tsx:268`](https://github.com/HexamindOrganisation/hexgate/blob/71860e73271932998c1c00547d9436586820a70c/platform/dashboard/src/routes/Policies.tsx#L266-L277) passes only `name`), and its re-sync only reloads the buffer when `name` changes — on a same-name switch it takes the `else if` branch and deliberately keeps a dirty buffer:

```ts
// EditorPane.tsx:82-91
if (name !== syncKey) { ... setDraft(drafts[name] ?? saved); }
else if (saved !== syncSaved) {
  const wasClean = draft === syncSaved;
  setSyncSaved(saved);
  if (wasClean) setDraft(saved);   // dirty buffer is KEPT
}

Trigger — Edit project A's policy.yaml without saving, then pick project B from the project picker, where B was visited in the last 5 minutes so its ["policy-files", B] query is still cached. filesQuery.isLoading is false, so the page never renders the loading state that would unmount the pane.

Break — active resolves to policy.yaml again (the pinned entry file, so always the same name), name is unchanged, wasClean is false, and A's text is kept. Save then fires PUT /v1/projects/<B>/policy-files/policy.yaml with project A's content, and the server recompiles B's agents (policy_modules/router.py:363-382).

Notice — Nobody. The inspector still shows B's stored resolved policy, because the page-level draft stays null (the onDraftChange effect's deps are all unchanged by the switch). The only signal is an "unsaved" badge, which reads as your own work in progress.

Reproduced against the real components: after the switch the editor held zzAAA_PROJECT_A: 1 and Save issued

PUT /v1/projects/proj-b/policy-files/policy.yaml
{"content":"zzAAA_PROJECT_A: 1\n"}

Policies.tsx:128-131 states this must not happen. Policies.test.tsx:188 passes only because its second project is uncached, so the pane unmounts.

Fix — key={projectId} on <EditorPane>, plus a regression test that switches back to a cached project with a dirty buffer.
```

quanghexa94:
Fixed in bc5883f9 — the cross-project write is gone.

**Root cause** was exactly as you diagnosed: `EditorPane` was keyed only by file name, so a switch to a cached project (no loading state to unmount the pane) kept the previous project's dirty buffer when both entry files are named `policy.yaml`.

**Fix:** `key={projectId}` on `<EditorPane>`, so it remounts on a project switch and discards the stale buffer.

**Regression test:** `Policies.test.tsx` — mocks the CodeMirror editor to a textarea (jsdom can't drive CodeMirror by keystroke), pre-seeds project 2 into the QueryClient so the switch hits no loading state (your cached-project precondition), dirties project 1's `policy.yaml`, switches to project 2, and asserts the editor shows project 2's content — not the dirty project-1 text. Verified it fails without the `key` and passes with it. (`renderWithProviders` now accepts a seeded client.)

309 tests green, fmt/lint/typecheck clean.

Good call on the screenshots too — I'll include screenshots/short clips in the description on the next front-end PR.

victorludvig (approved):
Approved!

PR #185, opened 2026-09-08
fix(redpanda): raise the record-size budget for message spans

victorludvig:
## What is Changing
Every size limit between the SDK and ClickHouse, raised together so a ~272 KiB LLM message span gets through. That 272 KiB is the SDK's own ceiling (256 input + 8 output + 8 system instructions) against a decision span's ~1 KiB — a ceiling, not a typical event, but sizing for the average would hold until the first RAG call and then drop a whole batch.

```mermaid
flowchart LR
  classDef sdkBox fill:#EAF2FF,stroke:#2F6FED,stroke-width:2px,color:#153E90
  classDef hexBox fill:#F3EAFB,stroke:#8E3FC7,stroke-width:2px,color:#4A148C
  classDef redpandaBox fill:#FCE8EC,stroke:#D6336C,stroke-width:2px,color:#8A1538
  classDef extBox fill:#FFF6E0,stroke:#E0A800,stroke-width:2px,color:#7A5900

  SP["one message span<br/>≤ 272 KiB<br/>256 input + 8 output + 8 sysinstr"]:::sdkBox
  S["SDK export<br/>MAX_EXPORT_BATCH_SIZE 64"]:::sdkBox
  P["nginx (host)<br/>client_max_body_size 32 MiB<br/>per-stage, outside this repo"]:::extBox
  R["OTLP/HTTP receiver<br/>max_request_body_size 32 MiB<br/>confighttp default 20 MiB"]:::hexBox
  B["batch processor<br/>send_batch_size 24<br/>send_batch_max_size 24"]:::hexBox
  X["kafka exporter<br/>producer.max_message_bytes 8 MiB (pre-compression)<br/>sending_queue bytes / 128 MiB · zstd"]:::hexBox
  T["topic hexgate.otlp.raw<br/>max.message.bytes 8 MiB (compressed)"]:::redpandaBox
  E["span-enricher<br/>max_partition_fetch_bytes 8 MiB"]:::hexBox
  Q["topic hexgate.otlp.dlq<br/>max.message.bytes 8 MiB"]:::redpandaBox

  SP --> S
  S -->|"64 × 272 KiB ≈ 17 MiB per POST"| P
  P --> R
  R --> B
  B -->|"24 × 272 KiB = 6.375 MiB per record"| X
  X --> T
  T --> E
  E -->|"rejected spans<br/>DLQ max_request_size 8 MiB<br/>(envelopes capped ~85 KiB by dlq.py)"| Q
```

Each hop rejects **whole** — the proxy and receiver drop the entire POST, the exporter and broker the entire record — so the narrowest limit decides what gets through, and it takes the unrelated decision spans batched alongside it down too.

| Hop | Setting | Value |
|---|---|---|
| SDK | `MAX_EXPORT_BATCH_SIZE` (`tracing/_senders.py`) | 64 spans (was OTel's 512) |
| Collector receiver | `http.max_request_body_size` | 32 MiB (was confighttp's 20) |
| Collector batch | `send_batch_size` / `send_batch_max_size` | 24 spans (was 512) |
| Collector exporter | `producer.max_message_bytes` · `compression` | 8 MiB · zstd |
| Collector exporter | `sending_queue` | `sizer: bytes`, 128 MiB |
| Collector | `memory_limiter` | 1024 / 256 MiB |
| Redpanda | `max.message.bytes` on `hexgate.otlp.raw` **and** `.dlq` | 8 MiB |
| Enricher | `max_partition_fetch_bytes` · DLQ `max_request_size` | 8 MiB |
| nginx (host) | `client_max_body_size` on `/v1/traces` — outside this repo, **applied to staging + prod 2026-09-09** | 32 MiB (was 8 MiB) |

Plus docs: `audit-pipeline.md` §4.1/§4.2/§4.3, and `platform/DEPLOY.md` §3 (the nginx limit) and §6 (re-running `redpanda-init` on an upgrade).

At `send_batch_size: 512` into a 1 MiB record, each span gets ~2 KiB. A message span is far past that, so the record is rejected and the 511 decision spans sharing it die with it. Raising one limit alone achieves nothing: a bigger topic just produces records the exporter refuses to send and the enricher can never fetch.

Three of the nine limits are less obvious than the topic and the record size:

- **`send_batch_size` comes down with `send_batch_max_size`.** The batch processor refuses to start unless `send_batch_max_size >= send_batch_size`, so both are 24; `send_batch_max_size` alone will not boot.
- **nginx is a seventh hop, and it was the narrowest.** It enforces `client_max_body_size` on `/v1/traces` *before* the collector authenticates, and both stages sat at 8 MiB — deliberately tighter than the box's `100M` default, since that route is the only bulk-upload endpoint and the limit bounds what an unauthenticated caller can make the proxy buffer. Ample for decision spans, but a ~17 MiB message export would have `413`d at the edge with the collector logging nothing. Raised to `32m` on staging and prod on 2026-09-09 (`401`/`200` verified after reload). It lives outside this repo, so no CI check can catch it drifting.
- **Queue memory, not record size, was the real blast radius.** The exporter's default queue holds 1000 *requests* with no notion of size: ~500 MiB of decision spans, but ~6.2 GiB once one request can be a 6.375 MiB message batch — on a box also running ClickHouse, Postgres and Redpanda for two stages. Now bounded at 128 MiB.

  That last one cuts both ways: the outage buffer for today's decision-only traffic drops from ~512k spans to ~134k, i.e. ~11 minutes of Redpanda downtime instead of ~43, and overflow sheds spans silently. Accepted rather than fixed with a bigger `queue_size`, since 128 MiB is already a fair share of the `memory_limiter` headroom (1024 − 256 MiB) on a two-stage box.

Two details behind the values, checked against upstream source rather than assumed: `producer.max_message_bytes` maps to franz-go's `ProducerBatchMaxBytes` and is measured **before** compression, so zstd buys no headroom against it (only the broker's limit sees the compressed batch); and the DLQ never needed 8 MiB — `dlq.py` envelopes top out at ~85 KiB — it gets it only so the two topics cannot drift.

Two follow-ups from review: #202 (the fixed 5 s shutdown budget now needs 8x more exports to drain the SDK queue) and #204 (export the queue metrics so a shed is visible, and back the queue with `file_storage`).

## Tests
- `make lint fmt-check test` — 2477 passed. New `test_sender_caps_the_export_batch_size` pins the export batch against OTel's default.
- `make platform-api-check` — 639 passed, 8 skipped. The new consumer test patches both aiokafka constructors and asserts the kwargs; the fakes the other lifecycle tests inject would hide their absence entirely.
- `hexgate-collector validate --config=config.yaml` passes, and fails on a deliberately misspelled key, so the Dockerfile gate is meaningful.

**Merged is not enough.** Every stage needs its topics altered and its collector + enricher restarted before message spans get through — `DEPLOY.md` §6 now carries the commands. `otlp_smoke.py` sends five small events in one sub-1-MiB export, so it would pass on a stage where none of that happened; the oversized smoke event that exercises this budget is PR 7.

### DIFF
--- a/docs/internals/audit-pipeline.md
+++ b/docs/internals/audit-pipeline.md
@@ -363,8 +363,8 @@ flowchart LR
   P["nginx (host)<br/>client_max_body_size 32 MiB<br/>per-stage, outside this repo"]:::extBox
   R["OTLP/HTTP receiver<br/>max_request_body_size 32 MiB<br/>confighttp default 20 MiB"]:::hexBox
   B["batch processor<br/>send_batch_size 24<br/>send_batch_max_size 24"]:::hexBox
-  X["kafka exporter<br/>producer.max_message_bytes 8 MiB (pre-compression)<br/>sending_queue bytes / 128 MiB · zstd"]:::hexBox
-  T["topic hexgate.otlp.raw<br/>max.message.bytes 8 MiB (compressed)"]:::redpandaBox
+  X["kafka exporter<br/>producer.max_message_bytes 8 MiB<br/>sending_queue bytes / 128 MiB · no compression"]:::hexBox
+  T["topic hexgate.otlp.raw<br/>max.message.bytes 8 MiB"]:::redpandaBox
   E["span-enricher<br/>max_partition_fetch_bytes 8 MiB"]:::hexBox
   Q["topic hexgate.otlp.dlq<br/>max.message.bytes 8 MiB"]:::redpandaBox
 
@@ -391,7 +391,7 @@ because it is enforced before the collector authenticates the token
 | receiver `http.max_request_body_size` | 32 MiB | confighttp defaults to 20 MiB and rejects the whole POST; the SDK exports ≤ 64 spans (~17 MiB worst case) |
 | `batch.send_batch_size` / `send_batch_max_size` | 24 spans | 24 × 272 KiB = 6.375 MiB, inside the 8 MiB record (down from 512). Decision traffic ships ~20× more records, all small. Both knobs: the processor refuses to start unless `send_batch_max_size >= send_batch_size`, and the max is what splits one oversized incoming request |
 | exporter `producer.max_message_bytes` | 8 MiB | client-side check *before* Redpanda sees the record — raising only the topic changes nothing |
-| exporter `producer.compression` | `zstd` | network and disk only — `max_message_bytes` maps to franz-go's `ProducerBatchMaxBytes`, measured *before* compression, so the hard client ceiling is 8 MiB uncompressed (30 spans). Only the broker's limit sees the compressed batch |
+| exporter `producer.compression` | `none` (configkafka's default) | `max_message_bytes` maps to franz-go's `ProducerBatchMaxBytes`, measured *before* compression, so zstd would buy no headroom — and the enricher's aiokafka has no zstd codec (`cramjam` is absent from `platform/api`'s lock), so a compressed batch would raise out of its poll loop. Enabling compression means shipping the codec and both images together |
 | exporter `sending_queue` | `bytes` / 128 MiB | the default is 1000 *requests*, which was ~500 MiB of decision spans but ~6.2 GiB of message batches |
 | `memory_limiter` | 1024 / 256 MiB | 384 MiB soft would refuse intake against 32 MiB bodies and a 128 MiB queue |
 | nginx `client_max_body_size` on `/v1/traces` | 32 MiB | **outside this repo, set per stage.** Enforced before the collector authenticates, so it is held below the box's `100M` default; was 8 MiB before message logging. See `platform/DEPLOY.md` §3 |
@@ -438,7 +438,8 @@ raw topic only so the two cannot drift apart; it does not need it. `dlq.py`
 quotes an oversized record as a 64 KiB preview (~85 KiB once base64'd), or
 its attributes at 32 KiB — never both in one envelope — so nothing it builds
 approaches even the 1 MiB default. See the record-size budget in §4.1 — this
-is the broker-side half of it, and it is enforced on the *compressed* batch.
+is the broker-side half of it, and it is enforced on the batch as sent —
+which, with the exporter's compression off, is the uncompressed size.
 
 Redpanda is a buffer, not a store: ClickHouse is the system of record, and the
 raw topic only needs to outlive an enricher restart or redeploy. It is
--- a/docs/internals/audit-pipeline.md
+++ b/docs/internals/audit-pipeline.md
@@ -351,6 +351,38 @@ there is no separate "message path", and batching is count-based, so a single
 set of limits has to be safe for the largest span type. A rejected batch drops
 all of its spans, decision spans included, which is why these move together:
 
+```mermaid
+flowchart LR
+  classDef sdkBox fill:#EAF2FF,stroke:#2F6FED,stroke-width:2px,color:#153E90
+  classDef hexBox fill:#F3EAFB,stroke:#8E3FC7,stroke-width:2px,color:#4A148C
+  classDef redpandaBox fill:#FCE8EC,stroke:#D6336C,stroke-width:2px,color:#8A1538
+  classDef extBox fill:#FFF6E0,stroke:#E0A800,stroke-width:2px,color:#7A5900
+
+  SP["one message span<br/>≤ 272 KiB<br/>256 input + 8 output + 8 sysinstr"]:::sdkBox
+  S["SDK export<br/>MAX_EXPORT_BATCH_SIZE 64"]:::sdkBox
+  P["Reverse proxy<br/>client_max_body_size 32 MiB<br/>⚠ outside this repo · nginx defaults to 1 MiB"]:::extBox
+  R["OTLP/HTTP receiver<br/>max_request_body_size 32 MiB<br/>confighttp default 20 MiB"]:::hexBox
+  B["batch processor<br/>send_batch_size 24<br/>send_batch_max_size 24"]:::hexBox
+  X["kafka exporter<br/>producer.max_message_bytes 8 MiB (pre-compression)<br/>sending_queue bytes / 128 MiB · zstd"]:::hexBox
+  T["topic hexgate.otlp.raw<br/>max.message.bytes 8 MiB (compressed)"]:::redpandaBox
+  E["span-enricher<br/>max_partition_fetch_bytes 8 MiB"]:::hexBox
+  Q["topic hexgate.otlp.dlq<br/>max.message.bytes 8 MiB"]:::redpandaBox
+
+  SP --> S
+  S -->|"64 × 272 KiB ≈ 17 MiB per POST"| P
+  P --> R
+  R --> B
+  B -->|"24 × 272 KiB = 6.375 MiB per record"| X
+  X --> T
+  T --> E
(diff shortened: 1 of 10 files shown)


guillaume-hexamind:
## Code review — 2 blocking findings

Both verified against the code in this branch, each with an executable repro. Scores are severity out of 100.

---

### 1. `platform/collector/config.yaml:125` — `compression: zstd` puts the enricher in a silent crash loop — **95/100**

**Description**

Enabling zstd on the kafka exporter makes every `hexgate.otlp.raw` record a zstd record batch, and the enricher's aiokafka client cannot decompress it.

`aiokafka` 0.14's codec gate is literally a `cramjam` import check (`aiokafka/codec.py`):

```python
try:
    import cramjam
except ImportError:
    cramjam = None

def has_zstd() -> bool:
    return cramjam is not None
```

`cramjam` is not a dependency of `platform/api` — `aiokafka>=0.12` is declared with no `[zstd]` extra, and `cramjam` appears 0 times in `platform/api/uv.lock`. The `zstandard` and `lz4` that *are* in that lock come from `clickhouse-connect`, and aiokafka never looks at them.

Verified in the repo venv by feeding a hand-built zstd v2 record batch to the same record class the consumer uses:

```
impl: <class 'aiokafka.record._crecords.default_records.DefaultRecordBatch'>
has_zstd False lz4 False snappy False gzip True
compression_type = 4
RAISED: aiokafka.errors.UnsupportedCodecError: Libraries for zstd compression codec not found
  ... default_records.pyx:377 in __iter__
  ... default_records.pyx:234 in _maybe_uncompress
  ... default_records.pyx:125 in _assert_has_codec
```

Nothing on the path catches it. `fetcher.py:1228` calls `res_or_error.getall(max_records)` bare, `getall` iterates the raising generator, and `getmany` → `fetched_records` adds no handler — so it escapes the poll loop at `jobs/enricher/consumer.py:177`, runs the `finally` (client shutdown), and escapes `run()`. `jobs/enricher/__main__.py` catches only `SchemaOutOfDate` and `TopicsMissing`, so the process dies on an unhandled exception.

**Failure scenario**

Deploy the new collector image. The first record it writes is zstd. The enricher crashes out of `run()`; `restart: unless-stopped` (`docker-compose.deploy.yml:300`) recycles it; it crashes again on the same batch, because no offsets were ever committed. Meanwhile `healthcheck: disable: true` (line 322) means the container still reports **Up** — the exact blind spot that service's own comment already flags for the wedged-consumer case. Nothing reaches ClickHouse and the raw topic's 3-day retention starts deleting spans.

This is not gated on message spans existing or on the topic `max.message.bytes` alter — ordinary decision traffic triggers it the moment the new collector starts producing.

**Suggested fix**

Either add the codec to the consumer side:

```toml
# platform/api/pyproject.toml
"aiokafka[zstd]>=0.12",   # or an explicit "cramjam>=2.8"
```
followed by `uv lock` in `platform/api/`.

Or drop the setting — the config comment right above it already says compression isn't bought for headroom against `max_message_bytes` (franz-go sizes `ProducerBatchMaxBytes` uncompressed), so `compression: none` costs the pipeline nothing but network and disk:

```diff
-      compression: zstd
```

Worth a test either way: assert `aiokafka.codec.has_zstd()` wherever the consumer's client kwargs are already pinned, so the dependency can't silently disappear again.

---

### 3. `hexgate/tracing/_senders.py:118` — hardcoding the batch size while the queue size stays env-derived can raise out of `configure()` — **45/100**

**Description**

`super().__init__(exporter, max_export_batch_size=MAX_EXPORT_BATCH_SIZE)` pins the batch size to 64 but leaves `max_queue_size` unset, and opentelemetry-sdk 1.38 still resolves that from the host application's environment (`sdk/trace/export/__init__.py:148`, `_default_max_queue_size` reading `OTEL_BSP_MAX_QUEUE_SIZE`) and then validates the pair:

```python
if max_export_batch_size > max_queue_size:
    raise ValueError("max_export_batch_size must be less than or equal to max_queue_size.")
```

Verified in the SDK venv:

```
--- pre-PR (no max_export_batch_size kwarg), env queue=32 batch=16 ---
OK  batch= 16 queue= 32
--- this branch (max_export_batch_size=64), same env ---
RAISED: ValueError max_export_batch_size must be less than or equal to max_queue_size.
--- env queue unset, batch=16 ---
OK  batch= 64 queue= 2048
```

So it is a regression introduced here, not a pre-existing one: that env pair worked before this commit. The trigger is narrow — the host app must set `OTEL_BSP_MAX_QUEUE_SIZE` below 64 **and** `OTEL_BSP_MAX_EXPORT_BATCH_SIZE` at or below it (setting the queue alone below 512 already failed before this PR).

**Failure scenario**

A customer whose process sets `OTEL_BSP_MAX_QUEUE_SIZE=32` and `OTEL_BSP_MAX_EXPORT_BATCH_SIZE=16` upgrades the SDK. `AuditSender.__init__` constructs the processor with no `try` on any hop of `configure()` → `get_or_create_sender` → `AuditSender.__init__`, so the `ValueError` escapes `configure()` and kills their process at startup. Same for the other two entry points, `tracing/usage.py:131` and `security/bans.py:219`.

This also contradicts the file's own stated design eight lines up: `_SPAN_LIMITS` pins every limit to `UNSET` specifically so customer `OTEL_*` vars can't reach the audit channel, and that comment names this exact hazard — *"a malformed value would raise out of `configure()`"*.

**Suggested fix**

Pin the queue size the same way the batch size is pinned — the new docstring already assumes 2048:

```diff
+MAX_QUEUE_SIZE = 2048
+"""Spans buffered before the processor drops the oldest. OTel's own default,
+pinned here for the same reason as MAX_EXPORT_BATCH_SIZE: a host application's
+OTEL_BSP_MAX_QUEUE_SIZE must not reach the audit channel — and a value below
+MAX_EXPORT_BATCH_SIZE would raise ValueError out of configure()."""
+
 ...
-        super().__init__(exporter, max_export_batch_size=MAX_EXPORT_BATCH_SIZE)
+        super().__init__(
+            exporter,
+            max_queue_size=MAX_QUEUE_SIZE,
+            max_export_batch_size=MAX_EXPORT_BATCH_SIZE,
+        )
```

`tests/tracing/test_sender.py` already reaches `_processor._batch_processor._max_export_batch_size`; extend that test to set both `OTEL_BSP_*` vars via `monkeypatch.setenv` and assert the sender still builds with batch 64 / queue 2048.

---

<sub>Two lower-severity items (the collector's 32 MiB receiver in front of an 8 MiB producer, and `memory_limiter` doubled with no container `mem_limit`) were left out as hardening notes rather than blockers. The fetch-buffer concern I initially raised on `consumer.py:135` was withdrawn: aiokafka's own docstring requires `max_partition_fetch_bytes` to be at least the broker's max message size or the consumer can wedge, so raising it to 8 MiB is the correct change.</sub>

victorludvig:
Both fixed — thanks, the zstd one would have taken prod down on deploy.

**1.** Dropped `compression: zstd` (f48c8ba) rather than adding the codec: repro confirmed, and compression was never a prerequisite here since the producer limit is measured uncompressed. Keeping it would also have made the fix depend on collector-vs-enricher deploy order. The comment records what enabling it later needs.

**3.** Pinned `MAX_QUEUE_SIZE = 2048`, both kwargs passed (17fce65), with a test that sets both `OTEL_BSP_*` vars and asserts batch 64 / queue 2048.

Hardening notes left alone: the 32 MiB receiver is intentional (the batch processor splits at 24 spans per record), and no compose file sets a `mem_limit` for the collector.

Both misses are folded into #205's review skill — a wire-compatibility lens, and no more "nothing in this repo sets that env var" for code shipping inside customers' processes.

guillaume-hexamind (approved):

PR #186, opened 2026-09-08
feat(platform-api): map hexgate.messages spans into LlmMessageEvent

victorludvig:

## What is changing
`LlmMessageEvent(AuditEnvelope)` in `schemas.py`, mirroring the `llm_message` columns of #183 including `run_id`; `capped_input_messages()` / `capped_output_messages()` / `capped_system_instructions()` in `jobs/enricher/enforcement.py` (key-redact like arguments, head+tail cap via the PR 1 helpers, truncate never reject, each returning the cut flag); `_message_fields()` and a `map_span` branch in `jobs/enricher/mapping.py` that ORs the enricher's cuts into the SDK's `truncated`; `as_json_value()` in `coerce.py` for the array-shaped attributes (the two bool flags reach the model raw — Pydantic coerces them). **`SCOPE_MESSAGES` is not added to `KNOWN_SCOPES` here** — the scope stays a loud `unknown_scope` DLQ reject until PR 4.

`jobs/enricher/dlq.py` now parses and redacts the three `gen_ai.*` message arrays before dead-lettering, the way it already does for `arguments` / `hint` / `attributes` — the DLQ is this PR's intended destination for every message span, and a tool-call message carries the same caller arguments a decision does.

`consumer.py` buckets events by type and commits the offset after the inserts, so a scope accepted before it has a bucket would be validated and then silently dropped with no DLQ record. Keeping registration for PR 4 means this PR can be reverted alone and main stays non-lossy. The DLQ redaction closes the gap that routing message spans there would otherwise open on a 30-day, no-ACL topic.

## Tests
`platform/api/tests/jobs/enricher/test_mapping.py` (calls `_message_fields()` directly: happy path, absent optional fields, SDK `truncated` kept, over-cap stored short, required attributes, `run_id` in/out); `test_enforcement.py` (cap boundary at and one byte over, head+tail marker, tool-call secret redacted, 8 KiB output/system caps); `test_coerce.py` (`as_json_value`); `test_dlq.py` (secret inside a message array redacted, bare-string field dropped).

### DIFF
--- a/hexgate/audit.py
+++ b/hexgate/audit.py
@@ -105,22 +105,62 @@ _SENSITIVE_ATTR_KEY_RE = re.compile(
 )
 _REDACTED = "[REDACTED]"
 
+# Keys whose string value may itself be serialized JSON that ``_redact`` must
+# look inside. In a ``gen_ai.*`` message a tool-call part carries the caller's
+# ``arguments``, and the raw OpenAI wire shape serializes them as a JSON
+# string, not an object — a string leaf, whose keys the substring match never
+# sees. Exact key names on purpose: parsing every string that happens to be
+# JSON would rewrite a user message whose content is JSON.
+TOOL_CALL_JSON_KEYS = frozenset({"arguments"})
 
-def _redact(value: Any, *, pattern: re.Pattern[str]) -> Any:
+
+def _redact(
+    value: Any,
+    *,
+    pattern: re.Pattern[str],
+    json_string_keys: frozenset[str] = frozenset(),
+) -> Any:
     """Return a copy of ``value`` with values under ``pattern``-matching keys replaced.
 
     Pure — never mutates the input, so the ``Decision`` the caller holds
-    keeps its full arguments; only the wire payload is redacted."""
+    keeps its full arguments; only the wire payload is redacted.
+
+    A string value under a key in ``json_string_keys`` that parses to a JSON
+    container is redacted inside and serialized back, so the field keeps the
+    shape the emitter chose while its nested keys still match."""
     if isinstance(value, dict):
         return {
             k: _REDACTED
             if isinstance(k, str) and pattern.search(k)
-            else _redact(v, pattern=pattern)
+            else _redact_json_string(
+                v, pattern=pattern, json_string_keys=json_string_keys
+            )
+            if k in json_string_keys and isinstance(v, str)
+            else _redact(v, pattern=pattern, json_string_keys=json_string_keys)
             for k, v in value.items()
         }
     if isinstance(value, (list, tuple)):
-        return [_redact(v, pattern=pattern) for v in value]
+        return [
+            _redact(v, pattern=pattern, json_string_keys=json_string_keys)
+            for v in value
+        ]
     return value
+
+
+def _redact_json_string(
+    text: str, *, pattern: re.Pattern[str], json_string_keys: frozenset[str]
+) -> str:
+    """``_redact`` applied inside a JSON-string leaf; non-JSON or scalar JSON
+    comes back untouched (nothing in it has a key to match)."""
+    try:
+        parsed = json.loads(text)
+    except ValueError:
+        return text
+    if not isinstance(parsed, (dict, list)):
+        return text
+    return json.dumps(
+        _redact(parsed, pattern=pattern, json_string_keys=json_string_keys)
+    )
 
 
 def _bounded_violations(violations: Sequence[str]) -> list[str]:
--- a/hexgate/tracing/semconv.py
+++ b/hexgate/tracing/semconv.py
@@ -74,7 +74,7 @@ HINT = "sec_ai.hint"
 ARGUMENTS = "sec_ai.arguments"
 ATTRIBUTES = "sec_ai.attributes"
 
-# --- Run attribution (SCOPE_AUDIT, plus RUN_ID on SCOPE_USAGE) -----------------
+# --- Run attribution (SCOPE_AUDIT; RUN_ID also on SCOPE_USAGE, SCOPE_MESSAGES) --
 # Omitted entirely when the emitter has no run to attribute: OTLP attributes
 # cannot carry null, and the platform's run_id is ``UUID | None`` — an empty
 # string is a 422. An absent attribute decodes to None; "" would not.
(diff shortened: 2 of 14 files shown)


guillaume-hexamind:
### Nested JSON-string tool-call `arguments` escape key redaction entirely — severity 72/100

**Where:** `platform/api/hexgate_api/jobs/enricher/enforcement.py:77` (`_capped_content`) and, live today, `platform/api/hexgate_api/jobs/enricher/dlq.py:70` (`_redacted_attributes`).

**What.** `redact` (`hexgate/audit.py:109`) matches dict *keys*; a string value is a leaf and comes back unchanged. The decision path works because `ARGUMENTS` is a top-level JSON string that `as_json_dict` parses first, so its keys become real keys. Inside a message array the equivalent payload sits one level deeper, and the raw OpenAI wire shape serializes `tool_call.arguments` as a **JSON string**, not an object. The string is never parsed, so the substring key match never reaches inside it.

Reproduced against this branch:

```python
capped_input_messages([{"role": "assistant", "parts": [
    {"type": "tool_call", "name": "login",
     "arguments": "{\"user\":\"bob\",\"password\":\"hunter2\"}"}]}])
# → [... "arguments": "{\"user\":\"bob\",\"password\":\"hunter2\"}"]   secret intact

capped_input_messages([{"role": "assistant", "parts": [
    {"type": "tool_call", "name": "login",
     "arguments": {"user": "bob", "password": "hunter2"}}]}])
# → [... "arguments": {"user": "bob", "password": "[REDACTED]"}]      redacted
```

Same on the DLQ path: a `hexgate.messages` span carrying the string-shaped payload produces an envelope with `hunter2` in plain text. That is the invariant `dlq.py:107` states — *"this topic has 30-day retention and no ACLs, so unredacted arguments must never reach it."* This path is reachable today, since `SCOPE_MESSAGES` is deliberately outside `KNOWN_SCOPES` and every message span is therefore DLQ'd as `unknown_scope`.

**Test gap.** Both new tests use the dict shape only — `test_dlq.py::test_when_a_message_span_is_rejected_then_dlq_redacts_inside_the_messages` and `test_enforcement.py:120`. The string variant is the one that fails.

**Two things that bound the severity**, for fairness:

- Nothing emits `hexgate.messages` yet (no producer anywhere outside `mapping.py` and tests), so this is latent, not leaking now.
- It is not new here: `capped_arguments({"body": '{"password": "hunter2"}'})` already returns the secret unredacted on the live decision path. This PR extends the existing gap to a new field rather than opening it.

It still scores high because the OTel GenAI semconv normalizes `tool_call.arguments` to an object while the provider wire format uses a string — so whether this bites depends entirely on whether the not-yet-written emitter normalizes, and this module's documented posture is that it explicitly does *not* trust emitters. Fixing it before the emitter ships is much cheaper than after, once the DLQ's 30-day retention has started accumulating.

**Suggested fix.** In `_redact` (or a wrapper at both call sites), attempt `json.loads` on string leaves whose key looks like `arguments`/`input`, and redact the parsed result when it yields a dict. A narrower option is to apply it only under `tool_call`-shaped parts. Either way, worth a test with the string-shaped `arguments` alongside the existing dict-shaped ones.

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code)</sub>

guillaume-hexamind (commented):
One blocking issue above; I approve of the logic otherwise

victorludvig:
> ### Nested JSON-string tool-call `arguments` escape key redaction entirely — severity 72/100
> **Where:** `platform/api/hexgate_api/jobs/enricher/enforcement.py:77` (`_capped_content`) and, live today, `platform/api/hexgate_api/jobs/enricher/dlq.py:70` (`_redacted_attributes`).
> 
> **What.** `redact` (`hexgate/audit.py:109`) matches dict _keys_; a string value is a leaf and comes back unchanged. The decision path works because `ARGUMENTS` is a top-level JSON string that `as_json_dict` parses first, so its keys become real keys. Inside a message array the equivalent payload sits one level deeper, and the raw OpenAI wire shape serializes `tool_call.arguments` as a **JSON string**, not an object. The string is never parsed, so the substring key match never reaches inside it.
> 
> Reproduced against this branch:
> 
> ```python
> capped_input_messages([{"role": "assistant", "parts": [
>     {"type": "tool_call", "name": "login",
>      "arguments": "{\"user\":\"bob\",\"password\":\"hunter2\"}"}]}])
> # → [... "arguments": "{\"user\":\"bob\",\"password\":\"hunter2\"}"]   secret intact
> 
> capped_input_messages([{"role": "assistant", "parts": [
>     {"type": "tool_call", "name": "login",
>      "arguments": {"user": "bob", "password": "hunter2"}}]}])
> # → [... "arguments": {"user": "bob", "password": "[REDACTED]"}]      redacted
> ```
> 
> Same on the DLQ path: a `hexgate.messages` span carrying the string-shaped payload produces an envelope with `hunter2` in plain text. That is the invariant `dlq.py:107` states — _"this topic has 30-day retention and no ACLs, so unredacted arguments must never reach it."_ This path is reachable today, since `SCOPE_MESSAGES` is deliberately outside `KNOWN_SCOPES` and every message span is therefore DLQ'd as `unknown_scope`.
> 
> **Test gap.** Both new tests use the dict shape only — `test_dlq.py::test_when_a_message_span_is_rejected_then_dlq_redacts_inside_the_messages` and `test_enforcement.py:120`. The string variant is the one that fails.
> 
> **Two things that bound the severity**, for fairness:
> 
> * Nothing emits `hexgate.messages` yet (no producer anywhere outside `mapping.py` and tests), so this is latent, not leaking now.
> * It is not new here: `capped_arguments({"body": '{"password": "hunter2"}'})` already returns the secret unredacted on the live decision path. This PR extends the existing gap to a new field rather than opening it.
> 
> It still scores high because the OTel GenAI semconv normalizes `tool_call.arguments` to an object while the provider wire format uses a string — so whether this bites depends entirely on whether the not-yet-written emitter normalizes, and this module's documented posture is that it explicitly does _not_ trust emitters. Fixing it before the emitter ships is much cheaper than after, once the DLQ's 30-day retention has started accumulating.
> 
> **Suggested fix.** In `_redact` (or a wrapper at both call sites), attempt `json.loads` on string leaves whose key looks like `arguments`/`input`, and redact the parsed result when it yields a dict. A narrower option is to apply it only under `tool_call`-shaped parts. Either way, worth a test with the string-shaped `arguments` alongside the existing dict-shaped ones.
> 
> 🤖 Generated with [Claude Code](https://claude.com/claude-code)

Fixed in 594aae8: redact now parses string values under arguments, redacts inside, and serialises back, at both the enforcement and DLQ call sites, with string-shaped tests beside the object-shaped ones; the pre-existing decision-path gap under arbitrary keys stays out of scope here.

guillaume-hexamind (approved):

PR #187, opened 2026-09-08
feat(sdk): add LlmMessageEvent and emit_llm_messages

victorludvig:
Design: [LLM message logging design](https://app.notion.com/p/3d4fb45dbae28141a3c4d32338a0d496) · [implementation spec](https://app.notion.com/p/3d4fb45dbae28109b82fd38cb03b9d11). Plan target: merge by **Thu 10 Sep**.

## What is changing
`hexgate/tracing/messages.py`: `LlmMessageEvent` (frozen dataclass, `SCOPE` ClassVar, `span_attributes()` serialising the three content fields as JSON and applying the PR 1 caps), `emit_llm_messages()` reusing `get_or_create_sender` and the `HEXGATE_LOCAL_MODE` gate, `HEXGATE_LOG_MESSAGES=0` opt-out read at emit time. Adds the scope to the tracer tuple at `hexgate/tracing/_senders.py#L178`.

Mirrors `tracing/usage.py` so message capture and token capture share one sender, one registry and one shutdown. The tracer tuple is hardcoded to three scopes; a scope missing there gets no tracer and fails at emit.

## Tests
`tests/tracing/test_messages_event.py` (attributes, caps applied, `truncated` flag), `test_messages_emit.py` (sender reuse, local mode, opt-out, never raises).

### DIFF
--- a/docs/reference/environment.mdx
+++ b/docs/reference/environment.mdx
@@ -33,6 +33,7 @@ raises a key-named error when the agent is built.
 | `HEXGATE_API_URL` | Platform endpoint. Defaults to Hexgate Cloud (`https://app.hexgate.ai`). Set to `http://localhost:8000` only when self-hosting locally — the key must be minted by whichever platform this points at. |
 | `HEXGATE_OTLP_ENDPOINT` | Where audit, usage and ban spans are exported (OTLP/HTTP). Defaults to `<HEXGATE_API_URL>/v1/traces`, which Hexgate Cloud and the reference deploy both serve. Set it only when the collector lives on a different host than the control plane, e.g. `http://localhost:4318/v1/traces` for a local `make collector-run`. |
 | `HEXGATE_LOCAL_MODE` | Truthy value forces local audit and shorts the platform auto-bind path, even when `HEXGATE_API_KEY` is set. `hexgate chat` sets this automatically. |
+| `HEXGATE_LOG_MESSAGES` | `0` (or `false` / `no` / `off`) switches off LLM message logging — the prompt delta and completion of every model call, exported under scope `hexgate.messages` — while decisions, usage and bans keep flowing. **Default on**: an audit log a customer has to discover and enable is not there when an incident needs it. Content is redacted and byte-capped by the SDK before export, retained 180 days, and stays under your controllership: it is your data, processed on your behalf. |
 | `HEXGATE_BIND_AGENTS` | Truthy value opts a code-built agent into the platform path at construction time (requires `HEXGATE_API_KEY` — the platform path is otherwise opt-in). |
 
 The key and the URL are coupled: a `fty_live_…` key only verifies against the
--- a/hexgate/audit.py
+++ b/hexgate/audit.py
@@ -76,8 +76,9 @@ MAX_VIOLATION_CHARS = 1024
 # Being in the repo is not the same as being in force: that change is
 # operational, so no message cap — 32 KiB or 256 KiB — is safe on a stage until
 # its topics have been altered and its collector and enricher restarted.
-# Nothing emits ``hexgate.messages`` yet, so declaring the constant here is
-# safe; wiring an emitter against a stage that has not been redeployed is not.
+# The emitter exists (``hexgate.tracing.messages``) but no adapter calls it
+# yet, so both stay inert; wiring an adapter hook against a stage that has not
+# been redeployed is what is unsafe.
 # Typical events stay a few KB; this is a ceiling, not a target.
 MAX_INPUT_MESSAGES_BYTES = 256 * 1024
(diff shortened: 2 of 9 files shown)


guillaume-hexamind:
Two findings from a review at `55fc854`, each verified by reading the source and running it. Neither blocks merging into `vl/feat/enricher_messages_scope` (nothing emits end-to-end yet), but both are cheaper to fix now than after the adapters land.

---

### 1. Redaction is silently skipped for non-dict messages — **72/100**

`_capped_json` (`hexgate/tracing/messages.py:65`) passes its value straight to `redact` → `cap_json_head_tail`, and neither descends into objects: `_redact` (`hexgate/audit.py:118`) returns anything that isn't a `dict`/`list`/`tuple` unchanged, and `_cap_leaves` (`hexgate/audit.py:273`) does the same for anything that isn't `str`/`dict`/`list`/`tuple`. `json.dumps(..., default=str)` then stringifies it, so the stored transcript is a `repr()`.

Reproduced at this head:

```
dataclass  → ["M(role='user', content='secret api_key=sk-live-ABC')"]   truncated: False
pydantic   → ["role='user' content='token=ghp_SECRET'"]
dict (ctl) → [{"role":"user","parts":[{"type":"tool_call","arguments":{"api_key":"[REDACTED]"}}]}]
```

This isn't limited to top-level message objects — any opaque value nested inside a dict message escapes the same way, and `tests/audit/test_message_caps.py:150` ("Framework message objects can carry a lock or a socket") shows the pipeline is knowingly fed non-JSON values already. The field types are `list[Any]`, and the only thing forbidding objects is prose in the module docstring, so the first adapter that forwards a framework's own message list defeats redaction with no error, no log and no flag.

It breaks an invariant this module states itself: blanking tool arguments *"in one place but not the other would leave the secret readable in the transcript."*

**Scope limit:** caps still hold. A 400 KB object message came out at 261,696 bytes against the 262,144 cap with `truncated=True` — `_cap_json_head_tail` measures `json.dumps(default=str)`, so the size is real and the preview wrapper catches it. This is a redaction hole only.

**Suggested fix.** Normalise to JSON-able containers *before* redaction, so `_redact` sees real keys rather than a string: a small step in `_capped_json` that converts a non-`dict`/`list` node via `model_dump()` (pydantic), `dataclasses.asdict()`, or `vars()`, falling back to `str()` only when none apply. The minimal alternative is a shape guard that refuses non-dict messages loudly instead of shipping them unredacted — but a guard that fires inside `emit_llm_messages`'s blanket `except` would be swallowed, so it needs to reject at the `LlmMessageEvent` boundary to be visible. Worth settling before any adapter hook is wired up.

---

### 2. Capture is on by default, and the opt-out is undocumented — **56/100**

`HEXGATE_LOG_MESSAGES=0` (`hexgate/tracing/messages.py:51`) is the only way to switch message-content capture off, and at this head the name appears nowhere outside `hexgate/tracing/messages.py` and `tests/tracing/test_messages_emit.py`. `docs/reference/environment.mdx` — the canonical table, which does list `HEXGATE_LOCAL_MODE`, `HEXGATE_BIND_AGENTS`, `HEXGATE_OTLP_ENDPOINT` — has no row for it. The PR touches zero doc files.

Separately, `docs/internals/audit-pipeline.md:613` still describes the opposite design:

> content is large, opt-in (nothing emits `hexgate.messages` yet, and capture stays off until an emitter ships)

This PR ships the emitter and makes capture default-on, so that line now contradicts the code on a privacy-relevant default. The adjacent comment in `hexgate/audit.py:79` *was* updated in this PR ("The emitter exists … but no adapter calls it yet"), so this looks like a missed spot rather than a disagreement.

No live exposure today — no adapter calls the emitter and the enricher does not yet accept the scope — which is why this is scored as a contract gap rather than a leak.

**Suggested fix.** Two edits: add a `HEXGATE_LOG_MESSAGES` row to `docs/reference/environment.mdx` stating that message-content capture is on by default and `=0` turns it off, and correct the parenthetical at `docs/internals/audit-pipeline.md:613`. Worth confirming the on-by-default choice itself with whoever owns privacy posture while it is still free to flip.

---

<sub>Reviewed with Claude Code.</sub>

guillaume-hexamind (commented):
Good to me, 2 comments to investigate above

victorludvig:
> Good to me, 2 comments to investigate above

1. Non-dict messages skipped redaction → _capped_json now round-trips each content field through JSON first (_as_json_data, pydantic model_dump / dataclass / mapping aware), so redaction and caps always walk plain data. Tests added for pydantic, dataclass and opaque leaves.
2. HEXGATE_LOG_MESSAGES undocumented → added to docs/reference/environment.mdx with the off spellings, default-on rationale and controllership note.

guillaume-hexamind (approved):

PR #189, opened 2026-09-08
feat(sdk): add the MessageCursor delta helper

victorludvig:
Design: [LLM message logging design](https://app.notion.com/p/3d4fb45dbae28141a3c4d32338a0d496) · [implementation spec](https://app.notion.com/p/3d4fb45dbae28109b82fd38cb03b9d11). Plan target: merge by **Fri 11 Sep**.

## What is changing
`MessageCursor` in `hexgate/tracing/messages.py`: per-`turn_key` high-water mark (count + prefix fingerprint) → `(new_messages, seq, resynced)`; tool results included, never filtered by role; shrink or in-place rewrite → full list with `resynced=True`; `reset(turn_key)` on run end. Pure functions over plain dicts.

Two design points worth a reviewer's attention, both settled by a grounded review on this branch:

- **`advance()` never raises.** It runs from framework hooks that re-raise into the customer's agent run, and fingerprinting is the one step here touching raw, un-flattened framework data — `json.dumps(sort_keys=True)` still throws on a non-scalar dict key (a tool returning `{date: value}`), mixed key types, or a cycle. Same rule `emit_llm_messages` and `AuditSender.emit` already keep, centralised by #105 so no adapter has to remember it. A failure degrades to a resync and leaves an unmatchable mark, so the *next* call resyncs too rather than slicing against a mark never established.
- **No LRU bound on the state map**, deliberately. An eviction would drop a list whose run is still going, and the cursor cannot tell that key from one it has never seen — so the comeback would go out as a *fresh turn*: full history again, at `seq=0`, not flagged `resynced`, with the turn's counter rewound past the gap detection the reader relies on. `reset()` on run end is the memory contract instead (a few hundred bytes per live list, no conversation content). An adapter that forgets `reset` leaks slowly and visibly; silently corrupting a transcript is the worse trade.

All four adapters compute the same delta; one implementation instead of four subtly different ones.

The non-obvious part is *what the fingerprint compares against*. Not "the list now" vs "the list last time" — those always differ, since the list grew. It is **the prefix of the list I am holding now** vs **the mark I stored for what I already emitted**:

```
stored: count=5, fp=hash(m0, m4)

append           [m0 m1 m2 m3 m4 m5 m6]
                 └── messages[:5] ──┘   hash(m0, m4) == fp   → emit [m5, m6]

summarise        [s0 m3 m4 m5 m6]
                 └─ messages[:5] ─┘     hash(s0, m6) != fp   → resync, emit all 5
```

A counter alone would be enough if these lists only ever grew. They don't: frameworks trim old turns or replace them with a summary to fit the context window. The trim case makes `messages[5:]` empty, so we would emit nothing, silently, forever. The summarise case is worse — the list can keep its length while holding different messages, so the slice returns real messages that are simply the wrong ones, with a hole before them and nothing to signal it. Hashing only the two endpoints keeps this O(1) per call instead of O(conversation); the blind spot (a rewrite strictly in the middle preserving length and both ends) is documented on `_fingerprint` and costs a stale message, not a corrupt slice.

## Tests
`tests/tracing/test_messages_cursor.py`: append of one, append of several (parallel tool results), trimmed prefix, same-length-different-content, two interleaved `turn_key`s, unchanged list (empty delta, no false resync), seq monotonic across a resync, key-order insensitivity, unserialisable message → resync not raise, and the call after a failure resyncing too.

### DIFF
--- a/hexgate/tracing/messages.py
+++ b/hexgate/tracing/messages.py
@@ -13,16 +13,20 @@ default — a log a customer has to discover and switch on is a log that is not
 there when an incident needs it — and ``HEXGATE_LOG_MESSAGES=0`` turns it off
 without touching the other streams.
 
-Which messages count as "new" is the adapter's job; this module only lays an
-already-derived event out on the wire — redacted, capped and serialised.
+Which messages are "new" is decided here too, by :class:`MessageCursor`: every
+adapter hook is handed the *whole* input list and has to work out what changed
+since its last call. One implementation shared by the four adapters rather than
+four subtly different ones.
 """
 
 from __future__ import annotations
 
 import dataclasses
+import hashlib
 import json
 import logging
 import os
+import threading
 from collections.abc import Mapping
 from dataclasses import dataclass, field
 from datetime import datetime, timezone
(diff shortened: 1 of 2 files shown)


hexgate/tracing/messages.py
```diff
@@ -0,0 +1,456 @@
...
+    @staticmethod
+    def _diff(
+        messages: list[Any], state: _TurnState
+    ) -> tuple[list[Any], bool, _TurnState]:
+        """Split ``messages`` against ``state`` and fingerprint it for next
+        time. ``messages[0]`` is canonicalised once and used for both
+        fingerprints: in steady state it is the same message every call, and
+        one of these lists can carry an inlined image."""
+        if not messages:
+            return [], state.count > 0, _FRESH_TURN
```
guillaume-hexamind: **An empty `messages` list resets the mark to the fresh-turn state, so the next call re-emits the whole history as an unflagged extension.**

`_FRESH_TURN` here is stored verbatim by `advance`, which leaves `count=0, fingerprint="0:"` — byte-identical to the state of a `turn_key` that has never been seen. The next call then takes the `state.count == 0` branch, `prefix == state.fingerprint`, and slices `messages[0:]`:

```
3 msgs        -> n=3 seq=0 resynced=False  state=(count=3, fp='3:ca73eb…')
empty         -> n=0 seq=1 resynced=True   state=(count=0, fp='0:')
same 3        -> n=3 seq=2 resynced=False  state=(count=3, fp='3:ca73eb…')   # duplicate
4th appended  -> n=1 seq=3 resynced=False                                    # recovers
```

The third row restates a, b, c with `resynced=False` and no gap in `message_seq`, so a reader has nothing to key duplicate-detection off. It is the outcome the class docstring rejects an LRU over — "the full history again … not flagged `resynced`" — though milder: `next_seq` keeps counting, so `seq == 0` still means "first event of the turn" and `system_instructions` are not re-attached. The duplication is one-time, not a stuck state.

**Whether this is a bug at all depends on a gap in the wire contract.** `semconv` defines `RESYNCED` as "an event that restates the whole list because the framework rewrote it rather than extending it". Under that wording the empty row on line 2 legitimately means *the list is now empty* — and a reader that implements it that way makes the mark reset correct and row 3 a faithful extension. A reader that treats an empty restatement as a no-op instead gets silent duplication. The contract does not say which, so both readers are conformant and they disagree.

**Suggested fix** — the same mark the exception path on line 410 already uses:

```suggestion
        if not messages:
            return [], state.count > 0, _TurnState(0, _UNMATCHABLE_FINGERPRINT, 0)
```

The comeback then goes out flagged `resynced`, which is consistent under *both* readings of the contract rather than only one of them — that is the reason to prefer it over pinning down the empty-resync semantics in `semconv`, though doing both would be better still.

Two smaller notes on the same path:

- Consider suppressing the event entirely for an empty list rather than emitting `resynced=True, messages=[]`, if an empty restatement is not meant to be meaningful. The `MessageDelta` docstring requires the caller to emit every delta it asks for, so that decision belongs here, not in the adapter.
- This branch has no test — the other branches of `_diff` each have a named one. Worth covering, since `LlmMessageEvent` carries `system_instructions` as its own field, so adapters are expected to split system messages out of the list they pass here: a call whose list holds only a system message hands over `[]`.

victorludvig: #1 (empty list resets the mark) — Fixed in 5c465b3: the empty branch now returns the same unmatchable mark as the fingerprinting-failure path, so the comeback is flagged resynced under either reading of the contract.

#2 (suppress the event entirely) — Declined: MessageDelta.messages is only the input delta, so suppressing the row would throw away output_messages, which is new on every call regardless.

#3 (no test for this branch) — Added in 5c465b3: test_when_the_list_is_empty_then_the_comeback_is_flagged and test_when_the_first_call_is_empty_then_it_is_not_a_resync.

guillaume-hexamind (commented):
One comment above, will approve then!

guillaume-hexamind (approved):

PR #190, opened 2026-09-08
feat(platform-api): add the session-scoped llm-messages read endpoint

victorludvig:
Design: [LLM message logging design](https://app.notion.com/p/3d4fb45dbae28141a3c4d32338a0d496) · [implementation spec](https://app.notion.com/p/3d4fb45dbae28109b82fd38cb03b9d11). Plan target: merge by **Mon 14 Sep**.

## Scope

This endpoint exists to populate the **Audit decision drawer** with the exchange around one decision — nothing else. It is an entity fetch: you already know which conversation you want and you get all of it, oldest first, so the drawer can anchor on the turn before the decision's timestamp and the turn after.

It is deliberately **not** a way to browse an agent's messages, which is why it has no `agent` / `user` / `model` / `window` filter. On a transcript a filter doesn't narrow, it puts holes in the middle: a session spans sub-agents and handoffs by design, and the only in-band signal for a missing row is a `message_seq` gap, which `schema.sql` defines as *the pipeline lost a row*. Long transcripts page (`limit`/`offset`) instead.

A project- or agent-wide messages read is a **follow-up**, and a different shape — an aggregation like `GET /projects/{id}/llm/summary`, returning counts and sizes rather than 272 KiB prompt bodies. Spec §V puts trace exploration in v1 non-goals, so that is a deliberate "not now".

## What is changing

`GET /v1/projects/{project_id}/audit/llm-messages` in `features/llm_messages/router.py`, cookie-authed like the other project-scoped dashboard reads. `list_llm_messages()` returns rows oldest first, ordered `(occurred_at, message_seq, event_id)`. `LlmMessageRow` / `LlmMessagePage` in `schemas.py`, with the three `gen_ai.*` content columns decoded back to objects and a zero `run_id` surfaced as `null`. `limit` clamps to 100, not the decision read's 200 — a decision row is about a kilobyte, a message row up to ~272 KiB.

### Scoped by `session_id`, `run_id`, or both — at least one, else 422

Two scopes rather than the one the spec names, because `session_id` is caller-supplied and most SDK users never set it: `HexgateContext.session_id` is `None` by default, `emit_llm_messages` folds that to `""`, and the column defaults to `""`. Requiring it would have left every transcript from an unnamed session stored and unreadable for its whole 180-day TTL — the common path, not an edge case. So `list_decisions` now returns `run_id` too (`AuditDecisionRow.run_id`, zero read back as `null`, `api.ts` in lockstep); that is the one change outside the slice, and it is the deferred half of #153, which said a per-run view was "unblocked by this PR, not shipped by it".

**Both fields are taken exactly as the caller sends them, blanks included** — a drawer forwarding a decision row should not strip an empty value, and `?session_id=&run_id=…` is the intended request. That matters for more than ergonomics: `session_id` is the second column of the storage sort key, so pinning it *even to `""`* keeps the scan inside one contiguous, time-ordered block. Measured on 4000 session-less rows of one run, same page: **195 rows / 64 MiB / 8 ms** with the blank sent, against **4036 rows / 923 MiB / 213 ms** without it. Omitting `session_id` means "I do not know the session" and accepts that cost. Documented on both OpenAPI parameters and in spec §III. A blank `run_id` likewise reads as absent, so a null one may be forwarded unchanged. Neither an empty session nor a zero `run_id` is a *scope* on its own — both match every unattributed row in the project — so the 422 still fires.

### Page and total are two queries, not `count() OVER ()`

A window function has no frame to prune, so ClickHouse buffered every matching row — `input_messages` included — before emitting the first, discarding the read-in-order short-circuit the `ORDER BY` was chosen for. `MAX_PAGE_SIZE` bounded the response but not the working set, and with `max_memory_usage = 0` the blow-up spends the server-wide budget rather than failing as a clean 503. Measured on a 2000-row session of 200 KB rows, same page: **1.05 GiB / 2067 rows** against **49.6 MiB / 106 rows** after the change, plus ~4 MiB for the count. The separate `count()` touches no content column. The `list_decisions` precedent does not carry over — ~1 KB rows, and its `ORDER BY occurred_at DESC` does not match that table's sort key, so no short-circuit is given up there. This also subsumes the old empty-page-past-the-end fallback.

Known remaining cost: `offset` is floored but not capped, so a deep page on a very long transcript still materialises `limit + offset` rows. No caller pages that deep today.

Also moves `_decode_json_column` from `features/audit/service.py` into `core/clickhouse.py` so both read paths share one decoder.

The Audit detail drawer (PR 14) and the OTLP smoke script (PR 7) both read a transcript back through this endpoint — it is the only way anything gets at the `llm_message` rows PR 4 started storing.

## Tests

`tests/features/llm_messages/test_llm_message_read.py` — 31 unit cases: ordering and the retention scope; session-only, run-only and both-scopes reads; the explicitly-empty session pinned rather than dropped; 422 on no scope, on a blank session alone, on a zero `run_id` and on a malformed one; a blank `run_id` falling through to the session scope; paging and the count query proving it reads no content column; JSON decoding including content that no longer parses; zero vs real `run_id`; the `UInt8` flags; `limit`/`offset` clamping; 503 when ClickHouse is down; and tenant isolation (401 anonymous, 404 unknown project, 403 non-member, 200 member).

Plus a `@pytest.mark.integration` round-trip against real ClickHouse — the mocked tests assert on the SQL *text* and never execute it, so a wrong column name or a broken `ORDER BY` passes all 31. Verified by mutation: a typo'd column fails with `UNKNOWN_IDENTIFIER`, a reversed `ORDER BY` fails. `tests/features/audit/test_audit.py` gains the two `run_id` cases on `list_decisions`.

`make fmt-check`, `make platform-api-check` and `pytest -m integration` all pass.


guillaume-hexamind:
Two issues found on review (scores are severity × confidence, both verified against `a4f43df`).

---

### 1. Blank `session_id` is a 422 even when a valid `run_id` is supplied — 65/100

`router.py:43` — `session_id: str | None = Query(default=None, min_length=1)` rejects `?session_id=&run_id=<uuid>` with `422 string_too_short` before `list_llm_messages` is ever called. Reproduced with the exact signature:

```
?session_id=&run_id=<uuid>  -> 422 string_too_short   # service never reached
?run_id=<uuid>              -> 200
```

This is the case the `run_id` scope was added for. `service.py:175` already treats a falsy `session_id` as absent (`if not session_id and ...`), so only the router turns it into an error — and the sibling endpoint disagrees too (`audit/router.py:260` declares `session_id: str | None = None`, no `min_length`). The parametrized 422 test covers `?session_id=` alone but never `?session_id=&run_id=<valid>`, which is the broken combination. A caller forwarding both fields off a decision row (`session_id=""`, per the docstring at `test_llm_message_read.py:131`) hits it immediately.

**Fix:** drop `min_length=1` and let `NoMessageScope` decide — it already returns 422 for blank-session-alone, so the existing tests keep passing:

```python
session_id: str | None = None,
```

---

### 2. `count() OVER ()` materializes the whole transcript on every page — 65/100

`service.py:188` — the window frame is the full partition, so ClickHouse buffers every matching row (`input_messages` included, capped at 256 KiB) before emitting the first one, regardless of `LIMIT`. `MAX_PAGE_SIZE = 100` bounds the response but not the server-side working set.

Measured on ClickHouse 24.10 with a table mirroring `llm_message`'s sort key, 2,000 rows × 200 KB, same page (`LIMIT 50 OFFSET 0`):

| query | peak memory | rows read |
|---|---|---|
| with `count() OVER ()` | **1.07 GiB** | 2,000 |
| without it | **27.3 MiB** | **52** |

The memory is only half of it: without the window function the read stops after 52 rows, because `ORDER BY occurred_at, message_seq, event_id` matches the storage sort key. `count() OVER ()` discards exactly the read-in-order short-circuit the docstring cites as the reason for that ordering. Paging doesn't relieve it — every page re-materializes the whole transcript.

Worth noting `max_memory_usage` is `0` here (no per-query cap), so this doesn't fail as a clean 503 on the offending request; it spends the server-wide budget and pressures concurrent queries.

Latent rather than live — no adapter calls the emitter yet and typical rows are a few KB — but the exposure is the RAG tail, i.e. the calls the 256 KiB cap was widened for. The `list_decisions` precedent doesn't carry over: ~1 KB rows, and its `ORDER BY occurred_at DESC` doesn't match that table's sort key, so no index short-circuit is being given up there.

**Fix:** plain page + separate `count()`. The count reads only key columns, and the existing "empty page past the end" fallback at `service.py:208` folds into it:

```python
result = client.query(
    f"SELECT {_LIST_COLUMNS} FROM {LLM_MESSAGE_TABLE} WHERE {where_sql} "
    "ORDER BY occurred_at, message_seq, event_id "
    "LIMIT {lim:UInt32} OFFSET {off:UInt32}",
    parameters=page_params,
)
...
total = int(
    client.query(
        f"SELECT count() FROM {LLM_MESSAGE_TABLE} WHERE {where_sql}",
        parameters=params,
    ).result_rows[0][0]
)
```

Returning `total` only on the first page is the cheaper variant if the second query shows up in profiles.

guillaume-hexamind (commented):
Good simple logic / 2 issues to check before merge

victorludvig:
Both fixed, thanks — the second one was the more interesting of the two.

**1. Blank `session_id` 422'ing with a valid `run_id`** — fixed in `fe323177`. Reproduced exactly as you described; `min_length=1` is gone and `NoMessageScope` decides, so blank-alone and no-scope still 422. It turned out to have a mirror I'd missed — `?session_id=sess_1&run_id=` 422'd too, because `UUID | None` rejects a blank before the scope check — so `run_id` is now parsed by hand and a blank reads as absent (`50f57252`).

**2. `count() OVER ()` materialising the whole transcript** — fixed in `a127d225`. Reproduced your numbers on the real table (1.07 GiB / 2000 rows read for a 50-row page) and went with the plain page + separate `count()`. Your point that the window function discards the read-in-order short-circuit the `ORDER BY` was chosen for was the part I'd missed — the code was arguing against itself, and the `list_decisions` comparison in my own comment was wrong for exactly the reason you gave.

That fix turned out to be incomplete, though: `run_id` isn't in the sort key, so the run-scoped read still had no short-circuit — 4036 rows / 923 MiB for the same page. Since a drawer forwards the decision row's fields verbatim, the request really is `?session_id=&run_id=…`, and binding that explicit blank (rather than treating `""` as absent) restores it: 195 rows / 64 MiB / 8 ms. That's `50f57252`, documented on the OpenAPI parameter and in spec §III so the drawer doesn't strip the blank.

Also added an integration test that round-trips the read against real ClickHouse — the mocked tests assert on SQL *text* and never execute it, so a typo'd column or a reversed `ORDER BY` passed all 31 of them.

guillaume-hexamind (approved):

PR #191, opened 2026-09-08
feat(sdk): capture OpenAI Agents messages in HexgateUsageHooks

victorludvig:
Design: [LLM message logging design](https://app.notion.com/p/3d4fb45dbae28141a3c4d32338a0d496) · [implementation spec](https://app.notion.com/p/3d4fb45dbae28109b82fd38cb03b9d11). Plan target: merge by **Mon 14 Sep**.

Stacked on #190 (`vl/feat/llm_messages_read`), not `main` — the read endpoint is what the second integration case verifies through.

## What is changing
`adapters/openai/usage.py`: `on_llm_start` stashes `(system_prompt, input_items)` per `turn_key`; `on_llm_end` emits usage **and** messages from one call site, then clears the stash. The runner already forwards `on_llm_start`. The Responses-API → `gen_ai.*` shape translation lives in the new `adapters/openai/messages.py`, so each of PRs 11–13 gets the same layer for its own framework.

`turn_key` is `f"{run_id}:{agent.name}"` — Hexgate's run id, not `id(context)`: that address is reused once the run context is freed, so two unrelated runs in one process would share a key and both restart `message_seq` at 0.

`TOOL_CALL_JSON_KEYS` gains `"response"`: frameworks stringify a tool's return value, so a tool serialising its own result landed a JSON string whose secrets redaction never opened.

The response hook carries no prompt, so the pair is needed. Only the delta reaches the wire — the input list is the whole conversation so far — and tool results ride with it, since a `policy_decision` row records a tool call but never its return value.

## Tests
`tests/adapters/openai/test_usage.py` (hooks) and `test_messages.py` (conversion); `tests/audit/test_message_caps.py` for the new redaction key. Two integration cases: one asserting the rows concatenate back to the conversation with contiguous `message_seq`, one reading the same transcript back through #190's `GET /v1/projects/{id}/audit/llm-messages` (cookie-authed, so it skips without `HEXGATE_SMOKE_EMAIL`/`_PASSWORD`).

Run locally against the full stack — Postgres, ClickHouse, Redpanda, platform-api, Collector, span-enricher:

| Suite | Result |
| --- | --- |
| `ruff check` + `format --check` | clean (hexgate, tests) |
| SDK unit | 2672 passed, 15 skipped |
| platform-api unit | 796 passed, 10 skipped |
| platform-api integration | 8 passed |
| SDK integration (full OTLP path) | 7 passed |

Known and tracked, not fixed here: #215 (should a handoff continue one transcript or start a new one — the handoff integration case lands with that decision) and #216 (`Agent.as_tool` nested runs get no hooks, so no usage or message rows).


guillaume-hexamind:
## Review — 2 issues

Verified against the branch tip, `hexgate/tracing/messages.py` / `semconv.py`, and `openai-agents==0.15.1`. Score is severity out of 100.

---

### 1. `resynced` is `true` on every row under a server-managed conversation — **55/100**

`usage.py:115` documents `input_items` as "the whole conversation, not a delta". That only holds when no `OpenAIServerConversationTracker` is in play.

`HexgateRunner.run/run_sync/run_streamed` forward `**kwargs` to `Runner.run`, so `conversation_id=` / `previous_response_id=` / `auto_previous_response_id=` reach `agents/run.py:542` and build a tracker. From then on `run_loop.py:1348` builds the prompt with `tracker.prepare_input(...)`, which returns **only the un-sent items** — `oai_conversation.py:421-441` sends the initial input once and `mark_input_as_sent` clears `remaining_initial_input`.

`MessageCursor` stores `count=1, fingerprint="1:digest(user,user)"` after turn 1; turn 2 hands it a short list starting with a tool output, the prefix fingerprint can never match, and `advance` returns `resynced=True` — on every turn, for the life of the run.

The content stays correct (the short list *is* the delta), so nothing is lost or duplicated. The flag is what's wrong: `semconv.py:46` defines `RESYNCED` as "restates the whole list", so a reader honouring that drops every earlier row and renders a two-message conversation. Latent today only because nothing consumes the column yet.

Same cause, minor: `run_loop.py:1387` permits an empty `filtered.input` when a tracker is present, and `MessageCursor._diff([])` stores `_TurnState(0, _UNMATCHABLE_FINGERPRINT, 0)`. Harmless here, since a resync is already the steady state in this mode.

**Suggested fix** — the hook can't see the tracker, but the runner sees the kwargs that create it. Pass the mode down and skip the diff when the framework is already sending deltas:

```python
# usage.py
_SERVER_CONVERSATION_KWARGS = frozenset(
    {"conversation_id", "previous_response_id", "auto_previous_response_id"}
)

class HexgateUsageHooks(RunHooks):
    def __init__(self, *, api_key: str, framework_sends_deltas: bool = False) -> None:
        ...
        self._framework_sends_deltas = framework_sends_deltas
```

In `_emit_messages`, when `self._framework_sends_deltas`, emit `new_input` as-is with `resynced=False` and take only the seq from the cursor (a `MessageCursor.advance_delta(key, count)`-style entry point, or a counter here), instead of diffing a list that was never the full history.

`_merge_hooks` already has `run_config` and `kwargs` in scope at all four call sites (`run`, `run_sync`, `run_streamed`/`arun_streamed` via `_launch_streamed`), so it becomes:

```python
def _merge_hooks(self, hooks: RunHooks | None, *, kwargs: dict[str, Any]) -> RunHooks:
    installed: list[RunHooksBase] = [
        HexgateUsageHooks(
            api_key=self.api_key,
            framework_sends_deltas=any(
                kwargs.get(name) for name in _SERVER_CONVERSATION_KWARGS
            ),
        ),
        _HexgateReachHooks(self),
    ]
```

---

### 2. `_resolve_model` ignores `RunConfig.model` — **50/100**

`agents/run_internal/turn_preparation.py:123-132 get_model` gives `run_config.model` precedence over `agent.model`, for both the `Model` and the `str` form:

```python
def get_model(agent: Agent[Any], run_config: RunConfig) -> Model:
    if isinstance(run_config.model, Model):
        return run_config.model
    elif isinstance(run_config.model, str):
        return run_config.model_provider.get_model(run_config.model)
    elif isinstance(agent.model, Model):
        return agent.model
    return run_config.model_provider.get_model(agent.model)
```

`usage.py:60 _resolve_model(agent)` reads only `agent.model`, and `runner.py:274/298` accepts and forwards `run_config`. So `HexgateRunner.run(agent, ..., run_config=RunConfig(model="gpt-4o-mini"))` against an `Agent(model="gpt-4o")` records `gpt-4o` — now in `llm_message` as well as `llm_usage`, so this PR doubles the blast radius of a pre-existing bug.

`RunContextWrapper` (`run_context.py:43-62`) carries `context`, `usage`, `turn_input`, `_approvals`, `tool_input` and no `run_config`, so the hook can't recover this on its own.

**Suggested fix** — inject `run_config` alongside the flag above and mirror `get_model`'s precedence:

```python
def _resolve_model(agent: Agent, run_config: RunConfig | None) -> str:
    configured = run_config.model if run_config is not None else None
    model = configured if configured is not None else agent.model
    if isinstance(model, str):
        return model
    if model is None:
        return "default"
    return getattr(model, "model", None) or type(model).__name__
```

with `HexgateUsageHooks.__init__` taking `run_config: RunConfig | None = None` and `on_llm_end` calling `_resolve_model(agent, self._run_config)`. The `"default"` placeholder for an unset model stays as documented.

guillaume-hexamind (commented):
2 comments above; the logic look good otherwise

victorludvig:
> ## Review — 2 issues
> Verified against the branch tip, `hexgate/tracing/messages.py` / `semconv.py`, and `openai-agents==0.15.1`. Score is severity out of 100.
> 
> ### 1. `resynced` is `true` on every row under a server-managed conversation — **55/100**
> `usage.py:115` documents `input_items` as "the whole conversation, not a delta". That only holds when no `OpenAIServerConversationTracker` is in play.
> 
> `HexgateRunner.run/run_sync/run_streamed` forward `**kwargs` to `Runner.run`, so `conversation_id=` / `previous_response_id=` / `auto_previous_response_id=` reach `agents/run.py:542` and build a tracker. From then on `run_loop.py:1348` builds the prompt with `tracker.prepare_input(...)`, which returns **only the un-sent items** — `oai_conversation.py:421-441` sends the initial input once and `mark_input_as_sent` clears `remaining_initial_input`.
> 
> `MessageCursor` stores `count=1, fingerprint="1:digest(user,user)"` after turn 1; turn 2 hands it a short list starting with a tool output, the prefix fingerprint can never match, and `advance` returns `resynced=True` — on every turn, for the life of the run.
> 
> The content stays correct (the short list _is_ the delta), so nothing is lost or duplicated. The flag is what's wrong: `semconv.py:46` defines `RESYNCED` as "restates the whole list", so a reader honouring that drops every earlier row and renders a two-message conversation. Latent today only because nothing consumes the column yet.
> 
> Same cause, minor: `run_loop.py:1387` permits an empty `filtered.input` when a tracker is present, and `MessageCursor._diff([])` stores `_TurnState(0, _UNMATCHABLE_FINGERPRINT, 0)`. Harmless here, since a resync is already the steady state in this mode.
> 
> **Suggested fix** — the hook can't see the tracker, but the runner sees the kwargs that create it. Pass the mode down and skip the diff when the framework is already sending deltas:
> 
> ```python
> # usage.py
> _SERVER_CONVERSATION_KWARGS = frozenset(
>     {"conversation_id", "previous_response_id", "auto_previous_response_id"}
> )
> 
> class HexgateUsageHooks(RunHooks):
>     def __init__(self, *, api_key: str, framework_sends_deltas: bool = False) -> None:
>         ...
>         self._framework_sends_deltas = framework_sends_deltas
> ```
> 
> In `_emit_messages`, when `self._framework_sends_deltas`, emit `new_input` as-is with `resynced=False` and take only the seq from the cursor (a `MessageCursor.advance_delta(key, count)`-style entry point, or a counter here), instead of diffing a list that was never the full history.
> 
> `_merge_hooks` already has `run_config` and `kwargs` in scope at all four call sites (`run`, `run_sync`, `run_streamed`/`arun_streamed` via `_launch_streamed`), so it becomes:
> 
> ```python
> def _merge_hooks(self, hooks: RunHooks | None, *, kwargs: dict[str, Any]) -> RunHooks:
>     installed: list[RunHooksBase] = [
>         HexgateUsageHooks(
>             api_key=self.api_key,
>             framework_sends_deltas=any(
>                 kwargs.get(name) for name in _SERVER_CONVERSATION_KWARGS
>             ),
>         ),
>         _HexgateReachHooks(self),
>     ]
> ```
> 
> ### 2. `_resolve_model` ignores `RunConfig.model` — **50/100**
> `agents/run_internal/turn_preparation.py:123-132 get_model` gives `run_config.model` precedence over `agent.model`, for both the `Model` and the `str` form:
> 
> ```python
> def get_model(agent: Agent[Any], run_config: RunConfig) -> Model:
>     if isinstance(run_config.model, Model):
>         return run_config.model
>     elif isinstance(run_config.model, str):
>         return run_config.model_provider.get_model(run_config.model)
>     elif isinstance(agent.model, Model):
>         return agent.model
>     return run_config.model_provider.get_model(agent.model)
> ```
> 
> `usage.py:60 _resolve_model(agent)` reads only `agent.model`, and `runner.py:274/298` accepts and forwards `run_config`. So `HexgateRunner.run(agent, ..., run_config=RunConfig(model="gpt-4o-mini"))` against an `Agent(model="gpt-4o")` records `gpt-4o` — now in `llm_message` as well as `llm_usage`, so this PR doubles the blast radius of a pre-existing bug.
> 
> `RunContextWrapper` (`run_context.py:43-62`) carries `context`, `usage`, `turn_input`, `_approvals`, `tool_input` and no `run_config`, so the hook can't recover this on its own.
> 
> **Suggested fix** — inject `run_config` alongside the flag above and mirror `get_model`'s precedence:
> 
> ```python
> def _resolve_model(agent: Agent, run_config: RunConfig | None) -> str:
>     configured = run_config.model if run_config is not None else None
>     model = configured if configured is not None else agent.model
>     if isinstance(model, str):
>         return model
>     if model is None:
>         return "default"
>     return getattr(model, "model", None) or type(model).__name__
> ```
> 
> with `HexgateUsageHooks.__init__` taking `run_config: RunConfig | None = None` and `on_llm_end` calling `_resolve_model(agent, self._run_config)`. The `"default"` placeholder for an unset model stays as documented.

Both fixed in `02afcb61` and `b3bbfaaa`, implemented as suggested — kept `resynced=False` in delta mode rather than adding a third state, since a reader concatenating rows gets the right transcript either way.

guillaume-hexamind (approved):

PR #192, opened 2026-09-08
test(platform-scripts): cover hexgate.messages in the OTLP smoke check

victorludvig:

## What is changing

`otlp_smoke.py` sends six `hexgate.messages` events alongside the five it had, and reads them back through the PR #190 endpoint — asserting per row that it landed, that `turn_key` / `message_seq` round-tripped, that `truncated` is set on exactly the over-cap events, and that those arrived at full size rather than clipped.

Five of the six carry an input message past the 256 KiB cap. Spec §IV.4 asked for one ~40 KiB event, but that number predates the cap it was written against (32 KiB, later rejected): at 256 KiB a 40 KiB message is not truncated at all, and even one *capped* span rides both limits it would need to cross. The record has to clear the exporter's 1,000,000 bytes **and** the broker's 1 MiB; four spans land exactly *on* the latter, five clear both (~1.25 MiB). `OVERSIZED_EVENT_COUNT` derives that from `MAX_INPUT_MESSAGES_BYTES` rather than hardcoding it.

`DEPLOY.md` §4 and `audit-pipeline.md` §4.1 updated — both carried a "check this by hand" note for exactly this gap.

PR #185 raised six record-size limits, and that is an **operational** change: merged is not enough, the topics have to be altered and the collector and enricher restarted. Nothing checked that they were. The old run's five events fit a sub-1-MiB export on any stage, so it passed either way.

Five oversized spans do not. On an unraised stage the record is rejected whole and every event reports `MISSING`. This run is the post-deploy check on staging and the release gate on prod, so it is where that has to be caught — and it is the only check anywhere that an over-cap payload is *degraded* rather than dropped.

## Tests

`make check-all` — SDK 2672 passed, platform-api 796 passed, ruff clean. (`dashboard-lint` aborts on a broken local Node install; no dashboard files here.)

**Ran green on both stages.** Staging:

```
exporting 11 events to https://app.staging.hexgate.ai/v1/traces  (session s_smoke_c0bc0ffd)
  decision:allow / deny / needs_approval   ok
  llm_usage · ban_enforcement · llm_message ok
  llm_message:oversized1..5                ok
PASS: all events landed
```

Prod, the release gate (session `s_smoke_a38ae7dd`) — same, all 11.

Both prove what no unit or integration test can: a ~1.25 MiB record survives the real collector, topic, enricher and ClickHouse on a deployed stage. Read back from prod directly, the six rows are `message_seq` 0–5 under one `turn_key`, five at exactly 262144 bytes with `truncated=1`.

Also measured through the real enricher mapping: the ordinary row lands `truncated=False`, each oversized row `truncated=True` at the cap.


guillaume-hexamind (approved):
Good logic.

No issues found with a /code-review high

PR #193, opened 2026-09-08
feat(sdk): capture LangChain messages in HexgateUsageCallbackHandler

victorludvig:
PR 11/15 of the LLM message logging stack ([spec](https://app.notion.com/p/LLM-message-logging-implementation-spec-3d4fb45dbae28109b82fd38cb03b9d11)). Stacked on PR 7's branch, not `main`.

**This is [#191](https://github.com/HexamindOrganisation/hexgate/pull/191)'s design applied to LangChain** — same hook pair, same stash-the-request/emit-on-response shape, same `MessageCursor` delta, same converters-split-from-hooks layout. Four things had to differ, and they are where the review effort is worth spending:

| | OpenAI (#191) | here |
|---|---|---|
| handler lifetime | fresh per run | one per proxy → needs `end_run` |
| `turn_key` name half | `agent.name`, per call | the proxy's, fixed → coarser |
| system prompt | its own hook argument | head of the message list → `split_system_instructions` |
| streamed completion | full `AIMessage` | `AIMessageChunk`, whose `type` is its class name |

## What is Changing

- **`adapters/langchain/usage.py`** — `HexgateUsageCallbackHandler` gains the request side: `on_chat_model_start` (and `on_llm_start` for bare prompts) stashes the message list under LangChain's `run_id`, `on_llm_end` correlates it, takes the delta against a `MessageCursor`, and emits usage and messages from one call site.
- **`adapters/langchain/messages.py`** (new) — `BaseMessage` → the OTel GenAI `role`/`parts` shape, following PR 10's split.
- **`adapters/_messages.py`** (new) — `as_dict()`/`text_part()` hoisted out of `openai/messages.py`. This is the decision the spec deferred to "when the second adapter lands" (§III SDK), which is this PR.
- **`adapters/langchain/agent.py`, `agents/factory.py`** — both run boundaries now drop the run's cursor state on the way out. Unlike the OpenAI adapter's per-run hooks, this handler lives as long as the proxy.

LLM prompts and completions are the stack's fourth audit stream, beside decisions, token usage and bans. This PR turns it on for LangChain — the second of four adapters. Until it lands, a LangGraph agent produces `policy_decision` rows with no record of what the model was asked or what it answered.

Two things that buys, per the spec's §Goal: a denied tool call can be read next to the exchange that caused it, and the transcript is the only place a tool's *return value* is stored at all — a decision row records that a tool was called, never what came back.

Three decisions worth pushing back on:

- **`turn_key` is the Hexgate run id**, not LangChain's `run_id`/`parent_run_id` — those are UUIDv7s minted per callback, so keying on either would re-send the conversation as a snapshot every turn. It is knowingly coarser than one message list: a run that calls a second model with its own list (a summariser node, an unwrapped sub-graph) files those calls under the same key and the two resync against each other. Complete and flagged, but the delta's compression is lost for that run.
- **A secret inside a *malformed* tool call is stored unredacted** — `invalid_tool_calls` carry the string LangChain could not parse, and `_redact_json_string` hands unparseable text back untouched. Kept anyway with its parse error, since no key-name rule can open a blob that is not JSON.
- **`model` comes from the request side** (`metadata["ls_model_name"]`), with `"default"` only as a last resort — every `"default"` row is one more known model collapsed into a single bucket on the Usage breakdown.

**Known limitations:** an agent invoked inside another one's graph leaks ~200 B per nested run (closing it needs `MessageCursor`'s eviction contract to change). Separately, a pre-existing hazard in `runtime/run_facts.py` — an abandoned `astream` generator restores `DETACHED` over a *concurrent* run's facts — predates this PR and affects all four adapters; filing separately. This diff is defended against it: `end_run` is handed its key at scope entry.

## Tests

`make lint`, `make fmt-check` and the full SDK suite pass: **2734 passed, 16 skipped**. Coverage: `messages.py`, `agent.py`, `_messages.py` 100%, `usage.py` 99%.

- **`test_messages.py`** (new) — role mapping incl. streamed chunks, tool calls and their content-block duplicates (promoted, orphaned, partially promoted, id-less, provider-run), tool results incl. legacy `FunctionMessage`, and the system-instruction split.
- **`test_usage.py`** — the delta, a resync after a trim, two runs on one handler, `end_run` from a foreign Context, the `HEXGATE_LOG_MESSAGES` opt-out, a missing stash, a failed call, a raising converter, the non-chat path.
- **`test_factory.py`** — the `create_agent` runtime drops its transcript state on both run methods.
- **`test_integration.py`** (`pytest -m integration`) — a tool-calling run's rows concatenate back to the conversation the agent saw, over the full OTLP path. Not run here; needs the live pipeline.


guillaume-hexamind:
## Code review — 2 findings

Reviewed at `713ab8e`. Scores are confidence that the issue is real and worth fixing, verified by running the PR's converter and `MessageCursor` against `langchain-core 1.3.2` / `langchain-openai 1.2.1`.

---

### 1. `custom_tool_call` is missing from `_TOOL_CALL_BLOCK_TYPES`, so that call is recorded twice — **90/100**

`hexgate/adapters/langchain/messages.py:36`

`_TOOL_CALL_BLOCK_TYPES` exists to drop a raw content block when the same call is also re-emitted from the standardized `tool_calls`. The comment justifies leaving out `server_tool_use`/`mcp_tool_use` because "the provider ran those itself, so they never reach `tool_calls`" — but `custom_tool_call` is not in that category.

In `langchain_openai/chat_models/base.py`, the `custom_tool_call` branch (4673–4679) is structurally identical to the `function_call` branch 27 lines above it (4646–4663): both append a raw content block **and** a `tool_calls` entry keyed on `output.call_id`.

```python
elif output.type == "custom_tool_call":
    content_blocks.append(output.model_dump(exclude_none=True, mode="json"))
    tool_call = {"type": "tool_call", "name": output.name,
                 "args": {"__arg1": output.input}, "id": output.call_id}
    tool_calls.append(tool_call)
```

`function_call` is in the frozenset; `custom_tool_call` isn't. Run through the PR's converter:

```
custom_tool_call: 2 part(s) -> ['custom_tool_call', 'tool_call']
function_call   : 1 part(s) -> ['tool_call']
```

One call, two parts, two different id fields — a reader counts the call twice, and on the completion side both copies eat the same 8 KiB budget. Reachable through `langchain_openai.tools.custom_tool` on the Responses API.

**Fix** — the block already carries `call_id`, so the existing drop condition at `messages.py:104` matches as soon as the type is in the set:

```diff
 _TOOL_CALL_BLOCK_TYPES = frozenset(
-    {"function_call", "tool_use", "tool_call", "invalid_tool_call"}
+    {"function_call", "custom_tool_call", "tool_use", "tool_call", "invalid_tool_call"}
 )
```

---

### 2. An empty stash still advances the cursor, burning seq 0 and losing the system-prompt lift for the whole run — **50/100**

`hexgate/adapters/langchain/usage.py:212`

When `prompt` is `None`, or carries `messages == []`, `_emit_messages` still calls `self._cursor.advance(turn_key, [])`. `MessageCursor._diff` returns `_TurnState(0, _UNMATCHABLE_FINGERPRINT, 0)` for an empty list (`tracing/messages.py:468`), so the mark is clobbered to `"?"`, and `advance` bumps `next_seq` regardless (`:455`). The stored `"?"` can never equal the `"0:"` prefix the next call computes, so the first *real* prompt resyncs at seq 1 — and because the system-instructions lift is gated on `delta.seq == 0` (`usage.py:215`), that run's system prompt never reaches `system_instructions`; it stays inline in `input_messages` for the rest of the run.

Reproduced against the real cursor:

```
empty advance -> seq 0  resynced False  msgs []
first real    -> seq 1  resynced True   n 2     # <- system prompt never lifted
second real   -> seq 2  resynced False  n 2
baseline (no empty advance): first real -> seq 0  resynced False
```

Scored 50 rather than higher because the trigger is narrow: with `HEXGATE_LOG_MESSAGES=0` at both ends, `_emit_messages` returns at `:201` and never advances, so it takes a genuine mid-call flip of that env var between `on_chat_model_start` and `on_llm_end` (a flip the module explicitly supports), or a start callback that failed and left no stash.

**Fix** — skip the call entirely rather than spending its seq:

```diff
         turn_key, messages, _ = prompt or _Prompt(self.turn_key(run_id), [], "")
+        # Nothing was stashed for this call, so there is no delta to measure.
+        # Advancing on an empty list would clobber the mark and spend the
+        # turn's seq, and seq 0 is the only event that lifts the system prompt.
+        if not messages:
+            return
```

This drops the completion for that one call too. If that matters more than the seq, the alternative is to leave `advance` alone and give the cursor a way to say "nothing observed" without spending a seq — but that means changing the `MessageDelta` contract ("the caller must emit an event for every delta it asks for"), so the early return is the cheaper trade.

**Worth noting:** this is not new to this PR. `adapters/openai/usage.py:206` pops with `_Prompt(None, [])` and feeds the result straight into `advance` at `:247`, with the same `seq == 0` gate on `system_instructions` at `:226` — same bug, same consequence. Fixing only the LangChain side leaves the other half.

guillaume-hexamind (commented):
1 blocking issue above, will approve afterwards

victorludvig:
> ## Code review — 2 findings
> Reviewed at `713ab8e`. Scores are confidence that the issue is real and worth fixing, verified by running the PR's converter and `MessageCursor` against `langchain-core 1.3.2` / `langchain-openai 1.2.1`.
> 
> ### 1. `custom_tool_call` is missing from `_TOOL_CALL_BLOCK_TYPES`, so that call is recorded twice — **90/100**
> `hexgate/adapters/langchain/messages.py:36`
> 
> `_TOOL_CALL_BLOCK_TYPES` exists to drop a raw content block when the same call is also re-emitted from the standardized `tool_calls`. The comment justifies leaving out `server_tool_use`/`mcp_tool_use` because "the provider ran those itself, so they never reach `tool_calls`" — but `custom_tool_call` is not in that category.
> 
> In `langchain_openai/chat_models/base.py`, the `custom_tool_call` branch (4673–4679) is structurally identical to the `function_call` branch 27 lines above it (4646–4663): both append a raw content block **and** a `tool_calls` entry keyed on `output.call_id`.
> 
> ```python
> elif output.type == "custom_tool_call":
>     content_blocks.append(output.model_dump(exclude_none=True, mode="json"))
>     tool_call = {"type": "tool_call", "name": output.name,
>                  "args": {"__arg1": output.input}, "id": output.call_id}
>     tool_calls.append(tool_call)
> ```
> 
> `function_call` is in the frozenset; `custom_tool_call` isn't. Run through the PR's converter:
> 
> ```
> custom_tool_call: 2 part(s) -> ['custom_tool_call', 'tool_call']
> function_call   : 1 part(s) -> ['tool_call']
> ```
> 
> One call, two parts, two different id fields — a reader counts the call twice, and on the completion side both copies eat the same 8 KiB budget. Reachable through `langchain_openai.tools.custom_tool` on the Responses API.
> 
> **Fix** — the block already carries `call_id`, so the existing drop condition at `messages.py:104` matches as soon as the type is in the set:
> 
> ```diff
>  _TOOL_CALL_BLOCK_TYPES = frozenset(
> -    {"function_call", "tool_use", "tool_call", "invalid_tool_call"}
> +    {"function_call", "custom_tool_call", "tool_use", "tool_call", "invalid_tool_call"}
>  )
> ```
> 
> ### 2. An empty stash still advances the cursor, burning seq 0 and losing the system-prompt lift for the whole run — **50/100**
> `hexgate/adapters/langchain/usage.py:212`
> 
> When `prompt` is `None`, or carries `messages == []`, `_emit_messages` still calls `self._cursor.advance(turn_key, [])`. `MessageCursor._diff` returns `_TurnState(0, _UNMATCHABLE_FINGERPRINT, 0)` for an empty list (`tracing/messages.py:468`), so the mark is clobbered to `"?"`, and `advance` bumps `next_seq` regardless (`:455`). The stored `"?"` can never equal the `"0:"` prefix the next call computes, so the first _real_ prompt resyncs at seq 1 — and because the system-instructions lift is gated on `delta.seq == 0` (`usage.py:215`), that run's system prompt never reaches `system_instructions`; it stays inline in `input_messages` for the rest of the run.
> 
> Reproduced against the real cursor:
> 
> ```
> empty advance -> seq 0  resynced False  msgs []
> first real    -> seq 1  resynced True   n 2     # <- system prompt never lifted
> second real   -> seq 2  resynced False  n 2
> baseline (no empty advance): first real -> seq 0  resynced False
> ```
> 
> Scored 50 rather than higher because the trigger is narrow: with `HEXGATE_LOG_MESSAGES=0` at both ends, `_emit_messages` returns at `:201` and never advances, so it takes a genuine mid-call flip of that env var between `on_chat_model_start` and `on_llm_end` (a flip the module explicitly supports), or a start callback that failed and left no stash.
> 
> **Fix** — skip the call entirely rather than spending its seq:
> 
> ```diff
>          turn_key, messages, _ = prompt or _Prompt(self.turn_key(run_id), [], "")
> +        # Nothing was stashed for this call, so there is no delta to measure.
> +        # Advancing on an empty list would clobber the mark and spend the
> +        # turn's seq, and seq 0 is the only event that lifts the system prompt.
> +        if not messages:
> +            return
> ```
> 
> This drops the completion for that one call too. If that matters more than the seq, the alternative is to leave `advance` alone and give the cursor a way to say "nothing observed" without spending a seq — but that means changing the `MessageDelta` contract ("the caller must emit an event for every delta it asks for"), so the early return is the cheaper trade.
> 
> **Worth noting:** this is not new to this PR. `adapters/openai/usage.py:206` pops with `_Prompt(None, [])` and feeds the result straight into `advance` at `:247`, with the same `seq == 0` gate on `system_instructions` at `:226` — same bug, same consequence. Fixing only the LangChain side leaves the other half.

Both fixed: custom_tool_call added to the frozenset in bbf4bd01 (it duplicated on the input side too, not just the completion), and the empty-stash advance in ebc526c3 here plus #229, which fixes it in MessageCursor itself so the OpenAI half you flagged is covered and PRs 12–13 can't reintroduce it.

guillaume-hexamind (approved):

PR #194, opened 2026-09-08
feat(sdk): capture Google ADK messages in HexgateUsagePlugin

victorludvig:
## What is Changing

PR 12/15 of the [LLM message logging implementation spec](https://www.notion.so/3d4fb45dbae28109b82fd38cb03b9d11). `HexgateUsagePlugin` gains the request-side `before_model_callback`: it stashes `LlmRequest.contents` and the system instruction by `invocation_id`, and `after_model_callback` emits token usage plus one `LlmMessageEvent` from the pair. `MessageCursor` turns ADK's rebuilt-every-call `contents` into the delta; `after_run_callback` resets it. New `adapters/google/messages.py` converts `Content`/`Part` to the GenAI `role`/`parts` shape, split out as in the OpenAI adapter.

Without it a denied Google ADK tool call has no transcript beside it — the fourth audit stream stops at the OpenAI adapter.

`invocation_id` is the `turn_key`: ADK rebuilds `contents` from the session on every call, so one invocation is one message list, and an `AgentTool`'s nested `Runner` gets an id of its own.

Three things the review changed:

- **Model name.** In SSE mode — which `hexgate serve` always uses — the only non-partial response is ADK's aggregate from `StreamingResponseAggregator.close()`, and it carries no `model_version`. The platform rejects both events on an empty `model` (`schemas.py`, `min_length=1`), so the stashed request model is the fallback.
- **Token double-count.** Every SSE chunk repeats the call's running counts, so usage now emits only on the completed response — one answer was being billed once per chunk.
- **Stash bound.** ADK skips `after_run_callback` when the agent loop raises or the caller stops iterating (a policy deny, an SSE disconnect), and this plugin lives as long as the runner, so `_pending` is LRU-capped like `_HexgateReachPlugin`'s map.

## Tests

`make check` green — 2718 passed.

**Unit tests:** `test_messages.py` (part/role conversion, tool calls and results, thoughts, system-instruction union), `test_usage.py` (delta across calls, interleaved invocations, resync, partial chunks, model fallback, usage counted once, stash bound, opt-out, conversion failure).

**Integration tests:** `test_integration.py` reads the run's `llm_message` rows back from ClickHouse and concatenates them to the conversation the agent saw. Run against the full local pipeline (platform-api, Collector, Redpanda, enricher): 7 passed, 1 skipped (needs a dashboard login).


guillaume-hexamind:
## Review: `HexgateUsagePlugin` message capture

Four findings, verified against the installed `google-adk` 1.32.0 and this branch. Scores are confidence that the finding is a real defect worth fixing. Suggested fix order is at the bottom.

---

### 1. The stash cap does not free the cursor for the runs it exists for — **80/100**

`hexgate/adapters/google/usage.py:141`

`after_model_callback` pops `_pending[key]` on the first completed response, so a run that ends abnormally *later* — a policy deny or an SSE client disconnect, the two cases the `_MAX_PENDING` comment names — has no pending entry left. `run_after_run_callback` is a plain statement after the event loop in `_exec_with_plugin` (`runners.py:963`), not in a `finally`, so it is skipped; and the LRU path (`usage.py:110`) only resets cursors for keys still in `_pending`. The `MessageCursor._turns` entry is therefore retained for the life of the plugin, which is the life of `HexgateRunner` / `GoogleServeDriver` (`streaming.py:139`).

`_HexgateReachPlugin` already handles exactly this shape — `before_tool_callback` has `except Exception: self._depth.pop(...); raise` with the comment "A raise here aborts the run, so after_run_callback never fires" (`runner.py:98-105`). The usage plugin has no equivalent for its cursor.

The leak is small (one `_TurnState` plus the key, a few hundred bytes — the prompt itself *is* correctly bounded), so this is low severity rather than a memory risk. But it is unbounded in a long-lived `serve` process, and `test_when_runs_end_abnormally_then_stashed_prompts_are_bounded` only ever calls `before_model_callback`, so it covers the half that already works.

**Fix:** keep the key and drop only the payload, so one cap bounds both halves — this also composes with finding 4's guard:

```python
prompt = self._pending.get(key) or _NO_PROMPT
if completed:
    self._pending[key] = _NO_PROMPT  # keep the key so the LRU can reach the cursor
```

`after_run_callback` still pops it on a normal end; the LRU now reaches abnormally-ended runs and resets both halves.

---

### 2. `invocation_id` alone is not unique per concurrent model call — **70/100**

`hexgate/adapters/google/usage.py:98`

`ParallelAgent._create_branch_ctx_for_sub_agent` (`parallel_agent.py:34-48`) does `invocation_context.model_copy()` and mutates only `branch` — `invocation_id` carries over — and `_merge_agent_run` runs the sub-agents as concurrent `TaskGroup` tasks (`parallel_agent.py:71-73`). `callback_context.invocation_id` resolves straight to it (`readonly_context.py:45-47`).

So sub-agent A stashes under `inv`, B overwrites it while A awaits the model, and A's `pop(inv)` returns **B's** prompt — A's `llm_message` row carries B's conversation. B's own response then finds no stash: its message event is silently dropped, and its usage event goes out with `model = llm_response.model_version or ""`, which in SSE mode is `""` because `close()` builds the aggregate without a `model_version` — rejected by `LlmInvocationEvent.model = Field(min_length=1)` (`platform/api/hexgate_api/schemas.py:670`), so the row is lost entirely. The shared cursor also interleaves two unrelated lists, flagging every event `resynced`.

Latent rather than live: hexgate never builds a `ParallelAgent` itself, so this only bites a caller who hands one to `HexgateRunner`. But it fails silently when it does. (`AgentTool` is fine — it builds its own `Runner` (`agent_tool.py:230`), so a fresh id, as the module docstring says.)

**Fix:** key the stash and the cursor per branch, e.g. `key = f"{callback_context.invocation_id}:{callback_context.agent_name}"`. That is consistent with `MessageCursor`'s own contract ("the framework's own identity for *one message list*" — sub-agents each keep their own list). Note `after_run_callback` then has to drop every key carrying that invocation prefix, not just the bare id.

---

### 3. A contentless streamed completion records no usage at all — **65/100**

`hexgate/adapters/google/usage.py:133`

In progressive mode `process_response` sets `partial = True` on every chunk and returns (`streaming_utils.py:308-310`), so the only non-partial response is `close()`'s aggregate — which returns `None` when `_parts_sequence` is empty (`streaming_utils.py:378`). `GoogleServeDriver` always runs `RunConfig(streaming_mode=StreamingMode.SSE)` (`streaming.py:215`) → `stream=True` (`base_llm_flow.py:1233`). A turn that produces zero content parts (prompt blocked, safety block with no output, empty candidates) therefore emits neither usage nor a transcript row, and orphans its stashed prompt. Google still bills the prompt tokens, and pre-PR they were recorded.

Narrow by construction — any text, any function call, or a mid-stream block after some output leaves `_parts_sequence` non-empty and the aggregate is produced. And the pre-PR behaviour wasn't right either (it emitted per chunk), so this is a coverage regression in the course of fixing a real bug rather than a straight one.

**Fix (optional, pick the cheap half):** stash the last-seen `usage_metadata` alongside the prompt and flush it from `after_run_callback` if the invocation never saw a completed response. If that state isn't judged worth it, a comment at the `completed` check recording that a contentless SSE turn is deliberately unbilled would stop the next reader re-deriving this.

---

### 4. Usage is not stash-guarded, so it can still double-count — **75/100**

`hexgate/adapters/google/usage.py:133`

`PROGRESSIVE_SSE_STREAMING` is `EXPERIMENTAL, default_on=True` (`_feature_registry.py:143`), but it is togglable via `ADK_ENABLE_PROGRESSIVE_SSE_STREAMING` / `adk --disable_features=...`. On the non-progressive path, `StreamingResponseAggregator.process_response` yields the merged-text aggregate (`streaming_utils.py:336-341`, `partial` never set) **and then** `yield llm_response` for the same final chunk (also `partial` unset, carrying the same cumulative `usage_metadata`). `base_llm_flow._call_llm_async` runs `after_model_callback` for every yielded response (`base_llm_flow.py:1245-1264`), so both are `completed` and `emit_llm_usage` fires twice for one model call — with a valid `model_version` on the second, so it isn't rejected platform-side either. That is precisely the double-billing this PR set out to fix.

Messages survive it only because the second response finds `_NO_PROMPT` (which the PR's own `test_when_a_response_has_no_stashed_request_then_no_event_is_emitted` demonstrates). The asymmetry is the bug.

**Fix:** apply the same guard to usage, so correctness doesn't depend on an upstream feature flag:

```python
if completed and usage is not None and prompt is not _NO_PROMPT:
    emit_llm_usage(...)
```

---

**Suggested order:** 4 first (one condition, restores the PR's own stated goal under a flag flip), then 2 (keying per branch costs nothing and it fails silently today), then 1 as a small follow-up. 3 is a judgement call.

guillaume-hexamind (commented):
A few comments to check, will approve after

victorludvig:
> ## Review: `HexgateUsagePlugin` message capture
> Four findings, verified against the installed `google-adk` 1.32.0 and this branch. Scores are confidence that the finding is a real defect worth fixing. Suggested fix order is at the bottom.
> 
> ### 1. The stash cap does not free the cursor for the runs it exists for — **80/100**
> `hexgate/adapters/google/usage.py:141`
> 
> `after_model_callback` pops `_pending[key]` on the first completed response, so a run that ends abnormally _later_ — a policy deny or an SSE client disconnect, the two cases the `_MAX_PENDING` comment names — has no pending entry left. `run_after_run_callback` is a plain statement after the event loop in `_exec_with_plugin` (`runners.py:963`), not in a `finally`, so it is skipped; and the LRU path (`usage.py:110`) only resets cursors for keys still in `_pending`. The `MessageCursor._turns` entry is therefore retained for the life of the plugin, which is the life of `HexgateRunner` / `GoogleServeDriver` (`streaming.py:139`).
> 
> `_HexgateReachPlugin` already handles exactly this shape — `before_tool_callback` has `except Exception: self._depth.pop(...); raise` with the comment "A raise here aborts the run, so after_run_callback never fires" (`runner.py:98-105`). The usage plugin has no equivalent for its cursor.
> 
> The leak is small (one `_TurnState` plus the key, a few hundred bytes — the prompt itself _is_ correctly bounded), so this is low severity rather than a memory risk. But it is unbounded in a long-lived `serve` process, and `test_when_runs_end_abnormally_then_stashed_prompts_are_bounded` only ever calls `before_model_callback`, so it covers the half that already works.
> 
> **Fix:** keep the key and drop only the payload, so one cap bounds both halves — this also composes with finding 4's guard:
> 
> ```python
> prompt = self._pending.get(key) or _NO_PROMPT
> if completed:
>     self._pending[key] = _NO_PROMPT  # keep the key so the LRU can reach the cursor
> ```
> 
> `after_run_callback` still pops it on a normal end; the LRU now reaches abnormally-ended runs and resets both halves.
> 
> ### 2. `invocation_id` alone is not unique per concurrent model call — **70/100**
> `hexgate/adapters/google/usage.py:98`
> 
> `ParallelAgent._create_branch_ctx_for_sub_agent` (`parallel_agent.py:34-48`) does `invocation_context.model_copy()` and mutates only `branch` — `invocation_id` carries over — and `_merge_agent_run` runs the sub-agents as concurrent `TaskGroup` tasks (`parallel_agent.py:71-73`). `callback_context.invocation_id` resolves straight to it (`readonly_context.py:45-47`).
> 
> So sub-agent A stashes under `inv`, B overwrites it while A awaits the model, and A's `pop(inv)` returns **B's** prompt — A's `llm_message` row carries B's conversation. B's own response then finds no stash: its message event is silently dropped, and its usage event goes out with `model = llm_response.model_version or ""`, which in SSE mode is `""` because `close()` builds the aggregate without a `model_version` — rejected by `LlmInvocationEvent.model = Field(min_length=1)` (`platform/api/hexgate_api/schemas.py:670`), so the row is lost entirely. The shared cursor also interleaves two unrelated lists, flagging every event `resynced`.
> 
> Latent rather than live: hexgate never builds a `ParallelAgent` itself, so this only bites a caller who hands one to `HexgateRunner`. But it fails silently when it does. (`AgentTool` is fine — it builds its own `Runner` (`agent_tool.py:230`), so a fresh id, as the module docstring says.)
> 
> **Fix:** key the stash and the cursor per branch, e.g. `key = f"{callback_context.invocation_id}:{callback_context.agent_name}"`. That is consistent with `MessageCursor`'s own contract ("the framework's own identity for _one message list_" — sub-agents each keep their own list). Note `after_run_callback` then has to drop every key carrying that invocation prefix, not just the bare id.
> 
> ### 3. A contentless streamed completion records no usage at all — **65/100**
> `hexgate/adapters/google/usage.py:133`
> 
> In progressive mode `process_response` sets `partial = True` on every chunk and returns (`streaming_utils.py:308-310`), so the only non-partial response is `close()`'s aggregate — which returns `None` when `_parts_sequence` is empty (`streaming_utils.py:378`). `GoogleServeDriver` always runs `RunConfig(streaming_mode=StreamingMode.SSE)` (`streaming.py:215`) → `stream=True` (`base_llm_flow.py:1233`). A turn that produces zero content parts (prompt blocked, safety block with no output, empty candidates) therefore emits neither usage nor a transcript row, and orphans its stashed prompt. Google still bills the prompt tokens, and pre-PR they were recorded.
> 
> Narrow by construction — any text, any function call, or a mid-stream block after some output leaves `_parts_sequence` non-empty and the aggregate is produced. And the pre-PR behaviour wasn't right either (it emitted per chunk), so this is a coverage regression in the course of fixing a real bug rather than a straight one.
> 
> **Fix (optional, pick the cheap half):** stash the last-seen `usage_metadata` alongside the prompt and flush it from `after_run_callback` if the invocation never saw a completed response. If that state isn't judged worth it, a comment at the `completed` check recording that a contentless SSE turn is deliberately unbilled would stop the next reader re-deriving this.
> 
> ### 4. Usage is not stash-guarded, so it can still double-count — **75/100**
> `hexgate/adapters/google/usage.py:133`
> 
> `PROGRESSIVE_SSE_STREAMING` is `EXPERIMENTAL, default_on=True` (`_feature_registry.py:143`), but it is togglable via `ADK_ENABLE_PROGRESSIVE_SSE_STREAMING` / `adk --disable_features=...`. On the non-progressive path, `StreamingResponseAggregator.process_response` yields the merged-text aggregate (`streaming_utils.py:336-341`, `partial` never set) **and then** `yield llm_response` for the same final chunk (also `partial` unset, carrying the same cumulative `usage_metadata`). `base_llm_flow._call_llm_async` runs `after_model_callback` for every yielded response (`base_llm_flow.py:1245-1264`), so both are `completed` and `emit_llm_usage` fires twice for one model call — with a valid `model_version` on the second, so it isn't rejected platform-side either. That is precisely the double-billing this PR set out to fix.
> 
> Messages survive it only because the second response finds `_NO_PROMPT` (which the PR's own `test_when_a_response_has_no_stashed_request_then_no_event_is_emitted` demonstrates). The asymmetry is the bug.
> 
> **Fix:** apply the same guard to usage, so correctness doesn't depend on an upstream feature flag:
> 
> ```python
> if completed and usage is not None and prompt is not _NO_PROMPT:
>     emit_llm_usage(...)
> ```
> 
> **Suggested order:** 4 first (one condition, restores the PR's own stated goal under a flag flip), then 2 (keying per branch costs nothing and it fails silently today), then 1 as a small follow-up. 3 is a judgement call.


All four reproduce — thanks, these were sharp. Fixed in e4d345ad.

- 1 — cursor leak. Confirmed: 2000 aborted runs left pending=0, cursor=2000. Took your tombstone: a completed call keeps its key with an empty marker, so the LRU still reaches the cursor. Invariant is now every cursor entry has a _pending entry — one cap bounds both. The old test only drove before_model_callback; the new one drives whole calls and fails with the fix reverted.
- 2 — ParallelAgent. turn_key is now invocation_id:agent_name. I'd dropped this at review partly on "the fix needs a private attribute" — wrong, agent_name is public. It also gives a transfer_to_agent target its own list, which is what it has, so a handoff starts at seq 0 with its own instructions instead of resyncing the parent's — matching §II and the OpenAI adapter.
- 3 — contentless streamed turn. Real, not fixed. Your flush hangs off after_run_callback, which finding 1 establishes is exactly what an aborted run skips — unreliable by construction. Took your second option: a comment at the completed check.
- 4 — usage double-count. Fixed, same guard as messages. Verified against the real aggregator: 1 emit on non-streaming, progressive and non-progressive (was 2 on the last).

Also hardened the review skill so fixes made during a review re-enter the gate — all three defects were in code that review's own fixes wrote (#205).

guillaume-hexamind (approved):

PR #195, opened 2026-09-08
feat(dashboard): show LLM messages in the Audit decision drawer

victorludvig:

## What is changing
`lib/api.ts`: `LlmMessageRow` + `listLlmMessages(projectId, sessionId)`. `routes/Audit.tsx`: an "LLM messages" `DrawerSection` between *Arguments* and *Envelope*, anchored on the selected decision — earlier turns collapsed, the anchor turn (latest message event before the decision) highlighted, the following turn expanded — with "transcript incomplete" / "history restated" / "truncated" markers. No new route.

The feature is decision context, not a trace explorer: an auditor opens a denied call and sees the prompt that produced it and the output that followed.

## Tests
- `Audit.test.tsx` with mocked rows covering a gap, a resynced row and a truncated field; `api.test.ts`. Acceptance: open the PR 7 smoke session on staging.

- manual tests:

https://github.com/user-attachments/assets/118fdb22-a86e-4e8b-93cd-4cf260859be2


platform/dashboard/src/routes/Audit.tsx
```diff
@@ -247,6 +309,18 @@ function DetailDrawer({
...
           </DrawerSection>
 
+          <DrawerSection label="Messages">
+            <LlmMessagesSection
+              rows={messagesQ.data?.rows ?? []}
+              total={messagesQ.data?.total ?? 0}
+              decisionOccurredAt={e.occurred_at}
+              offset={messagesQ.data?.offset ?? 0}
+              scoped={scoped}
+              isLoading={messagesQ.isPending && scoped}
```
quanghexa94: `isLoading = messagesQ.isPending && scoped`: a react-query query with `enabled: false` still reports status `pending` (`isPending === true`). If a decision is `scoped` but `projectId` is momentarily null — so `enabled: !!projectId && scoped` is false — this section shows "Loading…" with no request in flight and never resolves. In practice `projectId` is set once the audit table has rows, so it's a latent edge rather than a routine break, but consider also gating on `projectId` (or using `isFetching`) so a disabled query never presents as loading.

platform/dashboard/src/components/audit/messages.tsx
```diff
@@ -0,0 +1,464 @@
...
+  );
+  if (!scoped)
+    return note(
+      "No session or run id on this decision — nothing to scope a transcript to.",
+    );
+  if (isLoading) return note("Loading…");
+  if (isError) return note("Could not load the transcript.");
+  if (!rows.length) return note("No messages recorded for this session.");
+
+  const { earlier, anchor, next, later } = anchorTranscript(
```
quanghexa94: `anchorTranscript(rows, …)` is called directly in the render body, while its sibling `callSlots = useMemo(() => buildCallSlots(rows), [rows])` two lines up is memoized. Any drawer state change (e.g. toggling a `CollapsedTurns` section open/closed) re-runs the full transcript split over the whole page of rows each render, unlike `callSlots`. Cheap per call, but an unnecessary and inconsistent recompute — wrapping it in `useMemo(…, [rows, decisionOccurredAt])` would match the adjacent code.

platform/dashboard/src/components/audit/messages.tsx
```diff
@@ -0,0 +1,464 @@
...
+    </div>
+  );
+}
+
+function Part({ part }: { part: LlmMessagePart }) {
+  if (part == null || typeof part !== "object")
+    return <Pre>{String(part)}</Pre>;
+  const type = part.type;
+  if (type === "text") {
+    const content = part.content;
```
quanghexa94, line 128: For `type === "text"` the renderer hard-codes `part.content`. If any adapter emits the text under a different key (e.g. `text`), `content` is `undefined`, `JSON.stringify(undefined)` returns `undefined`, and React renders nothing — a blank `Pre` with no sign that prose was dropped. Unknown *types* get printed whole by the fallback at the bottom, but a known type with an unexpected field name does not. The tests only exercise `content`, so this shape mismatch would pass CI yet blank a real transcript. Couldn't verify against the actual gen_ai adapter part shape (not in this diff) — flagging as fragility: consider falling back to stringifying the whole part when `content` is undefined.

victorludvig:
Thanks @quanghexa94 — all three fixed in `38714b4d`.

1. `isLoading` now repeats every clause of `enabled`. Latent as you say, but the two should agree by construction — a disabled query stays `pending`, so any clause missing there is a permanent spinner.
2. `anchorTranscript` memoized and hoisted beside `buildCallSlots`. Fair catch on the inconsistency.
3. Fixed, though the premise is off: `content` *is* the key — spec `TextPart`, and `adapters/_messages.py:33` for every adapter. No adapter uses a different one. The real exposure is a row written by a different SDK version, since these are read across a 180-day TTL. Either way the silent blank was wrong, so a `text` part with no `content` now falls through to the whole-part printer, like everything else this file doesn't recognise. Test added.

platform/dashboard/src/routes/Audit.tsx
quanghexa94: Confirmed fixed in `3aa3b32` — `isLoading` now mirrors `enabled` (`&& !!projectId`). Thanks!

platform/dashboard/src/components/audit/messages.tsx
quanghexa94: Confirmed fixed in `3aa3b32` — now memoized on `[rows, decisionOccurredAt]` and hoisted above the early returns, matching `callSlots`. Thanks!

platform/dashboard/src/components/audit/messages.tsx
quanghexa94: Confirmed fixed in `3aa3b32` — a `text` part with no `content` now falls through to the whole-part printer instead of rendering blank. Thanks!

quanghexa94 (approved):
All three review comments addressed in `3aa3b32`, each fix matching what was suggested:
- `isLoading` now mirrors `enabled` (`&& !!projectId && scoped`), so a disabled query no longer shows "Loading…" forever.
- `anchorTranscript` is memoized on `[rows, decisionOccurredAt]` and hoisted, matching `callSlots`.
- A `text` part with no `content` falls through to the whole-part printer instead of rendering blank.

Core logic (`anchorTranscript` anchor/gap/duplicate handling, `buildCallSlots`) is correct and well covered by the new unit tests. LGTM 🚀

PR #196, opened 2026-09-08
feat(sdk): emit a per-run message event for Pydantic AI

victorludvig:
PR 13/15 of the LLM message logging stack ([spec](https://app.notion.com/p/3d4fb45dbae28109b82fd38cb03b9d11) · [design](https://app.notion.com/p/3d4fb45dbae28141a3c4d32338a0d496)). Rebased onto `main` after #193.

**This is the one adapter with no per-call hook.** OpenAI, LangChain and Google ADK all pair a request hook with a response hook and ship the delta between them. Pydantic AI has nothing short of `iter()`, so a run is the unit: one event, `message_seq=0`, no `MessageCursor`. Everything below follows from that, and it is where the review effort is worth spending.

| | per-call adapters (#191, #193) | here |
|---|---|---|
| unit | one LLM call | one agent run |
| delta | `MessageCursor` across calls | none — the run *is* the delta |
| `turn_key` | run id + agent name | the run id |
| completion | the response hook's argument | the last `ModelResponse` in the history |
| thinking | kept on input, dropped from output | dropped from both |

## What is Changing

- **`adapters/pydantic_ai/usage.py`** — `emit_run_messages()` beside `emit_run_usage()`, plus `_resolve_model()` shared by both so the `llm_usage` and `llm_message` rows cannot disagree about which model answered.
- **`adapters/pydantic_ai/messages.py`** (new) — `ModelMessage` → the OTel GenAI `role`/`parts` shape, following #191's converters-split-from-hooks layout. Reuses `adapters/_messages.py`'s `text_part`, hoisted by #193.
- **`adapters/pydantic_ai/agent.py`** — `_emit_run_events()` at the four run sites, inside the bound scope.

LLM prompts and completions are the stack's fourth audit stream, beside decisions, token usage and bans. Until this lands a Pydantic AI agent produces `policy_decision` rows with no record of what the model was asked or what it answered — and the transcript is the only place a tool's *return value* is stored at all, since a decision row records that a tool was called, never what came back.

## Five decisions worth pushing back on

- **`new_messages()`, not `all_messages()` — a deliberate departure from the spec's literal wording.** §II names `all_messages()`, but that replays whatever the caller passed as `message_history`: an N-turn chat stores turn 1 N times, O(n²) bytes, and the 256 KiB input cap then cuts the *newest* exchange first — measured at 86% of a 4 KiB tool result lost by turn 150. It also files one agent's messages inside another's row after a handoff. `new_messages()` serves the same section's actual rule, which `tracing/semconv.py` states verbatim: *never a snapshot of the whole conversation*. A session's rows still concatenate back to the conversation. **The spec doc needs a line either way this lands.**
- **The completion is the last `ModelResponse`, not the last message.** An `output_type=` agent — Pydantic AI's headline feature — delivers its answer through a `final_result` tool call, so its history ends on a `ModelRequest`. Taking `history[-1]` stored an empty completion for *every* structured-output run. Pydantic AI's own `AgentRunResult.response` scans backwards for the same reason.
- **`system_instructions` takes the *last* request's instructions.** They are re-rendered per request, not restated, so a dynamic `@agent.instructions` returns a different string each step — and first-wins reported a *previous* agent's system prompt after a handoff (verified: a billing agent's row claiming "You are TRIAGE").
- **`ThinkingPart` is dropped from the input side too**, which `openai/messages.py` says the adapters should agree on once #221 lands. The argument for diverging: there the input is one call's delta under a 256 KiB budget, here it is a whole run, so every past response's reasoning rides in it. **#221 needs a line recording that this adapter answered differently.**
- **Only usage is gated on run completion.** That guard (#76) exists because Pydantic AI reports 0 tokens until a run ends — a reason that does not transfer to prompts. An aborted stream still holds the prompt that was sent, and a run someone cut short is the one an incident review asks about. The emit is in a `finally`: as a trailing statement it never ran when the caller's body raised, which is how a stream is actually aborted. The transcript of an unfinished run records no completion — its last response is a mid-run tool call, not an answer.

Two smaller ones: `_resolve_model` falls through `AgentRun.result` (an `AgentRun` has no `.response` of its own, so an `iter` run overriding the model reported the configured one — a pre-existing usage bug this would otherwise have copied into a second column); and inline `BinaryContent` is stored as a `{media_type, bytes}` descriptor, since serialised whole its bytes expand to an escaped repr that filled the entire input budget with something nobody can read.

**Known limitation:** `message_seq` is always 0 and each run is its own `turn_key`, so a dropped span leaves no gap for a reader to detect. Recoverable by joining on `run_id`, which every row carries.

## Tests

`make lint`, `make fmt-check` and the full SDK suite pass: **2769 passed, 17 skipped**. Coverage: `messages.py` and `usage.py` 100%, `agent.py` 98%.

- **`test_messages.py`** (new) — the structured-output shape, a run with no response, re-rendered instructions, thinking on both sides, parallel tool returns folded into one message, retries with and without a tool name, builtin tools, multimodal prompts, the binary descriptor, and tool arguments in all three forms.
- **`test_usage.py`** — the emit, the `turn_key`, the `HEXGATE_LOG_MESSAGES` opt-out, a run that produced nothing, a history that cannot be read, and usage/messages agreeing on the model.
- **`test_agent.py`** — all four run methods emit the transcript; an aborted stream records it while suppressing usage.
- **`test_integration.py`** (`pytest -m integration`) — a tool-calling run's row read back from ClickHouse over the full OTLP path. Not run here; needs the live pipeline.


guillaume-hexamind:
### `last_response()` records a mid-run tool call as the completion on an aborted `run_stream` — 70/100

**What happens**

`run_messages()` (`hexgate/adapters/pydantic_ai/messages.py:158`) asks `last_response()` for the run's completion, which scans `reversed(history)` for the first `ModelResponse` and returns it unconditionally. Nothing checks whether that response is the run's *answer* or just an intermediate tool call.

This PR deliberately ungates the transcript from run completion (`_emit_run_events(result, usage=result.is_complete)`, `agent.py:154`), so incomplete runs now reach that code. On a multi-step run whose history ends on a `ModelRequest` carrying an ordinary tool return, the row comes out inverted.

Reproduced against real `pydantic_ai` 1.89.1 with `TestModel` and a `get_weather` tool, exiting the `run_stream` context without calling `get_output()`:

```
history:  ModelRequest[system-prompt, user-prompt]
          ModelResponse[tool-call]
          ModelRequest[tool-return]

INPUT:    user:  text 'weather in Paris?'
          tool:  tool_call_response(id=…get_weather, response='sunny in a')   ← its tool_call was hoisted out
OUTPUT:   assistant: tool_call(id=…get_weather)                               ← not a completion
```

The same run driven to completion is correct — the tool call stays in the input and the text answer is the output. So `output_messages` claims the model answered with a tool call, and the input carries a tool return whose matching call is no longer beside it (it is still recoverable from `output_messages`, but no longer where a reader joining calls to returns would look).

This lands squarely on the case the PR argues is worth logging — "a run someone cut short is the one an incident review asks about".

**Scope**

- `run_stream` aborted before `get_output()` on a multi-step run: reproduces every time.
- `iter`: does **not** reproduce. Driving `iter` to a break at 2/3/4/5 nodes never leaves the tool-return `ModelRequest` last (history jumps from 2 messages to 4), so `iter` aborts yield `input=[user]`, `output=[tool_call]` — unflattering, not corrupt.
- `run` / `run_sync`: unaffected, they only emit on success.

**Why the tests don't catch it**

`test_run_stream_still_records_the_transcript_when_the_caller_aborts` (`tests/adapters/pydantic_ai/test_agent.py:704`) drives `_FakeResult`, whose `new_messages()` returns a hard-coded clean `[ModelRequest, ModelResponse]`. It asserts that an emit happens, never what is in it. None of the 18 cases in `test_messages.py` uses a history ending on a real tool-return request.

**Suggested fix**

Hoist the last `ModelResponse` only when it really is the completion. The proxy already knows whether the run finished — `_emit_run_events` computes it for the `usage=` gate — so thread the same signal through:

```python
# agent.py
def _emit_run_events(self, result: Any, *, usage: bool = True) -> None:
    emit_run_messages(..., result, completed=usage, api_key=self._api_key)

# messages.py
def run_messages(
    history: list[ModelMessage], *, completed: bool = True
) -> tuple[...]:
    completion = last_response(history) if completed else None
```

An incomplete run then keeps every message in `input_messages`, in the order it happened, and emits an empty completion — which is the truthful record of a run that never produced one. A completed `output_type=` run still hoists correctly, so the `final_result` behaviour the module docstring describes is preserved.

The alternative — inspecting the history, and hoisting only when nothing but output-tool returns follows the last `ModelResponse` — works too, but `run_messages()` receives only the history and the output tool name is configurable (it is not always `final_result`), so it would need that name passed in as well. The completion flag is cheaper and uses a signal already at hand.

Worth adding a `test_messages.py` case for the aborted shape either way, since the fake in `test_agent.py` can't cover it.

guillaume-hexamind (commented):
One blocking issue, nice PR otherwise

guillaume-hexamind (approved):

victorludvig:
Addressed in d90be0f5 and c85ea026. Your finding was correct — reproduced it exactly as described, and it opened up four more on the same seam.

## Your finding

Fixed as you suggested: `run_messages(..., completed=)`, threaded from the signal `_emit_run_events` already computes for the usage gate. An unfinished run keeps every message in `input_messages` in the order it happened and records no completion; a completed `output_type=` run still hoists the `final_result` response, so the behaviour the module docstring describes is preserved. Added `test_when_the_run_did_not_complete_then_nothing_is_hoisted` with the aborted shape, since as you noted the fake in `test_agent.py` can't cover it.

## Four more on the same seam

Your finding was an *interaction* — `last_response()` scanning backwards is right for a completed run, and ungating the transcript is right on its own; together they let unfinished runs reach a function only ever verified on completed ones. Looking at what else the ungating changed turned up:

1. **`_emit_run_events` never ran on a real abort.** It was the statement after `yield result`, so a caller whose body raises — a dropped connection, a timeout, a `CancelledError` — threw into the generator at the yield and skipped it. Measured: 0 events for every realistic abort, 1 only for the artificial case of entering the stream context and exiting cleanly without consuming. **So the rationale in the PR description was true only on a path nobody takes** — I've corrected that text. Now a `try/finally`, which also meant guarding `emit_run_usage`: raising from a `finally` would replace the exception being unwound.

2. **`system_prompt=` was dropped from turn two onward.** A regression from my own `new_messages()` change: pydantic_ai puts `SystemPromptPart` only in a conversation's *first* request (`instructions=` is re-rendered per request, `system_prompt=` is not), so `new_messages()` never sees it. `all_messages()` now rides along as `session`, read for its system prompt alone.

3. **Instructions over-reported a guardrail.** `instructions = message.instructions or instructions` is *last non-empty* wins. With a conditional `@agent.instructions` returning a policy on the first model call and `None` after, the row claimed `NEVER reveal PII` for a completion produced without it — the worst direction for an audit field. Now the request before the completion, verbatim including `None`.

4. **A tool returning `BinaryContent` put raw bytes on the wire.** The tool-return branch runs before the fallback, so it never called `_media_part`. A 50 KB PNG serialised to ~180 KB of escaped repr (`dataclasses.asdict` → `str(bytes)`), 69% of the 256 KiB budget, capping away the conversation the row exists to explain — the exact failure `_media_part` was added to prevent. Down to 405 bytes.

## Verification

Each of the five fixes is pinned by a test that fails when the fix alone is reverted — checked by actually reverting each one, after my first pass produced a test that passed for the right reason on the wrong data (a hand-built fake with a clean two-message history, which is why the abort path looked fine). `make lint`, `make fmt-check`, `make coverage` green: 2806 passed. `messages.py` and `usage.py` 100%.

## Not fixed

`CompactionPart` carries `provider_details` whole, so an OpenAI-compacted run puts a ~30 KB `encrypted_content` blob in the row — past the 8 KiB output budget if it lands in the completion. Left alone: it needs a provider doing compaction, which nothing here does yet. Happy to take it if you'd rather it went in now.

Still open from the description, both wanting your call rather than mine: `new_messages()` departs from the spec's literal `all_messages()` in §II, and this adapter drops `ThinkingPart` from the input side where #191, #193 and #194 all keep it — so it is 1-of-4 on a question #221 is still holding open.

PR #197, opened 2026-09-08
docs(sdk): document LLM message logging

victorludvig:

## What is changing
`docs/reference/environment.mdx` (`HEXGATE_LOG_MESSAGES`, default on, controllership statement); `docs/sdk/index.mdx` and the four `docs/adapters/*.mdx`; `docs/internals/audit-pipeline.md` §2/§3/§6 plus the §9 gap-rate open item; `platform/DEPLOY.md` §4 smoke note.

Capture is on by default, so the opt-out and the data-controllership statement must be in the docs before the SDK release that follows this PR.

## Tests
Docs build; links checked.


guillaume-hexamind:
## Review: doc claims that contradict the code on this branch

This PR is documentation plus comment-only edits — no executable code paths change, so there is no runtime-regression risk in merging. Every finding below is a doc statement checked against the code on `vl/docs/message_logging` (head `310b5372`). Scores are severity out of 100.

---

### 1. Stale Google ADK `turn_key` in the §3.5 table — 80/100

`docs/internals/audit-pipeline.md:367` gives the ADK `turn_key` as `callback_context.invocation_id`. `hexgate/adapters/google/usage.py:71` returns `f"{invocation_id}:{agent_name}"`, and that module's own docstring says "`invocation_id` plus the agent name". The second commit of this PR corrected exactly this claim in `docs/adapters/google-adk.mdx` and missed the table. It also contradicts line 167 of the same file, whose example value is `"0b9c…:devops_agent"`.

Effect: someone debugging a `ParallelAgent` expects one key per invocation and cannot explain why two sub-agents produce two transcripts — the precise case the agent-name suffix was added to fix.

**Suggested fix:** in the table row, replace the cell with `callback_context.invocation_id` *plus* the agent name, noting that a `ParallelAgent`'s sub-agents each get a list of their own.

---

### 2. The "concatenate rows in time order" rule still misses handoffs — 65/100

`docs/concepts/audit-trail.mdx:34` gives the reader one exception to concatenation: `resynced`. There is a second. A handoff target starts a **new** `turn_key` at `message_seq` 0, unflagged, restating the conversation the source agent already logged — `docs/adapters/openai.mdx:135-139` says so directly ("It is handed the same conversation, so its first event restates what the source agent already logged", #215), and `docs/adapters/google-adk.mdx:141-145` says the ADK target begins at seq 0 "rather than resyncing the parent's", carrying the narrated rewrite of the parent's turns.

Following this page's rule on any run with a handoff silently duplicates the history — duplicated evidence that reads like a real exchange. `docs/internals/audit-pipeline.md:179-183` repeats the same incomplete rule.

**Suggested fix:** add a second exception naming handoffs, and tell the reader to group rows by `turn_key` before reading a session that contains one, rather than concatenating the session flat. Mirror it in the internals doc.

---

### 3. "Every hook is handed the *whole* input list" is false for OpenAI server-managed conversations — 60/100

`docs/internals/audit-pipeline.md:374` presents `MessageCursor` as the single delta mechanism. `HexgateUsageHooks._delta` (`hexgate/adapters/openai/usage.py:232-250`) bypasses the cursor entirely when `framework_sends_deltas` is set — under `conversation_id` / `previous_response_id` / `auto_previous_response_id` the SDK already sends only un-sent items, seq comes from a separate `_delta_seq` dict, and `resynced` is pinned false.

`docs/adapters/openai.mdx` documents this, but §3.5 is the reference for the wire contract and the cursor, so someone writing a fifth adapter or chasing a `sec_ai.message_seq` gap will look for cursor state that does not exist on those runs.

**Suggested fix:** add a short paragraph after the delta explanation naming the three kwargs, stating that the cursor is bypassed and seq counted separately, and that `resynced` stays false because the flag means a rewritten list, not a framework-supplied delta.

---

### 4. "Every adapter owes the cursor a `reset`… There is deliberately no LRU" — 50/100

Both halves are overstated as absolutes at `docs/internals/audit-pipeline.md:384`.

- The OpenAI adapter never calls `reset`. `hexgate/adapters/openai/usage.py:105-107` states that a fresh hooks instance per `run*` call "is how the cursor's 'reset on run end' contract is met".
- The ADK plugin *does* LRU-evict: `_MAX_PENDING = 256` (`hexgate/adapters/google/usage.py:85`), and line 118 calls `self._cursor.reset(stale)` on the evicted key — the very operation this paragraph says is deliberately never done. `docs/adapters/google-adk.mdx` says "LRU-capped" in this same PR.

The paragraph is accurate about the `MessageCursor` class; it is the framing as a rule over the adapters that is wrong.

**Suggested fix:** scope the sentence to the class — `MessageCursor` itself has no LRU, and each adapter retires its keys its own way (LangChain and ADK call `reset` on run end, OpenAI uses a per-run instance, ADK additionally LRU-caps its in-flight stash and resets the evicted key because ADK skips `after_run_callback` on a raise).

---

### 5. "records the conversation when a run completes" — 50/100

`docs/adapters/pydantic-ai.mdx:113`. `_emit_run_events` (`hexgate/adapters/pydantic_ai/agent.py:97`) calls `emit_run_messages` unconditionally and returns before `emit_run_usage` when `completed` is false — only usage is gated. Its docstring is explicit: "The transcript is still emitted — a run someone cut short is the one an incident review asks about." It runs from a `finally` on `run_stream` (`completed=result.is_complete`) and `iter` (`completed=run.result is not None`).

Recording aborted runs is the deliberate behaviour of the base branch's `fix(sdk): record the pydantic_ai transcript on the paths a run actually fails on`, and it is the case an incident review most wants. As written, someone investigating a cancelled stream concludes no transcript exists. (Note the boundary: `run` / `run_sync` emit after the await, not from a `finally`, so a raise on those two paths really does produce no row.)

**Suggested fix:** say the transcript is recorded at the end of a run *including one cut short*, since `run_stream` and `iter` emit from a `finally` — and mention the consequence, that an incomplete run's row carries an empty `gen_ai.output.messages` and no token counts, because its last response is a mid-run tool call rather than an answer.

guillaume-hexamind (commented):
Nice documentation details. You can check the different issues above before I approve

victorludvig:
All five fixed in e2035aea.

1. §3.5 ADK `turn_key` — now invocation id + agent name, with the ParallelAgent reason.
2. Concatenation rule — scoped per adapter: OpenAI handoff opens a key carrying the source's conversation, ADK one holding ADK's narrated rewrite, an unwrapped LangChain sub-graph shares the parent's key and resyncs against it, pydantic_ai mints a key per run whose rows *are* continuations. Also noted that grouping by `turn_key` doesn't resolve a resync, which lands inside the group.
3. Server-managed conversations — added the three kwargs, the cursor bypass, the separate seq counter and why `resynced` stays false.
4. `reset`/LRU — scoped to the class; each adapter's retirement spelled out.
5. pydantic_ai — records runs cut short; `run`/`run_sync` are the boundary.

Reviewing those fixes turned up three more, also fixed: "pydantic_ai reports 0 tokens until a run ends" is false (usage accumulates — measured), the empty-check guard doesn't fire from turn two onward because the system prompt alone populates it, and `system_prompt=` repeats on every row while dynamic instructions stay per-run.

guillaume-hexamind (approved):

PR #199, opened 2026-09-09
fix(sdk, collector): apply role-keyed top-level constraints, widen revocation staleness

guillaume-hexamind:
Two independent changes on one branch

## 1. `fix(sdk)` — a role-keyed policy's top-level `constraints:` was silently dropped

`load_policy_set_from_dict` validated only the values under `roles:`, so every other top-level key parsed and was discarded. An author wrote a role-wide fence, got no fence, and no error. The flat shape was never affected — there the whole payload validates as one `AgentPolicy` — so the same key meant two different things depending on whether the document declared roles.

- A file-level `constraints:` block now **unions** into every role. Union, not replace: a role dropping the file's fence would be fail-open, the same reason constraints union across `inherits`.
- Any **other** unrecognised sibling of `roles:` is now a `PolicySetError` naming the key. That closes the defect class rather than this one instance, and matches composed module policies, which are already `extra="forbid"` at every scope.
- The block's own shape is checked before the hoist — a bare string would otherwise splat into single characters, each a valid `str`.

Verified past the unit tests: the fence compiles into the **Rego** bundle per role, so this holds on the WASM engine and not only the pydantic path; and the resolve → build round-trip still loads, since `hexgate policy resolve` emits `{"roles": …, "_resolved": true}` for a multi-role result.

**Upgrade note** — anyone who wrote such a fence speculatively starts denying on upgrade, and the policy's compiled `source_hash` moves with it. Documented in `docs/policy/constraints.mdx`, alongside the note #155 wrote for the flat shape.

Found by a code review of #155 (already merged), then reproduced before being fixed.

## 2. `feat(collector)` — revocation `max_staleness` 2m → 1h, both knobs env-tunable

2m sat under the length of routine control-plane maintenance (minor-version upgrade, volume resize, slow restart), so a Postgres blip meant **total ingest rejection** — silently, because the deploy healthcheck is a TCP-connect probe that stays green while every request 401s, and refused spans never reach the Redpanda buffer that exists to absorb this.

- `poll_interval` stays 20s, so steady-state revocation latency is unchanged. The widened window applies only while refreshes are **failing**.
- Both knobs override via `HEXGATE_COLLECTOR_REVOCATION_POLL_INTERVAL` / `_MAX_STALENESS`, per stage, no image rebuild.
- The env default applies only when a variable is **unset**, so the deploy stack passes them with non-empty fallbacks: a blank value arrives as `0s` and the collector refuses to boot (verified against the built binary).

`collector-check` gained an accepted *and* a refused override case. The refusal is the load-bearing one — it can only fail if the env values actually reached the config struct, so a typo in either variable name fails the build instead of shipping a dead override.

## Testing

`make check-all` green: SDK 2570 passed, platform-api 638, dashboard 218, collector `ok` plus the three `validate` cases. The platform-api run matters here specifically — it compiles signed bundles through the same policy loader, so the new strict rejection could have broken bundle compilation. It didn't.

## Not in scope

Three further findings from the same review of #155, none addressed here:

- a named-export import (`caps.yaml#export`) drops the imported file's own top-level `boundary:` while a whole-file import of that same file raises — same fail-open class, reproduced;
- `agent.run` admission is audited outside `run_scope` on the LangChain / LangGraph / OpenAI paths but inside it on Google, so the row admitting a run has no `run_id` to join on;
- the CLI's constraint localization doesn't reach the file-level block, so a malformed expression there reports as a schema error rather than `default → <policy>`.

### DIFF
--- a/hexgate/security/models.py
+++ b/hexgate/security/models.py
@@ -166,7 +166,11 @@ class AgentPolicy(BaseModel):
     # instances, nothing reassigns a field), which is what makes memoizing
     # effective_tools safe. cached_property is a plain descriptor, not a field,
     # so pydantic must leave it alone.
-    model_config = ConfigDict(frozen=True, ignored_types=(cached_property,))
+    # extra="forbid": a mistyped field (``contraints:``) would otherwise be
+    # dropped in silence, and a dropped fence is fail-open.
+    model_config = ConfigDict(
+        frozen=True, ignored_types=(cached_property,), extra="forbid"
+    )
 
     version: int = 1
     inherits: list[str] = Field(default_factory=list)
--- a/hexgate/security/models.py
+++ b/hexgate/security/models.py
@@ -33,6 +33,12 @@ class BaseToolPolicy(BaseModel):
     milestone, these strings carry through verbatim.
     """
 
+    # extra="forbid" for the same reason AgentPolicy sets it: a mistyped field
+    # (``contraints:``) would otherwise be dropped in silence, leaving
+    # ``mode: allow`` with no fence at all. Inherited by FileToolPolicy and
+    # AgentTargetPolicy, so it covers every nested policy scope.
+    model_config = ConfigDict(extra="forbid")
+
     mode: PolicyMode = "deny"
     constraints: list[str] = Field(default_factory=list)
 
@@ -44,6 +50,8 @@ class BaseToolPolicy(BaseModel):
 
 class FileScope(BaseModel):
     """Restrict a file-oriented tool to explicit path patterns."""
+
+    model_config = ConfigDict(extra="forbid")
 
     allowed_paths: list[str] = Field(default_factory=list)
     denied_paths: list[str] = Field(default_factory=list)
--- a/Makefile
+++ b/Makefile
@@ -296,6 +296,12 @@ collector-check: ## Vet + test + build the collector, validate config.yaml (no o
 	cd platform/collector && go vet ./...
 	cd platform/collector && go build -o hexgate-collector ./...
 	cd platform/collector && ./hexgate-collector validate --config=config.yaml
+	# The two revocation knobs are env-overridable (see config.yaml). A valid
+	# pair must be accepted; an INVALID pair must be REFUSED — the refusal is
+	# what proves the env values reached the config struct, so a typo in either
+	# variable name fails here instead of shipping a dead override.
+	cd platform/collector && HEXGATE_COLLECTOR_REVOCATION_POLL_INTERVAL=5s HEXGATE_COLLECTOR_REVOCATION_MAX_STALENESS=30m ./hexgate-collector validate --config=config.yaml
+	cd platform/collector && (! HEXGATE_COLLECTOR_REVOCATION_POLL_INTERVAL=1m HEXGATE_COLLECTOR_REVOCATION_MAX_STALENESS=1s ./hexgate-collector validate --config=config.yaml >/dev/null 2>&1)
 
 # -------- Platform API (FastAPI control plane) --------
 #
--- a/Makefile
+++ b/Makefile
@@ -296,10 +296,8 @@ collector-check: ## Vet + test + build the collector, validate config.yaml (no o
 	cd platform/collector && go vet ./...
 	cd platform/collector && go build -o hexgate-collector ./...
 	cd platform/collector && ./hexgate-collector validate --config=config.yaml
-	# The two revocation knobs are env-overridable (see config.yaml). A valid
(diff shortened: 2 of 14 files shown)


hexgate/security/models.py
victorludvig, line 25: The PR rejects typos at the document level but not at the tool level, so this still loads silently with constraints=[] — an unlimited refund instead of a €500 cap:

```python
tools:
  refund_order:
    mode: allow
    contraints: ["args.amount <= 500"]   # typo: missing "s"
```

Fix: model_config = ConfigDict(extra="forbid") on BaseToolPolicy.

victorludvig (commented):
Hi @guillaume-hexamind, I approve the logic. One issue to fix regarding the BaseToolPolicy, and 3 optional cleanup to check:

1. compile_bundle swallows every exception → 200 OK with a nulled bundle. Highest leverage — it's what hides any policy error, including the ones this PR just made possible. Fix once, every future policy bug gets loud.
2. docs/internals/audit-pipeline.md:327,739 still say 2 min. This PR changes that exact constant. One-line fix, otherwise the doc is shipped wrong.
3. constraints.mdx:335 — source_hash doesn't move. It's sha256 of raw YAML bytes. The upgrade note tells operators to use it to find affected policies; they'll find nothing and conclude nothing changed. Point them at rego_hash/wasm_hash.

guillaume-hexamind:
Thanks @victorludvig — all four were valid. Addressed in 2295f58.

**1. `BaseToolPolicy` extra="forbid"** — fixed, and it was wider than the one case. `AgentPolicy` had the guard, nothing nested under it did, so a mistyped key was silently dropped at five scopes, each fail-open:

| Document | Was |
|---|---|
| `tools.<t>.contraints` | `mode: allow`, no cap at all |
| `tools.<t>.file_scope.alowed_paths` | empty `FileScope`, read as no path fence |
| `agents.<t>.vai` | `via` falls back to *both* transfer modes |
| `admission.contraints` | fence gone |
| `default_policy.contraints` | broadest fence gone |

`FileToolPolicy` and `AgentTargetPolicy` inherit from `BaseToolPolicy`, so your two-line fix on the base (plus `FileScope`) closes all five. One thing to flag: `ToolPolicy` is a union and the base arm now rejects `file_scope`, so there's a test pinning that a file tool still falls through to `FileToolPolicy` instead of failing the whole document.

**2. `compile_bundle` swallows everything** — you're right that this is what hides the rest, and the 200 is real: `PUT` → `compile_bundle` returns `None` → all three bundle columns nulled → committed → 200, enforcement silently on the pydantic fallback. Confirmed the dashboard doesn't cover for it either — Validate is its own button and Save never waits on it.

Fixed at the seam rather than in `compile_bundle`: the endpoint body is extracted to `_validate_policy_document`, and `PUT /agents/{name}` now runs the identical check `/validate` runs and returns **422** with its diagnostics. The two routes can no longer disagree about which documents load.

I deliberately left `compile_bundle` returning `None`: `recompile_project` uses it for its all-or-nothing fail-safe (R-POL-002) and `backfill_bundles` runs at boot, where raising would break startup. Your version — a distinct `PolicyCompileError`, caught in those two, propagated in the routes — is the better end state and catches the narrow band this misses (loads clean, fails at the opa stage). It also flips the contract `test_compile_bundle_returns_none_for_bad_policy` currently asserts, so I'd rather it be its own PR than ride along with a staleness default change. Happy to open it.

**3. `audit-pipeline.md`** — fixed at `:325` and `:609` (the mechanism description and the failure-mode table). Added the env-var names at `:325` too, since that's the operator-facing half.

**4. `source_hash` doesn't move** — correct, and this was the one actively sending operators the wrong way. It's `sha256(policy_yaml)`, computed before any parsing, so it's invariant across the change by construction. Verified against `main`:

| | `source_hash` | `rego_hash` | fence in Rego |
|---|---|---|---|
| `main` | `5dd21ad3…` | `3a41efaf…` | absent |
| this branch | `5dd21ad3…` | `06c44428…` | present |

Repointed at `rego_hash`/`wasm_hash`, with an explicit "`source_hash` will **not** move" so nobody repeats the diff. Note the error was pre-existing from #155 — this PR's new paragraph said "the same upgrade note applies" and inherited it.

**One extra**, found while wiring #2: `_validate_policy_document` called `parsed.get("roles")` with no mapping check, so a policy that's a bare list or a scalar 500'd. Pre-existing on `/validate`, but the save path routes through it now, so it returns a diagnostic instead. Test added.

`make check-all` green: SDK 2579, platform-api 645, dashboard 26 files, collector `ok` plus the three `validate` cases.

PR #200, opened 2026-09-09
feat(sdk): compose admission block (role→agent ingress access)

quanghexa94:
Adds `admission:` to the compose front-end — **agent-level ingress**: may this role *start/enter* this agent at all? It's the missing sibling of `reach:` (egress). The compose grammar (#175/#176) shipped tools/reach/mcp but not admission, even though the runtime (`AgentGate`, #124) already enforces it — so agent-level access couldn't be *authored* in compose. This closes that gap.

Off `main` (standalone) — the compose SDK it extends is merged, and it's independent of the open platform/dashboard PRs.

## What it does
- **`admission:` at any scope** — a single grant (it's about *this* agent), parallel to `reach`. In a role body it grants that role the right to run the agent; a top-level/agent boundary sets the ceiling.
- **Lowers to `agent.run`** — routed into `AgentPolicy.admission`, which the model already expands to the `agent.run` synthetic key. So the linker fold, closed-world admission, engine, rego, and signing paths are **unchanged** — this is pure front-end authoring.
- **Closed-world, like reach** — an ungranted role→agent is denied; a boundary that omits admission denies it even if a capability grants it; a boundary admission ceiling AND-s its constraint onto the grant.
- **Importable** — a capability file may carry `admission:` (leaf grant); import messages updated to list it.

```yaml
boundary:
  admission: { mode: allow }          # ceiling: admission is permitted at all
agents:
  billing_bot:
    roles:
      billing: { admission: { mode: allow } }   # only 'billing' may start billing_bot
      support: {}                                # closed-world deny
```

## Important files
- `hexgate/security/compose/grammar.py` — `admission` on `_GrantScope` + `BoundaryBlock`; `"admission"` reserved (block keyword).
- `hexgate/security/compose/lower.py` — lowers `admission:` into `AgentPolicy.admission` (→ `agent.run`), in both the grant and boundary paths.
- `hexgate/security/compose/imports.py` — import error messages list `admission`.
- `tests/security/compose/test_compose.py` — per-role gating, imported admission, boundary-ceiling constraint cap, closed-world deny, reserved-name rejection.

## Notes
- **No endpoint or runtime change.** `agent.run` already flows through resolve/check/preview generically; the platform graph endpoint already draws `role→agent` admission edges + role nodes; `AgentGate` already enforces it. So authoring admission lights up the graph and gates a served agent with nothing downstream to change.
- **Follow-up (dashboard):** the resolved view lists `agent.*` keys (reach today, admission now) as raw tool rows — worth grouping/labelling them as reach/admission in the editor's Resolved tab.
- 1314 SDK tests pass (compose + security + cli), ruff clean; review rounds fixed: stale docstring, weak reserved-name assertion, the `run.*`-at-admission caveat, and — folded in — the `_validate_run_refs` fail-open (it walked `tools` not `effective_tools`, so a `run.*` ref on an authored **admission or reach** grant skipped validation), now pinned by authored-path regression tests.

### DIFF
--- a/hexgate/security/compose/grammar.py
+++ b/hexgate/security/compose/grammar.py
@@ -10,7 +10,7 @@ Shape (scope by depth — a block keyword is a sibling of the ``agents``/``roles
 name-map, never a key inside it):
 
     version, import, export          # top-level only
-    boundary / tools / reach / mcp   # at any scope → applies to that scope
+    boundary / tools / reach / mcp / admission   # at any scope → that scope
     agents: { <name>: agent-body }   # top level only; body may add roles:
     roles:  { <name>: role-body }    # agent-body only
 
@@ -45,6 +45,7 @@ RESERVED_NAMES = frozenset(
         "tools",
         "reach",
         "mcp",
+        "admission",
         "agents",
         "roles",
     }
@@ -123,6 +124,10 @@ class BoundaryBlock(BaseModel):
 
     tools: dict[str, CeilingSpec] = Field(default_factory=dict)
     reach: dict[str, ReachSpec] = Field(default_factory=dict)
+    # The ceiling on ingress: may any role *start* this agent at all? Closed-world
+    # like reach — with no admission ceiling, an authored admission grant is
+    # intersected to deny, so a boundary must permit admission to allow it.
+    admission: CeilingSpec | None = None
 
 
 class _GrantScope(BaseModel):
@@ -138,6 +143,10 @@ class _GrantScope(BaseModel):
     tools: dict[str, GrantSpec] = Field(default_factory=dict)
     reach: dict[str, ReachSpec] = Field(default_factory=dict)
     mcp: dict[str, GrantSpec] = Field(default_factory=dict)
+    # Ingress grant: may this scope's role start/enter this agent? Lowers to the
+    # ``agent.run`` key the runtime AgentGate enforces. A single grant (it is about
+    # *this* agent), boundary-ceilinged like reach.
+    admission: GrantSpec | None = None
     imports: list[str] = Field(default_factory=list, alias="import")
 
     # Resolved leaf fragments spliced at this scope — populated by the import
--- a/hexgate/security/compose/grammar.py
+++ b/hexgate/security/compose/grammar.py
@@ -145,7 +145,10 @@ class _GrantScope(BaseModel):
     mcp: dict[str, GrantSpec] = Field(default_factory=dict)
     # Ingress grant: may this scope's role start/enter this agent? Lowers to the
     # ``agent.run`` key the runtime AgentGate enforces. A single grant (it is about
-    # *this* agent), boundary-ceilinged like reach.
+    # *this* agent), boundary-ceilinged like reach. Constraints here gate on
+    # identity/attributes, not ``run.*`` facts: admission fires at run entry,
+    # before a run accrues, so run counters read zero — a run budget belongs on a
+    # tool or reach, not admission.
     admission: GrantSpec | None = None
     imports: list[str] = Field(default_factory=list, alias="import")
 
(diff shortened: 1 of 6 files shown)


guillaume-hexamind:
### Test coverage gap: the new `run.*` validation fix isn't pinned by any test — **80/100**

**What's wrong**

Switching `_validate_run_refs` to `policy.effective_tools.values()` (`hexgate/security/policy_set.py:261`) is a real fix, but `test_run_ref_on_a_lowered_agent_key_is_validated` doesn't exercise it.

That test goes through `resolve_text`, and the compose path builds its result via `AgentPolicy.resolved(...)` (`linker.py:227`), which puts the folded `agent.run` key directly into `tools` and leaves the `admission` field `None`:

```
tools keys      : ['agent.run']
admission field : None
tools == effective_tools: True
```

With those two views identical on the compose path, the old `policy.tools.values()` walk already reached the key. Reverting line 261 alone: the new test still passes, and so does everything else (`2572 passed, 13 skipped`). Nothing in the repo fails without the change.

**What the fix actually closes** is the *authored* single-file path — an `AgentPolicy` fed straight to `PolicySet`, where `admission`/`agents` stay as model fields and never land in `tools`:

| authored shape carrying `run.bogus < 1` | before | after |
| --- | --- | --- |
| `admission: {...}` | accepted (fail-open) | rejected |
| `agents: {...}` (reach) | accepted (fail-open) | rejected |
| `tools: {...}` | rejected | rejected |

So this closes two fail-opens, not one — the reach half isn't called out in the PR description either.

**Suggested fix** — pin the real path with two tests in `tests/security/test_run_refs_validation.py`, next to the existing `test_policy_set_construction_rejects_an_unknown_run_path`:

```python
def test_run_refs_are_validated_on_an_authored_admission_grant() -> None:
    policy = AgentPolicy.model_validate(
        {"admission": {"mode": "allow", "constraints": ["run.definitely_not_a_path < 5"]}}
    )
    with pytest.raises(PolicySetError, match="unknown run"):
        PolicySet({DEFAULT_ROLE_NAME: policy})


def test_run_refs_are_validated_on_an_authored_reach_grant() -> None:
    policy = AgentPolicy.model_validate(
        {
            "agents": {
                "other": {
                    "as": ["handoff"],
                    "mode": "allow",
                    "constraints": ["run.definitely_not_a_path < 5"],
                }
            }
        }
    )
    with pytest.raises(PolicySetError, match="unknown run"):
        PolicySet({DEFAULT_ROLE_NAME: policy})
```

Verified both ways: they pass at `61d74fc`, and both fail when line 261 is reverted to `policy.tools.values()`.

Worth keeping the compose test as a front-end smoke test — it just shouldn't be the only thing standing between this fix and a silent regression.

guillaume-hexamind (commented):
One blocking issue, nice logic otherwise

quanghexa94:
Fixed — thanks for the precise trace, you're exactly right.

Added two regression tests on the **authored single-file path** in `tests/security/test_run_refs_validation.py` (next to `test_policy_set_construction_rejects_an_unknown_run_path`): an `AgentPolicy` with an `admission:` grant, and one with an `agents:`/reach grant, each carrying a bad `run.*` ref and fed straight to `PolicySet` → `PolicySetError`. I verified both **fail** when `_validate_run_refs` is reverted to `policy.tools.values()` and **pass** with the fix, so they actually guard it — while the compose test still passes either way (kept as the front-end smoke test, as you suggested).

Also confirmed the fix closes **both** fail-opens (admission *and* reach on the authored path); I've updated the PR description to call that out.

guillaume-hexamind (commented):
Good to me, just a ruff formatting to make before merge, see failing tests

quanghexa94:
it's good now Guillaume!

guillaume-hexamind (approved):

Issue #160, opened 2026-09-02
data model: add created_by and updated_by field in all tables


hexamindNR:
except a few tables, tables lack of field created_by, updated_by which are mandatory for auditability
for table devtoken a field user_id will be usefull id a user can create a token for an other user
Use Case a developper is leaving I want to invalidate all is token

PR #206, opened 2026-09-10
feat(platform-api, collector): soft-delete devtoken rows and filter revoked keys on every read path

### LINKED ISSUE #161: data model: add soft delete field for critical table.
for devtokens a deleted_at field is important to track history
it will be useful to update the cache for the go collector

### LINKED ISSUE #162: cache collector improvements
Change the cache retention du 1h at least
update query to fetch non deleted token
use soft delete to fetch deleted tokens instead of fetching all tokens

guillaume-hexamind:
Closes #161. Closes #162 (part 2; part 1 shipped in #199, part 3 stays deferred).

Revocation was a hard `DELETE`, so a revoked key left no trace — no timestamp, no revoker, nothing to show an auditor or to explain a support ticket six weeks later. `devtoken` now carries `revoked_at` / `revoked_by_user_id` and the row survives as the audit record.

**These two issues are one change.** Flipping the delete without the read-path filters is a full revocation bypass across OTLP ingest, bearer HTTP and WebSocket, so #162's "fetch non-deleted tokens" lands here rather than separately.

## What changed

| Surface | Change |
|---|---|
| `find_token_by_secret` | `revoked_at IS NULL`. Covers `_validate_sdk_token`, `require_project`, `ws_require_project` and `GET /v1/me/key` in one line — and stops a revoked key bumping its own `last_used_at` |
| `list_api_keys` | same filter; the dashboard's key list |
| `apiKeyQuery` (Go) | same filter. This is the OTLP ingest path's **only** revocation check |
| `delete_api_key` | becomes `revoke_api_key`, stamping who and when |

Naming follows `Ban` (`revoked_at` / `revoked_by_user_id`), not the `deleted_at` the issue text suggests — same idea, and a second name for it would fragment the convention. It also composes with #160's actor columns later.

## Reviewer notes

**A repeat revoke still 404s** and never moves the first timestamp. This diverges from `revoke_ban`, which is idempotent, and it is deliberate: `test_revoke_token_when_token_already_deleted_then_status_is_404` pins today's contract and the dashboard's `revokeToken` is written against it. Changing a status code inside a security fix is scope I did not want.

**No dashboard changes.** `list_api_keys` filters unconditionally — no `include_revoked` parameter. The audit value lands in the database here; surfacing revoked keys in the UI is a separate change with its own design question, and it would pull in `TokenListItem`, `api.ts` and `Tokens.tsx`.

**Scoped to `devtoken` only.** It is the one table with no `UniqueConstraint`, so a retained row costs nothing. Every other candidate (`Project`, `Agent`, `PolicyModule`, `RoleBinding`) would need a partial unique index on `deleted_at IS NULL`, which this codebase avoids because SQLite can't be relied on for it — see `Invitation`'s docstring.

**One drive-by fix, in its own commit.** `make postgres-init` ran `init_db()` without importing `hexgate_api.models`, so `SQLModel.metadata` was empty and `create_all` created **zero tables** — silently, exit code 0. My migration replay turned that into a loud `relation "devtoken" does not exist`, which is how it surfaced. The deploy stack's `api-init` already had the `import hexgate_api.models` workaround; the same one line is now here. Reviewable independently as `c333a23`.

## ⚠️ Deploy ordering — the migration goes first

`create_all` adds missing *tables*, never missing *columns*, so this schema change cannot ride along with the deploy.

| State | Result |
|---|---|
| Column present, old code | Fine — additive column nobody references |
| Column absent, new code | **Ingest outage.** The collector's first snapshot load raises `UndefinedColumn`, `start()` fails, the container refuses to boot. Every bearer-authenticated route 500s |

```bash
cd /srv/hexgate-<stage>
docker compose -p hexgate-<stage> --env-file platform/.env.<stage> \
  -f platform/docker-compose.deploy.yml exec -T postgres \
  psql -v ON_ERROR_STOP=1 -U hexgate -d hexgate \
  < platform/postgres/migrations/0001_devtoken_soft_delete.sql
make platform-up STAGE=<stage>
```

The migration is idempotent and safe to apply early. `DEPLOY.md` § 6 carries this; `make postgres-init` replays the directory locally.

`platform/postgres/migrations/` is a new directory, mirroring the existing hand-applied `platform/clickhouse/migrations/` convention (README Option B). Alembic is still the eventual answer — #160 step 3c is the change that will force it.

## Rollback

Revert the code, **leave the columns** — they are nullable and the old code never selects them. But rows soft-deleted while the new code ran reappear as live keys, so if any real revocation happened first:

```sql
DELETE FROM devtoken WHERE revoked_at IS NOT NULL;
```

That is the one manual step a rollback needs, and skipping it un-revokes keys.

## Testing

`make check-all` exits 0. `make collector-test-integration` passes against real Postgres + Redpanda.

- **6 new API tests** — row retained with the actor, second revoke keeps the first stamp, and one per bearer surface (`/v1/me/key`, `require_project`, `optional_api_key`, plus `last_used_at` not bumped). 652 API tests pass.
- **WebSocket** — a soft-delete twin of the existing 4401 test. The hard-delete one stays; "row genuinely gone" is still a distinct case.
- **The Go filter was mutation-tested.** Removing `WHERE revoked_at IS NULL` makes the integration suite **fail** (the collector logs `"keys": 1` — the revoked key still in the snapshot), so the revocation assertion is load-bearing rather than passing vacuously.
- `revokeAPIKey` in the Go harness is now an `UPDATE`; fixture teardown stays a `DELETE` (that is teardown, not revocation, and the row must leave before the FK deletes).

**`cache_test.go` cannot cover this**, contrary to the issue's suggestion: `fakeKeySource` returns a `map` and sits downstream of the SQL, so a "soft-deleted row" case there would assert nothing new. There is a 4-line string guard on the query constant instead, with the real coverage in the integration suite.

## Not in scope

Soft delete on other tables · revoked keys in the dashboard · `devtoken.user_id` + revoke-on-member-removal (#160 step 3a, lands on this same row) · Alembic · #162 part 3 incremental fetch · an index on `revoked_at` (the read is a deliberate full scan).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -189,6 +189,14 @@ postgres-init: postgres-up ## Create the platform-api tables on local Postgres (
 	@# stack's api-init one-shot imports it (docker-compose.deploy.yml).
 	cd platform/api && DATABASE_URL=$(POSTGRES_DSN) uv run python -c \
 		"import asyncio; import hexgate_api.models; from hexgate_api.core.db import init_db; asyncio.run(init_db())"
+	@# create_all adds missing TABLES, never missing columns, so a volume that
+	@# predates a column keeps 500ing at request time with no startup error.
+	@# Every file here is idempotent (IF NOT EXISTS), so replaying the whole
+	@# directory on each init is the cheapest honest local-dev migration story.
+	@# Deployed stacks apply these by hand — see platform/DEPLOY.md section 6.
+	for f in platform/postgres/migrations/*.sql; do \
+		docker exec -i hexgate-postgres psql -v ON_ERROR_STOP=1 -U hexgate -d hexgate < "$$f" >/dev/null; \
+	done
 
 .PHONY: postgres-stop
 postgres-stop: ## Stop Postgres (keeps the data volume)
--- a/Makefile
+++ b/Makefile
@@ -183,8 +183,12 @@ postgres-up: ## Start local Postgres and wait until healthy
 # fresh volume never surfaces as "no such table" on the first request.
 .PHONY: postgres-init
 postgres-init: postgres-up ## Create the platform-api tables on local Postgres (idempotent)
+	@# `import hexgate_api.models` is load-bearing: it registers every table on
+	@# SQLModel.metadata before create_all runs. Without it the metadata is
+	@# empty and this silently creates NOTHING — the same reason the deploy
+	@# stack's api-init one-shot imports it (docker-compose.deploy.yml).
 	cd platform/api && DATABASE_URL=$(POSTGRES_DSN) uv run python -c \
-		"import asyncio; from hexgate_api.core.db import init_db; asyncio.run(init_db())"
+		"import asyncio; import hexgate_api.models; from hexgate_api.core.db import init_db; asyncio.run(init_db())"
 
 .PHONY: postgres-stop
 postgres-stop: ## Stop Postgres (keeps the data volume)
--- a/Makefile
+++ b/Makefile
@@ -328,7 +328,7 @@ platform-api: ## Run the platform API dev server (FastAPI on :8000, SQLite)
 	cd platform/api && uv run uvicorn hexgate_api.main:app --reload --port 8000
 
 .PHONY: platform-api-pg
-platform-api-pg: postgres-up ## Run the platform API against local Postgres (starts PG first)
+platform-api-pg: postgres-init ## Run the platform API against local Postgres (starts PG first)
 	cd platform/api && DATABASE_URL=$(POSTGRES_DSN) uv run uvicorn hexgate_api.main:app --reload --port 8000
 
 .PHONY: platform-api-test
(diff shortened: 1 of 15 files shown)


victorludvig (approved):
LGTM, 3 nits about docs/comments to check, and merge conflict to resolve:

1. DEPLOY.md:252 says skipping the compensating DELETE on rollback un-revokes keys on HTTP, WebSocket and OTLP alike. Old code matches on secret, which is now masked, so HTTP and WS stay revoked. Only OTLP ingest and the dashboard token list are affected. Rewrite the sentence.
2. Stale hard-delete comments contradict the new mechanism: platform/collector/config.yaml:106-107 ("deletes its row rather than a flag"), docs/internals/audit-pipeline.md:323-324 ("absence = revoked"), and cache_test.go:97 (cites the removed delete_api_key).
3. test_ws_serve.py:189 is a near-verbatim copy of the test above it, differing in one line. Extract a helper and keep the hand-stamped revoked_at.

PR #209, opened 2026-09-10
feat(platform-api, dashboard): record who created and last updated every control-plane row

### LINKED ISSUE #160: data model: add created_by and updated_by field in all tables

guillaume-hexamind:
Closes #160.

> **Stacked on #206** (`gp/feat/devtoken_soft_delete` → `main`) — review that one
> first. This PR's base is its branch, so the diff here is only the actor-column
> work; it reuses #206's `revoked_at` / `revoked_by_user_id` and does not re-open
> the revocation filters.

Eleven of thirteen Postgres tables had no actor trail, so "who minted this key /
edited this policy / added this member" was unanswerable and unrecoverable after
the fact.

## The part that isn't paperwork

#160's own use case is a live hole: **a departing developer's API keys kept
working indefinitely, and no query could even enumerate them to revoke by hand.**
Keys are minted `ttl_seconds=None` and nothing linked one to a person.

`devtoken` gets **two** actors, because one column can't be both:

- `created_by_user_id` — who minted the key
- `owner_user_id` — **whose key it is**; this is what offboarding keys on

If an admin mints a key for Bob and only the creator is recorded, removing Bob
revokes nothing. So `TokenMintRequest` takes an optional `owner_user_id`, gated
on the caller being admin/owner **and** the target being a member of the
project's org — otherwise the key is parked where no sweep can reach it. Naming
your own id isn't delegation, so plain members still mint their own keys.

`remove_member` then revokes every live key the leaver owns, across every project
in the org. Two properties are load-bearing:

- the **last-owner guard runs first**, so a refused removal revokes nothing;
- the sweep **never commits** — it stamps in the caller's transaction so the
  revocations and the membership delete ride one commit. Revoking a colleague's
  keys but failing to remove their membership (or the reverse) is worse than
  either outcome alone.

Keys with **no** owner are untouched: a `deploy/provision.py` mint belongs to no
person, so nobody's departure should kill it.

Everything else is the sweep across the remaining tables — **13 routes, not the
22 estimated in #160**: of 23 mutating decorators, three write ClickHouse, one
writes nothing, and five already carried a trail. Six routes were *discarding*
the `User` their dependency already returns.

## Scope

| | +added | −removed | net | files |
|---|---:|---:|---:|---:|
| tests | 1813 | 26 | +1787 | 10 |
| platform: api source | 641 | 112 | +529 | 20 |
| platform: dashboard | 42 | 2 | +40 | 3 |
| migration (SQL) | 97 | 0 | +97 | 1 |
| docs (`DEPLOY.md`) | 20 | 6 | +14 | 1 |
| **TOTAL** | **2613** | **146** | **+2467** | **35** |

Tests are 69% of the additions. Most of that is the sweep's four boundaries —
another member's keys, another org's keys, ownerless keys, already-revoked keys —
each of which is a revocation bypass or an outage if wrong, plus six near-identical
"did this route stamp the caller" checks.

**For review, read the source first** — it's 780 added lines, and a good share of
that is comments:

```bash
git diff origin/gp/feat/devtoken_soft_delete...HEAD -- \
  platform/api/hexgate_api platform/dashboard platform/postgres
```

## ⚠️ Deploy order

`init_db()` is `create_all`, which creates missing **tables** and never adds
columns to a table that already exists.

```bash
make platform-migrate STAGE=<stage>   # FIRST — already step 1 of DEPLOY.md § 6
make platform-up STAGE=<stage>
```

Reversing them is an outage, but **not the same outage as #206** — don't debug
one expecting the other. `0001` kills OTLP ingest (the Collector's snapshot query
selects `revoked_at`, so it refuses to boot). `0002` leaves the Collector alone
and instead 500s every API route touching one of its nine tables, **reads
included — so the dashboard goes blank rather than degrading.** `DEPLOY.md` § 6
now spells out both signatures, and records that rolling back past `0002` needs
no compensating step (nullable, unreferenced by older code).

`git diff --stat platform/collector/` is empty, by design.

## Review notes

- **`remove_member` returns `MemberRemoval`, not `bool`.** It's an object, so
  `assert ok` is **truthy even when the removal failed** — the four existing
  assertions in `test_orgs.py` didn't fail loudly, they passed *vacuously*. All
  four now use `.removed`. This is the most likely silent regression in the PR.
- `DELETE /orgs/{id}/members/{uid}` stays **204**; the revoked count goes to a log
  line and to the confirm dialog as static text, not onto the wire (the dashboard
  is written against 204).
- `recompile_project` / `backfill_bundles` deliberately touch neither
  `updated_at` nor `updated_by_user_id` — a derived bundle rebuild isn't an
  authored edit, and stamping it would replace an agent's real author with
  whoever last edited a shared policy module. Commented at `_apply_bundle` and
  pinned by a test, because it's the kind of thing a later reader "fixes".
- Invite-accept attributes the new membership to the **inviter**, not the
  invitee: "who let this person in" is the audit question, and the invitee is
  already `user_id` on that row.
- `emails_for_user_ids` moved from the bans slice to members — a plain `User`
  read, and a third feature now resolves actor ids for display.
- Dashboard scope is one **Owner** column on `Tokens.tsx` (minter in the tooltip)
  plus a corrected remove-member dialog, which previously promised only lost
  *access*. Exposing the trail on the other read schemas is left for when a
  screen asks; `BanRead` is the pattern.

## Decisions you may want to push back on

- **`updated_by` ships, but is not a change log.** It overwrites, so it answers
  "who touched this last", never "what did the policy say before Tuesday". That's
  a reason not to *rely* on it, not to skip a two-line column operators ask for
  constantly. It lands on exactly the five tables mutated in place; hard deletes
  (`delete_module`) still record nothing, because there's no row left to stamp.
- **NULL means "no human actor" — no sentinel.** `create_all` emits these as FKs
  to `user`, so a `"system"` string couldn't be stored. **No historical actor can
  be backfilled** — the information never existed, so pre-migration rows read
  NULL.
- **The first-boot seed writes NULL** for two reasons: it has no human actor, and
  `ensure_default_seed` inserts the `Organization` before the `User` in one
  commit, so attributing seed rows would rest on the unit of work ordering
  inserts by FK dependency.
- **Columns declared per table, not via a mixin.** Nothing here puts a
  `Field(foreign_key=...)` in a shared mixin, and sharing a declaration that
  produces a `ForeignKey` across mapped classes wants `declared_attr` to be safe.

## Verified

`make check-all` exits 0 — SDK 2579, platform-api 699 (+46), dashboard 218,
collector, lint, fmt, typecheck. Each of the three commits was checked out in a
clean worktree and is independently green.

**Against a real Postgres**, not just the SQLite suite:

- `0002` applied to a volume that predated the columns — all 21 columns plus the
  `ix_devtoken_owner_user_id` index landed, the `updated_at` backfill correctly
  took `created_at`, and a replay is a clean no-op.
- The offboarding story end-to-end over HTTP on asyncpg: admin mints a key
  **owned by** a developer → 200 → admin removes the developer → **401** → the
  row survives with `owner_user_id` = the leaver, `revoked_by_user_id` = the
  admin, secret masked. The developer keeps their own personal org.

Five guards were **negative-tested** — each confirmed to fail when the invariant
it protects is broken: the sweep gaining a commit; `_apply_bundle` stamping
`updated_at`; a model column with no `ALTER TABLE`; a typo'd `ADD COLUMN`
(`created_by_user_i` is valid SQL); and a migration that was never applied.

## Follow-ups

1. **Control-plane `audit_event` log** — the change history columns can't give,
   and the only thing that covers hard deletes. Referenced from `models.py` and
   `policy_modules/service.py`; worth filing before this merges.
2. "This will revoke N keys" preview on the remove-member dialog.
3. Dashboard owner-picker on the mint dialog (delegated mint is API-only here).
4. Revoking keys when a `User` is deactivated (`is_active = False`) — different
   lifecycle; an org-scoped sweep is the wrong shape for a platform-wide disable.
5. A `created_by` / `updated_by` SQLModel mixin, once verified in isolation.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -189,6 +189,14 @@ postgres-init: postgres-up ## Create the platform-api tables on local Postgres (
 	@# stack's api-init one-shot imports it (docker-compose.deploy.yml).
 	cd platform/api && DATABASE_URL=$(POSTGRES_DSN) uv run python -c \
 		"import asyncio; import hexgate_api.models; from hexgate_api.core.db import init_db; asyncio.run(init_db())"
+	@# create_all adds missing TABLES, never missing columns, so a volume that
+	@# predates a column keeps 500ing at request time with no startup error.
+	@# Every file here is idempotent (IF NOT EXISTS), so replaying the whole
+	@# directory on each init is the cheapest honest local-dev migration story.
+	@# Deployed stacks apply these by hand — see platform/DEPLOY.md section 6.
+	for f in platform/postgres/migrations/*.sql; do \
+		docker exec -i hexgate-postgres psql -v ON_ERROR_STOP=1 -U hexgate -d hexgate < "$$f" >/dev/null; \
+	done
 
 .PHONY: postgres-stop
 postgres-stop: ## Stop Postgres (keeps the data volume)
--- a/Makefile
+++ b/Makefile
@@ -183,8 +183,12 @@ postgres-up: ## Start local Postgres and wait until healthy
 # fresh volume never surfaces as "no such table" on the first request.
 .PHONY: postgres-init
 postgres-init: postgres-up ## Create the platform-api tables on local Postgres (idempotent)
+	@# `import hexgate_api.models` is load-bearing: it registers every table on
+	@# SQLModel.metadata before create_all runs. Without it the metadata is
+	@# empty and this silently creates NOTHING — the same reason the deploy
+	@# stack's api-init one-shot imports it (docker-compose.deploy.yml).
 	cd platform/api && DATABASE_URL=$(POSTGRES_DSN) uv run python -c \
-		"import asyncio; from hexgate_api.core.db import init_db; asyncio.run(init_db())"
+		"import asyncio; import hexgate_api.models; from hexgate_api.core.db import init_db; asyncio.run(init_db())"
 
 .PHONY: postgres-stop
 postgres-stop: ## Stop Postgres (keeps the data volume)
--- a/Makefile
+++ b/Makefile
@@ -328,7 +328,7 @@ platform-api: ## Run the platform API dev server (FastAPI on :8000, SQLite)
 	cd platform/api && uv run uvicorn hexgate_api.main:app --reload --port 8000
 
 .PHONY: platform-api-pg
-platform-api-pg: postgres-up ## Run the platform API against local Postgres (starts PG first)
+platform-api-pg: postgres-init ## Run the platform API against local Postgres (starts PG first)
 	cd platform/api && DATABASE_URL=$(POSTGRES_DSN) uv run uvicorn hexgate_api.main:app --reload --port 8000
 
 .PHONY: platform-api-test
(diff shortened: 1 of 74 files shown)


victorludvig (approved):
Approved, optional housekeeping/cleanup:

DEPLOY.md:262-264's wrong failure signature; 0002:55,64 pointing at models.py:274/383 that this PR moved to 326/450; test_actor_columns.py:88 "Eleven of thirteen" (nine of fourteen); deps/tokens.py:118 "all three bearer-auth surfaces" (now four); 0001:19-22 citing a foreign_key="user.id" no field declares; Makefile:197-199 missing || exit 1; test_tokens.py:549 pinning an outcome that contradicts actor_fk_column's docstring; 27294c8 being a merge commit.

PR #217, opened 2026-09-14
docs: add pr-description and write-adr agent skills

quanghexa94:
Add two on-demand agent skills under `.claude/skills/` so anyone using an agent on this repo drafts PRs and ADRs in the conventions we already follow, instead of re-deriving them each time. Pure tooling/docs, no code paths touched.

## Design

Two `SKILL.md` files, each loading on task intent (matched by the `description` frontmatter), following the same shape as the existing `integration-tests` skill.

| Skill | When it loads | What it encodes |
| --- | --- | --- |
| `pr-description` | opening/drafting a PR body | objective, design, important files, test plan, try it, notes; scale-to-change; the "write the test plan yourself" guard against rubber-stamping. Points at real anchors: `deploy/demo_policies`, `docs/adr/`, and the `type(scope):` subject in CLAUDE.md. |
| `write-adr` | recording a durable decision | the style `docs/adr/` already uses: `R-<AREA>-NNN` ids, Status, Applies-to globs, MUST/MUST-NOT decision, a deliberately un-compressed Why, and the supersede-chain pattern (`R-AGENT-001` -> `R-AGENT-002`). |

The `write-adr` skill is descriptive, not prescriptive: it was written by reading the ten existing records and matching them, so new ADRs stay consistent with what is there.

## Important files

| File | What to look at |
| --- | --- |
| `.claude/skills/pr-description/SKILL.md` | the section order and the template; check the anchors match the repo |
| `.claude/skills/write-adr/SKILL.md` | that the id scheme, Applies-to, and supersede guidance match `docs/adr/` |

## Test plan

Markdown only, so no suite to run. Verified by hand: `git grep` for `co-authored-by` / `generated with claude` / `noreply@anthropic` across tracked files returns nothing, so the skills impose no attribution convention (each contributor keeps their own). Confirmed both files render (fenced template uses a four-backtick outer fence so the inner block shows correctly) and that the paths they cite (`deploy/demo_policies`, `docs/adr/`) exist.

## Notes

- Deliberately unopinionated on PR/commit attribution trailers, so this does not impose one contributor's setting on the team.
- Two further ideas were considered and left out for now: pulling an in-repo AI-docs reference that both skills derive from, and a design-doc skill. Easy follow-ups if wanted.


.claude/skills/pr-description/SKILL.md
```diff
@@ -0,0 +1,131 @@
...
+   technical shape: the key idea, the main type or function, what it reuses vs.
+   what is new, and the one or two decisions worth knowing. A small table works
+   well when the PR adds several related things (lints, endpoints, flags). Keep
+   it to what a reviewer needs, not a full design doc. Link the design doc or the
+   relevant ADR (`docs/adr/R-…`) if there is one.
+3. **Important files (required when the diff spans more than a few files).** A
+   `| File | What to look at |` table naming the files that carry the design and
+   what to check in each. Highest-leverage section: it turns a flat file list
+   into a review order. Leave out the mechanical files.
+4. **Test plan (required for anything that changes behaviour).** What you
```
victorludvig, line 67: Here I would put it like that:
- unit tests: <unit tests you ran>
- integration tests: <integration tests you ran> / are they local integration tests (against local db), or against staging?
- end-to-end: <what you ran>
Then any additional tests you would want to run

.claude/skills/pr-description/SKILL.md
```diff
@@ -0,0 +1,131 @@
...
+- **Point at the files that matter.** In a 13-file diff, name the three that
+  carry the design; leave the ten mechanical ones out.
+- **Write for someone without the background.** The reviewer may not have been in
+  the design discussion.
+
+## Section order
+
+Use these in this order. Drop the ones a given PR does not need.
+
+1. **Objective (required).** One short paragraph; make the first sentence stand
```
victorludvig, line 51: Not sure about this section, since the PR title should already contain the objective.
We can directly talk about ##What is changing

.claude/skills/pr-description/SKILL.md
victorludvig, line 67: More than a test plan, it's also about non-regressions:
> What tests did you add to make sure that someone starting fresh on this code will be able to spot a regression?

So I would rename it Tests rather than Test plan.

.claude/skills/pr-description/SKILL.md
```diff
@@ -0,0 +1,143 @@
...
+  job is the motivation and the approach the diff cannot show.
+- **Scale length to the change.** A one-file fix gets a paragraph. A subsystem
+  gets the full skeleton. Do not pad a small PR into a big one.
+- **Make it scannable.** Short paragraphs, bullet lists, and tables beat prose walls.
+- **Point at the files that matter.** In a 13-file diff, name the three that
+  carry the design; leave the ten mechanical ones out.
+- **Write for someone without the background.** The reviewer may not have been in
+  the design discussion.
+
+## Section order
```
victorludvig, line 47: Following up on our meeting this morning, do you think we could add a Risk Tier, with one of these risk tiers: low, medium, high, and then a short explanation why.
We should define what is low, medium, high, for example:
- low: secondary feature not impacting the core functionalities, e.g. AI Act reports
- medium: feature that could impact the core functionalities, e.g. make the platform slower, or a core UI page
- high: core impacted, e.g. public APIs, security/auth, migration...

Then the review effort can be adjusted according to risk tier.

PR #220, opened 2026-09-15
fix(platform-scripts): apply each store's migrations independently

guillaume-hexamind:
> **Design note:** [Migration stack — design note for reviewers](https://app.notion.com/p/3ddfb45dbae281edb8f8f1ebc669a999) — why this work exists, how schema delivery worked before, the four invariants these changes establish, and what is deliberately deferred (the Postgres startup guard, Alembic). Read that first if you want the whole picture; this PR is its first wave.

`platform-migrate` was two sequential `for` loops in one target, each ending `|| exit 1`. Any
failure inside the Postgres loop ended the *target*, so the ClickHouse loop never ran — and
nothing said so.

Both stages hit this during the #209 rollout, from two unrelated causes:

- **staging** — the Postgres run hung on `0002` (lock contention with the live API, no
  `lock_timeout`), was interrupted, and took every ClickHouse migration with it.
- **prod** — `0003` failed with `relation "policy_file" does not exist` on a stack predating
  #179, same result.

In both cases the error on screen was a Postgres error, so a Postgres problem got fixed. The
real damage surfaced later as the api crash-looping on `SchemaOutOfDate: llm_message` — a
symptom with no visible connection to its cause.

## What changes

The loops move to `platform/scripts/migrate.sh`, invoked the same way `env-secret.sh` already
is (`Makefile:391`). Tracking per-store status across two loops inside a recipe would mean one
backslash-joined logical line carrying `rc` state, which is worse than the 14 lines it
replaces. The script:

- **applies the two stores independently** — a Postgres failure is reported, ClickHouse still
  runs;
- **stops within a store at the first failure** — `0003` may depend on `0002` having landed;
- **ends with a summary and a nonzero exit** if either did not complete:
  `migrate(prod): postgres=FAILED clickhouse=ok`.

`set -e` is deliberately absent, with a comment saying so — adding it would restore exactly the
bug this removes.

**Postgres DDL now runs under `lock_timeout=10s`.** `ADD COLUMN` takes ACCESS EXCLUSIVE and the
non-concurrent `CREATE INDEX` takes SHARE, so one session idle in a transaction blocked the
staging run indefinitely, and then blocked every query queued behind it. Ten seconds and a
named error is the better failure. `HEXGATE_MIGRATE_LOCK_TIMEOUT_MS=0` waits forever, for a
quiesced stack running a long index build.

This is a real behaviour change: a migration that previously *eventually* succeeded against a
busy stage now fails after ten seconds. That is why `DEPLOY.md` § 6 gains a `stop api enricher`
step before the migrate, keeping `collector` and `redpanda` up so OTLP keeps buffering through
the window.

Also in scope:

- `postgres-init` gains the `|| exit 1` its deploy counterpart already had. Without it a
  mid-loop failure is swallowed and make exits on the *last* iteration's status.
- psql output is no longer discarded on the deploy path — the backfill `UPDATE n` counts are
  the only evidence the `updated_at` fill did anything. `postgres-init` keeps `>/dev/null`:
  it runs on nearly every local workflow and is usually a no-op. The asymmetry is commented.
- Each file is announced **after** it succeeds, not before. The `applying 0002` line left on
  staging's hung terminal named a file that had *not* been applied.
- New `make clickhouse-migrate` for local dev. The only documented way to pick up a new
  ClickHouse table was `clickhouse-reset`, which wipes — so a developer whose volume predated a
  migration was pointed at destroying data.

## Scope

| | +added | −removed | files |
|---|---:|---:|---:|
| `platform/scripts/migrate.sh` (new) | 73 | 0 | 1 |
| `Makefile` | 29 | 15 | 1 |
| `platform/DEPLOY.md` | 20 | 0 | 1 |
| **TOTAL** | **122** | **15** | **3** |

No Python, no TypeScript, no SQL. **The migrations applied are byte-identical** — this changes
the runner, not what it runs.

## Verified

`make check-all` exits 0 — SDK 2653, platform-api 767, dashboard 218, collector, lint, fmt,
typecheck.

Each behavioural claim above was run, not reasoned about:

- **`make clickhouse-migrate`** against a real local ClickHouse — all three files applied,
  exit 0, clean no-op on an already-migrated volume.
- **`make postgres-init`** twice — clean replay, exit 0 both times.
- **The new `|| exit 1`, negative-tested.** A deliberately broken `.sql` dropped into the
  middle of the directory makes `postgres-init` exit **2**; without the guard the identical run
  exits **0**.
- **`lock_timeout`, negative-tested.** Against a table held under ACCESS EXCLUSIVE by another
  session: with no timeout the `ALTER` was **still blocked after 6s** (the staging hang,
  reproduced); with `PGOPTIONS='-c lock_timeout=1000'` it failed in **1s** with
  `canceling statement due to lock timeout`.
- **Cross-store isolation**, exercised against the committed script with only the two runner
  functions stubbed — `apply_dir`, the summary and the exit logic are the shipped code:

  | Scenario | Result |
  |---|---|
  | Postgres fails on `0002` | Postgres stops after 1; **ClickHouse applies all 3**; exit 1 |
  | All succeed | both `ok`; exit 0 |
  | ClickHouse fails on `0003` | Postgres `ok`; ClickHouse stops after 2; exit 1 |
  | No `STAGE` argument | exit 1 |

  The first row is the prod scenario.

`shellcheck` has not been run against the script — it is syntax-checked with `bash -n` only.

## Out of scope

This makes the runner honest; it does not make the migration files safer or the coverage
better. Three follow-ups, each independent of this one and of each other:

1. **`to_regclass` guards** on every migration, so a file whose table `create_all` has not
   built yet skips instead of failing — the prod half of the story above.
2. **`--refresh-package hexgate`** in the API Dockerfile. The image installs the SDK as a
   built wheel behind a persistent cache mount, and uv keys that cache on a version that has
   not moved since `5eace3d`, so both stages shipped a stale SDK and the enricher died at
   import.
3. **Executing the migrations in CI.** Nothing runs them today — `test_postgres_smoke.py:74`
   regexes `ADD COLUMN` out of the files and checks the names against a `create_all` database,
   so a migration that cannot execute at all passes.

All three are specified in the design note above, with the rollout evidence behind each.


platform/scripts/migrate.sh
```diff
@@ -0,0 +1,73 @@
...
+}
+
+run_postgres() {
+  compose exec -T -e PGOPTIONS="-c lock_timeout=$LOCK_TIMEOUT_MS" postgres \
+    psql -v ON_ERROR_STOP=1 -U hexgate -d hexgate <"$1"
+}
+
+run_clickhouse() {
+  compose exec -T clickhouse sh -c \
+    'clickhouse-client --user "$CLICKHOUSE_USER" --password "$CLICKHOUSE_PASSWORD" --multiquery' \
```
quanghexa94: **`--database` divergence between local and deploy runners.** This deploy runner passes no `--database`, but the new `make clickhouse-migrate` target passes `--database hexgate_audit`. The same migration files therefore run against different default databases on the two paths.

Harmless today — all three files are fully qualified (`hexgate_audit.policy_decision`, `hexgate_audit.llm_message`, …), so nothing depends on the default DB. The risk is future: an author who writes an unqualified `ALTER TABLE llm_message …` would see it pass locally and fail on deploy — a divergence that passes local review but breaks the stage. Cheap fix: add `--database hexgate_audit` here to match local.

quanghexa94: **No ClickHouse-side timeout analogous to Postgres `lock_timeout`.** The design premise is "fail in seconds with a nameable error, don't hang," but that guarantee only covers Postgres. A ClickHouse DDL blocked on an in-flight mutation/merge would hang `run_clickhouse` indefinitely, and it runs *after* Postgres — so there's no timeout to fail fast. The DEPLOY.md "stop the writers and re-run" remedy doesn't apply here. Lower probability given today's lightweight ALTER/CREATE files, but it's the same failure class the Postgres timeout was added to eliminate, left asymmetric.

platform/scripts/migrate.sh
```diff
@@ -0,0 +1,73 @@
...
+}
+
+# Applies every file in a directory through `runner`, stopping at the first
+# failure. One line per file AFTER it succeeds, so an interrupted run shows what
+# completed rather than what it was about to attempt.
+apply_dir() {
+  local label="$1" dir="$2" runner="$3"
+  local applied=0 f
+  for f in "$dir"/*.sql; do
+    if [ ! -e "$f" ]; then
```
quanghexa94, line 52: **An empty/globless migrations dir is reported as a store FAILURE.** If `"$dir"/*.sql` ever matches no files, the loop runs once with the literal unexpanded glob, `[ ! -e "$f" ]` is true, `apply_dir` returns 1, and the run prints `migrate(<stage>): <store>=FAILED` and exits 1 — despite nothing being wrong. That directly contradicts the "read the last line of the run" guidance this PR adds to DEPLOY.md: an operator would conclude a migration broke when there were simply none to apply. Not reachable today (both dirs are populated), but a latent trap; a store with zero files should be `ok`, not `FAILED`.

platform/DEPLOY.md
```diff
@@ -227,19 +232,35 @@ the files are idempotent, so a release that adds no column replays them to no
 effect. Reversing the two is an outage, not a slower path (see below). Skip it
 only on a first-ever deploy, where there is no existing schema to alter.
 
+**Read the last line of the run, not the last file it named.** The two stores
+are applied independently — a Postgres failure no longer skips ClickHouse — and
+the run ends with `migrate(<stage>): postgres=… clickhouse=…` plus a nonzero
+exit if either did not complete. A run that ends without that summary was
+interrupted, and you cannot assume either store finished.
+
+Postgres statements run under `lock_timeout=10s`. If you skip the `stop` above,
```
quanghexa94, line 261: **The 10s `lock_timeout` defense relies on the operator remembering the separate `make platform-stop-writers` step** — `platform-migrate` does not depend on it. Skipping the stop turns the timeout into a `FAILED` Postgres run that must be redone. I see from the design note that auto-stop-then-migrate wiring is *deliberately deferred*, so flagging this only for completeness rather than as a change request — it's a known, accepted footgun.

quanghexa94 (commented):
Code review (high effort). No correctness bugs — the core independence logic (`pg_rc`/`ch_rc` captured separately, ClickHouse runs regardless of Postgres, nonzero exit if either failed) is correct, and `compose()` matches the Makefile's `DEPLOY_COMPOSE`. Four lower-severity findings inline; none are merge blockers, but #1 and #2 are one-line fixes that close latent traps the PR's own docs could mislead an operator into.

platform/scripts/migrate.sh
guillaume-hexamind: Good catch, and it's a stronger finding than the comment claims — fixed in e9b79cd.

You hedged on the future risk, but I went to check whether `CLICKHOUSE_DB: hexgate_audit` (set in **both** compose files) already made the divergence moot. It doesn't. That variable *creates* the database; it does not make it the user's default:

```
$ clickhouse-client --user hexgate --password … --query "SELECT currentDatabase()"
default
```

So the failure you predicted reproduces exactly, in both directions:

| Path | Same unqualified `ALTER TABLE llm_message …` |
|---|---|
| deploy runner (no `--database`) | `Code: 60. DB::Exception: Could not find table: llm_message. (UNKNOWN_TABLE)` |
| `make clickhouse-migrate` (`--database hexgate_audit`) | succeeds |

Applied your fix, reading from the container's own env rather than a literal so it matches how the credentials on that line already work:

```bash
clickhouse-client --user "$CLICKHOUSE_USER" --password "$CLICKHOUSE_PASSWORD" \
  --database "$CLICKHOUSE_DB" --multiquery
```

Safe on a first-ever deploy — `init/schema.sql` opens with `CREATE DATABASE IF NOT EXISTS hexgate_audit`, and `platform-migrate` runs `up -d --wait clickhouse` before the script. Verified against a live container: the exact `sh -c` string parses, `currentDatabase()` is now `hexgate_audit`, and all three real migration files replay clean (rc=0) on an already-migrated volume.

The durable version of this is PR 4 (executing the migrations in CI), which would catch an unqualified statement outright rather than relying on the two runners staying in sync.

guillaume-hexamind: Not taking this one, because the load-bearing claim doesn't hold — but it's worth recording why, since it's a reasonable thing to assume.

**ClickHouse does not hang indefinitely here.** It has a default lock timeout, checked on the running server:

```
$ SELECT name, value FROM system.settings WHERE name = 'lock_acquire_timeout'
lock_acquire_timeout    120
```

It throws after 120s rather than blocking forever, so this isn't the same failure class as the Postgres side — the staging incident was an *unbounded* wait, which is what made it indistinguishable from a wedged terminal.

The mutation/merge part is also weaker than it looks: all four audit tables are `ReplacingMergeTree`, **not** `Replicated`, and ClickHouse's own docs for `alter_sync` say it "does nothing to alters of not `Replicated` tables." `ADD COLUMN` on a non-replicated MergeTree is metadata-only and doesn't wait on merges.

```
$ SELECT name, engine FROM system.tables WHERE database='hexgate_audit'
ban_enforcement   ReplacingMergeTree
llm_invocation    ReplacingMergeTree
llm_message       ReplacingMergeTree
policy_decision   ReplacingMergeTree
```

So the real residue of your point is narrower: Postgres is bounded **explicitly** at 10s, ClickHouse **implicitly** at 120s, and nothing in the file says so. That's a documentation gap, not an asymmetric hang risk. I'd rather note the 120s default in a comment than add a second knob — `--lock_acquire_timeout=10` does work on this client version (verified, reads back as `10`) if you'd prefer the explicit symmetry, but it changes behaviour for a case that is already bounded.

Happy to add the comment in this PR if you want it here rather than deferred.

platform/scripts/migrate.sh
guillaume-hexamind: Confirmed — I ran the shipped `apply_dir` verbatim against an empty directory and your trace of the control flow is exactly right: unexpanded glob, `[ ! -e "$f" ]` true, `return 1`, store reports `FAILED`. And you're right that it contradicts the "read the last line of the run" guidance this PR adds.

One thing to fold in before fixing it, though — a **missing** directory returns the identical code and an identically-shaped message:

```
case A: directory exists but is EMPTY        → rc=1
case B: directory does NOT exist (wrong cwd) → rc=1
```

`migrate.sh` uses relative paths (`platform/postgres/migrations`), so running it from anywhere but the repo root lands in case B — and that one *should* fail loudly. So "zero files should be `ok`" can't be a one-liner; it needs `[ -d "$dir" ] || return 1` first, then empty-but-present treated as `ok` with a note, so the wrong-cwd guard survives.

Not taking it in this PR: it needs someone to empty a populated migrations directory to reach, and I'd rather not touch the failure-reporting path on a change whose whole purpose is making that path trustworthy. Will do it with the distinction above next time this file is opened — PR 2's guards are the likely moment. Shout if you'd rather have it now.

platform/DEPLOY.md
guillaume-hexamind: The mechanism is exactly as you describe — `platform-migrate: _require-stage-env` is the whole dependency list, and nothing depends on `platform-stop-writers`.

One correction worth making, because it changes what you should expect from me: the design note does **not** say auto-stop wiring was deliberately deferred. Its "what is deliberately not being done yet" section covers only the Postgres startup guard and Alembic. You read a decision into a gap that nobody had actually made — so thanks for raising it, it deserves a real answer rather than the pass you generously gave it.

The answer is: keep them separate, deliberately. Wiring `platform-migrate: platform-stop-writers` would take a control-plane outage every time someone runs a migrate to confirm there's nothing new to apply — and since the files are idempotent, that's a normal thing to do. Trading a recoverable footgun for a guaranteed outage on the no-op path is the wrong direction.

What's actually missing is that the remedy lives in DEPLOY.md § 6 instead of arriving with the error. Better fix: have `migrate.sh` recognise a `lock_timeout` failure and print `run 'make platform-stop-writers STAGE=<stage>' and re-run` on the spot. I'll do that as a follow-up rather than widen this PR, and record the not-auto-wired decision in the design note so the next reviewer doesn't have to infer it either.

quanghexa94 (approved):
Approving. The runner change is well-scoped and correctly reasoned — byte-identical migrations, disciplined deferral of the bigger rewrites (Alembic, startup guard), and the store-independence logic (`pg_rc`/`ch_rc` captured separately, ClickHouse runs regardless of Postgres, nonzero exit if either failed) is right.

My earlier finding #1 (the `--database` divergence between the deploy runner and `make clickhouse-migrate`) is resolved — `--database "$CLICKHOUSE_DB"` resolves to `hexgate_audit`, matching the local target, so an unqualified migration can no longer pass locally and fail on deploy. Nicely commented too.

Two latent, non-blocking follow-ups (not reachable with today's populated dirs / lightweight CH DDL, so fine to defer):
- `apply_dir` treats an empty/globless migrations dir as a store FAILURE — a store with zero files should read `ok`, not `FAILED`, to match the "read the last line" guidance.
- No ClickHouse-side timeout analogous to the Postgres `lock_timeout`; a blocked CH DDL could still hang the run.

CI is green across the board. Note the branch is BEHIND main — rebase/merge before landing.

quanghexa94 (approved):

PR #222, opened 2026-09-16
fix(platform-api): union policy graph to the most-permissive verdict

quanghexa94:
The policy graph had two bugs surfaced by the compose demo: a reach edge between two agents showed **deny** even though the policy allows it, and the graph's role filter only ever listed **default** while the Playground showed three roles. Both trace to the same place — how the graph unions across roles.

## What it does

- **Union to the most-permissive verdict.** The union view (no role filter) merged duplicate edges to the *strictest* verdict across roles. Agent reach/admission keys are closed-world (an unlisted key denies regardless of `default_policy` — R-AGENT-002), so a role that lacks a grant materializes a phantom `deny` that masked the real `allow` from a role that has it. Edges now merge to the **most-permissive** verdict (`allow > approval_required > deny`), tagged with the role(s) at that verdict — matching how tool edges already read. The per-edge `roles` list and `?role=` filter still give the exact per-role picture.
- **Report every role.** The graph response gains a `roles` field: the union of every role across all agents, independent of the `?role=` filter. The dashboard's role picker now sources its options from that, instead of a generic `*` resolve that only ever returned `default`. The `/graph` route drops its separate resolve round-trip entirely.
- **No filter flicker.** `usePolicyGraph` keeps the previous graph on screen through a role-change refetch (`keepPreviousData`), so the dropdown and canvas don't blank out on each switch.

## Try it

Open **Policies → graph** (or the full-page `/graph`) on the seeded compose demo. The `support_bot → billing_bot` reach edge now reads **allow** (green), and the role filter lists `billing`, `default`, `support` — the same roles the Playground offers.

## Important files

- `platform/api/hexgate_api/features/policy_modules/service.py` — `_graph_from` most-permissive merge + `roles` union collection.
- `platform/api/hexgate_api/schemas.py` — `PolicyGraphResponse.roles`.
- `platform/dashboard/src/components/policy_files/PolicyGraphDialog.tsx` — role picker sources from `graph.data.roles`.
- `platform/dashboard/src/lib/policy_files.ts` — `keepPreviousData` on `usePolicyGraph`.
- `platform/dashboard/src/routes/Graph.tsx` — drops the wrong `*` resolve.

## Notes

- Backend: two new tests (`test_policy_graph_union_keeps_most_permissive_reach`, `test_policy_graph_returns_all_roles`); `policy_modules` suite 61 passed.
- Reviewed with `/code-review high` — three findings fixed (a `tsc -b` fixture break, the filter flicker, a stale comment).


guillaume-hexamind (approved):
Good to me

PR #223, opened 2026-09-16
fix(dashboard): resolved-tab agent dropdown, inline folder create, legible graph edges

quanghexa94:
Three small quality-of-life fixes in the Policies editor, on top of the graph-correctness work in #222.

## What it does

- **Resolved tab: agent picker is now a dropdown.** The selector was a free-text input tucked in a thin bar above the inspector — easy to miss, easy to mistype. It's now a `<select>` listing `* (generic)` plus every agent the entry file declares, so you pick a real agent instead of typing its name. Agent names are parsed client-side (`composeAgentNames`, a `js-yaml` read of the entry's `agents:` keys — the authoritative source, since agents can only be declared in the entry), sourced from the entry's unsaved buffer so a just-added agent shows up before Save (debounced; the last good list is kept through a transiently-invalid mid-edit so the options don't flicker).
- **File tree: folders are created inline.** Clicking **New folder** used to pop a bare text input in a bar at the top of the explorer — detached from where the folder lands, and not obviously a folder. Now an inline folder-styled row (chevron + editable name) appears in the tree at the target spot: at the root for a top-level folder, or **inside** the target folder (which force-opens and stays open so the new folder is visible). You type just the leaf name; Enter commits, Escape cancels.
- **Policy graph: edges are legible at rest.** Resting edges were thin (`1.6`, and `1.1` for deny) and deny edges were dimmed to `0.32` opacity, so red/deny edges in particular were hard to see unless focused. Resting width now sits close to the focused `3.4` (`2.6` allow, `2.2` deny) and non-focus deny opacity is up to `0.6`. The focus state still stands out via its glow + full stroke.

## Try it

Open **Policies** on the seeded compose demo. In the **Resolved** tab, the Agent selector is a dropdown of the declared agents (`support_bot`, `billing_bot`) plus `* (generic)`. Click **New folder** (top bar, or a folder's hover button) — a new folder appears inline where you name it. Open the **graph** — edges read clearly at a glance, and the red `deny` edges are visible without hovering.

## Important files

- `platform/dashboard/src/lib/imports.ts` — new `composeAgentNames` helper (+ tests).
- `platform/dashboard/src/routes/Policies.tsx` — derives the picker's `agentNames` from the entry buffer.
- `platform/dashboard/src/components/policy_files/InspectorTabs.tsx` — the Resolved-tab dropdown.
- `platform/dashboard/src/components/policy_files/FileTree.tsx` — the inline new-folder row.
- `platform/dashboard/src/components/policy_files/PolicyGraphDialog.tsx` — resting edge width + deny opacity.

## Notes

- Tests: `composeAgentNames` unit tests; InspectorTabs agent-select test; FileTree inline-create (root + subfolder), keeps-parent-open, and Escape-cancel tests.
- Reviewed with `/code-review high` twice — findings fixed: entry-buffer source regardless of active tab, debounced + sticky parse, and the created-folder-stays-visible fix. All hook updates use the render-phase adjust-state pattern (lint-clean).


guillaume-hexamind (approved):
Look good to me; rebase before merge

PR #224, opened 2026-09-16
fix(platform-scripts): guard each migration against a table create_all has not built

guillaume-hexamind:
> Stacked on #220 (`gp/feat/migrate_runner`) — review that one first; this PR's diff is only the three SQL files plus two comment edits.
>
> **Design note:** [Migration stack — design note for reviewers](https://app.notion.com/p/3ddfb45dbae281edb8f8f1ebc669a999). This is PR 2 of that plan.

`ADD COLUMN IF NOT EXISTS` guards the **column**, not the table. So on a prod stack predating
#179, `0003` failed with:

```
ERROR:  relation "policy_file" does not exist
```

and because `platform-migrate` stopped there, every ClickHouse migration was skipped with it —
surfacing days later as the api crash-looping on `SchemaOutOfDate: llm_message`. #220 fixes the
skipping half. This fixes the half that made the file fail in the first place.

The Makefile already described this bug as an operating procedure — *"on a first-ever deploy …
this fails on the missing table. Skip it there."* That is a rule which depends on the operator
knowing how far behind a stage is. This PR makes it a property of the files instead.

## What changes

Each table's statements move inside a `to_regclass` guard:

```sql
DO $$
BEGIN
  IF to_regclass('public.policy_file') IS NULL THEN
    RAISE NOTICE 'policy_file absent -- create_all builds it complete; skipping';
    RETURN;
  END IF;
  ALTER TABLE policy_file ADD COLUMN IF NOT EXISTS created_by_user_id varchar;
  ALTER TABLE policy_file ADD COLUMN IF NOT EXISTS updated_by_user_id varchar;
END $$;
```

The skip is **correct, not merely convenient**: when the table is absent, `create_all` is about
to build it complete, actor columns included. The migration genuinely has nothing to do — today
it just cannot say so.

Two structural moves inside `0002`, both because the statements fail on a missing table exactly
like an `ALTER` does:

- the `CREATE INDEX` moves into the `devtoken` block;
- the four backfill `UPDATE`s move from the trailing "Backfills" section into their own table's
  block.

Also retired, since the guards make them false: the "upgrades only … skip it there" comment in
the Makefile, and the matching sentence in `DEPLOY.md` § 6.

`to_regclass` over `information_schema` because it returns NULL rather than raising, is
schema-qualified, and is one function call. A shared `hexgate_add_column()` helper was
considered and rejected — it puts a maintained function into every production database and
makes each file depend on a prior migration having run, so a file would no longer be
independently pasteable into `psql`. The verbosity is the cheaper cost.

## Scope

| File | Change |
|---|---|
| `0001_devtoken_soft_delete.sql` | 1 guard block |
| `0002_actor_columns.sql` | 9 guard blocks; index + 4 backfills moved inside |
| `0003_policy_file_actor_columns.sql` | 1 guard block |
| `Makefile` | the caveat comment |
| `platform/DEPLOY.md` | § 6, one sentence |

No Python, no TypeScript, no runner changes. **The set of columns after a full replay is
unchanged** — asserted below, not assumed.

## Verified

Against a live local Postgres. **The negative test ran first**, so the guards are shown to be
what makes it pass rather than something that was always green:

| # | Scenario | Before this PR | After |
|---|---|---|---|
| A | all 3 files vs an **empty** database | **all 3 FAIL** — `relation "devtoken" / "organization" / "policy_file" does not exist` | **all 3 exit 0** |
| B | simulated pre-release schema (`create_all`, drop the added columns, replay) | — | exit 0, guards taken |
| C | replay a second time | — | exit 0, `ix_devtoken_owner_user_id` count still 1 |
| D | migrated vs fresh `create_all` column set | — | **IDENTICAL** (name + type, all 14 tables) |

**The load-bearing check for the backfill move.** Moving `UPDATE … SET updated_at = created_at`
inside the same `DO` block that adds `updated_at` is the one genuinely risky part of this diff —
it depends on plpgsql preparing each statement on first execution rather than parsing the whole
block up front. Seeded a row with `created_at` three days old, replayed, and read it back:

```
updated_at=2026-09-13 09:49:02.051988+00  created_at=2026-09-13 09:49:02.051988+00
```

The backfill runs. It is not silently skipped.

**CI regex, explicitly checked** — `test_postgres_smoke.py:74` greps these files, and the plan
flagged indentation as a risk. It still matches 23 columns across 10 tables and 1 index, and
`pytest tests/core/test_postgres_smoke.py` passes against Postgres.

`make postgres-init` twice: clean, exit 0 both times. `make check-all` exits 0 — SDK 2690,
platform-api 802, dashboard 318. All four probe databases dropped afterwards; `pg_database`
confirms none remain.

## Deliberately not in scope

The **nullability divergence** — `organization.updated_at`, `organization_member.updated_at`,
`project.updated_at`, `role_binding.created_at` are nullable on migrated stages and `NOT NULL`
on fresh ones. I re-confirmed it is present and **unchanged** by this PR (the run in row D shows
exactly those four and nothing else). It is a defect in *what* the migrations add, not in
whether they can run, and it wants PR 4's test to find it before `0004` fixes it.

One caveat worth stating plainly: a guard converts *"this migration failed"* into *"this
migration did nothing"* for the missing-table case. Correct when `create_all` is about to build
the table; wrong if a table is missing for some **other** reason — a botched restore, a dropped
table. The `NOTICE` is the mitigation, and PR 5's Postgres startup check is the real backstop,
since a schema that ends up short would still refuse to boot.


platform/postgres/migrations/0002_actor_columns.sql
```diff
@@ -21,63 +21,142 @@
...
+-- two disagree wherever search_path leads elsewhere: the guard reports "absent"
+-- for a table the ALTER would have found, and the columns are skipped forever
+-- while create_all never adds them to a table that already exists.
 
 -- organization -------------------------------------------------------------
-ALTER TABLE organization ADD COLUMN IF NOT EXISTS created_by_user_id varchar;
-ALTER TABLE organization ADD COLUMN IF NOT EXISTS updated_at timestamptz;
-ALTER TABLE organization ADD COLUMN IF NOT EXISTS updated_by_user_id varchar;
+DO $$
+BEGIN
```
quanghexa94, line 44: **Nit (structural, non-blocking): the guard scaffold is copy-pasted 11× with the table name hand-repeated.** Each block repeats `DO $$ BEGIN / IF to_regclass('<t>') IS NULL THEN RAISE NOTICE / RETURN ... END $$`, with the table name typed twice — once in `to_regclass()` and again in the `ALTER`s. So "every table is guarded" is enforced by duplication, not a shared mechanism. A future block where the `to_regclass` name diverges from the `ALTER` target (or picks up a stray `public.` on only one side) would silently skip forever, and no test catches it.

I see the PR body already weighed and rejected a shared `hexgate_add_column()` helper (it would put a maintained function in every prod DB and break per-file pasteability into `psql`) — that's a reasonable call, so flagging this only so the tradeoff is on the record, not as a change request. PR 4's replay-from-empty test would be the natural place to catch a future name-divergence.

platform/postgres/migrations/0002_actor_columns.sql
```diff
@@ -21,63 +21,142 @@
...
+  IF to_regclass('organization') IS NULL THEN
+    RAISE NOTICE 'organization absent -- create_all builds it complete; skipping';
```
quanghexa94, line 46: **Latent: a missing table is an expected `NOTICE`, indistinguishable from real schema drift.** If a later edit adds a column to a migration but forgets it in `models.py`, then on a stage where that table is absent the guard `RETURN`s, `create_all` builds the table *without* the column, the run exits 0, and the gap stays invisible until a runtime read of the missing column errors. Verified all current columns DO match `models.py`, so this is latent, not active.

This is the accepted tradeoff the PR body states plainly — a guard converts "this migration failed" into "this migration did nothing" — and **PR 5's Postgres startup check is the real backstop** (a schema that ends up short refuses to boot). Noting it so the dependency on PR 5 is explicit: until it lands, this failure mode exits 0 silently.

quanghexa94 (approved):
Approving. Tightly scoped PR 2 of the migration-stack plan — per-table `to_regclass` guards so a migration skips instead of erroring on a stage where `create_all` hasn't built the table yet. No correctness bugs.

I cross-checked all 11 guard blocks: every `to_regclass('X')` matches the `ALTER`/`UPDATE`/`CREATE INDEX` target inside its block (including `devtoken` = `ApiKey.__tablename__`), the guarded columns and the `owner_user_id` index all match `models.py`, the unqualified-name/search-path reasoning is right, and the backfill `UPDATE`s kept their semantics after moving inside the `DO` blocks. The risky move (backfill inside the same block that adds the column) was negative-tested in the PR body. CI green.

Two low-severity, latent observations inline — both non-blocking, and both already answered by the design's PR-5 startup guard. Fine to defer.

platform/postgres/migrations/0002_actor_columns.sql
guillaume-hexamind: Agreed, and taking it as a PR 4 item rather than a change here.

Confirmed the state you describe: 11 blocks, the table name hand-written 3× per block (`to_regclass`, the `NOTICE` string, the statements). I parsed all 11 and cross-checked each `to_regclass('X')` against every `ALTER TABLE` / `UPDATE` / `CREATE INDEX … ON` target inside its own block — 11/11 match, no stray `public.` on either side. So it is latent, as you say.

One correction to the remedy, because it matters for how PR 4 gets written: **replay-from-empty would not catch this.** From an empty database every guard skips, so a typo'd `to_regclass` name is indistinguishable from a correct one — both paths `RETURN`. The divergence only manifests where the table *exists* and the guard wrongly says it doesn't, i.e. the scenario-B shape in the PR body (`create_all` → drop the added columns → replay → compare column sets).

So PR 4 gets two things instead of one:

1. A static consistency check, no live Postgres needed, runs in the default suite: parse each `DO` block, assert the `to_regclass` argument equals the set of table names in the statements it guards. ~15 lines, fails at PR time rather than only under a particular replay shape. Strictly stronger than any replay test for this failure mode.
2. The replay test covers scenario B, not just empty → replay.

Thanks — the tradeoff being on the record is the useful part, and this sharpened where the test has to sit.

platform/postgres/migrations/0002_actor_columns.sql
guillaume-hexamind: Checked this one against the suite, and the specific mechanism is already covered — though it pointed at a real gap next to it.

`test_hand_applied_migrations_match_the_live_schema` asserts `expected - live == ∅`, where `expected` is scraped from the migration SQL and `live` is `information_schema`. In CI that database is a **fresh `postgres:16` with only `create_all` applied** — the `platform-api-postgres` job (`tests.yml:141`) never replays the migrations. So the test is effectively asserting *migrations ⊆ models.py*: a migration column that `models.py` forgot makes CI red immediately, guards or no guards. That is true before and after this PR, so the exit-0-silently window you describe does not open.

What the guards genuinely add is the narrower case the PR body states — a table missing for some *other* reason (botched restore, dropped table) — where PR 5's startup check is indeed the backstop. Agreed on keeping that dependency explicit.

The gap worth naming is the **inverse** direction, which nothing covers today: a column in `models.py` with no migration. Fresh stacks get it from `create_all`, migrated stacks never do, and it surfaces as a runtime `UndefinedColumn` — exactly the shape of the `llm_message` incident, on the Postgres side. `expected - live` is one-directional and structurally cannot see it. Pre-existing and unchanged here; adding it to PR 4's scope alongside the guard-name check from the other thread.

PR #226, opened 2026-09-16
feat(dashboard): modernize the playground into an agentic chat

quanghexa94:
Rebuilds the Playground into a modern, agentic chat surface — closer to Claude/OpenAI — on top of the editor-UX work in #223.

## What it does

- **Immersive layout, no side panel.** The old 3-column grid (left config / chat / right decisions) is gone. It's a single centred chat column; agent name + live connection status live in the **top-left corner** (links to the agent, shows `reconnecting…`/`offline`/`no agent serving`), and the **Acting as** role picker — now a compact pill → popover — sits in the **top-right** with a reset button.
- **Modern composer.** A pill-shaped input with a **circular send button inside** it (up-arrow), that shrinks when empty.
- **Agent turn = steps + tool calls.** The assistant message reads as a compact timeline: reasoning (muted), each tool call as a step card, then the final answer with a blinking stream caret. User turns are right-aligned bubbles.
- **Floating decision deck.** Streamed policy decisions collapse into a card pile bottom-right with a count badge; it **opens on hover or tap** (keyboard-reachable, `aria-expanded`) into the full list — verdict, args, output, timing per card — instead of claiming a permanent panel. Hidden while an approval is pending so it can't overlap Approve/Deny.
- **Thinking animation.** Ported the hexkit **goo/metaball loader** (three dots fused via an SVG filter) + a shimmering "Thinking…" label + a blinking caret, retargeted to this app's `--primary`/`--foreground` tokens, with a `prefers-reduced-motion` guard. Plus a subtle message-entrance rise.

## Try it

Open **Playground** with an agent served (`hexgate serve`). Send a message — watch the goo loader, the tool-call steps, and the decision deck fan out (hover on desktop, tap on touch). The whole thing adapts to light/dark and the blue/plum accent.

## Important files

- `platform/dashboard/src/routes/Playground.tsx` — the rewrite.
- `platform/dashboard/src/index.css` — goo / shimmer / caret / entrance keyframes (`.pg-*`), reduced-motion guarded.
- `platform/dashboard/src/routes/Playground.test.tsx` — new 9-test component suite (mocks the live hook).

## Notes

- The `usePlayground` hook and its wire protocol are untouched — this is a view-layer change. The existing hook tests still pass; the new suite covers send, the acting-as control, tool steps, the thinking loader, the deck (count + click-toggle + hidden-during-approval), and the offline notice.
- Reviewed with `/code-review high` — four interaction findings fixed (deck/approval overlap, hover-only decisions, per-token scroll-yank, mid-session offline hint).
- Screenshots to follow once eyeballed on a running instance.


guillaume-hexamind:
## Code review — `platform/dashboard/src/routes/Playground.tsx`

Reviewed at `067a389`. Tests and types are green on the head: `vitest run src/routes/Playground.test.tsx` → 10/10, `tsc -b --noEmit` clean, `eslint` reports only two pre-existing `set-state-in-effect` warnings on effects this PR didn't introduce.

Seven findings below, each scored /100 by severity. Findings 1–4 were re-derived against the layout arithmetic (sidebar width, flex min-size resolution, box measurements) rather than read off the diff. All line numbers are `platform/dashboard/src/routes/Playground.tsx` at the head commit.

---

### 1. Approval stack lost its height cap — composer and lower Approve/Deny get clipped · **78/100**

`ApprovalPromptStack` was `max-h-[45%] overflow-y-auto scrollbar-thin` before this PR; at `Playground.tsx:721` it is now `mb-3 space-y-2 rounded-xl border … p-3` — no cap, no scroll container.

That stack lives in the footer of a `h-screen flex flex-col overflow-hidden` root (`:134`). The transcript is `flex-1 overflow-y-auto` (`:174`), so its automatic minimum size resolves to 0 and it shrinks away entirely; the footer is a plain flex item whose `min-height: auto` is its content height, so it grows past the viewport and the root clips it — with no scrollbar anywhere to recover. `h-screen` also means `<main className="flex-1 overflow-y-auto">` in `AppShell.tsx:340` won't scroll to rescue it, since the root's border box never exceeds `main`.

Measuring the actual classes: `p-3` (24) + header (~20) + the `mt-2 max-h-40` args pre (168, `:809`) + `mt-2 h-7` buttons (36) ≈ 248px per card, plus ~56px of stack chrome and an 80px composer band. Three concurrent approvals at the full args cap ≈ 820px, which overflows an 800px viewport; two cards clip on a window under ~650px tall. The file's own comment at `:702` notes N concurrent approvals are expected from parallel tool calls. When it hits, the buttons are unreachable and the approvals fail closed at TTL expiry.

**Suggested fix:** restore the cap on the stack wrapper — `max-h-[45vh] overflow-y-auto scrollbar-thin` alongside the new `rounded-xl border … p-3` styling.

---

### 2. Decision deck covers the composer's Send button below ~1344px window width · **80/100**

The deck is `absolute bottom-24 right-4 z-20 … sm:bottom-6 sm:right-6` (`:584`) with an `h-[52px] w-[216px]` handle (`:620`), so it occupies 24–240px from the right and 24–76px above the root's bottom edge.

The composer's Send button is `absolute right-2 grid size-9` (`:422`) inside a `mx-auto w-full max-w-2xl px-4` container. Root width is window − 240 (sidebar, `AppShell.tsx:269`; the root's `-mx-8` cancels `main`'s `px-8`), so Send's right edge sits at `(rootWidth − 672) / 2 + 24`. The two overlap horizontally whenever that margin is under 216px — **window narrower than ~1344px with the sidebar expanded, ~1160px collapsed.** That catches 1366×768 and 1280×800, the two most common laptop widths, plus any non-maximized window.

Vertically they overlap too: the composer band is `py-4` + `h-12`, and the 36px Send button is centred in the 48px input, so it occupies 22–58px above the bottom against the deck's 24–76px. `z-20` beats the composer band's `z-10`, and the deck is later in DOM regardless — so a click on Send toggles the deck's `pinned` state instead of sending.

The PR already mutes the deck to keep Approve/Deny clickable (`:215`); the composer needs the same consideration. Note the mobile branch is fine — `bottom-24` (96px) clears the ~80px composer band.

**Suggested fix:** lift the deck above the composer band (`sm:bottom-24`), or right-align it outside the composer's max-width, or mute it while the composer is focused / non-empty.

---

### 3. `atBottomRef` is never re-armed on send, so auto-follow stays stuck off · **62/100**

`atBottomRef` is declared `true` at `:122`, written **only** in the transcript's `onScroll` at `:172`, and read at `:124`; `submit()` (`:128`) never touches it.

Concrete case: the user scrolls up to re-read an earlier turn (ref → false), then types and hits Enter. Their own message and the whole streamed reply land off-screen with no movement at all — the send looks like a no-op, and nothing re-arms the ref until they manually scroll back down. Sending should always re-pin to the bottom.

(The reset path self-heals — when the transcript's content shrinks, browsers clamp `scrollTop` and fire a `scroll` event, which flips the ref back to true against the empty state. Only the send path needs fixing.)

**Suggested fix:** `atBottomRef.current = true;` in `submit()`, before `sendChat`.

---

### 4. `aria-hidden` container whose `<button>` stays in the tab order · **55/100**

When `muted`, the deck wrapper gets `aria-hidden={true}` plus `pointer-events-none opacity-0` (`:582`–`:585`), but the handle button at `:616` has no `disabled` and no `tabIndex={-1}`. Neither `opacity-0` nor `pointer-events-none` removes an element from the tab order, so a keyboard user tabbing past the composer during a pending approval lands on an invisible, `aria-hidden` control — focus vanishes visually, screen readers announce nothing, and Enter still flips `pinned`. This is the axe `aria-hidden-focus` rule.

The expanded panel is already handled correctly (`tabIndex={shown ? 0 : -1}` at `:593`); only the handle is missing it. Worth noting the new test at `Playground.test.tsx:189` can't catch this: `queryByRole` excludes `aria-hidden` subtrees, so it passes regardless.

**Suggested fix:** add `disabled={muted}` (or `tabIndex={muted ? -1 : 0}`) to the handle button.
---

### Checked and clean

The `stateIcon`/`verdictColor` extraction is behaviour-preserving; role derivation and the `sendChat(text, { roles })` wiring are untouched; `hasSteps` is safe (`reasoning` is a non-optional `string` defaulting to `""` in `lib/playground.ts`); the removed lucide imports (`Radio`, `Send`, `User`, `MessageSquareCode`) have no remaining references; and the reduced-motion guard covers every new keyframe.

One non-bug note: the "Playground docs" link and the page's `<h1>` now live only on the empty state, so both disappear once a session starts — intentional?

guillaume-hexamind (commented):
A couple of auto review comments above.

Otherwise all seem good, will approve once addressed

quanghexa94:
All four addressed in `1834b80c`. Dashboard suite green (335/335), `tsc` clean, prettier clean, eslint unchanged (only the two pre-existing `set-state-in-effect` warnings).

1. **Approval stack cap** — restored a cap: the header stays pinned and the cards scroll in a `max-h-[45vh] overflow-y-auto scrollbar-thin` region, so N concurrent approvals can't overflow the root and push the composer / Approve-Deny out of reach.
2. **Deck vs Send** — the deck now mutes while the composer has text (the same mechanism it already uses for a pending approval), so a Send click can't land on it. I chose *non-empty* over *focused*: an empty composer has nothing to send (Send is disabled), and muting on mere focus would hide the deck through most of a session.
3. **Auto-follow on send** — `atBottomRef.current = true` in `submit()` before `sendChat`, so a message sent after scrolling up (and its streamed reply) re-pins to the bottom.
4. **Deck handle a11y** — the muted handle is now `disabled`, so it leaves the tab order and Enter can't flip `pinned`. Added a DOM-level test (since `getByRole` can't see it under `aria-hidden`) plus a deck-muted-on-composer-text test.

Non-bug note: left the "Playground docs" link + `<h1>` on the empty state only, as in the current design — happy to keep them visible mid-session if you'd prefer.

guillaume-hexamind (approved):

PR #227, opened 2026-09-16
build(platform-api): rebuild the SDK wheel on every image build

guillaume-hexamind:
> Stacked on #224 (`gp/feat/migration_table_guards`), itself on #220. Review those first — this PR's own diff is **one file**.
>
> **Design note:** [Migration stack — design note for reviewers](https://app.notion.com/p/3ddfb45dbae281edb8f8f1ebc669a999). This is PR 3 of that plan.

Not a migration bug. It blocked the same deploy twice, from the same root cause, and belongs
in the same batch.

The API image installs the SDK as a **built wheel** (`--no-editable`) behind a persistent
BuildKit cache mount. uv keys its cache for a path dependency on the package **version** —
never on source content. That version has sat at `0.3.1` since `5eace3d` while `ade047e`
(#186), #187 and #189 all changed `hexgate/`. So a rebuild re-ran `uv sync`, uv found `0.3.1`
in cache, and the image shipped a wheel built from pre-#186 source:

```
ImportError: cannot import name 'TOOL_CALL_JSON_KEYS' from 'hexgate.audit'
```

Two things made it hard to see in the field, and both are worth knowing:

- **Only the enricher crashed**, though `api` and `enricher` share one image. The API's import
  graph never reaches the symbol; the enricher's `dlq.py` and `enforcement.py` do.
- **The enricher's healthcheck is `disable: true`**, so a crash-looping container reported
  "Up" the whole time.

## What changes

```dockerfile
uv sync --frozen --no-dev --no-editable --refresh-package hexgate
```

Scoped to the one package, so every third-party dep still comes from the cache mount. The
added build cost is one wheel build of a pure-Python package.

Plus a **build-time assertion** that the installed wheel matches the source copied into the
image. A flag on its own reads as redundant and gets deleted by the next person tidying the
Dockerfile — which is exactly how this returns. The comment explains why it is load-bearing;
the assertion makes removing it fail the build instead of a production container.

## Verified

**The bug, reproduced from scratch** in a throwaway path-dependency project with a fixed
`UV_CACHE_DIR` — same shape as the image, no relationship to this repo:

| Build | Action | Installed symbols |
|---|---|---|
| 1 | `uv sync --no-editable`, source has `CONST_A` | `CONST_A` |
| — | add `CONST_B`, **leave version at 0.1.0** | — |
| 2 | delete venv, plain `uv sync --no-editable` | **`CONST_A` only** ← the enricher `ImportError` |
| 3 | delete venv, sync **`--refresh-package`** | **`CONST_A`, `CONST_B`** ← the fix |
| 4 | add `CONST_C`, delete venv, plain sync again | **`CONST_A`, `CONST_B`** ← stale again |

Build 4 is the one I added beyond the plan, and it settles a real question: `--refresh-package`
repairs the cache entry, so a plain sync immediately afterwards looks fine. **The flag has to be
permanent, not a one-off repair** — the next source change without a version bump goes stale
again.

**The real image**, built `--platform linux/amd64`:

- `grep -c TOOL_CALL_JSON_KEYS` on the installed `site-packages/hexgate/audit.py` → **`1`**
  (it was `0` on the stale builds).
- The enricher's actual import — `from hexgate_api.jobs.enricher import dlq, enforcement` →
  `enricher imports OK`.

**The assertion, tested in both directions** — a check that cannot fail is worse than none:

| | Result |
|---|---|
| Good build | `SDK wheel matches the source in this image`, build succeeds |
| Source perturbed after sync | `Files ./audit.py and …/hexgate/./audit.py differ` → **build exits 1** |

Two details the plan flagged as unverified, now settled against the built image:

- **`python` had to become `uv run --no-sync python`.** The plan's draft used bare `python`; at
  that stage the venv is not on PATH and it raises `ModuleNotFoundError: No module named
  'hexgate'`. The plan's snippet would have silently needed fixing.
- **`diff -r` on the package would fail for a benign reason** — `hexgate/CLAUDE.md` is in the
  source tree and not packaged. The assertion compares `.py` files only.

`make check-all` exits 0 — SDK 2690, platform-api 802, dashboard 323. All probe images and the
scratch project deleted.

## Out of scope

The **enricher healthcheck** (`disable: true` in `docker-compose.deploy.yml`) is why a
crash-looping enricher reported "Up" through both incidents. Already a known gap, but this
rollout is the second time it hid a crash-loop, and it deserves its own issue rather than a
quiet fix here.


quanghexa94 (commented):
Code review (high effort). No correctness bugs — `--refresh-package hexgate` targets the right package, the `if ! diff …; exit 1; fi` structure is sound, `diff -r`'s asymmetric reporting genuinely covers both directions, the `-x` excludes nest correctly, and an `import hexgate` failure surfaces its own traceback rather than the misleading message. Verified against the base image (GNU diffutils 3.8), the packaging config (`packages.find include = ["hexgate*"]`, no `package_data`), and the source tree (only non-`.py` file under `hexgate/` is `CLAUDE.md`, excluded). Nicely reproduced-from-scratch and both-directions-tested. Two low-severity fragility notes inline; neither blocks.

platform/api/Dockerfile
```diff
@@ -47,8 +47,27 @@ COPY hexgate ./hexgate
...
-    uv sync --frozen --no-dev --no-editable
+    uv sync --frozen --no-dev --no-editable --refresh-package hexgate
+
+# Guards the flag above: fail the build, not the deploy, if the installed wheel
+# drifts from the source. `uv run` because the venv is not on PATH here. Compared
+# both ways and across every file type, so a module missing from the wheel, one
+# deleted from the source but retained in it, and drifted package data all fail
+# the build. CLAUDE.md is in the source tree but deliberately not packaged.
+RUN INSTALLED="$(uv run --no-sync python -c 'import hexgate, os; print(os.path.dirname(hexgate.__file__))')" \
+    && if ! diff -r -x 'CLAUDE.md' -x '__pycache__' /app/hexgate "$INSTALLED"; then \
```
quanghexa94, line 65: **Fragility (non-blocking): the guard diffs the whole `hexgate/` tree against a hardcoded exclusion allowlist.** It compares the entire source tree to the installed package, excluding only `CLAUDE.md` and `__pycache__`. Any *future* source file that setuptools legitimately doesn't package will break **every** `platform-api` image build until someone adds another `-x`.

Concrete trigger: adding `hexgate/py.typed` — the expected PEP 561 marker for a typed library, and **not** auto-included by setuptools without `package_data`. A routine SDK edit would then fail all API builds with "installed hexgate SDK differs from the source" — an image failure caused by an unrelated change. Not reachable today (the only non-`.py` file is `CLAUDE.md`), but this guard's whole job is to protect future edits, so it's the one edge that undercuts its own purpose.

Deeper fix: verify against the wheel's own `RECORD`/manifest (what packaging actually ships) rather than the whole source tree with a maintained exclusion list. Cheaper stop-gap: a comment here saying "add an `-x` when a new non-packaged file lands under hexgate/ (e.g. py.typed)" so the next person isn't surprised.

platform/api/Dockerfile
```diff
@@ -47,8 +47,27 @@ COPY hexgate ./hexgate
...
 RUN --mount=type=cache,target=/root/.cache/uv \
-    uv sync --frozen --no-dev --no-editable
+    uv sync --frozen --no-dev --no-editable --refresh-package hexgate
+
+# Guards the flag above: fail the build, not the deploy, if the installed wheel
+# drifts from the source. `uv run` because the venv is not on PATH here. Compared
+# both ways and across every file type, so a module missing from the wheel, one
+# deleted from the source but retained in it, and drifted package data all fail
+# the build. CLAUDE.md is in the source tree but deliberately not packaged.
+RUN INSTALLED="$(uv run --no-sync python -c 'import hexgate, os; print(os.path.dirname(hexgate.__file__))')" \
```
quanghexa94, line 64: **Minor: a corrupted `INSTALLED` value reports as wheel/packaging drift.** If `uv run --no-sync python -c …` ever emits anything to stdout before the path (a uv notice, a deprecation print), `INSTALLED` becomes a non-directory string, `diff -r` exits **2** ("trouble", not "differ"), and the guard prints the same "stale wheel / unpackaged file" message — sending the operator toward `--refresh-package` / pyproject packaging instead of the real cause. Low probability given `--no-sync`. Cheap hardening: assert `INSTALLED` is a directory before the diff, or branch on `diff`'s exit code (1 = differ, 2 = error).

quanghexa94 (approved):
Approving. PR 3 of the migration-stack plan — one-file Dockerfile change adding `--refresh-package hexgate` so a source change without a version bump can't ship a stale cached wheel, plus a build-time assertion that the installed wheel matches the source in the image. No correctness bugs: the flag targets the right package, the `if ! diff …; exit 1; fi` structure is sound, `diff -r` covers both directions, and the excludes nest correctly. Well reproduced-from-scratch and both-directions-tested. CI green.

The two inline notes are latent fragility, non-blocking:
- the whole-tree diff against a hardcoded `-x` allowlist will break all API builds the day a `py.typed` (or any non-packaged file) lands under `hexgate/`;
- a corrupted `INSTALLED` value makes `diff` exit 2 and misreport as wheel drift.

Both worth a follow-up but neither blocks. Note this is stacked on #224, which should land first.

PR #228, opened 2026-09-16
test(platform-api): execute the migrations in CI against real Postgres and ClickHouse

guillaume-hexamind:
> Stacked on #227 (`gp/feat/sdk_wheel_refresh`) → #224 → #220. Review those first.
>
> **Design note:** [Migration stack — design note for reviewers](https://app.notion.com/p/3ddfb45dbae281edb8f8f1ebc669a999). This is PR 4, the one that turns "we found it in prod" into "CI found it".

Nothing in CI has ever run a migration file. The check that *looks* like coverage —
`test_postgres_smoke.py` — regexes `ADD COLUMN` out of the SQL and asserts those names exist
in `information_schema`, on a database `create_all` had just built. **The columns are there
because the models declare them.** A migration that cannot execute at all passes it.

That is how `relation "policy_file" does not exist` reached a prod deploy.

## What changes

Three tests per store, each on its own scratch database so they cannot order-couple:

| Test | Asserts |
|---|---|
| **Replay from empty** | Every file, in order, against a database with no tables. The executable form of #224's guard convention. |
| **Equivalence** | `create_all` → snapshot A. `create_all`, drop what the migrations add, replay → snapshot B. **A == B**, on columns, types, nullability *and* indexes. |
| **Replay twice** | Idempotency — asserted in every migration header, verified nowhere. |

Plus `0004_actor_column_not_null.sql`, and the CI wiring (a ClickHouse service alongside the
Postgres one that job already had).

## The defect the equivalence test found on its first run

```
only after migrating=[('organization','updated_at','timestamp with time zone','YES'), …]
only in create_all=[('organization','updated_at','timestamp with time zone','NO'),  …]
```

`create_all` emits `NOT NULL` for `organization.updated_at`,
`organization_member.updated_at`, `project.updated_at` and `role_binding.created_at` — the
models declare them non-`Optional`. `0002` added them as bare `timestamptz` and backfilled
values but never issued `SET NOT NULL`, **so staging and prod are nullable on all four where
every fresh database is not.**

Exposure is low: every application write goes through the model's `default_factory`, so the
API cannot store a NULL. A manual `INSERT` or a restore could, and such a row would fail to
hydrate. `0004` closes it, re-running the backfill first so each block is correct standalone
rather than only in sequence after `0002`.

It sat in two production databases and no amount of *reading* the files surfaced it.

## Two things the plan assumed that turned out to be wrong

Both were found by running, and both changed the implementation:

- **`exec_driver_sql` cannot run these files.** SQLAlchemy routes it through asyncpg's
  prepared-statement path, which rejects a multi-command string with `cannot insert multiple
  commands into a prepared statement`. asyncpg's own `execute` uses the simple query protocol
  and takes a whole file. (Splitting on `;` is not an option either — every Postgres migration
  is now `DO $$ … END $$;`.)
- **The ClickHouse files cannot be split on `;`.** `--multiquery` is a clickhouse-client
  feature; the HTTP interface takes one statement per request, so the test has to split.
  Neither naive approach survives: `--` comments carry prose full of semicolons, *and*
  `COMMENT 'SDK-truncated JSON snapshot; may be lossy'` puts one inside a string literal.
  `split_statements` tracks both states and has its own database-free unit test.

## Verified — every test shown red before green

A passing suite proves nothing unless the tests have been seen to fail. Each was broken
deliberately:

| Test | Broken by | Failed with |
|---|---|---|
| Replay from empty | removing #224's guard from `0003` | `UndefinedTableError: relation "policy_file" does not exist` |
| Equivalence (PG) | not yet writing `0004` | the four-column nullability mismatch above |
| Replay twice | `IF NOT EXISTS` → bare `ADD COLUMN` | `DuplicateColumnError: column "revoked_at" … already exists` |
| Equivalence (CH) | `message_seq UInt32` → `UInt64` in the migration | `only migrated=[('llm_message','message_seq','UInt64')] only fresh=[… 'UInt32')]` |

**The ClickHouse row is the one worth reading.** It passed at first *despite* the injected
drift — `schema.sql` already builds every table, so `0003`'s `CREATE TABLE IF NOT EXISTS` is a
no-op and its column definitions never execute. The test was proving nothing. It now drops
what the migrations add before replaying them, mirroring the Postgres design, and then it
catches the drift. This is the enforceable form of `0003`'s "keep this byte-identical to
`init/schema.sql`" header, which nothing checked.

`make check-all` exits 0 — SDK 2690, platform-api 803, dashboard 323. All scratch databases
dropped; `pg_database` and `SHOW DATABASES` confirm none remain.

## Note on `0004`'s guard style

`0004` uses `to_regclass('organization')` **unqualified**, matching the `search_path` fix on
#224's branch rather than the `public.`-qualified form currently on this base. When #224
merges, the whole directory is consistent. Writing it qualified would have added a fourth file
carrying a bug that was already fixed one branch up.

## Scope note

The plan permitted splitting ClickHouse into a PR 4b. I kept it here — `clickhouse/0003` is
the migration that took both stages down, and the whole argument of this PR is that
unexecuted SQL is unverified SQL.

## Out of scope

A migration ledger and the `create_all`/migration split itself — that is PR 7 (Alembic). This
PR makes the current design *verifiable*; it does not replace it.


quanghexa94 (commented):
Code review (high effort). Strong PR — the Postgres side is solid (0001–0003 already `to_regclass`-guarded, empty-DB replay passes, and `0004` correctly closes the fresh-vs-migrated nullability divergence), and `split_statements`' doubled-quote handling works. One finding I'd treat as blocking (#1): a regex hole silently guts the ClickHouse equivalence test for 5 of 6 `run_*` columns — the exact 'passes but proves nothing' failure this PR exists to eliminate. I traced it against `clickhouse/migrations/0002` and confirmed it. The other four are low-severity test-hygiene / naming follow-ups. Details inline.

platform/api/tests/core/test_migrations.py
```diff
@@ -0,0 +1,455 @@
...
+@clickhouse_only
+def test_clickhouse_migrations_replay_onto_a_fresh_schema() -> None:
+    with scratch_clickhouse_database("hexgate_ch_replay") as database:
+        with clickhouse_client() as client:
+            run_clickhouse_script(client, CLICKHOUSE_INIT_SCHEMA.read_text(), database)
+            for path in migration_files(CLICKHOUSE_MIGRATIONS):
+                run_clickhouse_script(client, path.read_text(), database)
+
+
+CH_ADD_COLUMN_RE = re.compile(
```
quanghexa94: **🔴 Blocking: the ClickHouse equivalence test silently skips 5 of the 6 `run_*` columns.** `CH_ADD_COLUMN_RE` matches one `ALTER TABLE …` up to a literal `;`, with each clause body captured by the non-greedy `[^;]*?` — which cannot cross a semicolon. But `clickhouse/migrations/0002` line 30 is `COMMENT 'RunFacts.id; zero when…'` — a semicolon **inside a string literal**. So on the `policy_decision` ALTER (6 columns in one statement) the match terminates at that in-string `;`, and `CH_COLUMN_NAME_RE.findall(clauses)` returns **only `run_id`**.

`strip_clickhouse_objects` therefore drops only `run_id` and leaves `run_tool_calls`, `run_llm_calls`, `run_denials`, `run_total_tokens`, `run_elapsed_ms` in place from `schema.sql`. When the equivalence test replays `0002`, those five `ADD COLUMN IF NOT EXISTS` are no-ops, their migration DDL never executes, and `A == B` holds for them **regardless of whether `0002` and `schema.sql` agree**. That's the same 'passes but proves nothing' hole this PR sets out to close — on the very table (`policy_decision`) whose ClickHouse migration took both stages down.

Not a prod bug (the migrations are consistent today), but the guarantee is quietly false for those columns. Fix options: strip `--` comments and string literals before the ADD-COLUMN regex (reuse the comment/string state-tracking that `split_statements` already has), or collect column names across the whole `ALTER` body rather than terminating the capture at the first raw `;`.

platform/api/tests/core/test_migrations.py
```diff
@@ -0,0 +1,455 @@
...
+
+
+def run_clickhouse_script(client, sql: str, database: str) -> None:
+    """Apply one file, retargeted onto a scratch database.
+
+    Every statement is qualified ``hexgate_audit.<table>`` — the deploy runner
+    relies on that — so pointing a file elsewhere rewrites the prefix rather
+    than switching the session's default database.
+    """
+    retargeted = sql.replace(f"{CLICKHOUSE_DATABASE}.", f"{database}.")
```
quanghexa94, line 357: **Low: `run_clickhouse_script` retargets only `{CLICKHOUSE_DATABASE}.` (with the trailing dot),** so `schema.sql`'s leading `CREATE DATABASE IF NOT EXISTS hexgate_audit;` runs unretargeted. Benign no-op in CI, but a developer who points `HEXGATE_CLICKHOUSE_HOST` at a shared/dev ClickHouse ends up creating/touching the real `hexgate_audit` database outside the scratch sandbox, and it's never dropped. Compounds with the gating point below.

platform/api/tests/core/test_migrations.py
```diff
@@ -0,0 +1,455 @@
...
+)
+
+CLICKHOUSE_DATABASE = "hexgate_audit"
+AUDIT_TABLES = ("policy_decision", "llm_invocation", "llm_message", "ban_enforcement")
+
+pg_only = pytest.mark.skipif(
+    "postgres" not in os.environ.get("DATABASE_URL", ""),
+    reason="set DATABASE_URL to a Postgres DSN to run (see `make postgres-up`)",
+)
+clickhouse_only = pytest.mark.skipif(
```
quanghexa94: **Low: the ClickHouse tests are gated only by `HEXGATE_CLICKHOUSE_HOST`, not the repo's opt-in `integration` marker** (`pyproject`: `integration: requires a running ClickHouse (opt-in)`), which exists for exactly this. A developer who exports `HEXGATE_CLICKHOUSE_HOST` for ordinary platform work would then silently run scratch `CREATE/DROP DATABASE` against their dev ClickHouse during a normal `make platform-api-check`, outside the intended gate. Marking these `integration` would align them with the rest of the suite (and, with the retarget point above, keep stray writes off a dev instance).

platform/api/tests/core/test_migrations.py
```diff
@@ -0,0 +1,455 @@
...
+#
+# The store whose migration took both stages down. It has no create_all twin —
+# init/schema.sql is the greenfield path — so the equivalence test compares a
+# volume built from schema.sql alone against one built from schema.sql plus the
+# migrations. That is the enforceable form of 0003's "keep this byte-identical
+# to init/schema.sql" header, which nothing checked.
+
+
+@contextmanager
+def clickhouse_client(database: str | None = None) -> Iterator[object]:
```
quanghexa94: **Low (altitude): `clickhouse_client` re-implements connection/env parsing already centralized in `core/clickhouse.get_clickhouse`,** with its own literal host/port/user/password defaults (`8124`, `hexgate`, `hexgate-dev-password`). If the platform's real connection defaults change, this copy diverges silently and the tests connect somewhere other than the app does. Reusing the shared builder would keep them in lockstep.

.github/workflows/tests.yml
```diff
@@ -140,8 +140,10 @@ jobs:
 
   platform-api-postgres:
     # The platform-api job runs the whole suite on SQLite, never exercising
-    # asyncpg. This job sets DATABASE_URL and runs only the opt-in smoke test
-    # — the sole coverage that the schema builds on Postgres.
+    # asyncpg. This job sets DATABASE_URL and runs the opt-in Postgres tests:
+    # the schema-builds smoke test, and the migration tests that EXECUTE
+    # platform/{postgres,clickhouse}/migrations rather than pattern-matching
```
quanghexa94, line 145: **Low (naming): the job's display name `Platform API (Postgres smoke)` is now stale** — this job runs migration-execution tests against real Postgres **and ClickHouse**, not just a Postgres smoke test. You updated the comment right above but not the `name:`. The sole CI coverage for `clickhouse/migrations` now hides behind a check name that says the opposite; anyone scanning the check list for ClickHouse coverage concludes there is none. Suggest something like `Platform API (migration execution)`.

guillaume-hexamind:
All five review points applied in b18646d.

**The blocking one — you were right, and it mattered.** `CH_ADD_COLUMN_RE` terminated at the first raw `;`, and `0002`'s *first* column carries `COMMENT 'RunFacts.id; zero when…'`. So the parser returned `run_id` alone; the other five `run_*` columns were never dropped, replayed as no-ops, and drift in them could not fail the test. Reproduced your scenario — `run_denials UInt32` → `UInt64`:

- before: `1 passed`
- after: `only migrated=[('policy_decision','run_denials','UInt64')] only fresh=[…'UInt32')]`

Column collection now runs statement-at-a-time through `split_statements`, so no pattern hunts for its own terminator in raw text. Added a database-free test pinning all six columns — my original negative test only passed because `message_seq` happened to be a column the broken regex reached, which is exactly how this got through.

The rest:

- **`CREATE DATABASE`** — skipped rather than retargeted (the prefix rewrite never matched it, no trailing dot). Verified: `SKIPPED: ['CREATE DATABASE IF NOT EXISTS hexgate_audit']`, no remaining statement names the real database.
- **`integration` marker** — adopted. CI needed a *second* step rather than a widened selector: `-m integration` across the whole suite pulls in tests wanting a seeded schema this job doesn't build, so it's scoped to `tests/core/test_migrations.py -m integration`.
- **Connection defaults** — now from `get_settings()`. Did not reuse `get_clickhouse` itself: it's `lru_cache`d onto a single database and these tests need several scratch ones, so `database` stays a parameter. Noted in the docstring.
- **Job name** — `Platform API (migration execution)`. ⚠️ This changes a CI check name; if branch protection references the old string, that rule needs updating or the required check will read as permanently missing.

`make check-all` green (SDK 2738, platform-api 804, dashboard 323). Scratch databases confirmed dropped from both stores.

quanghexa94 (approved):
Approving — the update resolves everything from my last review, including the blocker.

- **#1 (was blocking): fixed.** The semicolon-terminated mega-regex is gone; `clickhouse_objects_added` now splits via the comment/string-aware `split_statements` and collects every ADD COLUMN name per statement. I verified all six `policy_decision` columns are now collected, and `test_every_added_clickhouse_column_is_collected` pins the exact bug against regression. The equivalence test now actually exercises the DDL it claimed to.
- **#2 fixed** — `run_clickhouse_script` skips `CREATE DATABASE`, so nothing escapes the scratch sandbox.
- **#3 fixed** — the ClickHouse tests are gated by the repo's `integration` marker.
- **#4 fixed** — job renamed `Platform API (migration execution)`.
- **#5 addressed** — `clickhouse_client` builds from `get_settings()`; keeping it separate from the `lru_cache`d `get_clickhouse` singleton is well-reasoned and documented.

Also confirmed the anchored `^` matchers are safe: `split_statements` strips `--` comments inline and `.strip()`s each statement. CI green. This is PR 4 of the plan and the one that turns "found it in prod" into "CI found it" — nicely done. Rebase onto main before landing (currently BEHIND).

PR #230, opened 2026-09-17
feat(platform-scripts): seed llm_message transcripts alongside decisions

victorludvig:
## What is Changing

`platform/scripts/seed_audit.py` seeds a second table. Each run's decisions now also produce the conversation they came out of — one `llm_message` event before each decision, carrying the completion that asked for the tool, plus one after the last for what followed, so a run of N decisions has N+1 message events.

- Rows come from the API's own `_llm_message_row`, the same way the decision rows already reuse `_decision_row` — the seed names no column beyond splicing `received_at` in.
- `turn_key` is the run id plus the agent name, as the adapters build it, so `message_seq` is contiguous within one message list.
- `clear()` drops both tables together; leaving transcripts behind would show a conversation for a decision that no longer exists.
- `_validate_columns()` now fails fast, naming `make clickhouse-migrate`, on a database that predates migration 0003 — otherwise it surfaces as a mid-seed insert error with half the rows written.

The Audit decision drawer (#195) reads a decision's transcript from `llm_message`. Seeded decisions alone leave its new section empty for every row, so there is currently no way to look at that feature against realistic data — only unit tests with mocked rows, and the OTLP smoke run's six synthetic events.

The imperfections are seeded deliberately, so every marker the drawer can render has data behind it rather than only a test fixture: a `message_seq` gap, a resynced turn, and a truncated field. One 120-turn conversation is seeded for the same reason — normal runs stop at four decisions (the anomaly detector's threshold bounds them), so nothing else crosses the drawer's 50-row page, and the path where it has to fetch the transcript's tail instead of its head would never be taken.

## Tests

`ruff check` and `ruff format --check` clean. Run against local ClickHouse:

```
cd platform/api && uv run python ../scripts/seed_audit.py --number_users 3
→ Inserted 1048 decision rows (1020 normal + 28 anomalous) and 1313 llm_message rows.
```

Verified in the seeded rows: 39 `turn_key`s with a non-contiguous `message_seq`, 45 resynced, 16 truncated, and one 121-turn session whose deny decision anchors at turn 113 — past the first page, which is the case the drawer's windowing exists for.


guillaume-hexamind:
## Review findings

Three issues, each scored out of 100. Findings 1 and 2 share a root cause and one fix; finding 3 is independent. Scores reflect a verification pass — each claim below was reproduced by running `build_rows()` against this branch.

---

### 1. `MESSAGE_LEAD_SECONDS` exceeds the minimum gap between decisions — **70/100**

`_run_timestamps` spaces decisions by `rng.randint(1, SECONDS_BETWEEN_DECISIONS_MAX)` — as little as 1 second — while every turn's message row is stamped `decision.timestamp - MESSAGE_LEAD_SECONDS` (2s) in `_transcript_rows`.

`anchorTranscript` (`platform/dashboard/src/lib/llm_messages.ts`) keeps the **last** turn with `occurred_at <= decidedAt`. So whenever two consecutive decisions are 2s apart or less, turn *i+1* lands at or before decision *i* and steals its anchor: the drawer labels the **next** completion "This call" and pushes the real one into `earlier`.

Measured on `build_rows(target, 3, 1)`: **40 of 754 consecutive decision pairs (5.31%)** fall within the lead, against the 4.44% predicted by `2/45`.

**Suggested fix** — make the spacing floor exceed the lead:

```python
SECONDS_BETWEEN_DECISIONS_MIN = MESSAGE_LEAD_SECONDS + 1
# ...
timestamp += timedelta(
    seconds=rng.randint(SECONDS_BETWEEN_DECISIONS_MIN, SECONDS_BETWEEN_DECISIONS_MAX)
)
```

`_MAX_RUN_DURATION` derives from the max alone, so the detector-separation invariant is unaffected. A sturdier variant: pass the previous decision's timestamp into `_transcript_rows` and clamp the lead to half the real gap, so no future caller can reintroduce this.

---

### 2. Anomaly runs pack 20–50 decisions into 6 distinct timestamps — **75/100**

`generate_anomalies` draws `anomaly_base - timedelta(minutes=rng.randint(0, 5))`, which has minute granularity — 6 possible values for `REQUESTS_PER_ANOMALY_MIN..MAX` (20–50) requests. By pigeonhole most decisions in a spike share a timestamp, each still becomes its own turn at `t - 2s`, and `anchorTranscript`'s `<=` then resolves **every** decision in a same-minute group to that group's last turn.

Measured: **30 of 36 anomaly decisions** are not last in their timestamp group, so all 30 anchor on a later turn than their own; 12 land on a completion whose `tool_call` parts don't name the decision's `tool_name`.

This ranks above finding 1 despite the shared cause: anomaly decisions are precisely the rows an auditor opens the drawer on.

**Suggested fix** — give the spike seconds-resolution spread with the same min-spacing guarantee, reusing the existing generator:

```python
ANOMALY_WINDOW_SECONDS = 300
start = anomaly_base - timedelta(seconds=ANOMALY_WINDOW_SECONDS)
timestamps = list(_run_timestamps(rng, start, requests))
```

At 50 requests x >=3s that spans <=150s, comfortably inside the detector window and well above `_ANOMALY_MIN_REQUESTS`.

---

### 3. `--clear` fails half-applied on any dev database predating migration 0003 — **55/100**

`clear()` unconditionally loops `("policy_decision", LLM_MESSAGE_TABLE)` running `ALTER TABLE ... DELETE`, and the `--clear` CLI branch never calls `_validate_columns`. `llm_message` only exists from `init/schema.sql` on a fresh volume or a hand-applied `0003_add_llm_message.sql`, so on an existing local volume the first mutation runs to completion (`mutations_sync=2`) and the second raises a raw `UNKNOWN_TABLE`.

The data outcome is actually fine — with no `llm_message` table there are no transcripts to delete — but `make seed-audit-clear` now exits non-zero with a driver traceback and never prints `Seed rows cleared.`, where previously it worked on any DB with `policy_decision`.

Related: `_validate_columns`'s docstring says it guards "a local ClickHouse that predates it [and] has the decision table and not this one", but `client.query("DESCRIBE TABLE ...")` **raises** `DatabaseError` for a missing table rather than returning an empty description, so the friendly `run \`make clickhouse-migrate\`` error only ever fires for a table that exists but is short a column.

**Suggested fix** — one helper covers both:

```python
def _table_exists(client: Client, table: str) -> bool:
    return bool(client.query(f"EXISTS TABLE {table}").result_rows[0][0])
```

Skip missing tables in `clear()`, and gate `_validate_columns` on it so the absent-table case raises the intended message instead of a traceback.

guillaume-hexamind (commented):
good, will approve once claude issues addressed

PR #233, opened 2026-09-17
feat(sdk): child.as_tool() for native agent-as-tool sub-agent reach

quanghexa94:
Native sub-agents were invisible tool closures: the parent held no handle, so a delegation was gated as a *plain* tool, not a reach edge, and couldn't be enumerated or registered. This adds a first-class **agent-as-tool** construct.

## What it does
`HexgateAgent.as_tool(*, name=, description=)` returns a policy-gated delegation tool you mount in a parent's `tools=`. On call it decides `agent.tool:<child>` under the **parent's** policy, then runs the child's own enforced `ainvoke` under the inherited role — control returns with the child's answer.

- This is the **only native reach mode**. `create_agent` has no handoff / `subagents=` parameter: handoff (control *transfers*) needs a seam LangChain doesn't expose, so it exists only on OpenAI/Google agents (authored with their SDKs, read + gated by hexgate — not authored through `create_agent`).
- `HexgateAgent.subagents` is a derived property — enumeration/registration read the edges straight off `tools=`, nothing extra stored.
- Guards thread through the delegation; a reach denial short-circuits ahead of the pipeline (never logged as a successful call); a child governance refusal renders as `{"ok": False}` output and counts as an error, not a success.
- Fail-loud collisions: a generated delegation name clashing with another tool, or two edges onto the same target agent, raise `ValueError`.

## Try it
```python
from hexgate import create_agent

billing_bot = create_agent("gpt-4o", name="billing_bot", tools=[refund_order])
support_bot = create_agent(
    "gpt-4o",
    name="support_bot",
    tools=[view_orders, billing_bot.as_tool()],   # agent-as-tool (via="tool")
    # billing_bot.as_tool(name="delegate_refund")  # override the LLM-facing name
)
# support_bot's policy gates `agent.tool:billing_bot`; billing_bot enforces its own policy.
```

## Important files
- `hexgate/agents/factory.py` — `as_tool()`, the `subagents` property, coverage-aware reach warn
- `hexgate/adapters/langchain/tools.py` — `SubagentTool` (reach gate → child `ainvoke`, result coercion)
- `hexgate/agents/subagents.py` — `SubagentEdge`

## Notes
- Handoff on native is out of scope (needs graph composition / turn-ownership).
- A reach *denial* is audited but does not yet increment the `run.denials` counter — a cross-cutting reach-gate follow-up, not native-specific.


hexgate/agents/factory.py
```diff
@@ -604,7 +678,22 @@ def enforce_policy(
         # a guard Halt(NEEDS_APPROVAL).
         wrapped: list[ToolSpec] = []
         for tool_spec in self.tools:
-            if isinstance(tool_spec, BaseTool):
+            if isinstance(tool_spec, SubagentTool):
```
victorludvig, line 681: You lock down a support agent: deny everything, allow view_orders only. You mount billing_bot.as_tool() so it can get refunds issued. You haven't written an agents: block yet.

A customer says "refund my order." The model calls delegate_to_billing_bot. It goes through, and billing_bot issues the refund. You thought support was read-only. And the audit dashboard shows nothing, because nothing called decide().

Every other tool in that same tools= list is denied. The delegation isn't, because skipping GuardedTool here means the only check left is ReachGate — and that one switches itself off when the policy has no agents: block (tools.py:319).

The comment's reason doesn't hold: GuardedTool would decide delegate_to_billing_bot, a plain name, which falls to default_policy — policy.py:65 closed-world-denies agent keys only. That's the fallback the docs already promise (agent-level-enforcement.mdx:110).

Fix: decide agent.tool:<child> when declares_tool_reach(), else the plain tool name — same as openai/tools.py:135-154. run_guarded_async already takes policy_key/policy_args for this, and it brings back the audit row.

hexgate/agents/factory.py
```diff
@@ -604,7 +678,22 @@ def enforce_policy(
...
+                # A SubagentTool gates itself through ReachGate (agent.tool:<child>,
+                # honoring declares_reach). Routing it through GuardedTool would
+                # decide the wrong key, always-on, and closed-world-deny delegation
+                # whenever the policy has no agents block. Inject the SAME enforcer
+                # the binding holds so a refresh reaches it.
+                wrapped.append(
+                    tool_spec.model_copy(
```
victorludvig, line 690: Take the global guard from the guards doc (docs/concepts/guards.mdx:58):

@before_tool                    # every tool
def block_secrets(call):
    if looks_like_a_secret(call.args):
        return Halt(reason="Refused: remove the credential and resend.")

After enforce_policy, "every tool" means every tool except the sub-agent hop.

A customer pastes an API key into the chat. The model relays the request and puts it in the task argument of delegate_to_billing_bot. block_secrets would catch that on any ordinary tool. It doesn't run here, so the credential goes into the child agent and on to its model and tools. Same for an @after_tool(observe=True) logger: the delegation never shows up in the guard log, so there's no trace afterwards.

What makes it easy to miss is the ordering. create_agent(guards=[...]) on its owhread_into_subagent_tool covers exactly that. It's the second step that wipes it— so it works when you try it, then stops once you add the policy. Revert the production change and that test still goes green.

Cause: model_copy(update={...}) overwrites unconditionally, so pipeline and approval_handler are replaced with whatever this enforce_policy call was given — None when the caller didn't
re-supply them. Four lines down, GuardedTool.wrap does the opposite: "Idempotentugh unless explicitly overridden" (tools.py:154, resolved_pipeline = pipeline ifpipeline is not None else tool.pipeline).

create_agent(tools=[view_orders, child.as_tool()], guards=[...]).enforce_policy(policy)

view_orders.pipeline -> populated, guard fires
delegation.pipeline  -> None,      guard never fires

Fix: fall through instead of overwriting, same as GuardedTool.wrap. (create_agen— _bind_policy forwards guards=, factory.py:1064.)

victorludvig:
Hi @quanghexa94, 2 edge cases to check, plus this out-of-diff potential issue:


Out-of-diff, follow-up: approval_required on a reach target never prompts via the CLI

Not a change for this PR — the file isn't touched here — but this PR introduces the tool class that falls through it, so flagging it while it's fresh.

_apply_approval_handler (hexgate/agents/loader.py:377) re-stamps the approval handler onto GuardedTools only. A SubagentTool hits the bare append and keeps approval_handler=None, so _cleared_async refuses without asking. In practice: you write the doc's agents: { billing-bot: { mode: approval_required } }, run hexgate chat, and get a flat refusal instead of a prompt — while ordinary tools prompt fine in the same session, so it reads as a bad policy. The if not touched warning that would have caught it stays quiet, because any ordinary tool on the agent sets touched.

Fails closed, so it's a dead feature rather than a hole. One isinstance arm fixes it; _apply_decision_observer just above has the same gap.

quanghexa94:
Thanks @victorludvig — all three fixed, SDK suite green (2138 pass). Rebased on latest main.

**1. Fail-open (ungated delegation under a deny-default with no `agents:` block).** You're right, the comment's reasoning was wrong. `SubagentTool` now decides through `run_guarded_async` like the OpenAI adapter (`adapters/langchain/tools.py`): `agent.tool:<child>` when `enforcer.policy.declares_tool_reach()`, otherwise the **plain tool name** — so a deny-default agent denies the delegation and it produces a `decide()` audit row. New tests: deny-default + no `agents:` block denies and records the denial; a handoff-only policy name-gates (consistent with the OpenAI adapter, not a silent closed-world deny of every as-tool).

**2. Guard bypass after `enforce_policy`.** The `model_copy(update=...)` overwrote `pipeline`/`approval_handler` unconditionally; it now falls through like `GuardedTool.wrap` (`factory.py`), keeping the tool's existing pipeline unless the call supplies a new one. So `create_agent(guards=[...]).enforce_policy(policy)` keeps the guard on the delegation hop. New regression test.

**3. Out-of-diff (`approval_required` on a reach target never prompts).** Added `SubagentTool` arms to `_apply_approval_handler` and `_apply_decision_observer` (`agents/loader.py`), so the CLI's approval handler and decision observer reach the delegation tool — the reach approval now prompts and decision events surface.

hexgate/agents/factory.py
winlp4ever: Fixed — and you're right, the comment's reasoning was wrong. `SubagentTool._arun` now decides through `run_guarded_async` like the OpenAI adapter: `agent.tool:<child>` when `enforcer.policy.declares_tool_reach()`, otherwise the **plain tool name** — so a deny-default agent with no `agents:` block denies the delegation and it produces a `decide()` audit row instead of slipping through. New tests: deny-default + no `agents:` block denies and records the denial (`facts.denials == 1`, `tool_calls == 0`); a handoff-only policy name-gates rather than silently closed-world-denying every as-tool.

hexgate/agents/factory.py
winlp4ever: Fixed. The `model_copy(update={...})` was overwriting `pipeline`/`approval_handler` unconditionally; it now falls through like `GuardedTool.wrap` — keeps the tool's existing pipeline/approval_handler unless this `enforce_policy` call supplies new ones. So `create_agent(tools=[view_orders, child.as_tool()], guards=[...]).enforce_policy(policy)` keeps the guard on the delegation hop. Added a regression test that halts the delegation after a guard-less re-enforce.

victorludvig (approved):
Approved!

PR #235, opened 2026-09-17
feat(platform-api): record the operator's AI Act classification per agent

victorludvig:
Risk Tier: low (no thorough review required)

Design: [AI Act — proposal](https://app.notion.com/p/3d4fb45dbae2815885eaffb9698f0793) · [evidence report mockup](https://claude.ai/code/artifact/72a0d0f0-2e00-4416-a389-da470f75cc2e) — PR 1 of 5.

## What is changing
A new `agent_classification` table and `GET|PUT /v1/projects/{project_id}/agents/{name}/classification` on the agents router, holding what the operator asserts about an agent under the AI Act: intended purpose, their role, risk tier, Annex III point, the named oversight owner, and the compliance-checker revision they relied on — stamped with who recorded it and when. Completeness is computed (purpose, role, tier, oversight owner, plus the Annex III point when the tier is `high_risk`); a partial entry is accepted and reported as incomplete rather than rejected. GET prefills the purpose from the registered manifest description when no entry exists yet.

It is section 1 of the AI Act evidence report — the only new stored data the report needs. The entry lives in its own table rather than on `Agent` because it is a different person's statement on a different cadence and needs provenance that `Agent.updated_at` cannot carry. The service records the assertion and judges nothing about its substance: no check that a tier fits the agent's tool set, no warnings. The closed sets on tier and role are a wire-shape guard only — completeness compares the tier by exact string, so a typo must be a 422 rather than a silently not-high-risk entry.

One thing to know about the recorder's name:

- Every entry records who saved it. Deleting that person's account sets the name to null and keeps the entry — the same as every other actor column.
- We do it this way so a GDPR erasure request is never blocked by a database error. The alternative, refusing the delete, would make a legal obligation fail.
- The cost: the entry then asserts a risk tier with nobody named against it. The report should show that as needing re-confirmation, not as a valid assessment.
- #246 covers the better fix — anonymise the user row instead of deleting it, so the name goes but the trail stays intact.

## Tests
`platform/api/tests/features/agents/test_classification.py`: the round trip with server-side provenance, the completeness rule in both directions on `high_risk`, replace-not-merge semantics, blanks and unknown keys rejected, an assertion that looks inconsistent still recorded, recovery from a lost first-save race, an erased recorder leaving the entry unattributed, and tenant isolation. `make check-all` green.


guillaume-hexamind (approved):
No issues found running code review high

Logic looks good to me

PR #237, opened 2026-09-17
feat(platform-api): render the AI Act evidence annex to PDF

victorludvig:
## What is changing
The AI Act evidence report renders through an HTML template and a print stylesheet via WeasyPrint, replacing the LaTeX/tectonic renderer: `@page` margin boxes for the running footer, `thead` repeated where a table breaks, `break-after: avoid` on section headings, fixed column widths so a long value wraps inside its column. Output is PDF/A-2b and is byte-identical across renders of one annex. Also drops the worktree's `.vscode/` and `Specs/` from the repo — the spec now lives in [Notion](https://app.notion.com/p/3e2fb45dbae281f794fefa40ff4a1a8d) as the canonical copy.

The document is a design, not a dump, and CSS paged media is where that design is expressible — the mockup's hierarchy, restrained rules and small-caps article references carry over directly. It also retires the hand-written `escape_tex()` and its fuzz suite: Jinja autoescaping applies at interpolation, so there is no call site left to forget.

## Tests
`make check-all` green. HTML assertions for what the document contains; PDF assertions, read back out of the pages, for how it is set — the running footer on every page, a repeated header where the decision sample breaks, a 400-character token wrapping inside its column, a hostile string in every annex field round-tripping, two renders agreeing byte for byte, and a walk over the annex asserting every value reaches the document (with the deliberate omissions listed and reasoned). Both the golden annex and a real stored one were rendered and read page by page.

**CI has not run on this branch.** `tests.yml` triggers on `pull_request: branches: [main]`, so a PR targeting a feature branch never runs the gate — see #250. That is awkward here, because this is the PR that adds the WeasyPrint apt dependencies to that workflow. Verified locally instead: `make check-all` green with Pango installed, and the workflow's package list compared literally against `platform/api/Dockerfile` — identical, seven packages.


guillaume-hexamind:
## Code review — 3 findings

Reviewed the render layer and verified each finding by rendering `golden_annex.json` through `build_html` (importing `html.py` directly, since WeasyPrint isn't installed locally). Scores are /100.

The two headline fixes in 309f45a3 both hold up: `require_project_admin` matches the three sibling routes so `.pdf` no longer routes around the ban-enforcement boundary, and `_mark_unrepresentable` emitting `&lt;U+XXXX&gt;` survives HTML parsing as visible text.

---

### 1. `_matrix_roles` silently drops authorisation columns — **78/100**

`render/html.py:174` takes the column order from `rows[0]["cells"]` and only falls back to `matrix["roles"]` when there are no rows, so `roles` is ignored whenever a row exists.

Deleting `support` from row 0's cells in the golden annex removes `>support</th>` from the document entirely — the header and every tool's Allow/Deny verdict — while the signed annex still declares the role. A whole authorisation column disappears from an evidence document with nothing to indicate it.

The docstring justifies reading from cells on the grounds that a drifted `roles` list would otherwise be "a failed render". I tested that: substituting `lambda mx: list(mx["roles"])` and adding a role present in `roles` but in no cell renders fine — `report.html.j2:195` already does `row.cells.get(r)` with an `{% else %}—{% endif %}` fallback, which produces exactly the em-dash the docstring is asking for. So the current implementation trades a risk that doesn't exist for silent data loss that does; the rejected alternative is strictly safer.

Worth noting `any_other_tool.cells` is built from `matrix.defaults`, a separate dict, so that row can lose a column independently.

```python
def _matrix_roles(matrix: dict) -> list[str]:
    """Every role the annex mentions, declared order first."""
    ordered = list(matrix.get("roles") or [])
    seen = set(ordered)
    for row in matrix.get("rows") or []:
        for role in row.get("cells") or {}:
            if role not in seen:
                seen.add(role)
                ordered.append(role)
    return ordered
```

---

### 2. The "nothing is silently dropped" guard doesn't cover numbers — **62/100**

`COUNTER_LABELS` (`render/html.py:43`) is a hardcoded list of 8 keys, and `test_every_annex_string_reaches_the_document` walks only `isinstance(value, str)` — int and float leaves are never asserted.

Adding `secret_guard_bypasses: 999` to `activity.counters` leaves section 3's stat row at 8 tiles and the string `999` appears nowhere in the document, with the suite green. That's the regression class the test's own docstring says it exists to prevent.

The asymmetry is one-directional: *removing* a counter does fail loudly (`UndefinedError: 'dict object' has no attribute 'denials'`) via `StrictUndefined`. Only addition is silent. The one other `COUNTER_LABELS` reference (`test_render.py:214`) checks label wording, not key coverage.

```python
def test_every_counter_has_a_tile() -> None:
    assert set(sample_annex()["activity"]["counters"]) == {k for k, _ in COUNTER_LABELS}
```

---

### 3. Module-level WeasyPrint import makes one route a whole-API boot dependency — **50/100**

`main.py:17` imports `render.router` → `render/__init__.py` → `render/pdf.py:34` `from weasyprint import CSS, HTML`, which dlopens Pango at import time. An environment without the native stack fails to start the entire control plane rather than 500-ing one endpoint.

Deployment is genuinely covered — `Dockerfile:45` installs `libpango-1.0-0 libpangoft2-1.0-0 libharfbuzz0b` in the final runtime stage (not the builder), and `.github/workflows/tests.yml:133` matches — which is why this is scored as hardening rather than a live defect. The exposure is every other path: host runs, a future image, a rebased base image. A report route shouldn't be able to take down the API.

Moving the import into `render_pdf` confines the blast radius to the route:

```python
def render_pdf(html: str) -> bytes:
    from weasyprint import CSS, HTML
    ...
```

---

<details>
<summary>Checked and cleared</summary>

- **Timestamps.** `_dt` dropping the offset looked like a bug under the "Occurred (UTC)" headers, but no non-UTC value can reach it: operator-supplied period bounds go through `query_scope._ensure_utc` (`astimezone(utc)`), `generated_at` is `utcnow()`, `recorded_at` is `DateTime(timezone=True)` read via asyncpg, and the ClickHouse rows go through `_as_utc` over `DateTime64(3,'UTC')`. All five `| dt` call sites are UTC by construction.
- Route shadowing between `/reports/{rpt_id}.pdf` and the sibling `/reports` / `/reports/{rpt_id}/annex`.
- `Content-Disposition` — filename comes from `report.id` after a successful lookup, so a hostile path param 404s first.
- Tenant isolation — `get_report` filters on `project_id`, so a cross-project id reads as absent.
- `pill-neutral` has no rule of its own, but `.pill`'s base already uses `--flat-bg`/`--flat-fg`, so the fallback renders as intended.
- Jinja precedence in `78.0 / roles | length` — filters bind tighter than `/`, so this parses as intended.
- Generated HTML is well-formed (nesting check over the golden annex).

</details>

guillaume-hexamind (commented):
Big PR, but no major macro issues found.

A couple of comments to investigate above

victorludvig:
All three fixed in this branch.

1. **`_matrix_roles`** now returns the union — `matrix["roles"]` in declared order, then any role only a row's `cells` carries. A declared role with no cell renders as the template's existing em-dash instead of the column vanishing. Covered by `test_when_a_role_has_no_cell_then_its_column_still_renders`.
2. **Counter coverage**: added `test_every_counter_has_a_tile`, asserting `set(annex["activity"]["counters"]) == {k for k, _ in COUNTER_LABELS}`. The added-counter case now fails loudly, matching the removal case `StrictUndefined` already caught.
3. **WeasyPrint import** moved into `render_pdf`, so a missing native stack costs this route a 502 instead of the API's boot. The two tests that monkeypatched `render.pdf.HTML` now patch `weasyprint.HTML`.

Tests: `make check-all` green; platform-api integration 11 passed; SDK integration against the local OTLP stack 9 passed, 1 skipped (needs dashboard smoke credentials).

guillaume-hexamind (approved):

PR #238, opened 2026-09-17
feat(sdk): tabulate a resolved policy set as an authorisation matrix

victorludvig:
**Risk Tier**: Low (no thorough review required)

Design: [AI Act — proposal](https://app.notion.com/p/3d4fb45dbae2815885eaffb9698f0793) · [evidence report mockup](https://claude.ai/code/artifact/72a0d0f0-2e00-4416-a389-da470f75cc2e). Plan and section numbering: `Specs/ai_act_evidence_report.md` (untracked), PR 2 of 5.

## What is changing

New `hexgate/security/matrix.py`: `authorisation_matrix(PolicySet) -> Matrix`, a pure function rendering roles x tools as `allow`/`approval`/`deny` plus constraint text, exported from `hexgate.security`. No existing behaviour changes.

Its actual output, for a policy with a mixin, a role-wide fence, a permissive-default role, an empty role and one reachable agent:

| Tool | `default` | `admin` | `billing` | `observer` | `support` |
|---|---|---|---|---|---|
| `agent.handoff:ledger_bot` | Deny | Deny | Allow<br>`run.tool_calls < 50` | Deny | Deny |
| `delete_database` | Deny | Approval | Deny | Deny | Deny |
| `lookup_order` | Allow | Allow | Allow<br>`run.tool_calls < 50` | Deny | Allow |
| `refund_order` | Deny | Allow | Allow<br>`(run.tool_calls < 50) and (args.amount <= 500)` | Deny | Approval<br>`args.amount <= 50` |
| *any other tool* | Deny | Allow | Deny | Deny | Deny |
| *any unnamed agent* | Deny | Deny | Deny | Deny | Deny |

- `default` leads because every undefined role falls back to it; it is a column, not the table.
- The same tool answers differently per role (`refund_order`). At call time `combine_role_verdicts` takes the permissive union across a caller's roles; the matrix deliberately does not, because each column is that role's own standing grant.
- The mixin `read_only` has no column — it is folded into `support` and `billing` but cannot be assigned to anyone.
- `admin` sets `default_policy: allow`, which is why *any other tool* exists: without that row the grid would imply Deny and assert a control nothing enforces.
- *any unnamed agent* stays Deny even for `admin` — agent keys are closed-world, so a permissive default cannot grant a delegation path the policy never named.
- `observer` is declared but empty, so it gets a column of denies rather than vanishing; a missing row reads as "not covered" instead of "grants nothing".

No PDF and no endpoint here — those are PRs 4 and 3.

Section 2 of the evidence report needs a standing grant table per agent; `analyzer.py` lints and the engines answer one proposed call, so nothing tabulated. Cells resolve through `get_tool_policy` and carry every condition the engine checks — the role-wide fence then the tool's own, in `evaluate_tool_call`'s order — so a cell can understate a grant but never overstate one.

## Tests

`tests/security/test_matrix.py`, 40 cases: the five shapes the spec names, the reach keys, constraint and file-scope rendering, parity with the dashboard's TypeScript resolver (with the deliberate divergences documented), and a differential check that no cell claims more than the engine grants across a corpus of argument, `run` and `ctx` shapes. `make check-all` green.


guillaume-hexamind (approved):
All good to me. Better to open an issue if you can so that we can keep the duplication of logic in mind for later.

victorludvig:
> All good to me. Better to open an issue if you can so that we can keep the duplication of logic in mind for later.

I opened the issue: https://github.com/HexamindOrganisation/hexgate/issues/247

PR #239, opened 2026-09-17
feat(platform-api): assemble and sign the AI Act evidence report

victorludvig:
**Risk surface.** Secondary additive feature except for four shared audit read functions (`summarize`, `summarize_llm_invocations`, `list_decisions`, `list_ban_enforcements`), which gain keyword arguments defaulting to today's behaviour — dashboard call sites are unchanged. The new endpoint is the only caller of the new paths. Heaviest operation is report generation; it is bounded by `REPORT_QUERY_SETTINGS` (1 GiB / 30 s per query), measured at 847 MB peak on a 3.2M-row slice. Revert is the single commit; the one new table is additive.

## What is changing

`features/ai_act/` assembles one signed evidence document per project — cover plus the spec's five sections — from the resolved policy bundle and the three audit tables, and serves it over `POST /v1/projects/{id}/ai-act/report`, `GET …/reports` and `GET …/reports/{rpt_id}/annex`. The stored JSON's exact bytes are the signed artifact, kept verbatim with their digest. `audit.summarize` and `summarize_llm_invocations` gain an additive `distinct_events` flag, and their list reads a `with_total` flag; both default to today's behaviour, so the dashboard is untouched.

What this PR produces is the report's **JSON scaffolding**: the cover and five sections as structured data, stored verbatim on an `AiActReport` row with its SHA-256 digest and Ed25519 signature. It is the canonical artifact — the only thing the signature covers. (The spec calls this document the *annex*, and that name survives in the endpoint path and the column.)

Signing the JSON rather than a rendered file is deliberate: a signature is over bytes, and a PDF is not byte-reproducible — engine and font versions and embedded timestamps all move it — so two renders of one report could disagree and look tampered with. The JSON's bytes are fixed (`canonical_bytes`: set key order, no whitespace).

Turning it into something a human hands an auditor — an A4 PDF — is #237, stacked on this branch. That renderer reads the stored JSON and nothing else, so re-rendering an old report reproduces it, and the engine stays swappable without touching a signature. The dashboard tab that drives all of this is #236.

Roadmap item: give an operator one button that produces evidence of the controls in place and the events recorded, without asserting anyone is compliant. Counts are over distinct `event_id`s because `ReplacingMergeTree` dedup is eventual and a double-counted retry is the misleading figure Art. 99(5) penalises.

Both upstreams are merged, so nothing is stubbed: section 2 tabulates the compiled bundle through `hexgate.security.matrix` (#238), and section 1 reads `AgentClassification` (#235) and reuses that slice's `missing_fields`, so the report and the classification endpoint cannot disagree about whether an agent is complete.

## Tests

- Golden annex over a seeded slice of the three tables — section 2 built from a real `PolicySet` through the real tabulator, section 1 from a real `AgentClassification`.
- The condition-generated gap list, matrix sourcing, period clamping, signature verification through the section 5 recipe, and tenant isolation.
- `make check-all` green.
- Exercised against a live ClickHouse at 1.6M and 3.2M decision rows: peak query 847 MB under the 1 GiB ceiling, whole report 1.4 s — against 2.58 GiB and an OOM-killed container before the `with_total` and ceiling changes.


guillaume-hexamind:
## Code review — 2 findings

Two issues worth addressing before merge. Scores are severity out of 100; each was verified against the code on this branch.

---

### 1. Ban-enforcement records escape their admin-only gate — 82/100

`platform/api/hexgate_api/features/ai_act/router.py:68`, `:110`, `:129`

`GET /v1/projects/{id}/audit/ban-enforcements` is the only admin-gated read in the audit router (`features/audit/router.py:296`), and all three `/bans` CRUD routes are `require_project_admin` too. The comment above that endpoint is explicit that this is deliberate: *"Ban enforcements read on their own admin-gated endpoint (like ban CRUD)."*

All three AI Act endpoints gate on `require_org_member`, which is `require_project_admin` minus the role check. The annex embeds up to `BAN_ENFORCEMENT_LIMIT = 500` full `BanEnforcementRow`s under `activity.ban_enforcements.rows` — `user_id`, `session_id`, `ban_type`, `ban_id`, `reason` (confirmed present in `tests/features/ai_act/golden_annex.json`: `{"user_id": "u_mallory", "ban_id": "ban_abc123", "reason": "offboarded", …}`).

So a plain org member POSTs `/ai-act/report`, gets the rows inline in the 201 body, and `GET …/reports/{rpt}/annex` re-serves them indefinitely on an append-only row. They now read who was banned and the free-text `reason` — the data the bans and audit APIs deliberately withhold from non-admins.

Scoped at 82 rather than higher because it is intra-tenant, and `/audit/decisions` already exposes `user_id` / `session_id` / `arguments` to org members. The marginal disclosure is the ban itself and its reason.

**Suggested fix** — simplest is to gate the feature at the level of its most sensitive contents:

```python
# features/ai_act/router.py
from hexgate_api.deps.project import require_project_admin

@router.post(".../ai-act/report", ...)
async def api_generate_ai_act_report(
    ...,
    membership: tuple[User, OrganizationMember] = Depends(require_project_admin),
)
```

…and the same on the two GETs. If org members should keep access to the report, the alternative is to keep `activity.ban_enforcements.total` and the counters but drop `rows` for non-admins — though note that changes the annex bytes and therefore the digest, so the caller's role would become part of what is signed. Gating the endpoints keeps one document per period.

---

### 2. `list_reports` loads every stored annex blob to compute a byte length — 50/100

`platform/api/hexgate_api/features/ai_act/service.py:734` and `features/ai_act/router.py:52`

```python
select(AiActReport).where(...).order_by(AiActReport.generated_at.desc())
```

No `limit`/`offset`, and `annex_json` is a plain `str` column with no deferred loading (`models.py:618`). `_summary` uses the loaded document for exactly one thing:

```python
annex_bytes=len(report.annex_json.encode("utf-8")),
```

So `GET …/ai-act/reports` pulls every annex a project has ever generated into the API process and re-encodes each one, to produce an integer. An annex carries one matrix per agent, 40 sampled decisions with their `arguments` / `attributes` blobs, and up to 500 ban rows — tens to low hundreds of KB each. Nothing dedups on `(project, period)` or rate-limits generation, so the row count only grows.

**Suggested fix** — push the length into SQL and paginate:

```python
async def list_reports(
    session: AsyncSession, project_id: str, *, limit: int = 25, offset: int = 0
) -> list[tuple[AiActReport, int]]:
    cols = (
        AiActReport.id, AiActReport.project_id, AiActReport.period_start, ...,
        sa.func.octet_length(AiActReport.annex_json).label("annex_bytes"),
    )
```

…selecting the summary columns explicitly rather than the whole entity, so `annex_json` never leaves Postgres on this path. A stored `annex_bytes` column written at generation time works equally well and keeps the read trivial. Either way the endpoint wants `limit`/`offset` like the audit reads have.

---

Not raised here: the per-query `REPORT_QUERY_SETTINGS` ceiling checks out — I verified every read on this path goes through `client.query` (9 sites in `audit/service.py`, 1 in `llm_invocations/service.py`, no `.insert`), so nothing escapes the wrapper via `__getattr__`, and `_read_events` runs its queries serially in one `to_thread`, so a generation peaks at 1 GiB rather than N × 1 GiB. A per-project cooldown or in-flight guard would still be worth a follow-up, but the measured OOM is closed.

guillaume-hexamind (commented):
2 issues to investigate; will approve once done

victorludvig:
Both fixed in `ce8782d4`. Thanks — the first one was a real hole and my tests couldn't have caught it.

**1. Ban rows escaping the admin gate**

- All three endpoints now `require_project_admin`. I took the gating option, not the drop-`rows`-for-non-admins one, for the reason you gave: per-caller filtering puts the caller's role inside what the signature covers, so one period would have two digests.
- Added `test_plain_org_member_cannot_read_the_report` — a real `member` of the project's org, 403 on all three routes. Verified it fails against the old gate (a member got **200** on the history read), so it's a regression test rather than decoration.
- Added `test_project_admin_can_read_the_report` so `admin` is enough and owner isn't required.
- Why it survived four review rounds: every auth test I had used the org owner, who passes an admin check, so the dependency was untested in the only direction that mattered. I'd also compared this slice against the four audit reads it reuses and never checked the fifth — the one whose data it embeds.
- Note for #236: the AI Act tab is now admin-only. Nothing else calls these endpoints today (`grep ai-act platform/dashboard/src` is empty).

**2. `list_reports` loading every annex**

- `annex_bytes` is now a stored `int` on `AiActReport`, set at signing time from `len(canonical_bytes)`. Took the stored-column option over `octet_length` — `octet_length` isn't available on the SQLite test path, and the count is a property of the signed artifact anyway, fixed when the signature was taken over exactly those bytes.
- `list_reports` selects explicit columns into a `ReportSummary` NamedTuple, so `annex_json` never leaves Postgres. Compiled against the pg dialect to confirm the projection.
- `limit`/`offset` added, clamped `max(1, min(limit, 200))` / `max(0, offset)` to match `features/audit/router.py`. Worth flagging: unclamped, `LIMIT -1` is an executor error on Postgres but silently returns everything on SQLite, so it would have 500'd only in production while the dev suite stayed green.
- New tests: history rows carry no annex and `annex_bytes` equals the served byte length; pagination; the clamps; and a guard that `_SUMMARY_COLUMNS` order matches `ReportSummary._fields`, since that mapping is positional.
- Why I missed it: I measured the generation path against a live ClickHouse and never looked at the Postgres read. The cost wasn't a query — it was a `len()` on a column I'd assumed was lazily loaded.

Agreed on the follow-up: no per-project cooldown or in-flight guard, and nothing dedups on `(project, period)`. Happy to take that as a separate issue.

`make check-all` green. Two dashboard CodeMirror tests flaked on timeout during one run under load; they pass on a clean re-run and this PR touches no dashboard files.

guillaume-hexamind (approved):

PR #234, opened 2026-09-17
feat(sdk): enumerate_subagents — per-framework sub-agent introspection

quanghexa94:
Reads a parent agent's outgoing sub-agent edges — `target` name, `via` (tool or handoff), and an optional live `child` handle — as a list of `SubagentLink`. Pure introspection: no rewrapping, no policy, no behavior change. This is the reader PR 3 (manifest refs) and PR 4 (recursive registration) build on.

**Stacked on #233** (`subagents=` construct) — base is `lhq/feat/subagents-construct`, so the diff here is PR 2 only. Merge #233 first.

## What it does
- `enumerate_subagents(agent) -> list[SubagentLink]`, dispatched by module (mirroring `create_manifest`):
  - **HexgateAgent** → the `subagents=` registry from #233.
  - **OpenAI Agents** → `Agent.handoffs` (handoff) + `Agent.as_tool()` origins (tool).
  - **Google ADK** → `sub_agents` (handoff) + `AgentTool` on `tools` (tool).
  - raw LangGraph graph / Pydantic AI / unknown → `[]` (the child is a tool closure with no handle).
- Targets canonicalize the same way the reach gate does (`canonical_agent_name`, nameless → `default`), so enumeration agrees with what is actually governed.
- Reuses the adapters' existing origin readers (`_agent_tool_target`) for the as-tool/AgentTool cases; a bare OpenAI `Handoff` recovers its live target from `_agent_ref` so PR 4 can recurse.

## Design note
`SubagentLink{target, via, child}` is intentionally distinct from the authoring `SubagentEdge`: an OpenAI as-tool origin exposes only a name, so `target` is always present and `child` is a live handle only when there is one to recurse into. `child` is excluded from eq/hash (`compare=False`) — a framework agent is unhashable, and a link's identity is its `(target, via)`, so `set(...)` dedups by edge.

## Important files
- `hexgate/agents/enumeration.py` — `SubagentLink` + `enumerate_subagents` + per-framework readers (lazy SDK imports).
- `hexgate/agents/__init__.py`, `hexgate/__init__.py` — exports.
- `tests/agents/test_enumerate_subagents.py` — 12 tests.

## Try it
```python
from hexgate import enumerate_subagents, create_agent, as_tool

billing, _ = create_agent(model, tools=[refund_order], name="billing_bot")
support, _ = create_agent(model, tools=[...], name="support_bot", subagents=[as_tool(billing)])
[...]

PR #240, opened 2026-09-17
feat(sdk): record sub-agent reach edges on the manifest + via-specific warn

quanghexa94:
Records each agent's sub-agent reach edges on its manifest, so registration and the dashboard can see the agent graph — and makes the native "reach unenforced" warning name exactly what it can't enforce.

**Stacked on #234** (`enumerate_subagents`) — base is `lhq/feat/enumerate-subagents`, so the diff here is PR 3 only. Merge #233 then #234 first.

## What it does
- Adds `SubagentRef{name, via}` and `subagents: list[SubagentRef] | None` to `AgentManifest`. `create_manifest` populates it from `enumerate_subagents` (PR 2) for every framework.
- The field is **Optional-`None`** (not `[]`), so an agent with no sub-agents is dropped by `compute_manifest_hash`'s `exclude_none=True` and hashes exactly as before — dedup continuity holds. An agent that *has* sub-agents records them and gets one new manifest version (intended).
- **No DB migration:** `AgentVersion.manifest` is a JSON column, so the new field rides in that blob. The platform `AgentManifest` schema mirrors `SubagentRef`/`subagents` so the edges parse, persist, and surface in the read view (`AgentManifestView`) for the dashboard.
- **Via-specific native warn (SDK-only):** the factory computes the *uncovered* reach from the keys it already reads (`_uncovered_reach` → `(handoff_targets, unmounted_tool_targets)`), and `warn_if_reach_unenforced` names the via + targets ("handoff … has no seam"; "agent-as-tool … not mounted via subagents=") instead of a generic message. Fully-covered configs stay quiet; opaque WASM bundles fall back to the generic warning.

## Important files
- `hexgate/manifest/models.py` — `SubagentRef` + `subagents` field.
- `hexgate/manifest/builder.py` — `create_manifest` populates `subagents` from `enumerate_subagents` (via a `_build_base_manifest` split).
- `hexgate/agents/factory.py`, `hexgate/security/agent_gate.py` — via-specific uncovered-reach warning.
- `platform/api/hexgate_api/schemas.py` — mirrored `SubagentRef` + `subagents` (read view inherits it).

## Test plan
- SDK: manifest round-trips refs (native `subagents=` + OpenAI handoffs); stays hash-stable when absent; via-specific warn names the uncovered handoff/tool targets and not the mounted one.
- Platform: `register_manifest` stores `subagents`, dedups a sub-agent-less re-register (created=False, same hash), a sub-agent edge yields a new hash/version, and the manifest read view surfaces the edge.
- Verified: SDK `tests/agents/ tests/register/ tests/adapters/*` and platform `tests/` — SDK 1634 + platform 803 pass; `ruff check`/`format` clean on both.

## Notes
- Enumeration/records only what each adapter can actually execute; a granted `agent.handoff:*` on native surfaces via the via-specific warn (no silent no-op).
- Depends on #234 (`enumerate_subagents`) and #233 (`subagents=` construct).


quanghexa94:
⚠️ This PR was **not really merged** — GitHub auto-marked it merged as an *empty no-op* during a stacked-PR rebase mishap (the head branch briefly equaled its base, both at `3c2ffe87`). None of this PR's actual content merged anywhere; `main` is untouched. The real change lives on the same branch and continues in **#244**.

PR #241, opened 2026-09-17
feat(cli): register the whole sub-agent tree (register_tree) + serve/register flags

quanghexa94:
Makes `hexgate register` / `hexgate serve` cover the **whole sub-agent tree**, not just the root — so each sub-agent lands as its own agent with a starter, dashboard-editable policy. That's the chain that makes editing `billing_bot` in the dashboard actually gate a delegated refund.

**Stacked on #240** (manifest `subagents[]`) — base is `lhq/feat/manifest-subagents`, so the diff here is PR 4 only. Merge order: #233 → #234 → #240 → this.

## What it does
- Adds `register_tree(agent, ...)`: a memoized, **post-order** walk over `enumerate_subagents` (PR 2). Each sub-agent with a live handle is registered *before* its parent, idempotently (`post_manifest` dedups by content_hash) and cycle-safely.
- **Collision handling:** two different agents reached under one name would conflate under one `(project, name)`. The walk memoizes `name → (id(agent), dump)`; a same-*object* revisit (a cycle, or a shared node) is recognized by identity and deduped, while a *different* agent with the same name raises `AgentTreeCollision` (a `HexgateError`). `force=True` keeps the first-seen and skips the colliding one (with a warning) — it does not overwrite, because a cycle-safe memoized walk can't re-post a name.
- `on_register(name, result)` fires per posted node, so serve prints its "Registered / unchanged" line for every node.
- **Wiring, both default off:** `hexgate register --register-subagents [--force]` and `hexgate serve --register-subagents` (`auto_register_subagents=` on `build_runtime_from_local_agent`, reusing the root manifest it already built). A mid-tree POST failure surfaces as a clean `HexgateError` at serve startup, not a traceback.

## Important files
- `hexgate/cli/register/register.py` — `register_tree` + `AgentTreeCollision`.
- `hexgate/cli/register/main.py`, `hexgate/cli/serve.py`, `hexgate/cli/_common.py` — the opt-in flags + serve wiring.

## Test plan
- `tests/register/test_register_tree.py` (10): post-order, deep tree, diamond/shared-node memoized once, collision raises + `force` skips-with-warning + identical-dedup, cycle-to-root-with-override is not a spurious collision, root-manifest reuse, `on_register` per node, CLI flag selects `register_tree` vs `register_agent`.
- Verified: `tests/cli/ tests/register/` (184) and the full SDK suite (2747) pass; `ruff check`/`format` clean.

## Notes
- Deferred (documented): partial registration on a mid-tree failure (inherent to N-post registration; opt-in), no serve `--force` (serve should refuse a genuinely-ambiguous same-name tree), and concurrent/background posting (a future startup-latency optimization).
- Completes the 4-PR core track; PR 5 (dashboard renders the registered tree) is the optional follow-on.


quanghexa94:
⚠️ This PR was **not really merged** — GitHub auto-marked it merged as an *empty no-op* during a stacked-PR rebase mishap (the head branch briefly equaled its base, both at `3c2ffe87`). None of this PR's actual content merged anywhere; `main` is untouched. The real change lives on the same branch and continues in **#245**.

PR #244, opened 2026-09-17
feat(sdk): record sub-agent reach edges on the manifest + via-specific warn

quanghexa94:
Record the enumerated reach edges on the manifest so registration and the dashboard can see the agent graph, and make the "reach unenforced" warning name the exact via + target.

## What it does
- Adds `SubagentRef{name, via}` and `AgentManifest.subagents: list[SubagentRef] | None`, populated in `create_manifest` from `enumerate_subagents` for every framework.
- The field is `Optional`-**None** (not `[]`): an agent with no sub-agents is dropped by the content-hash's `exclude_none`, so it hashes exactly as before — **no DB migration** (the platform stores the manifest as a JSON blob).
- **Via-specific warn**: on native, `warn_if_reach_unenforced` names what the mounted edges don't cover — declared `agent.handoff:*` (no native seam) or an `agent.tool:*` with no mounted `child.as_tool()` edge. `via` defaults to `["tool","handoff"]`, so a target that is both mounted *and* granted as a tool is quiet (its handoff bit is a permissive-default artifact); a handoff-only grant on a tool-mounted child still warns.

## Try it
```python
from hexgate import create_agent, create_manifest

billing = create_agent("gpt-4o", name="billing_bot", tools=[refund_order])
support = create_agent("gpt-4o", name="support_bot", tools=[billing.as_tool()])
create_manifest(support).subagents   # [SubagentRef(name="billing_bot", via="tool")]
```

## Important files
- `hexgate/manifest/models.py` — `SubagentRef`, `AgentManifest.subagents`
- `hexgate/manifest/builder.py` — populate from `enumerate_subagents`
- `hexgate/security/agent_gate.py` — via-specific `warn_if_reach_unenforced`
- `hexgate/agents/factory.py` — `_uncovered_reach` (default-via aware)
- `platform/api/hexgate_api/schemas.py` — platform manifest mirror gains `subagents`

## Notes
- An agent that *does* have sub-agents gets one new manifest version the first time it re-registers after this lands (the manifest genuinely gained edge data); steady-state re-registers dedup as before.


victorludvig:
I approve the logic, please check these 2 findings:

### 1. The new via-specific warning never fires in SaaS — and the generic text it falls back to says reach is unenforced when it *is* enforced

`hexgate/agents/factory.py:814` → `hexgate/agents/factory.py:736` → `hexgate/security/agent_gate.py:97-105`

**Trigger** Run a native agent with a mounted `child.as_tool()` against the platform, where the
policy arrives as a signed bundle (`hexgate/security/binding.py:174-176`).
**Break** `_uncovered_reach` bails to `None` for anything that isn't a `PolicySet`, and
`PolicyBundle` exposes no `effective_tools` — so the `uncovered is None` branch always wins and
logs *"reach is not enforced on the native adapter yet … handoffs will proceed without a reach
check"*. That is false: `declares_tool_reach()` is answered from the signed manifest
(`hexgate/security/bundle.py:327-330`), so `SubagentTool._arun` **does** decide
`agent.tool:<target>` and **does** emit an audit row.
**Notice** Nobody corrects it — the log line *is* the output, and it asserts the opposite of what
the gate does. The PR's entire new behaviour is dead on the one path the deployed product takes.

Not a regression (identical in `c60ac791`), and the fallback is acknowledged in the diff's own
comment — but the comment treats it as an edge case when it is the production case.
`PolicyBundle.declares_tool_reach()` is readable, so at minimum the message can stop claiming the
tool seam is unchecked.

### 2. Both new tests pass with their subject reverted

`tests/agents/test_subagents.py:573-575` and `platform/api/tests/features/agents/test_agents.py:1371-1372`

**Trigger** A later commit touches `warn_if_reach_unenforced`'s call site, or the platform schema
default.
**Break** Swapping `handoff_targets=uncovered[1]` / `tool_targets=uncovered[0]` at
`hexgate/agents/factory.py:743-744` inverts every remediation the warning gives (it tells you to
mount a handoff-only target via `as_tool()`) — **50 tests still pass**, because the three
assertions are independent substring checks over the whole `caplog.text` and never tie a name to
its clause. Separately, `subagents: Optional[...] = Field(default_factory=list)` at
`platform/api/hexgate_api/schemas.py:665` makes every sub-agent-less manifest hash differently —
**all 46 platform tests still pass**, because line 1372 explicitly permits `[]`, the exact value
that breaks continuity, and line 1371 is a tautology (the same object hashed twice).
**Notice** Nobody — green suite. On prod the second one mints a spurious `AgentVersion` for every
already-registered agent on its next register.

quanghexa94:
Thanks @victorludvig — both addressed. SDK 1435 + platform 126 green.

**1. Dead warning on the SaaS path + false generic text.** Right on both counts. Native now **never** emits the generic "reach not enforced … handoffs proceed without a check" message: the call site (`hexgate/agents/factory.py`) only warns when `_uncovered_reach` returns a readable result (a local `PolicySet`). For a signed `PolicyBundle` it returns `None`, and native *does* enforce agent-as-tool reach there (`SubagentTool._arun` reads `declares_tool_reach()` off the bundle), so we stay silent instead of asserting the opposite of what the gate does. The via-specific warn remains a local-policy dev hint — the bundle is opaque to per-target enumeration, and the platform is authoritative for reach config on that path.

**2a. Warn test passed with the via swapped.** Fixed — the assertions now tie each name to its own clause: `agent-as-tool reach to ['refunds'] … not mounted` and `handoff reach to ['escalation'] … no seam`, plus negative checks that neither is misfiled under the other via. Swapping `handoff_targets`/`tool_targets` now fails it.

**2b. Hash-continuity test permitted `[]`.** Fixed — it now asserts the stored value `is None` (not `in (None, [])`), so a schema default of `[]` fails; and I added an explicit check that a `subagents: []` manifest hashes **differently** from the None one, documenting exactly why the field must default to None. FWIW the current head already declares `subagents` as `= None` on both the SDK (`manifest/models.py`) and platform (`schemas.py:665`) — the `default_factory=list` isn't in this revision — but the test now locks that in so a revert can't pass green.

victorludvig (approved):

PR #245, opened 2026-09-17
feat(cli): register the whole sub-agent tree (register_tree) + serve/register flags

quanghexa94:
Register a parent **and its whole sub-agent tree** in one call, so every reachable agent lands as its own dashboard-editable agent — not just the root.

## What it does
`register_tree(agent, *, force=False, ...)` walks `enumerate_subagents` post-order (children before parents, so a parent manifest only references already-registered children):
- Memoized by object identity → cycle- and diamond-safe; idempotent (`post_manifest` dedups by content hash).
- Fail-loud on a **collision** (two different agents reachable under one name) with `AgentTreeCollision`; `force=True` keeps the first-seen one and skips the colliding node — while still registering that node's own distinct sub-agents.
- `on_register(name, result)` fires per posted node.

Wired behind opt-in flags (default off):
- `hexgate register --register-subagents [--force]`
- `hexgate serve --register-subagents [--register-subagents-force]`

## Try it
```bash
hexgate register app:support_bot --register-subagents      # posts billing_bot, then support_bot
hexgate serve app:support_bot --register-subagents          # same, at serve startup
```

## Important files
- `hexgate/cli/register/register.py` — `register_tree`, `AgentTreeCollision`
- `hexgate/cli/_common.py` — serve-path wiring (reuses the root manifest)
- `hexgate/cli/serve.py`, `hexgate/cli/register/main.py` — the flags

## Notes
- Only a link with a live handle is walked; a name-only edge (bare `Handoff` / as_tool origin) is already recorded on the parent manifest.


tests/cli/test_serve.py
```diff
@@ -493,6 +493,39 @@ def test_build_runtime_skips_auto_register_when_disabled(
...
+        description=None,
+        approval_handler=None,
+        auto_register=True,
+        auto_register_subagents=True,
+        auto_register_subagents_force=True,
+        console=Console(),
+    )
+
+    assert captured["register_tree_kwargs"]["force"] is True
     assert captured["get_agent_name"] == "customer_bot"
```
victorludvig: ▎ This line is a context line in the diff, but the new test's body was inserted above it, so it's now the last assertion of test_build_runtime_threads_subagents_force_into_register_tree instead of test_build_runtime_skips_auto_register_when_disabled. Both tests still pass (get_agent_name is set by the policy fetch, which runs either way), so nothing flags it — but the disabled-auto-register test lost the check, and this assertion is unrelated to force-threading. Suggest moving it back.

victorludvig (commented):
LGTM, one potential issue to check before merging

tests/cli/test_serve.py
quanghexa94: Good catch — fixed. `assert captured["get_agent_name"] == "customer_bot"` was the tail of `test_build_runtime_skips_auto_register_when_disabled`; my force-threading test got inserted above it and orphaned it. Moved it back to the disabled-auto-register test (where it verifies the policy fetch still runs), and `test_build_runtime_threads_subagents_force_into_register_tree` now ends on its own `force is True` assertion. 26/26 in `test_serve.py` pass.

quanghexa94: Follow-up: the first push after this fix tripped the `ruff format --check` gate (the moved assertion + its comment ran past the line limit). Fixed by trimming the comment to keep it a clean single line — all checks green now (SDK, Dashboard, Platform API, migrations). Ready for merge.

victorludvig (approved):

PR #252, opened 2026-09-21
fix(sdk): support Google ADK toolsets in manifests and wrappers

### LINKED ISSUE #248: fix(sdk): hexgate raises TypeError on any Google ADK agent using a BaseToolset (e.g. skills)
## Summary

`hexgate` raises `TypeError` on any Google ADK agent whose `tools` list contains a `BaseToolset` — which is every agent using ADK's Agent Skills (`SkillToolset`). Both the registration path and the wrap path fail, so such an agent can neither be registered nor policy-wrapped.

This is not specific to skills: it affects **every** `BaseToolset` (MCP toolsets, `BigQueryToolset`, anything from `from_config`). ADK skills just make it a common path, since `skills=` in an ADK agent config produces a `SkillToolset`.

## Reproduction

Verified against `google-adk 1.32.0` (the version in the repo venv; skills require >= 1.25.0).

```python
from google.adk.agents import Agent
from google.adk.skills import load_skill_from_dir
from google.adk.tools.skill_toolset import SkillToolset

def issue_refund(order_id: str, amount: float) -> str:
    """Issue a refund."""
    return "ok"

skill = load_skill_from_dir("./skills/refunder")   # any dir with a SKILL.md
ts = SkillToolset(skills=[skill], additional_tools=[issue_refund])
agent = Agent(name="support", model="gemini-2.0-flash", instruction="help", tools=[ts])

from hexgate.manifest.google import create_google_manifest
create_google_manifest(agent)          # raises

from hexgate.adapters.google.tools import wrap_tools
wrap_tools(list(agent.tools), enforcer) # raises
```

Actual output:

```
agent.tools types: ['SkillToolset']

=== hexgate create_google_manifest ===
RAISED: TypeError <google.adk.tools.skill_toolset.SkillToolset object ...> is not a callable object

=== hexgate wrap_tools ===
RAISED: TypeError Cannot install policy on tool <...SkillToolset object ...>:
        expected google.adk BaseTool or callable, got SkillToolset.
```

## Root cause

`BaseToolset` is **not** a `BaseTool` and is not callable — verified: `issubclass(BaseToolset, BaseTool)` is `False`, and `BaseToolset` defines no `__call__`. Its members are `get_tools`, `get_tools_with_prefix`, `process_llm_request`, `close`, `get_auth_config`, `from_config`, `_is_tool_selected`.

Two call sites assume otherwise:

1. **`hexgate/manifest/google.py`** — `tool = entry if hasattr(entry, "_get_declaration") else FunctionTool(func=entry)`. A toolset has no `_get_declaration`, so it falls into the `FunctionTool(func=...)` branch and ADK rejects the non-callable.

2. **`hexgate/adapters/google/tools.py`** (`_normalize`) — accepts `BaseTool` or `callable`, raises otherwise. A toolset is neither.

## Expected behaviour

A `BaseToolset` should be expanded into its constituent tools, each of which is an ordinary `BaseTool` that the existing `wrap_tool` already handles.

## Suggested fix

Toolset expansion is `async` (`BaseToolset.get_tools` is a coroutine) while `wrap_tools` and `create_google_manifest` are sync, so eager expansion is awkward and — importantly — **loses tools that appear only later** (see below). The suggested approach is a `GuardedToolset` that composes the inner toolset and gates on every `get_tools()` call:

```python
class GuardedToolset(BaseToolset):
    def __init__(self, inner, enforcer, *, approval_handler=None, pipeline=None):
        super().__init__()
        self._inner = inner
        self._enforcer = enforcer
        self._approval_handler = approval_handler
        self._pipeline = pipeline

    async def get_tools(self, readonly_context=None):
        return [
            wrap_tool(t, self._enforcer,
                      approval_handler=self._approval_handler,
                      pipeline=self._pipeline)
            for t in await self._inner.get_tools(readonly_context)
        ]

    async def process_llm_request(self, *, tool_context, llm_request):
        await self._inner.process_llm_request(
            tool_context=tool_context, llm_request=llm_request
        )

    async def close(self):
        await self._inner.close()
```

`wrap_tools` then routes on type. Note the return type widens from `list[BaseTool]` to `list[BaseTool | BaseToolset]`; `agent.model_copy(update={"tools": ...})` in `wrap_google_agent` accepts both, since ADK's `ToolUnion` is `Callable | BaseTool | BaseToolset`.

### Two things that are easy to get wrong

- **`process_llm_request` must be delegated.** `SkillToolset.process_llm_request` is what injects the skills system instruction and the skills XML. If the wrapper doesn't forward it, skills silently stop being used by the model with no error. Worth an explicit test.

- **Gate on every `get_tools()` call, not once.** `SkillToolset.get_tools()` returns `self._tools + dynamic_tools`, where `_resolve_additional_tools_from_state` reads state key `_adk_activated_skill_{agent_name}` (written by `LoadSkillTool.run_async`) and exposes tools named in the activated skill's `frontmatter.metadata["adk_additional_tools"]`. **Activating a skill can therefore enlarge the tool surface mid-run**, and a wrapper that gates only the wrap-time set lets those through ungated.

## Notes

- ADK skills are `@experimental(FeatureName.SKILL_TOOLSET)` and emit a `UserWarning`; worth pinning a tested ADK range in the version-compat matrix.
- Full analysis, including the manifest/policy design for skills, is in `plans/skills/` (`README.md` §2.1, `google-adk.md`).

mikemikimike:
## Summary

- Support Google ADK `BaseToolset` entries in agent wrapping and manifest generation.
- Re-resolve and policy-gate toolsets on every tool resolution while preserving prefixes and ADK lifecycle hooks.
- Add async manifest resolution for toolsets and regression coverage for dynamic and skills-style toolsets.

Fixes #248

Issue: https://github.com/HexamindOrganisation/hexgate/issues/248

## Tests

- `uv run --frozen pytest tests/adapters/google/test_tools.py tests/adapters/google/test_wrapper.py tests/register/test_manifest.py -q` (41 passed)
- `uv run --frozen ruff check hexgate tests`
- `uv run --frozen ruff format --check hexgate tests`
- `python -m compileall -q hexgate tests`
- `uv run --frozen pytest tests/` (2825 passed, 25 failures matching clean `main` on Windows, 18 skipped, 20 deselected)


victorludvig:
Hello @mikemikimike, did you run integration tests?

guillaume-hexamind (changes requested):
Solid PR overall, but would really need your input on an important related question to be able to merge. Do you think Anthropic models will stay relevant and take over OpenAi and Gemini models? And will solve the next maths millenium problem? Eager to hear your take on the subject

mikemikimike:
Thanks for asking. I think Anthropic models will stay important, but I don’t think they will completely replace OpenAI and Gemini. Each model has its own strengths.  
About solving the next Millennium Prize problem, maybe one day, but I don’t think current AI can do it reliably yet.

mikemikimike:
Thanks for the question. I expect Anthropic to remain relevant, but I do not expect one provider to reliably replace OpenAI and Gemini across all workloads; the trade-offs are workload-dependent. Current AI systems also cannot reliably solve Millennium Prize problems, so any claim that they will do so would be premature. That question is outside the scope of #248. The SDK change and its tests are unchanged; please re-review the technical change and clear the changes-requested state if no code changes are needed.

PR #253, opened 2026-09-21
feat(sdk): add skill definitions to the agent manifest

guillaume-hexamind:
## What

Adds the manifest models that let an agent carry its skills: `SkillResources`, `SkillDefinition`, and `AgentManifest.skills`.

**Nothing populates the field in this PR.** This is the wire contract alone, so it can land and deploy ahead of any producer — the ADK and deepagents discovery lanes, the platform mirror, and the dashboard all build behind it.

## Two shapes that are load-bearing

**`skills` is `list[SkillDefinition] | None = None`, not `= []`.**
`compute_manifest_hash` (`platform/api/hexgate_api/features/agents/service.py`) dumps with `exclude_none=True`, precisely so a new optional field with a `None` default leaves existing digests untouched. A `[]` default is not `None`, so it would enter the canonical JSON and change the digest of **every manifest already registered in every project** — minting one redundant `AgentVersion` per agent on the next `hexgate register`.

**`resources` is `SkillResources | None = None`.**
`None` means the framework does not enumerate L3 contents at all; an instance with empty lists means it does and the skill ships none. These are genuinely different facts — deepagents' discovery is frontmatter-only and never walks `references/`, `assets/` or `scripts/`, while ADK does enumerate them. With a non-optional default both serialize identically and no consumer can render them apart.

## Invariants the validators enforce

All three degrade rather than reject — a manifest that cannot register is worse than one that under-reports, and the operator still sees every tool and the skills themselves.

| Invariant | Behaviour |
| --- | --- |
| `MAX_SKILLS = 200` | truncate + warn |
| `MAX_RESOURCES_PER_SKILL = 100` | truncate + warn |
| `name` unique within the agent | collapse + warn, last declared wins |

**On the uniqueness collapse.** The platform mirror plans `UNIQUE (agent_version_id, name)` on the rationale that both frameworks dedupe upstream. They don't: ADK's per-toolset `_skills` is a name-keyed dict, but an agent holding two skill toolsets — a bundled library plus a user-level one — merges them by concatenation. That is exactly the override idiom, and unguarded it surfaces as an `IntegrityError` on `hexgate register` at the moment someone merges libraries. Collapsing here, in the one place every producer shares, means the platform's constraint can never fire. Last wins because an override is declared after what it overrides; the winner keeps the first occurrence's position, so the ordering the manifest hash sees stays stable.

## Changes

| File | Change |
| --- | --- |
| `hexgate/manifest/models.py` | two models, one field, two validators, two constants, `_truncate` + `_dedupe_by_name` helpers |
| `hexgate/manifest/__init__.py` | exports |
| `hexgate/__init__.py` | re-export beside `AgentManifest` |
| `tests/register/test_manifest.py` | `TestSkillDefinitions` — 9 tests |

SDK-only: no framework import in the diff, and no manifest builder changed.

## Verification

- `make test-one T=tests/register/test_manifest.py` — 20 passed
- `make check` (lint + fmt-check + full suite) — 2855 passed, 18 skipped (pre-existing, `srt` binary not installed)

`make check` rather than `check-all`, since no platform or dashboard surface is touched.

## Deploy ordering — read before merging the first producer

This PR is safe to merge and deploy in any order; it adds a field nobody populates. The gate applies to the **producers**: the platform mirror must be **deployed** before the SDK release that carries ADK or deepagents discovery.

The platform's `AgentManifest` (`schemas.py:643`) sets no `model_config` and so takes pydantic's default `extra="ignore"`. An SDK emitting `skills` against an un-mirrored platform does **not** 500 — the field is silently dropped. Verified against the platform model: the digest is byte-identical with the field absent, populated, or `[]`. So this is a data-loss window, not an outage: skills registered in it are absent from the platform, and because `register_manifest` treats a matching hash as a no-op, an agent whose *only* change is its skill set registers as unchanged. Self-healing after the upgrade, at the cost of one extra `AgentVersion`.

## For the platform mirror PR

Three things this PR deliberately leaves to it:

1. **The `resources | None` distinction must survive into the column** — store `None` as SQL `NULL`, not `{}`.
2. **The digest-side guard belongs there, as a schema parity test.** The hash-continuity test in this PR pins only the SDK's half (that `skills` never serializes as `[]`); the digest is computed in the platform project over the platform's mirror, so nothing asserted here can reach it. No parity test exists anywhere under `platform/api/tests/` today, and `schemas.py:616` asserts the mirror only in a comment. It is feasible: `platform/api` already depends on the SDK (`pyproject.toml:112`, editable path dep), so the test can import both models and diff their field sets — it fails today with `missing on platform: ['skills']`, for the right reason.
3. **Drop the "both frameworks enforce it upstream" rationale** for the unique constraint — it isn't true (see above). The constraint is still right; the SDK now guarantees it, so the DB is a genuine backstop rather than the first line of defence. A duplicate-names test against the DB will no longer see duplicates arrive from a real manifest.

## Rollback

Additive and SDK-only. Nothing reads the field yet, so `git revert` is clean.


victorludvig (approved):
LGTM, my manual and automated reviews did not find any issue.

PR #254, opened 2026-09-21
fix(sdk): expand Google ADK toolsets in the manifest and policy gate

### LINKED ISSUE #248: fix(sdk): hexgate raises TypeError on any Google ADK agent using a BaseToolset (e.g. skills)

guillaume-hexamind:
> **Design doc:** [Agent Skills in Hexgate — ADK & deepagents feasibility](https://app.notion.com/p/3e2fb45dbae281a383efc8d1d97c7a22) — the Notion page (Tech / Product › Docs) specifying how skill logic works in hexgate. This PR implements part of it.

## What

`hexgate` raised `TypeError` on any Google ADK agent whose `tools` list contained a
`BaseToolset`. Both the manifest path and the policy gate now **expand** a toolset
into its constituent tools instead of rejecting it.

- `create_google_manifest` records a toolset's member tools as ordinary `ToolDefinition`s.
- `wrap_tools` wraps a toolset in a new `GuardedToolset` whose every yielded tool is
  policy-gated. It re-resolves the inner toolset on each call rather than snapshotting
  once, so tools a toolset only exposes later in a run are gated too.

Closes #248.

**Skill-agnostic by design.** The fix keys off `BaseToolset` alone — no `SkillToolset`
import, no skill concept — so it covers ADK's MCP toolset, `BigQueryToolset` and
anything built `from_config`. `BaseToolset` has been present throughout the supported
range, so this works against the pinned `google-adk>=1.14` floor with no pin bump.

## Three details that are load-bearing

- **`get_tools` delegates to the inner's `get_tools_with_prefix`, not `get_tools`.**
  That is the method ADK itself calls (`llm_agent.py:185`) and the one that applies the
  inner's `tool_name_prefix`. Delegating to `get_tools` would apply the *wrapper's* unset
  prefix instead, silently dropping the inner's — tools would reach the model unprefixed
  and a policy written against `mymcp_search` would stop matching. The wrapper also
  deliberately does **not** set its own `tool_filter` / `tool_name_prefix`, which would
  double-apply the prefix.
- **The wrapper sets `_use_invocation_cache = False`.** ADK's `@final`
  `get_tools_with_prefix` memoizes per `invocation_id`. A wrapper leaving the flag at its
  default would cache the first result and short-circuit before ever delegating, losing
  tools added mid-invocation. This fails *closed* (the tool is never offered), so it would
  have been a silent bug with no error.
- **`get_auth_config` is delegated.** ADK populates credentials through it *before*
  calling `get_tools` or executing any tool, so omitting it breaks the credential flow of
  any authenticated toolset — the ADK MCP toolset being the obvious case.

## Error handling

The manifest path degrades to a warning when a toolset cannot enumerate offline (a live
MCP connection, say): the agent still registers, minus that toolset's tools. The adapter
path deliberately does **not** swallow — ADK already catches exceptions from toolset
resolution and returns `[]`, so a second layer would hide bugs with no benefit.

`wrap_tool` and `_normalize` are untouched: a direct `wrap_tool(toolset)` still raises its
clear error.

## Verification

- `make check` — ruff check + ruff format --check + full SDK suite: **2863 passed**
  (18 pre-existing opt-in/env skips).
- 13 new tests in `tests/adapters/google/test_toolsets.py`, 4 in `tests/register/test_manifest.py`,
  using a hand-rolled fake toolset rather than `SkillToolset` so they pass on the pinned floor.
- Tests were **mutation-checked** rather than trusted green: reintroducing the
  `get_tools_with_prefix` → `get_tools` bug fails 2 adapter tests + 1 manifest test, and
  flipping `_use_invocation_cache` back to `True` fails 2 adapter tests.
- `#248` reproduced end to end: before, `create_google_manifest` raised; after, the manifest
  lists the toolset's tools, the agent carries a `GuardedToolset`, the allowed tool runs and
  the denied one returns the `[policy_denied]` marker.

`make framework-compat` is not required — no framework contract the probes assert changes here.

## Scope

SDK only. No platform, dashboard, or wire-format change; no migration. Single commit, so
`git revert` restores the previous behaviour exactly (which is the `TypeError`, so reverting
re-opens #248 but breaks nothing that was working).


victorludvig:
I approve the logic, 2 edge cases to check:
### Code review — PR #254 — 2 findings

1. **A skill that registers its own tools slips past the policy gate entirely.**
   `hexgate/adapters/google/tools.py:188-193`

   **Trigger:** An agent uses a toolset that adds its tools to the LLM request itself — `ComputerUseToolset` today, or any third-party one following the pattern ADK documents.

   **Break:** The wrapper passes that call straight to the inner toolset, which hands ADK its *raw* tools. ADK dispatches from that list, so the gated copies are built and thrown away — the tool runs even under a deny-all policy.

   **Notice:** Nobody — the call returns a normal result, and nothing is audited, because no policy decision was ever made.

   Worth fixing here: before this PR a toolset raised `TypeError`, so it failed closed and loud. Reach is narrow today (that one ADK toolset, still experimental), but the PR claims to cover any toolset.

2. **A skill that can't be reached at register time gets permanently denied.**
   `hexgate/manifest/google.py:82-91`

   **Trigger:** First `hexgate register` for an agent with a skill/MCP toolset, while the MCP server isn't up.

   **Break:** Register succeeds, but the skill's tools are missing from the manifest, so the one-time starter policy denies them all — and re-registering later never fixes it, because policy is only ever generated once.

   **Notice:** Nobody — those tools just return `[policy_denied]`, which looks like a policy you wrote on purpose.

   Also fires when the server *is* up but needs auth: registration connects without credentials, since those only exist at run time.

guillaume-hexamind:
Thanks — both are real issues with the toolset seam. Finding 1 is fixed in 81bc457f; finding 2 I've confirmed in detail but I'd like to argue it belongs in a follow-up rather than here. One correction to the first one's specifics.

## 1 — fixed, but the ComputerUseToolset case wasn't actually exposed

The mechanism is real and I've fixed it. But I couldn't reproduce the specific break you described, so I measured it rather than guessing.

ADK's `base_llm_flow._process_agent_tools` (L435-457) does, per toolset entry:

1. `toolset.process_llm_request(...)` — inner may write `tools_dict` itself
2. `_convert_tool_union_to_tools(...)` → our `get_tools_with_prefix` → gated copies
3. `tool.process_llm_request(...)` per gated copy → `append_tools` → `tools_dict[name] = tool`

So our gated copies register **after** the inner's raw ones and overwrite same-named keys. Reproducing all three shapes against the wrapper as it stood:

| Shape | final `tools_dict` | gated? |
|---|---|---|
| A — direct-register, no prefix (**ComputerUseToolset as shipped**) | `click` | ✅ gated |
| B — direct-register **+ `tool_name_prefix`** | `click`, `cu_click` | ❌ raw `click` ungated |
| C — registers a tool `get_tools()` never returns | `click`, `secret_exec` | ❌ `secret_exec` ungated |

`ComputerUseToolset.process_llm_request` iterates the same `self._tools` that `get_tools` returns and sets no prefix, and `ComputerUseTool._get_declaration()` returns a real declaration — so `append_tools` does write the key and our gated copy wins. That one was already covered.

B and C are genuinely exposed, and your point that this is a **regression** stands either way: before this PR a toolset raised `TypeError`, so it failed closed and loud.

**Fix:** `GuardedToolset.process_llm_request` now snapshots `tools_dict`, delegates, then re-gates every entry the inner added or replaced. Identity comparison, so entries the inner never touched are left alone and never double-wrapped. Three tests added; the two bypass tests fail if I revert the fix.

One judgment call worth flagging: re-gating whatever a third-party toolset wrote is slightly more invasive than "wrap what `get_tools` returns", and could surprise a toolset that deliberately registers a pass-through helper. I think ungated dispatch is clearly worse, but say the word if you'd rather scope it to unprefixed-name collisions only.

## 2 — confirmed, but I'd like to do it as a follow-up

You're right, and I traced it through:

- `agents/service.py:483` — `if agent_created:` seeds the starter policy; the docstring is explicit that re-registers "leave the operator's edited policy alone"
- `_default_policy_for_manifest` builds its buckets from `manifest.tools`, and the generated policy is `default_policy: mode: deny`

So a toolset that can't enumerate at **first** register → absent from the manifest → absent from the starter policy → permanently `[policy_denied]`, and re-registering never repairs it. Your auth sub-point holds too: we pass no `ReadonlyContext` and credentials only exist at run time.

Two things that make me want to keep it out of this PR:

- It fails **closed**, and it isn't introduced here — before this PR that agent couldn't register at all. It's a new failure *mode* only because registration now succeeds.
- The real fix is platform-side (policy is never regenerated) plus a CLI-side loud summary / `--fail-on-partial`. That breaks A1's SDK-only scope and its clean-revert property.

So I'd keep the swallow and open a follow-up for "registration under-reports a toolset → operator has no way to notice or repair". Happy to file it — or if you'd rather block this PR on at least the CLI-visibility half, I'll add that here instead.

One thing I did fix now, since it's the same code path: the warning *does* reach stderr, but its traceback led with `RuntimeError: no running event loop` — my own `_run_sync` control-flow probe — which buried the real cause. Unchained.

Also merged `main` in (253 landed and touched `test_manifest.py`) rather than rebasing, so your review anchors survive. `make check`: 2875 passed.

victorludvig (approved):
Approved, consider opening an issue for the follow-up so that it's not forgotten.

guillaume-hexamind:
Filed finding 2 as #257 so it does not get lost. Still happy to pull the CLI-visibility half into this PR instead if you would rather not merge with it outstanding — otherwise this is ready for another look.

PR #256, opened 2026-09-21
feat(platform-api): persist agent skills from the manifest

guillaume-hexamind:
Implements **PR M2 — Platform: accept and persist skills** (`plans/skills/PRM2.md`). Now targets `main`, since M1 (#253) has merged — the effective diff here is platform-API only.

## What changes

The platform accepts a manifest carrying `skills`, persists one row per skill per agent version, and surfaces them on the dashboard read path.

- **`schemas.py`** — mirrors M1's `SkillResources` and `SkillDefinition`, and adds `AgentManifest.skills: Optional[list[SkillDefinition]] = None`.
- **`models.py`** — new `Skill` table, modelled on `Tool`: `UNIQUE (agent_version_id, name)`, JSON `resources` / `allowed_tools` / `additional_tools`.
- **`agents/service.py`** — `_create_skills`, called from `register_manifest` immediately after `_create_tools`, so a version and its skills commit together. `manifest.skills is None` writes nothing.
- **`agents/router.py`** — 409 on a manifest declaring two skills of one name.
- **`core/ids.py`** — `skl_` row-id prefix.
- **`tests/test_actor_columns.py`** — `skill` joins `tool` in `EXEMPT_TABLES`.

## Decisions worth a reviewer's attention

**No SQL migration.** `skill` is a brand-new table and `core/db.py` runs `SQLModel.metadata.create_all`, which adds missing *tables* on every stage. `COLUMNS_NEEDING_A_MIGRATION` is for columns added to tables that already shipped, so nothing goes there either.

**No caps or dedupe mirrored.** The SDK's `MAX_SKILLS` / `MAX_RESOURCES_PER_SKILL` and its same-name collapse stay SDK-side: it truncates before it sends, and the platform's job is to accept what it is given. Hashing is no argument either way — `compute_manifest_hash` runs over what the platform *parsed*, so it would stay self-consistent under a cap.

**`resources` distinguishes SQL NULL from `{}`.** NULL means the framework does not enumerate L3 contents; `{}` means it does and the skill ships none. M6 renders the two differently. The column is `JSON(none_as_null=True)` — without that, SQLAlchemy persists Python `None` as the JSON scalar `'null'` and `WHERE resources IS NULL` matches zero rows.

## Hash continuity (gate G2)

`compute_manifest_hash` uses `exclude_none=True`, so a pre-skills payload and the same payload carrying `skills: null` hash identically: an old SDK re-registering against the new platform matches its existing version rather than minting a duplicate. `test_manifest_hash_is_unchanged_by_the_new_field` is the guard; `test_re_register_with_new_skills_creates_a_new_version` is its flip side.

**This must be *deployed*, not merely merged, before the SDK release containing A2** — the first PR that actually populates `skills`. Until then `extra='ignore'` silently drops the field on an older platform.

## Review fixes (second commit)

Four findings, each verified against running code before being fixed:

1. **Duplicate skill names 500'd.** Nothing in `platform/api` registers an exception handler, so the `UNIQUE` `IntegrityError` reached Starlette's default handler. Now pre-checked in `_create_skills` → `DuplicateSkillNameError` → **409**, the shape `create_project` already uses and for its stated reason. The constraint stays as the backstop, now asserted directly.
2. **`resources=None` was stored as JSON `'null'`, not SQL NULL** (`JSON.none_as_null` defaults to `False`). The two stayed distinguishable so the goal held, but the column comment misdescribed the storage and a later `IS NULL` query would have matched nothing. Fixed while nothing reads the column, and pinned with a raw-SQL assertion.
3. **A docstring justified skipping caps by a digest the SDK computes over the manifest.** No such digest exists — `compute_manifest_hash` is platform-only, and the SDK's `content_hash` is a per-`SKILL.md` body digest. Decision unchanged, reason corrected.
4. **The no-skills-key test was vacuous.** It asserted `!= 422` behind an unauthenticated bearer; FastAPI resolves dependencies before validating the body, so it 401'd with the schema never reached — an empty manifest passed it too. Now authenticated, asserting **201** and zero skill rows. The sibling legacy-shape test carried the same misconception in a comment asserting the opposite order, so it is corrected rather than left to be copied again.

## Out of scope

Starter-policy generation from skills (M5), dashboard rendering (M6 — it reads `AgentManifestView.manifest.skills`, which this PR populates with no further endpoint work). No policy behaviour changes.

## Tests

Ten cases in `tests/features/agents/test_agents.py`, extending `_sample_manifest` with an optional `skills` argument: row persistence and field round-trip, `skills=None` writing nothing, the compat guarantee, the hash guard, NULL vs `{}` resources (ORM *and* SQL), duplicate names at both the service and HTTP layers, the UNIQUE backstop, the manifest read path, and a genuine new version on a skill change.

Each fix was checked to be load-bearing by reverting it and confirming the matching test fails.

Verified after merging `main`: `make check-all` exits 0 — SDK 2855 passed, platform API 907 passed, dashboard and collector green.


victorludvig (approved):
LGTM

